PR #1, opened 2021-01-04
Adding Support for ArcOS

dt-arrcus:
Few Notes:

I have followed all the steps on your https://netsim-tools.readthedocs.io/en/latest/platforms.html#contributing-new-devices page.  Thanks for making it straight forward to add a new device type.  This PR passes the tests you laid out in that document.

A customer wishing to use ArcOS nodes in these topologies will need to reach out to their Arrcus Rep for access to both the Vagrant Box file and Ansible collection for ArcOS.   I added a note to that in the install.md file

Since ArcOS doesn't support a gather_facts module, I wrote a quick playbook so the collect configs would still be supported.  Not sure where you wanted to store that so I added a new directory called `vendor_specific_tasks`.  Now that I type that its sort of a terrible name,  so if you need that renamed let me know.

### DIFF
--- a/ansible/config/arcos.yml
+++ b/ansible/config/arcos.yml
@@ -1,18 +1,13 @@
-
 ---
 - name: Push Candidate Config
   template:
     src: '{{ config_template }}'
-    dest: '/tmp/.{{ inventory_hostname }}.xml'
+    dest: '/tmp/.{{ inventory_hostname }}.txt'
 
 - name: Apply candidate config
   arrcus.arcos.arcos_config:
-    src: "/tmp/.{{ inventory_hostname }}.xml"
+    src: "/tmp/.{{ inventory_hostname }}.txt"
     load_operation: merge
     comment: "initial config pushed by netsim"
-
-- name: Remove temp file
-  file:
-   state: absent
-   path: '/tmp/.{{ inventory_hostname }}.xml'
+    delete_src_file: true
 
--- a/ansible/config/arcos.yml
+++ b/ansible/config/arcos.yml
@@ -1,0 +1,40 @@
+---
+- name: ArcOS config pre 2.10
+  block:
+    - name: Push BGP Candidate config
+      template:
+        src: '{{ config_template }}'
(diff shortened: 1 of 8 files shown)


ipspace:
Thanks a million. As I don't have access to Arrcos images (and no time to play with them ;) please try to implement these ideas:

**In "ansible/config/arcos.yml"**

* The config modules usually accept Jinja2 templates. If arrcus.arcos.arcos_config works the same way you could skip the first step
* Looks like Arrcos config file is a text file. If that's the case, please use _txt_ not _xml_ extension (if you need an extra temporary file - see above)

**in "vendor_specific_tasks/arcos/collect_config.yml" **

Is that really the simplest way to collect information from Arrcos device? Creating a file and then reading a file and then... seems way too convoluted.

After you have a streamlined version of both tasks I'll merge the pull request and rearrange the configuration gathering to be more modular (there might be other platforms that don't return configuration data in fact gathering).

dt-arrcus:
Thanks for the feedback.  

Today, sadly, we don't accept a template as an input source, so the template rendering is a two-step process.  Hopefully, that will change in the future, and once it does, I will update this repo as well.

Agreed on the file extension.  That was a leftover relic from when I was pushing XML config, but I am using CLI spec here, as you correctly pointed out.  Also, I can streamline the config gathering right from bash.  I will work on those and add them to this PR.

ipspace:
Agreed on the file extension. That was a leftover relic from when I was pushing XML config, but I am using CLI spec here, as you correctly pointed out. Also, I can streamline the config gathering right from bash. I will work on those and add them to this PR.

Don’t overcomplicate things - it’s just that I don’t like too many temporary files (and each Ansible task takes forever to execute).

Thank you! Ivan

dt-arrcus:
Made the following changes:

- in arcos_config module, added a `delete_src_file` attribute, so there is no longer a need for a 2nd step of deleting the temp file.  Note once that module supports templates directly I will clean up more.
- Added the ability to get the running-config from ArcOS device in a single task.

PR #5, opened 2021-04-01
adding support for OSPF in ArcOS

dathelen:
Added:
- OSPF config template for OSPF in ArcOS
- VB template for ArcOS VMDK

Currently, no support for unnumbered interfaces in ArcOS.  But I will add when/if that comes.


ipspace:
Thanks a million for making this happen!

PR #10, opened 2021-05-05
Fix running vSRX under AMD CPUs.

### LINKED ISSUE #9: Running with an AMD CPU
I had to configure the following in `netsim/templates/provider/libvirt/vsrx-domain.j2`, otherwise the vSRX got stuck on
```Booting [/packages/sets/pending/boot/os-kernel/kernel]...```

This sets the `--cpu` parameter for `virt-install` to `qemu64` instead of the default `host`.

```
{{ name }}.vm.provider :libvirt do |domain|
    domain.default_prefix = "#{cwd}_"
    domain.cpus = 2
    domain.memory = 4096
    domain.disk_bus = "ide"
 +  domain.cpu_mode = "custom"
end
```

I was wondering if you have any experiences with AMD/intel differences in this regard. I could submit a PR that insert this line whenever a AMD CPU was used. The only problem is that I have no way to test whether this happens for all/most/any AMD setups except for mine because I have no other system available.

Kircheneer:
Optionally add in `domain.cpu_mode = "custom"` in the `vm.provider` section of the Vagrantfile for vSRX VMs. This forces the cpu mode to `qemu64` which seems to work for running vSRXs, although the performance is quite bad.

I'ved used the transform method of the Provider class for this because it seemed a right use case for it. I can move it elsewhere if you want.

Fixes #9

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -49,8 +49,8 @@ class Provider(Callback):
     return self._default_template_name
 
   def transform(self,topology):
-    if "processor" not in topology:
-      topology["processor"] = platform.processor().split()[0]
+    if "processor" not in topology.defaults:
+      topology.defaults.processor = platform.processor().split()[0]
 
   def dump(self,topology):
     template_path = self.get_template_path()
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -5,6 +5,7 @@
 # Provider class and replacing or augmenting its methods (most commonly, transform)
 #
 
+import platform
 import os
 import sys
 import importlib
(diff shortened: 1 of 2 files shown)


ipspace:
Nice fix. Love it. The only thing I would change would be to use topology.defaults.processor - earlier stages of the data model transformation are extremely annoyed if an unknown top-level key appears (to save the users from hard-to-find typos), while the **defaults** are not so strictly checked.

Also, by adding a value to **defaults**, you could use a local **topology-defaults.yml** file that would define the processor for multiple topologies.

Kircheneer:
Like that?

ipspace:
Thank you, that's it.

Just FYI (and you already did it that way), coming from other programming languages like Pascal and JS, I prefer the Box syntax ;) 

The only catch is that if a dictionary key doesn't exist and you use it, Box creates an empty sub-dictionary, and that doesn't look nice in final YAML, that's why I still use `if "something" in dict` instead of `if dict.something`

PR #11, opened 2021-05-05
Fix LLDP configuration for Junos

Kircheneer:
My vSRX VMs came up without LLDP enabled. I'm not well versed enough in Junos to know is this is generally applicable so a sanity check there would be great.


ipspace:
You're absolutely right. Thanks for the fix!

PR #12, opened 2021-05-06
Fix NXOS under AMD CPUs

Kircheneer:
#9

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -47,6 +47,7 @@ The Vagrantfile templates for individual network devices were derived from the f
 
 **Notes:**
 
+* There is a Virtualbox based Vagrant box available for the Nexus 9300v which can be converted using the Vagrant plugin `vagrant mutate`
 * Nexus 9300v takes almost exactly a minute after SSH becomes available on the management interface to "boot" the Ethernet linecard (module 1) and activate Ethernet interfaces. Leave it alone for a while even when Vagrant tells you the VM is ready.
 * If you're experiencing high CPU utilization with Cisco CSR, [set halt_poll_ns to zero](https://codingpackets.com/blog/kvm-host-high-cpu-fix/).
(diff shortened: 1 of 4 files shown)


Kircheneer:
So, while reviewing this I realized that my platform detection code only worked properly under Windows. I've implemented this for my Ubuntu 20.04 VM and I think it should run well under most modern AMD systems. I've put in a check for MacOS as well but I don't have a device for testing. We could also assume Intel in case of MacOS - I'm not well versed in what kind of hardware Apple devices have built-in.

Kircheneer:
Don't look into this too closely yet - I will need some more time until its ready.

ipspace:
```
uname -v
Darwin Kernel Version 19.6.0: Mon Apr 12 20:57:45 PDT 2021; root:xnu-6153.141.28.1~1/RELEASE_X86_64
```

```
$ python3
Python 3.9.1 (default, Jan  8 2021, 17:17:17)
[Clang 12.0.0 (clang-1200.0.32.28)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.processor()
'i386'
>>>
```

OSX is OK - Apple uses Intel CPUs, and whatever is reported is NOT anywhere close to AMD.

Kircheneer:
I've done the following things:
- Change the Vagrantfile template for NXOS so that it
  - Runs the Cisco supplied Vagrant box properly
  - Runs properly under my AMD system
- Add a short hint in the docs to the Cisco supplied Vagrant box and how to use it
- Work in your comment about the logic swapping

The only unknown is: Does your blog-post built Vagrant box still work under these settings?

ipspace:
Of course it broke my setup. What else could one expect ;) 

Got it to work with your **mutate** trick (which I love) and a tiny change in the template (and I suspect even my original box would work with that change).

Kircheneer:
Go thank Dinesh Dutt for that, he showed me that trick! Just keep in mind the default credentials for the Cisco supplied box are admin:admin - hence my adding the username to the initial configuration for NXOS.

PR #14, opened 2021-05-08
Expand install libvirt to install Ansible collections

Kircheneer:
I noticed that a couple of Ansible collections have to be installed as well in order to use all of your tools. This should take care of that installation.


ipspace:
At least the "legacy" NXOS and Junos modules should be in Ansible distro, so I'm wondering why you don't have them. 

In any case, the change makes perfect sense (and is along the lines of what Ansible team is preaching), but if we do install network device collections, we should do it for all supported devices (+ IOS, EOS).

Also, it looks like the "changed when" test would report no changes when one collection is installed but another one is not (because "already installed" string would be printed).

Finally, please submit a PR against the dev_0.6.1 branch, not against the master branch.

PR #15, opened 2021-05-09
Expand install libvirt to install Ansible collections

Kircheneer:
Re #14:

> At least the "legacy" NXOS and Junos modules should be in Ansible distro, so I'm wondering why you don't have them.

Doesn't that have something to do with the introductions of collections in Ansible 2.9? I thought they moved everything but the ansible-base to Ansible galaxy.

> In any case, the change makes perfect sense (and is along the lines of what Ansible team is preaching), but if we do install network device collections, we should do it for all supported devices (+ IOS, EOS).

Fixed. Providing a [requirements.yml](https://docs.ansible.com/ansible/latest/galaxy/user_guide.html#install-multiple-collections-with-a-requirements-file) for Ansible collections.

> Also, it looks like the "changed when" test would report no changes when one collection is installed but another one is not (because "already installed" string would be printed).

This is fixed by just using `ansible-galaxy collection install -r requirements.yml`

> Finally, please submit a PR against the dev_0.6.1 branch, not against the master branch.

Of course. I wasn't able to find dev_0.6.1, hence the 0.6.2 as with the other PR.

### DIFF
--- a/install.libvirt
+++ b/install.libvirt
@@ -11,6 +11,7 @@
   - name: Install apt dependencies
     apt:
       pkg:
+        - ansible
         - libvirt-dev
         - qemu-kvm
         - vagrant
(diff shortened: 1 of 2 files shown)


ipspace:
I wouldn't fix the collection versions (unless that's a requirement), or we'll have to bump them up continuously. This is a test environment, installing _latest_ versions by default should be OK (and everyone can change that if they so desire).

As for "Ansible collections" - you don't have to use them (yet), old playbooks still work. The only difference is that you get network module fixes at the speed of new Ansible releases if you don't download individual collections from Ansible Galaxy.

PR #17, opened 2021-05-10
Fix Ansible install docs

Kircheneer:
The `wget` command downloaded the HTML page rather than the raw script. I've updated that and included the `requirements.yml` file in the download because otherwise the command would fail.


ipspace:
Nice catch. Will merge this and then modify it to create the requirements file from Ansible playbook, so it will still be just a single file to download.

PR #18, opened 2021-05-17
Remove all unused imports

Kircheneer:
A little cleanup pass over the imports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,13 @@
 .vagrant
 .DS_Store
 
+# Ignore editor files
+.vscode/
+.idea/
+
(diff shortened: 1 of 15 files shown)


ipspace:
Thank you! I guess that's the right way of doing things... but of course I'll pay for that when I'll add some new functionality and forget to add the import at the top ;))

Please submit the PR against dev_0.6.3. It's not something that should be urgently fixed, so it doesn't make sense to merge it straight into master branch.

Kircheneer:
At some point I'll (probably) start remembering to change the merge target branch. Promise 👍

PR #47, opened 2021-11-09
Fix vagrant-mutate installation instructions

grigorenkoae:
As per documentation https://github.com/sciurus/vagrant-mutate#vagrant-mutate-1


ipspace (approved):
Thanks a million!

PR #49, opened 2021-11-22
VyOS support

ssasso:
Support for VyOS device type (https://vyos.io/).
Supported features:

- Initial config
- Config fetch
- BGP Module
- OSPF Module

Documentation is updated as well.


ipspace:
Great job. Thanks a million!

PR #52, opened 2021-11-29
Fix: IOS-XE syntax for domain lookup

ssasso:
`no ip domain-lookup` command is not available on newer Cisco IOS-XE versions.
The updated command is `no ip domain lookup`, which is available also on older versions.

Tested on:
* 03.15.00.S.155
* 17.03.03


ipspace:
Thank you!

PR #56, opened 2021-12-09
WIP: Support link without IP configuration

ssasso:
In certain cases it might be required to have links configured without IP addresses
(and without all the "ip unnumbered" stuff)

With this patch, it should be possible to specify **prefix** as _false_ on the link attributes. With this, the _ipv4_ or _ipv6_ keys will not be present on the link augmented attributes, hence skipping all the IP address configuration part of the templates.

This would be useful i.e. when testing VLANs, or M(c)LAG peer link or - in general - when you want a link but you would like to configure every IP parameter on your own.

### DIFF
--- a/docs/links.md
+++ b/docs/links.md
@@ -43,6 +43,8 @@ A dictionary describing an individual link contains *node names* as well as *add
 * **role** -- link role, used to select custom addressing pool or specific configuration module behavior.
 * **bandwidth** -- link bandwidth (used to configure interface bandwidth).
 
+If you need a link without any IP configuration (i.e., if you need to test VLANs on it), you can specify **prefix** as *false*.
+
 Links could contain [additional attributes](#custom-attributes-in-node-and-link-data) like *delay* and [module-specific attributes](modules.md#module-specific-node-and-link-attributes). Additional (custom) attributes have to be defined in **defaults.link_attr** list to differentiate them from misspelled node names. 
 
 ### Example
--- a/netsim/addressing.py
+++ b/netsim/addressing.py
@@ -235,7 +235,7 @@ def normalize_af(af: str) -> str:
   return 'ipv4' if af == 'ip' else af
 
 def parse_prefix(prefix: str) -> typing.Dict:
-  if isinstance(prefix,bool) and prefix is False:
+  if not prefix:
     return {}
   supported_af = ['ip','ipv4','ipv6']
   prefix_list: typing.Dict = {}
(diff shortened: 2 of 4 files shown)


ipspace:
Great idea. I would add the explanation to the **prefix** bullet. No worries about that, I can do it after merging the PR.

I wonder whether it might make sense to support empty **prefix** dictionary, so your test would be simplified as 'if not prefix` (calling routing checks whether **prefix** key exists before calling **parse_prefix**, so there's no chance of a false positive). Also, typing of the `prefix` input parameter has to be fixed (I can do that after merging).

More importantly, I would love to see a test case. Create a sample topology in `tests/topology/input` directory, and run `create-transformation-tests.sh` in `tests` directory to create desired results in `tests/topology/expected` directory.

ipspace:
... forgot to mention, you can run tests with `run-tests.sh` executed from within the `tests` directory before pushing the commit.

ssasso:
I will add the topology test that I did.

About the support for empty *prefix* dictionary, you mean to change it in *augment/links.py*, row 74?

Sth like:
```
  if 'prefix' in link:
    if not link['prefix']:
        pfx_list = {}
    else:
        pfx_list = addressing.parse_prefix(link.prefix)

```
?

ipspace:
Don't change augment/links.py -- that code checks whether there's a prefix specified on the link and if so calls the parsing routine. That's a decent enough demarcation, the calling code shouldn't care what's in the prefix.

Just replace your test in netsim/addressing.py (`if isinstance(prefix,bool) and prefix is False:`) with `if not prefix:` and then you can get unnumbered link with an empty dictionary (**prefix:**) or an empty string (**prefix: ""**) or a False value (**prefix: False**).

ssasso:
Additional comment: on Ubuntu 20.04 **pytest** seems to be the one for python2, while **pytest-3** the one for python3. While running this one, I got "4 passed, 57 warnings".

ipspace:
"on Ubuntu 20.04 pytest seems to be the one for python2" -- I installed pytest with pip, so I got **pytest** as a command. Will change that to **python3 -m pytest**

The end result is correct -- as long as the test suite doesn't throw an error you're good to go. Have to clean up the warnings -- they are generated by all the "let's check this particular error" test cases.

ipspace:
Merged, polished links.md, adjusted the squashed merge commit and force-pushed resulting dev branch to github.

PR #57, opened 2021-12-14
Add sros srlinux support (using Containerlab)

jbemmel:
These changes add 'srlinux' and 'sros' to the list of supported devices, for initial/BGP/OSPF use cases

Some comments:
* Dependency on nokia.grpc Ansible Galaxy collection: https://galaxy.ansible.com/nokia/grpc; Ansible Galaxy has its own way of managing dependencies (outside requirements.txt I believe)
* SRLinux uses gNMI with its native YANG model
* For SROS 2 incremental provisioning methods are provided: OpenConfig(default) and Nokia native, both using gNMI
 + The driver attempts to detect when OpenConfig cannot be used (in case of OSPF for example)
 + For now, OpenConfig use is set to 'False' because of a pending PR for nokia.grpc (to support 'origin')
* 'regular' CLI scripts for SROS and SRLinux are included for completeness, but not currently referenced
 
Note that the incremental method is particularly suited to a GitOps workflow, and could likely be applied to other vendor devices as well


ipspace:
Thanks a million -- that's a huge chunk of work. It looks perfect, and I trust you thoroughly tested the configuration bits ;). Also, I love the improved error reporting.

It would be nice if you could add a sample topology in `tests/integration` (a few nodes, OSPF + BGP, maybe IPv6 as well) so whoever wants to run a quick check whether they can bring up SR/Linux or SR/OS can do so.

There are just a few changes in the common code I'm worried about 

* Several times in `netsim/modules/__init__.py` you changed `get(x,[])` to `get(x,[]) or []`. That would be a nice safeguard in case x is None or an empty dictionary, but I don't think we could ever get to that point. Did you encounter edge cases that prompted you to add that? If you did, then I have to detect them earlier on (or change the code not to create empty dictionaries), and add them to the "this must fail" tests.
* There's no need for `nodes|default([])` in clab.j2. The transformation code makes sure nodes is a non-empty lists. Good catch on the links though.
* Can you use the same license for multiple SR-OS nodes? If so, it would make sense to make it global -- have to check how to do that in the easiest possible way. Likewise for `binds` -- I would expect them to be pretty static, so maybe they should go into topology details. In any case, these settings (if needed) have to be documented somewhere or we'll have a few unhappy users.

Finally, why did you change calls to `containerlab` to `custom_containerlab`? There is no `custom_containerlab` in my _containerlab_ installation.

Thanks again for great work! Ivan

jbemmel:
Hi Ivan,

I'm still working on
https://github.com/jbemmel/netsim-examples/tree/jvb_add_sros_srlinux_bgp_addpath_example/BGP/Multipath_sros,
wanted to get the core bits out the door first

I came across the empty list issue when I built a topology with a single
SROS node (while developing the Ansible scripts). I repaired a few other
places out of an abundance of caution, probably more conservative than
needed (but I figured it wouldn't hurt)

Yes, one can use a single license file, and in Containerlab you can put
that under the 'kinds' section (which NetSim isn't using right now). Would
be an easy addition.
The binds do typically have to be per node, as soon as there is 1 unique
file all of them must be defined at the lowest level.

Roman recently added a feature to enable passwordless SSH on
Containerlab (on my suggestion), for SRL nodes at least; the PR isn't part
of the main release yet so I had to use a custom built binary.
Using 'custom_containerlab' as a binary serves as a reminder (and also
because NetSim-Tools doesn't make it particularly easy to set the PATH to
binaries...)

I would suggest we wait for the Containerlab release with this passwordless
SSH feature, before we release this code (using the regular
'containerlab' binary)

Regards,
Jeroen

ipspace:
Thanks for the hint - I never tried a "single node, no links" topology. Will add two tests to cover that.

Waiting for you to tell me to merge it. No rush; I probably won't push another release out this year... unless I get bored and implement MPLS with LDP+BGP-LU and push that out.

jbemmel:
> Thanks for the hint - I never tried a "single node, no links" topology. Will add two tests to cover that.
> 
> Waiting for you to tell me to merge it. No rush; I probably won't push another release out this year... unless I get bored and implement MPLS with LDP+BGP-LU and push that out.

This is ready to be merged, I've implemented gNMI also for SRLinux and no longer depend on a pending new release for Containerlab or special binary

ipspace:
Thank you!

Merged as a single commit and amended it (to have a clean history):

* remove unnecessary `or []` from **modules/__init__.py**
* fix **clab.j2** (it crashed when there were no groups defined)
* fix IPv6 table in **platforms.md** (extra cells due to vertical bar in a new line).

Force-pushed the resulting **dev** branch.

PR #71, opened 2021-12-16
Merge branch 'dev' into dev

petercrocker:
Suggested changes:

- two cumulus j2 templates are missing `do write` at the end, causing two issues:
    - changes aren't saved, only left as running config in vtysh, so don't get reflected in cumulus `net show` commands
    - if you configure both OSPF and BGP module settings, the second config.sh file overwrites the first, resulting in only one of two modules being configured
 
- sort ansible.cfg settings, which will make it easier to add/diff them when they change in the future

- add 'ansible_managed = # Ansible Managed File' to ansible.cfg. This allows you to put {{ ansible_managed }} in the top of any j2 template, making it clear to someone looking at the resulting templated file that it's auto-generated by ansible, not user-editable. this setting doesn't harm anything in netsim_tools, but makes it easier to extend.

- add 'filter_plugins = ./filter_plugins' to ansible.cfg. I use an arista-provided filter plugin (aka an ansible filter like "| ipaddr" ) called natural_sort which properly sorts numbers like 010 before 11. this setting doesn't harm anything in netsim_tools, but makes it easier to extend.

- change 'roles_path=./roles' so that i can add ansible roles in the roles/ directory instead of cluttering up the root netsim-tools folder. this setting doesn't harm anything in netsim_tools, but makes it easier to extend.

- change default libvirt mem for cumulus image from 768m to 1024m. In the past, 768 was enough for the last 5 years, but it's just at a tipping point where certain more complex configs will exhaust the memory. 

- change default libvirt mem for linux image from 512m to 1024m. If you want to do any kind of testing on the linux boxes and install packages, you can quickly run out of memory.

As a related point, maybe it would be better to expose the VM default settings as user-configurable, as opposed to being compiled in to the python package? For example, netlab allows the -defaults option, but this doesn't expose VM settings.

- change bridge name in vagrant-libvirt.xml to be virbr9, instead of vibr2. Less likely to collide with an existing bridge on a user's system (which it does on mine!)

- correct MAC addresses on lines 22-23 of vagrant-libvirt.xml. Ivan you need to work on your hex counting ;) it's 08, 09, 0A not 08, 09, 10! The result is the ansible playbooks using the hosts.yml file created won't connect to the vms if they get those bad mac addresses.

- added more MAC addresses to the vagrant-libvirt.xml. Otherwise a user can create a sim with a limit of about 10 devices.


petercrocker:
Closing, replacing the PR

PR #72, opened 2021-12-16
Print exception to stdout such that user can understand what went wrong

jbemmel:
In my case: netlab wasn't in the PATH


ipspace:
Thank you. After merging I'll add the full command (with arguments) that failed to the printout (who knows when that might come handy). 

We need something similar when running Ansible playbooks. Will add.

ipspace:
Added more info on which command was executed and which **netlab** command failed. Pushed updated commit to the **dev** branch.

Also: please don't submit a PR against master branch.

PR #74, opened 2021-12-17
[WIP] Support Vagrant 2.2.19 and vagrant-libvirt > 0.7.0

ssasso:
**WORK IN PROGRESS _ DO NOT MERGE (yet)**

Starting from vagrant-libvirt version 0.7.0, every network is deleted on vagrant destroy, including the management network.

This behaviour is related to the *always_destroy* parameter default value, which is documented on:
* https://newreleases.io/project/github/vagrant-libvirt/vagrant-libvirt/release/0.7.0
* https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1381

This is not what we want on **netsim-tools**, because the management network is configured on the installation phase with all the DHCP static leases.

A PR has been opened on vagrant-libvirt to have a new option to better control this: https://github.com/vagrant-libvirt/vagrant-libvirt/pull/1425

This PR in netsim-tools allows to use the new proposed option.
A gem version check has been introduced on the Vagrantfile template, so we need to be sure on the vagrant-libvirt version before merging this.

Additionally: the *vm.guest" type *:other* is not usable anymore on newer Vagrant versions. This has been removed from the *routeros* template.

### DIFF
--- a/netsim/templates/provider/libvirt/Vagrantfile.j2
+++ b/netsim/templates/provider/libvirt/Vagrantfile.j2
@@ -7,6 +7,9 @@ Vagrant.configure("2") do |config|
     {{ name }}.vm.provider :libvirt do |domain|
       domain.management_network_mac = "{{n.mgmt.mac}}"
       domain.qemu_use_session = false
+      if Gem.loaded_specs['vagrant-libvirt'].version > Gem::Version.new('0.7.0')
+        domain.management_network_keep = true
+      end
     end
 {%   endif %}
 {%  include n.device ~ "-domain.j2" %}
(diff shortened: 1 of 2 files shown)


ipspace:
I think it would be much better to check whether 'vagrant-libvirt' network is present and create it if it's not on **netlab up** instead of fighting the ever-changing windmills.

Right now, the **libvirt.py** provider has a **start_lab** hook that sets LLDP-enabled flag on Linux bridges. We could:

* Change the **start_lab** hook into **post_start_lab**
* Add **pre_start_lab** hook
* In **pre_start_lab** hook execute **virsh net-list**, check for **vagrant-libvirt** and create it if needed.

Makes sense?

ssasso:
That makes a lot of sense! This will also allow, in the future, to use the **pre_** and **post_** start hooks also for other tasks, if needed.
I'll submit a different PR in the next hour for that, and if you like it, we can delete this one.

PR #75, opened 2021-12-18
Pre and Post start hooks for libvirt LAB UP

ssasso:
Updated code to run pre_start and post_start hooks on **netlab up**.

This is required to have also a check on the libvirt management network bridge (pre_start)
 and creating it if not already present. This is required starting from vagrant-libvirt
 0.7.0, which destroys the bridge at every vagrant destroy.

Additionally: the *vm.guest* type :other is not usable anymore on newer Vagrant versions. This has been removed from the routeros template.

This PR replaces https://github.com/ipspace/netsim-tools/pull/74 .


ipspace:
Looks good to me. Thanks a million!

PR #85, opened 2021-12-22
Add SROS/SRLinux ISIS and SR-MPLS support

jbemmel:


ipspace:
I'm guessing this one is not yet finished -- the tests are failing (it seems like you're using attributes without defining them), and it needs to be rebased as the **dev** branch changed in the meantime.

jbemmel:
I think in this case it's the test that is wrong:
[image: image.png]

I added extra global attributes to the sr module:
[image: image.png]

The test assumes sr has no attributes defined. I'll see if I can fix that

Jeroen

ipspace:
There are a number of problems with this PR:

* You did not start with the then-current copy of the **dev** branch (I had to fix a few things after the last PR), so I'm seeing all sorts of changes that have already been included in the **dev** branch. Please rebase.
* You added global SR values that might be appropriate for your device(s) but not for other platforms. We can't have global defaults in global settings that are tailored for a specific box. Right now I'd put those defaults into the SR Linux Jinja2 config template. That is obviously not the ideal solution; I have to work on a way to have device-specific module defaults.
* Similar with **clab.j2** -- you added **type** attribute and set it to a default value that is appropriate for SR Linux, but not for (for example) cEOS or FRR. You made sure **type** is set for SR/OS and SR Linux with **provider_type** setting, so you should use `{% if 'type' in n %}` in clab.j2.
* I would love to see a more structured approach to attributes (obviously both approaches work, call me religious ;). For example "sr.label_range: 50000-60000" or "sr.label_range: [ 50000,60000 ]" or

```
sr:
  label_range:
     start: 50000
     size: 1000
```

In any case, if we want to go and chase those nerd knobs, we have to put together a list of everything that should be configurable, and the figure out how best to structure it.

Finally, it looks like it's high time I write a "testing" document ;) There's a script that generates expected results, so whenever a test fails you run that one and do a diff to figure out what has changed.

jbemmel:
I thought the SR Global Block (SRGB) range would be applicable to all
platforms, every vendor seems to have a different default range.

From RFC8402:

SR Global Block (SRGB): the set of global segments in the SR domain.
   If a node participates in multiple SR domains, there is one SRGB for
   each SR domain.  In SR-MPLS, SRGB is a local property of a node and
   identifies the set of local labels reserved for global segments.  In
   SR-MPLS, using identical SRGBs on all nodes within the SR domain is
   strongly recommended.  Doing so eases operations and troubleshooting
   as the same label represents the same global segment at each node.
   In SRv6, the SRGB is the set of global SRv6 SIDs in the SR domain.


I could have left it at default for SROS and SRLinux too, but as nerd
knobs go I think this one is justified.

However, unlike 'AS' for example we wouldn't want nodes to be able to
override this global setting (as it would defeat the purpose)


[image: image.png]


I forgot to include the attributes in 'no_propagate', could/should? be
implied by the fact that the module does not have any 'node' or 'link'
level attributes


Regards,

Jeroen

jbemmel:
Created a new, clean PR

PR #88, opened 2021-12-26
Cleanup stub link logic

jbemmel:
Add 'node_count' to links to side step discussions about what constitutes a 'stub' link...

### DIFF
--- a/docs/release/1.0.md
+++ b/docs/release/1.0.md
@@ -2,7 +2,7 @@
 
 ## New Devices
 
-* [Jeroen van Bemmel](https://github.com/jbemmel) contributed support for Nokia SR Linux and SR OS (including initial device configuration, OSPF, BGP, and SR-MPLS).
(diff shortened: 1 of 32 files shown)


ipspace:
This makes perfect sense. However:

* **node_count** is not propagated into node interface data, which is OK, we should check **l.neighbors** in device configuration templates. I have no idea why `netsim/ansible/templates/initial/srlinux.j2` works -- it should crash with undefined value.
* **node_count** could be greater than 2 for multi-access links (`clab.j2` change). Have to test whether containerlab supports that, if not we have to check node count and generate errors in `clab.py`

jbemmel:
Updated Containerlab to support LAN links, requires post-stop on provider to cleanup bridges

Tests are failing, it's a simple matter of missing 'node_count' attributes everywhere - not sure how to regenerate the data. Manual updating is a pain...

ipspace:
I'll take it from here. Thanks a million!

jbemmel:
Need to add a check to only create/delete bridges when there are more than 2 nodes on a link

PR #95, opened 2021-12-29
BFD module and L2-only port support

jbemmel:
* Adds BFD module with support by srlinux and sros
* Allows 'ipv4: False' to model 'split' links with an L2 service across a network

Prefix pools need more work to be generalized to support this use case; the change included here only enables a pool consisting of a single prefix (which should be possible)

Patches all templates to validate ipv4/ipv6 addresses using their respective filters, allowing 'ipv4: False' to be skipped

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 recursive-include netsim/ansible *
+recursive-include netsim/extra *
 recursive-include netsim/templates *
 recursive-include netsim/install *
 recursive-include netsim *.yml
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Interested? [Read the documentation](https://netsim-tools.readthedocs.io/) and [
 
 ## Releases
 
-The latest release is [release_1.0.5](https://github.com/ipspace/netsim-tools/releases/tag/release_1.0.5).
+The latest release is [release_1.0.6](https://github.com/ipspace/netsim-tools/releases/tag/release_1.0.6).
 
 The last release containing the old (obsolete) CLI commands is [release_0.9.3](https://github.com/ipspace/netsim-tools/releases/tag/release_0.9.3).
 
--- a/docs/netlab/initial.md
+++ b/docs/netlab/initial.md
@@ -4,11 +4,13 @@
 
 The Ansible playbook invoked by **netlab initial** command deploys device configurations in two steps:
 
-* Initial device configurations[^1]
-* Module-specific device configurations[^2]
+* Initial device configurations[^itag]
+* Module-specific device configurations[^mtag]
+* Custom configuration templates[^ctag]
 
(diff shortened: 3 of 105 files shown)


ipspace:
Somehow you managed to crash the IPv6 address allocation. The IPv6 test case fails -- it should result in ipv6 addresses assigned to interfaces, and all we get is 'ipv6: false'.

Also, the BFD parameter names are unnecessarily long, but I can fix that afterwards.

jbemmel:
> Somehow you managed to crash the IPv6 address allocation. The IPv6 test case fails -- it should result in ipv6 addresses assigned to interfaces, and all we get is 'ipv6: false'.

Must be due to the global search & replace - I'll look into it

ipspace:
If I understood correctly, there are three independent changes in this PR:

1. clab.j2 is broken because I forgot to fix it when migrating to nodes-as-a-dictionary (plus you correctly removed **nodes_map** from the cleanup code)
2. bfd support
3. support for nodes with no addresses.

Please submit separate PRs for (1) and (2) and I'll merge them. We'll solve (3) with #97, for the moment you can implement split links with **prefix: False** on the link and **ipv4: prefix** on node-on-link. Here's an example:

```
nodes:
  r1:
  h1:
  h2:

links:
- prefix: False
  r1:
  h1:
    ipv4: 10.42.42.1/24
- prefix: False
  r1:
  h2:
    ipv4: 10.42.42.2/24
```

ipspace:
Finally:

* Some IS-IS changes started appearing out-of-the-blue. They might be a good candidate for another PR, or you could implement a plugin for the moment.
* Please use reasonable easy-to-type attribute names. How many people will get `DesiredMinTxInterval_us` right on the first try? ;) Also, why would `loop_free_alternate: True` be better than `lfa: True`? It's like having `bgp.autonomous_system` instead of `bgp.as`. People who know what loop-free alternate is also know what LFA stands for ;)
* No job is finished until the paperwork is done. If you decide to add a module (and BFD was a really good candidate), please write at least some minimum documentation.

jbemmel:
> nodes:
>   r1:
>   h1:
>   h2:
> 
> links:
> - prefix: False
>   r1:
>   h1:
>     ipv4: 10.42.42.1/24
> - prefix: False
>   r1:
>   h2:
>     ipv4: 10.42.42.2/24

Could we do something like this?
```
subnets:
- name: net1
  prefix: 10.42.42.0/24

links:
- subnet: net1
  r1:
  h1:
   ipv4: 1
- subnet: net1
  r4:
  h2:
   ipv4: 2
```

ipspace:
We're thinking along the same lines. Answered in #97.

jbemmel:
See https://github.com/ipspace/netsim-tools/pull/100 for BFD-only PR

PR #100, opened 2021-12-30
Add BFD module (for IS-IS on SRLinux and SROS)

jbemmel:
Including some minimal documentation, and a (simple, global) config test


ipspace:
* Simplified BFD configuration parameter names
* Changed units to msec -- checked multiple platforms, and it seems that nobody apart from SR Linux uses microseconds (and even milliseconds are probably pushing it on virtual devices)
* Slightly polished the documentation
* Amended the commit and force-pushed the changed commit back to the **dev** branch.

PR #120, opened 2022-01-05
Fix links to same node

jbemmel:


ipspace:
Thanks for the fix. Looks like Box changed its behavior and passes references on assignments (at least sometimes ;) instead of creating copies. It was about time, the old behavior was driving me mad (see also _wrong_ versus _different_ discussion).

Will change the requirements to make sure we get Box version 5.4.1 or higher in **pip install**.

Now I have to fix LAN code along the same lines to survive `interfaces: [ r1, r1, r1 ]`

PR #121, opened 2022-01-05
Cleanup topology defaults for sros and srlinux

jbemmel:
- Change default ansible connection sros and srlinux, grpc only used for very specific tasks; Ansible sometimes uses the underlying connection at times when it's least expected


ipspace:
Nice to see you used squash merge -- I like to keep the history as informative as possible.

It would also be nice if you could summarize the commit message the next time; someone looking at it in the future doesn't need the list of all minor commits you made in the feature branch.

PR #125, opened 2022-01-05
Fix isis templates

jbemmel:


ipspace:
Wow, you're fast ;)

A few comments:

* We need another check for **external** role that would delete **isis** attribute from the interface
* After that, the initial interface loop should be `for l in interfaces if 'isis' in l`... but then we'd need to make sure at least one **isis** attribute is always present on the interface (or just set **isis** to an empty dictionary)
* I'd like to have **network_type** attribute with value **point-to-point** instead of **network_type_p2p**. Doesn't matter (much) for IS-IS, but OSPF has three network types and I'd like to make that configurable.
* Unnumbered OSPF interface with no neighbors makes no sense (thus the "!= 1" test)

ipspace:
One more thing: **unnumbered** attribute should not be checked when figuring out is-is network type, as it could be set on multi-access link.

jbemmel:
> One more thing: **unnumbered** attribute should not be checked when figuring out is-is network type, as it could be set on multi-access link.

The logic was in the FRR template, I tried not to change it

jbemmel:
Implemented all requested changes, 'passive' flag is always defined for interfaces that have IS-IS enabled

You may want to double-check the nxos template, the if/endif nesting seemed wrong

ipspace:
@jbemmel Thank you. Great job!

I cleaned up the **network_type** logic a bit to allow the attribute to be set (or reset) from the topology file. There were also a few minor gotchas in device configuration templates.

Amended the commit and force-pushed it back into **dev** branch. You might want to rebase your local repository.

PR #126, opened 2022-01-06
Update OSPF templates; also fixes 'module' references to be defined

jbemmel:
Add a test case for external links in OSPF


ipspace:
Thanks for the fixes and the extension of the OSPF test case -- I did the same addition to the integration test, but not to the regular test case.

`module|default([])` is nice, but not necessary - BGP configuration template would not be executed without 'bgp' in `module`.

PR #133, opened 2022-01-07
Remove 'isis' if interface has no IP

jbemmel:


ipspace:
Thanks a million, nice catch.

I think the tests can be simplified. To start with, you could use `... l.get('ipv4',False)` considering that nonempty string (IP address) evaluates as True in logical conditions.

Also, the test for **unnumbered** is not be needed. If an interface has **unnumbered** attribute, it also has **ipv4** and/or **ipv6** attribute, depending on what's configured on the loopback interface. Run `addressing-lan.yml` and `addressing-p2p.yml` test cases. I usually run them with `netlab create -o yaml:nodes` to get interface data in easy-to-read format.

And yeah, I made the same error ;)

PR #145, opened 2022-01-10
clab add OVS bridge support

jbemmel:
Also modifies the Linux bridges to forward LLDP, LACP, 802.1X and other multicast protocols

Use case for OVS: OSPF over multi-access, for some reason the Linux bridge does not forward Hello packets

Default to Linux standard bridges, OVS requires separate package


ipspace:
Interesting, in my case IOSv didn't want to send hello packets onto unnumbered multi-access network at all.

PR #154, opened 2022-01-17
Bug fix: fortinet ospf area id now correctly set

petercrocker:
Bug fix for Fortinet OSPF:
- Removed caveat from caveats.md
- Corrected fortinet ansible templates to use `interfaces` instead of `links`
- Uncommented fortinet ansible task _Configure OSPF network statement_ now that it's got correct syntax and works

Tested this on the example topology called fortinet.yml


ipspace:
Ah, now I understand what broke your Ansible tasks. Apologize for that -- I fixed the Jinja2 templates, but forgot about those task lists. Merging...

PR #155, opened 2022-01-17
Platforms: Cumulus ip unnumbered support

petercrocker:
Platforms: Add support for IP unnumbered interfaces on Cumulus

- Mod `initial/cumulus.j2` to set unnumbered interfaces to have loopback IP address
  - Reference: https://support.cumulusnetworks.com/hc/en-us/articles/202796476-RETIRED-OSPF-Unnumbered-Sample-Configurations
- Mod `topology-defaults.yml` to allow ip unnumbered and ospf unnumbered
- Add cumulus box to `tests` @ipspace please verify these two changes, as I can't test since I don't have all the images mentioned

### DIFF
--- a/netsim/ansible/templates/initial/cumulus.j2
+++ b/netsim/ansible/templates/initial/cumulus.j2
@@ -70,10 +70,18 @@ interface {{ l.ifname }}
  description Stub interface
 {% endif %}
 {% if l.ipv4 is defined %}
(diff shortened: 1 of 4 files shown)


ipspace:
Thanks for the fix. Will merge the PR, fix the tests, run integration tests, amend the commit and force-push it back.

Issue #148, opened 2022-01-13
BGP and OSPFv2/v3 router-id generation


jbemmel:
Currently, sros and srlinux use the ipv4 loopback ip as router-id.
This means that loopback.ipv4 must always be defined.
Should this determination be moved to core logic (bgp module)?
Same for the rr cluster-id (currently hardcoded to loopback.ipv4 in several places)

Furthermore, the bgp template generates both ipv4 and ipv6 peerings to the same neighbor; should there be a preference (in the BGP logic) and do we need a way to disable peering on ipv4 or ipv6?

ipspace:
Interesting. Never thought about that. What happens when you try to run OSPF and BGP in an IPv6-only network?

We could add "router_id" (or should it be just "id"?) calculation to BGP and OSPF modules, which would also make it configurable via topology file. I wonder what happens if you use "0.0.0.id" as router ID (in case it's not configured and the loopback interface is not set). Implementations that understand it's just a 32-bit number won't have a problem (plus we'd avoid any potential overlap with lab addressing).

As for parallel ipv4 and ipv6 peerings, that's the sanest way of configuring multi-vendor dual-stack BGP. From the topology transformation perspective I have to check what happens in v6-only network. From the BGP design perspective, if someone wants to have something different (like Cumulus running IPv4 AF over IPv6 LLA session), they could always write a plugin ;)

jbemmel:
Although one could configure a different router_id for BGP and OSPF, I think it makes the most sense if it is the same.
I could add the 'loopback.ipv4|default(...)' logic to the templates, but I would prefer to have the 'initial' module do the honors
(or perhaps the 'augment' logic where loopbacks get calculated)

In my ipv6-only topology only the ipv6 BGP session comes up. If you say it's the sanest way to try both by default that's fine, but I'd still like a flag to say "don't configure ipv4 peers for bgp"

Perhaps
```
bgp:
 peering:
  ipv4: False
```

ipspace:
Ah, so you configured IPv4 loopbacks in an IPv6-only network (to get router IDs) and all of a sudden you have a bunch of IBGP IPv4 sessions. Yeah, that sucks ;)

I like the idea of controlling which sessions one wants to have (and we could extend it to EVPN or L3VPN sometime), but I think we need to have IBGP and EBGP parts, for example:

```
bgp:
 peering:
  ibgp:
   ipv4: False
```

jbemmel:
[...]

PR #159, opened 2022-01-18
BGP flags to control peering and address families

jbemmel:
Tests will fail, needs a bulk-add of these new flags to 'expected output'

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.j2
+++ b/netsim/ansible/templates/bgp/srlinux.j2
@@ -63,7 +63,7 @@ updates:
 #}
    neighbor:
 {% for n in bgp.neighbors %}
-{%  for af in ['ipv4','ipv6'] if n[af] is defined %}
+{%  for af in ['ipv4','ipv6'] if n[af] is defined and bgp.peering[n.type][af] %}
 {{ neighbor_global(n,af) }}
 {%  endfor %}
 {% endfor %}
--- a/netsim/ansible/templates/bgp/srlinux.j2
+++ b/netsim/ansible/templates/bgp/srlinux.j2
@@ -25,6 +25,7 @@ updates:
 #}
    group:
 {% for g in ['ibgp-ipv4','ibgp-ipv6','ebgp'] %}
+{% set bgp_type = g[:4] %}
 {% set ibgp_rr = (g!='ebgp' and bgp.rr|default(0)) %}
    - group-name: {{g}}
      admin-state: enable
@@ -35,6 +36,11 @@ updates:
      timers:
       connect-retry: 1
       minimum-advertisement-interval: 1
+
+{%   for af in ['ipv4','ipv6']  %}
+     {{ af }}-unicast:
+      admin-state: {{ 'enable' if bgp[af] is defined and af in bgp.address_families[bgp_type] else 'disable' }}
+{% endfor %}
 
 {%   if bgp.community[ g[:4] ]|default([]) %}
 {%   set list = bgp.community[ g[:4] ] %}
--- a/netsim/ansible/templates/bgp/sros.gnmi.j2
+++ b/netsim/ansible/templates/bgp/sros.gnmi.j2
@@ -22,8 +22,8 @@ updates:
     - group-name: "{{c}}"
       admin-state: enable
       family:
-       ipv4: {{ 'ipv4' in bgp }}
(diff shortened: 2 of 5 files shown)


ipspace:
**mypy** failed, which means you have errors in your code, and just looking at **add_if_peering** I wondered how it could work as you use **n** in function arguments and **node** later on.

Anyway, I don't understand why you want to have two parameters (**address_families** and **peering**). What's the difference? When would you use one and not the other?

From where I look, we only need one parameter: which address families do we want to have configured between IBGP or EBGP neighbors. What am I missing?

jbemmel:
In my SRv6 example, I use IPv6 for the iBGP peering session with the RR, but only exchange ipv4 routes (with ipv6 next hops). So
```
bgp:
 peering:
  ibgp: { 'ipv6': True }
 address_families: 
  ibgp: [ ipv4 ]
```
One parameter relates to the address to pick for connecting to peers (ipv4 or ipv6, could include other corner cases like ebgp on loopbacks but not really needed right now). The address_families are about what gets exchanged between the peers ( ipv4, ipv6, evpn, label-ipv4, etc. )

For most labs, I would expect the defaults (ipv4+ipv6 for both) would be fine

ipspace:
When we started discussing this idea in #148 I wrote

> As for parallel ipv4 and ipv6 peerings, that's the sanest way of configuring multi-vendor dual-stack BGP. From the topology transformation perspective I have to check what happens in v6-only network. From the BGP design perspective, if someone wants to have something different (like Cumulus running IPv4 AF over IPv6 LLA session), they could always write a plugin ;)

... and what you're proposing is exactly the "they could always write a plugin" stuff. You solved your particular scenario, but once we open that can of worms we have to do a lot of stuff to get it right:

* Which sessions do you want to set up (any combination of IPv4 and IPv6)
* Which AF do you want to transport over which session?
* What do you want to do for IBGP versus EBGP?

Plus, in most implementations I had the dubious "privilege" of working with, running IPv4 AF over IPv6 or vice versa required tons of hacking with route maps to get the next hops right. Your scenario probably works only because you're running it over SRv6.

I suggest you implement this as a plugin to support your specific scenario, and if we ever get to the point where we really need a tight control of address families to support some core functionality, we could revisit the idea.

However, please do feel free to add the plugin under `netsim/extra` directory. It would be great to start collecting contributed plugins.

jbemmel:
The problem is that - unlike the BGP anycast example - the flags to be set here directly impact the bgp Jinja2 template logic. The way BGP peering and families are currently modeled (that is, implicitly), we need more than a plug-in that would make a few adjustments to the topology. How would a plug-in modify the topology to implement ipv6 peering with an ipv4 address family?

The code in bgp.py that removes bgp neighbors when they don't have an ip address version for which peering is enabled, is optional; it would be fine to leave the neighbors in place, and filter them out in the templates. I understand if you prefer not to complicate the core logic to support an edge case, but I do believe we should model desired BGP peering ip versions and address families explicitly

ipspace:
OK, here's how I would tackle this issue in a plugin:

* Let BGP module do its magic
* In **post_transform** module hook:
  * Remove all IPv4 sessions from **node.bgp.neighbors** (we don't want them)
  * Add **address_family: [ ipv4 ]** to IPv6 neighbors (extra magic to get multi-AF session)
* Let BGP configuration template to its magic -- you'll get a mesh of BGP sessions between IPv6 endpoints with IPv6 AF enabled on them
* In your configuration template add activation of IPv4 address family on sessions between IPv6 endpoints

Makes sense?

jbemmel:
As I understand it, in this proposal the user would add
```
plugins: [ bgp_ipv4_af ] # (say)
```
and then the 'regular' bgp j2 templates would be adapted to use bgp.address_family if defined?

For this use case, I would personally prefer
```
bgp.address_family: [ ipv4 ]

nodes:
 some_special_node:
  bgp.address_family: [ ipv4, ipv6 ]
```

The problem with plugins, is that I see no obvious way to parameterize them; you enable them globally, or not. In my opinion, BGP address families are critical and warrant proper ("first class citizen") modeling, as opposed to an optional one-off plugin or afterthought

ipspace:
Give me a day or two to write a PoC ;)

jbemmel:
Subsumed by https://github.com/ipspace/netsim-tools/pull/176

PR #166, opened 2022-01-20
Add plugin for removing bgp ipv4 peering

jbemmel:
'Hardcoded' to remove all ipv4 neighbors

Could be generalized if we had a mechanism for parameterizing plugins, e.g.
```
plugins: 
 filter-bgp-peering:
  address_family: [ipv6]
```

Note that the need for this plug-in went away with the router_id update, but it may still be a useful example


jbemmel:
@ipspace I'll leave it up to you, merge it or abandon it I don't mind either way

jbemmel:
Let's leave this one be for now

PR #167, opened 2022-01-20
Use router_id for sros and srlinux

jbemmel:
Implements https://github.com/ipspace/netsim-tools/issues/148

I noticed the other templates check for a valid ipv4 address, assuming that is unnecessary

Note that IS-IS also needs a router-id (per protocol instance)


ipspace:
You're absolutely right, checking for valid IPv4 address is unnecessary (belt-and-braces approach to configuration templating), more so as I implemented the check within Python code in the meantime.

PR #172, opened 2022-01-21
Platforms: Cumulus Linux v5.0 using NVUE YAML Config

petercrocker:
Platform support for Cumulus Linux >= 5.0.0

Device name: **cumulus_nvue**

Notes:
* Feature parity with `cumulus` device platform (Cumulus Linux <=4.4.0)
* Full Initial, OSPF, BGP template support
* Docs updated
* Tested against `tests/module/bgp-session-test.yml` with all devices set to `cumulus_nvue`


petercrocker:
Ok @ipspace this one is ready to go!

PR #173, opened 2022-01-21
setting a static loopback address always gets overridden by an IP from the loopback pool

petercrocker:
**Issue:** setting a static `nodes.<device>.loopback.[ipv4||ipv6]` address always gets overridden by an IP from the loopback pool

Example static loopback:
```yaml
nodes:
  ext1:
    loopback:
      ipv4: 172.16.1.1/32
```

**Result:** 

in `host_vars/ext1/topology.yml`:

```yaml
loopback:
  ipv4: 10.0.0.1/32
```

**Fix:** Add a check to only assign loopback from pool addresses if it's not already statically defined.

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -16,19 +16,21 @@ from .. import addressing
 # Starting with release 1.1, the nodes data structure is a dictionary. Convert
 # lists of dictionaries or lists of strings into a unified dictionary structure
 #
+
+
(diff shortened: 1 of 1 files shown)


petercrocker:
@ipspace I didn't put the junos additions on a different branch, so it got sucked in to this PR. So this PR is for two items:

1) Fix the issue listed above
2) Add an SR template for junos

PR #174, opened 2022-01-21
Model stub interfaces as regular interfaces, not loopbacks; set passive flag

jbemmel:
Fixes https://github.com/ipspace/netsim-tools/issues/171


ipspace:
Thank you!

Issue #142, opened 2022-01-09
Module code and configuration priority


ipspace:
Current code already reorders modules based on module dependencies (#83), but that's not good enough when we have modules that are technically independent but have to be executed in order. For example:

* BGP python code should be run before IS-IS or OSPF python code to calculate link roles (and cause IGP to be disabled **external** links)
* BFD code should probably be run before IS-IS or OSPF code

We don't have that problem right now as BGP code marks link roles in **link_pre_transform**, and IGP modules use those roles in **node_post_transform**, but it's still there...

On the other hand, we might need specific ordering of IGP- versus BGP modules in configuration process. Can't see where that could happen, but who knows... (any ideas @ssasso @jbemmel @petercrocker?)

Implementation details:

* We need **priority** attribute for modules (similar to **requires**) with two attributes: **code** and **configuration**.
* **code** priority should be used when ordering the list of global modules to be executed at various module hook points
* **configuration** priority should be used when ordering the list of modules within node data (to ensure the configuration templates are executed in order).
* Higher priority is better (because we'll use zero as the default value)

Both parameters should modify the **modules.sort_module_list** behavior.

PR #176, opened 2022-01-21
Add srv6 module (with initial support on sros)

jbemmel:
I've tried to minimize the deltas required to implement this, this PR proposes to add an 'address_families' knob for bgp to provide more control over the enabled protocols (to be extended to include evpn in the near future)

### DIFF
--- a/docs/module/srv6.md
+++ b/docs/module/srv6.md
@@ -1,0 +1,57 @@
+# Segment Routing over IPv6 (srv6) Configuration Module
+
+This configuration module configures SRv6 over IS-IS and (optionally) BGP on Nokia SR OS.
+
+Supported SRv6 features:
+
+* End SIDs (nodes) and End-x SIDs (links), all statically allocated
+* IS-IS routing protocol with SRv6 TLVs
+* Dynamic BGP Service SIDs
+
+The module currently depends on IS-IS module and will trigger a configuration error if the **isis** module is not enabled in the network topology.
+
+## Parameters
+
+* Locator: an IPv6 address range to allocate to SRv6 nodes; each node is assigned a unique /64 prefix from this range
+
+## Example
+
+We want to create a four-router SRv6 network.
+
+All devices run IS-IS and SRv6:
+```
+module: [ isis, srv6 ]
+```
+
+We want to use unnumbered P2P interfaces and dual-stack deployment with IPv6 configured on loopback addresses:
+
+```
+addressing:
+  p2p:
+    unnumbered: true
+  loopback:
+    ipv6: 2001:db8:cafe::/48
+
+srv6.locator: 2001:db8:aaaa::/48  # Must not overlap with interfaces
+```
+
+```
+provider: clab
+defaults.device: sros # Type must support FP4 chipset
+
+nodes:
+  e1:
+  e2:
+  c1:
+  c2:
+```
+
+The devices are connected into a square topology:
+
+```
+links:
+- e1-c1
+- e1-c2
+- e2-c1
+- e2-c2
+```
--- a/netsim/ansible/tasks/deploy-config/sros.yml
+++ b/netsim/ansible/tasks/deploy-config/sros.yml
@@ -55,7 +55,7 @@
 - block:
   - block:
     - name: Enable Open Config YAML modules (retry to give Containerlab a chance to finish configuration)
-      when: enable_open_config is not defined and 'ixr-ec' != type
+      when: enable_open_config is not defined and 'ixr-ec' != clab.type
       tags: [ initial ]
       nokia.grpc.gnmi_config:
         prefix: configure
--- a/netsim/ansible/templates/bgp/sros.gnmi.j2
+++ b/netsim/ansible/templates/bgp/sros.gnmi.j2
@@ -13,7 +13,7 @@ updates:
      ebgp: {{ 16 if 'ixr' in clab.type else 64 }}
     min-route-advertisement: 1 # Be aggressive about sending updates
     connect-retry: 5           # Retry frequently, suitable in DC context
-    client-reflect: {{ 'True' if bgp.rr|default('') else 'False' }}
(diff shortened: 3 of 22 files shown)


ipspace:
So you sent me on a wild chase through SRv6 drafts/RFCs 😉

To be precise, we have three dimensions of nerd knobs, and I **REALLY** don't want to implement them in the core functionality. The core should be an opinionated but extensible/modifiable. So we have:

* Transport sessions -- so far IPv4 and IPv6, although I'm sure there's someone out there running BGP over CLNS.
* Address families enabled on individual transport sessions (IPv4, L2VPN, L3VPN, EVPN... on IPv4, IPv4 or IPv6 on IPv6)
* Next hops for enabled address families.

In a non-SRv6 setup, enabling IPv4 AF on IPv6 transport session would still result in IPv4 next hops in BGP updates on IPv4 AF, but you'd have to set next hops with a route map (or vice versa).

In SRv6, due to draft-ietf-bess-srv6-services-05, IPv4 prefixes are announced with IPv6 next hops according to rfc8950 to implement _Global IPv4 over SRv6 Core_.

I would say it's an SRv6 services configuration issue, not a BGP configuration issue. If SRv6 module is configured to offer IPv4-as-a-service over SRv6, then it's the job of SRv6 module to remove the extraneous IPv4 neighbors from the **nodes** data structures, and the job of SRv6 configuration templates to configure BGP accordingly.

We could make SRv6 dependent on BGP, or use #142 to make sure SRv6 hooks are executed after BGP hooks, and SRv6 is configured after BGP.

jbemmel:
The PR only adds bgp.address_families as an explicit parameter, it does not configure it. It changes the logic from "if peering over ipv4, enable ipv4 AF" to "if ipv4 AF is enabled, ..." - nothing more complicated than that

At https://github.com/jbemmel/netsim-examples/tree/sr-v6/routing/sr-v6 you can see a sample topology, it includes the required options to enable IPv4 as the only address family for ibgp (though it could be left at default too)

The srv6 module handles bgp nexthop configuration knobs to enable ipv4-over-srv6. There is no need for the srv6 module to modify bgp peerings, if the user wants ipv4 peers too (for whatever reason - say bgp ipv4 towards the Linux hosts) then those could happily co-exist.

In terms of 'nerd knobs', this module configures the ones required for its specific use case; it will only affect users who enable the module. The complexity is opt-in, people who want to keep things simple can still get that - by doing absolutely nothing

ipspace:
We have tons of conflicts based on other work we all did while this one was being discussed. Please rebase and resolve the conflicts first. Have to see where you're using **bgp.address_families** parameter, and it's a bit hard to figure that out now ;)

jbemmel:
It's here: https://github.com/ipspace/netsim-tools/pull/176/files#diff-f0ab3c85f96152c6480c96448755d668d7ebc1cb26552f18bde1da39e3371b47R27

I have resolved the conflicts, but I noticed I pulled in some NVUE related changes in default-topology - not sure where those came from

ipspace:
> In terms of 'nerd knobs', this module configures the ones required for its specific use case; it will only affect users who enable the module. The complexity is opt-in, people who want to keep things simple can still get that - by doing absolutely nothing

Not true. We added default BGP settings that will be propagated into every single node and won't be used by any of the device configuration templates... and then someone will notice those settings, tweak them, and be totally baffled when they don't work. If anyone complains I'll simply send them your way ;)

Also, based on my reading of your templates, you're creating ibgp-v4 and ibgp-v6 groups and **enable all configured address families on all groups**. I wonder what the end result of that would be in a simple BGP+OSPF (or IS-IS) network, or a simple dual-stack BGP+IS-IS network... but who am I to have an opinion on that.

Merging the PR... may the Force be with us ;))

jbemmel:
My assumption was that the other device templates would be updated to take their hint from bgp.address_families, rather than the existence of an ipv4/ipv6 peering address. I agree there may need to be some tweaking - for example by calculating effective active address families per peer, if there is both an ipv4 and an ipv6 session then perhaps ipv4 should be enabled only for ipv4 etc.

Let's try it like this and if we decide we don't like it after all, we can change it

ipspace:
> My assumption was that the other device templates would be updated to take their hint from bgp.address_families, rather than the existence of an ipv4/ipv6 peering address.

And I've been saying since "forever" (not that anyone would listen) that in typical BGP deployments there's no need for extra complexity. Either you're running a single-protocol network and have one BGP transport session and AF, or you're running a dual-stack network and have two BGP transport sessions and two AF (one per transport session). Will document those assumptions as part of BGP module description ;)

In most cases, there's no need for the extra complexity, even in EVPN or MPLS/VPN networks. You have IPv4 transport session and you enable extra address families on top of it. Problem solved. Not many vendors support running EVPN or MPLS/VPN on top of IPv6 because VXLAN runs on IPv4 and very few vendors implemented LDPv6 anyway.

The only reason we're talking about this stuff that I can see at the moment is because you wanted to build an IPv4-only network on top of SRv6.

In short, everything beyond the basic BGP setup should be handled outside of the BGP module **because it's not needed to implement a sane BGP network but to support something heaped on top of BGP used as a kitchen sink**, and there's nothing stopping another module or plugin from adjusting BGP neighbors data structure in any way they see fit.

> Let's try it like this and if we decide we don't like it after all, we can change it

Here we go. As you can see, I don't like it. I will leave the **address_families** attribute in the list of BGP attributes (no harm there), but will move the default values onto SROS device definition (because SROS is the only one implementing them at the moment). I will make sure the SROS node data stays intact (so your examples will work), but node data for all other devices won't be polluted with settings that are not implemented.

In any case, should we ever decide we need a more structured approach, we have to solve the question of three dimensions of BGP transport/AF/NH conglomerate. Will open an backburner issue for that.

jbemmel:
Resolves https://github.com/ipspace/netsim-tools/issues/146

PR #179, opened 2022-01-21
Update Nokia grpc plug-in install instructions to use git repo directly

jbemmel:
See https://github.com/ipspace/netsim-tools/issues/168, Galaxy package hasn't been updated in 2 years


ipspace:
Awesome, thank you!

Issue #181, opened 2022-01-23
Data model needed to support RFC 8950 (extended BGP next hops)


ipspace:
With RFC 8950 (and 5549, but nobody ever implemented that one apart from Cumulus for unnumbered EBGP) it became possible to freely mix BGP transport sessions, address families, and next hop AFIs. We might need a data model (in the future once EVPN, L3VPN and SRv6 are implemented on a wide range of devices) to support that Jenga tower of Cluedo-like combinatorial complexity.

We need to be able to specify:

* What happens on IBGP versus EBGP sessions
* Which transport sessions we'd like to have (assuming they are possible according to addressing setup)
* Which address families we'd like to run over which transport session
* What next hop we'd like to use for individual address families.

The only generic way to address that at the moment would be a giant data structure _type_._transport_._af_._next hop_. For example, the default setup would be:

```
bgp.af_map:
  ibgp:
    ipv4: [ ipv4 ]
    ipv6: [ ipv6 ]
  ebgp:
    ipv4: [ ipv4 ]
    ipv6: [ ipv6 ]
    unnumbered: [ ipv4, ipv6 ]  # to support implicit RFC 5549 used by Cumulus on EBGP unnumbered
```

Standard BGP-LU would be:

```
bgp.af_map:
  ibgp:
    ipv4:
      ipv4:
        nexthop: mpls
    ipv6:
      ipv6:
        nexthop: mpls
  ebgp:
    ipv4: [ ipv4 ]
    ipv6: [ ipv6 ]
```

What we'd need to support 6PE would be (where **mpls** stands for _Labeled Unicast_)

````
bgp.af_map:
  ibgp:
    ipv4:
      ipv4: True
      ipv6:
        nexthop: mpls
  ebgp:
    ipv4: [ ipv4 ]
    ipv6: [ ipv6 ]
    unnumbered: [ ipv4, ipv6 ]  # to support implicit RFC 5549 used by Cumulus on EBGP unnumbered
````

What @jbemmel would like to see for his SRv6 example would be:

```
bgp.af_map:
  ibgp:
    ipv6:
      ipv4:
        next_hop: ipv6
  ebgp:
    ipv4: [ ipv4 ]
```

Hopefully noone in his right mind would tweak that map by hand, it should be done by other modules/plugins. Also, you might have noticed "please keep me sane" shortcuts (list instead of nested dictionaries) that would be expanded at the evaluation time.

[...]

PR #182, opened 2022-01-24
ibgp-over-ebgp plugin

jbemmel:
A separate plugin to augment the bgp.neighbor lists with additional ebgp peers, based on an optional 'underlay_as' attribute per node

The module assumes bgp.as is the ibgp AS, and the regular bgp module is used to configure the ibgp peerings.

In the resulting bgp.neighbors list, a 'local_as' attribute is added for eBGP peers. The bgp templates will have to be updated to apply this (optional) attribute where needed.

Fixes https://github.com/ipspace/netsim-tools/issues/163

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -3,6 +3,7 @@
 ## Cisco IOS
 
 * Cisco IOSv release 15.x does not support unnumbered interfaces. Use Cisco CSR 1000v.
+* Cisco CSR 1000v does not support interface MTU lower than 1500 bytes or IP MTU higher than 1500 bytes.
 
 ## Cumulus Linux in ContainerLab
 
--- a/docs/links.md
+++ b/docs/links.md
@@ -277,9 +277,17 @@ links:
   mtu: 1500
 ```
 
-### Device MTU
+When the node **mtu** parameter is not specified, its default value is fetched from **defaults.interfaces.mtu** or **defaults.devices** setting.
 
-**mtu** parameter specified within device defaults is copied into node data and applied to all interfaces without a specified MTU. To change the device default use **defaults.devices** setting, for example:
+For example, to build a lab using 8K jumbo frames, use:
+
+```
+defaults.interfaces.mtu: 8192
+```
+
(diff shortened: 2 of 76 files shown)


ipspace:
In a word: **NO**.

You managed to reinvent Enterprise MIB. There's a reason you won't find "if device == Nokia then" in OpenConfig YANG data models.

Creating a separate copy of the BGP logic is bad (what if we change it to support things like #181), hard-coding plugin name into the BGP module is as appropriate as hard-coding the IP address of the database server into your web application, or hard-coding the device or vendor name in a standard YANG model.

I would suggest you start with another BGP node attribute (for example, **underlay_as**) which can be set in nodes or groups (don't try to overload the meaning of bgp.as or bgp.as_list). If you'd like to have **ibgp_over_ebgp** attribute with two autonomous systems, then process that in the **init** code to set global or node **bgp.as** and **bgp.underlay_as** (or similar).

In the **post_transform** code iterate over all links, identify situations where neighbor **underlay_as** are different and add EBGP sessions with **local_as** to node **bgp.neighbor** list, effectively _augmenting_ what BGP module does not _replacing_ it.

As for the actual configuration, you could either modify exiting BGP templates, or have configuration templates in the plugin directory and add them to the node **config** attribute (see https://github.com/ipspace/netsim-examples/tree/master/routing/anycast-mpls/bgp-anycast). _Local AS_ might be a useful addition to the BGP module, so in this particular case it's probably best to have it in the BGP template.

jbemmel:
I added the 'print' in the bgp module just to highlight the point where the as_list model breaks down, as it only keeps the last seen as per node. The reference to the plug-in name can be removed, I was just thinking you may want to put some sort of warning there (or even an error - as the config is ambiguous)

I ended up where I did because I started modeling bgp.as_list in a wrong way (with duplicate members) and then tried to retrofit that model. I can see you prefer to not change bgp.as_list, I'll see what I can do

ipspace:
Yet again, you picked up a few of my commits. Please rebase and resubmit the PR. Thank you!

jbemmel:
Superseded by https://github.com/ipspace/netsim-tools/pull/184

PR #184, opened 2022-01-24
ebgp local as plugin

jbemmel:
Clean PR to address https://github.com/ipspace/netsim-tools/pull/182#issuecomment-1020332550


ipspace:
Great job. I would still call the plugin 'ibgp-over-ebgp' as that's what it does (it is not a generic EBGP local AS implementation), but it doesn't matter. Merging...

PR #185, opened 2022-01-25
Platforms: ansible 2.9 compatibility and OSPF fix for cumulus_nvue

petercrocker:
Platforms:
- cumulus: remove FQDN from ansible task names for Ansible 2.9.X compatibility
- cumulus_nvue: 
  - fix formatting of network statement to include subnet mask of loopback
  - fix j2 formatting of area id to handle integer-formatted area ids (e.g. `area 1` vs. `area 0.0.0.1`)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -67,7 +67,7 @@ groups:
     node_data:
       bgp.anycast: 10.42.42.42/32
 
-nodes: 
+nodes:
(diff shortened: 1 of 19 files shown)


ipspace:
Thank you!

Changed ansible.netcommon.ipaddr into ipaddr (good habits die hard ;), pushed amended commit back to dev.

PR #186, opened 2022-01-25
add junos static hostname mapping

petercrocker:
Platforms:
- junos: Add static hostname mappings (Issue https://github.com/ipspace/netsim-tools/issues/139)


ipspace:
Thank you!

Had to add the "if loopback then use that otherwise use the first interface" logic. Pushed the amended commit to dev.

PR #187, opened 2022-01-26
BGP knobs/enhancements to add node-within-link level bgp.advertise_default_route and prefixes for export policies (without blackhole route) + various bits that probably should be separated out

jbemmel:
* Add link level BGP flag bgp.advertise_default_route to advertise a default route to the given eBGP peer
* Fix export policies to eBGP peers
* Enable rapid-withdraw by default on SR Linux
* Don't prepend global iBGP AS in eBGP AS path (also for FRR)


ipspace:
This makes perfect sense, but I'd like to understand what exactly you're trying to achieve: advertise IGP prefixes into BGP (in which case I'd prefer the attribute to be called **advertise** to make it consistent with other similar attributes) or create BGP prefixes without supporting RIB entries (AFAIK FRR is one of the few BGP daemons willing to do that).

Also, now that we're fixing this bit, I would also make it multi-AF. No need for rework, I can take your commit and tweak it as needed, I just need to know what the end-goal is.

ipspace:
As for "default route originate" attribute -- that makes perfect sense, but it needs to be done right:

* Specifying it on a link makes no sense -- who is going to advertise default route to whom? It has to be specified on interface (node data within link).
* I'd add a node-level attribute to originate default route to EBGP and/or IBGP peers. How often does it make sense to advertise the default route just to one EBGP peer (not saying it doesn't, just that it might be more common to advertise it to _all_ EBGP peers).

Finally, _pretty please_, don't mix a bunch of divergent things into a single pasta dish with semi-relevant title. Your future self (and everyone else) will be grateful for sane commit titles. This is not about SR OS enhancements, this is about adding two new core BGP attributes.

Unfortunately, I can't find a good way to unravel this one into its constituent parts, so once we agree what needs to be done, I'll merge and amend the commit, implement it on a few more platforms, and then ask you to fix the SR-* parts. OK?

jbemmel:
We can leave this PR as a reflection of intent, for discussion purposes, and I'll re-submit individual PRs as appropriate

We seem to be missing an attribute level: we have 'global', 'node' and 'link' but not 'link.node' (i.e. node within the context of a link). bgp.advertise_default_route is an example, there may be others. It's not a link level attribute, but the tool wouldn't let me put it under link.node unless I declared it as link attribute.

I briefly played with using "bgp.advertise=default" (to avoid having to define yet another attribute), but I figured this would overload things with new semantics (and you've already clarified that that is undesirable, elsewhere)

jbemmel:
Regarding 'export': I needed 'originate' but without creating a local black hole route. In my example, I sent 2 prefixes down from the spine ( 0.0.0.0/1 and 128.0.0.0/1 just to override a "real" default route at existing Linux hosts ), and I needed an export policy to pass those on at the leaves. 'originate' obviously doesn't work for that.

No redistribution or relation to IGP (although perhaps a broader import/export policy framework would include that too)

ipspace:
> We can leave this PR as a reflection of intent, for discussion purposes, and I'll re-submit individual PRs as appropriate

Perfect. Thank you!

> We seem to be missing an attribute level: we have 'global', 'node' and 'link' but not 'link.node' (i.e. node within the context of a link). bgp.advertise_default_route is an example, there may be others. It's not a link level attribute, but the tool wouldn't let me put it under link.node unless I declared it as link attribute.

Agreed. Will fix that. In the meantime, if it's declared as a link attribute, then it can be used within node data, so you can proceed with that part and I'll move it to interface attributes when they're done. Also, we can keep it interface-level at the moment, and if someone starts asking for more we can add that (not that anyone ever asked for anything ;).

> I briefly played with using "bgp.advertise=default" (to avoid having to define yet another attribute), but I figured this would overload things with new semantics (and you've already clarified that that is undesirable, elsewhere)

Thank you. Overloads usually end as a pile of technical debt.

ipspace:
> Regarding 'export': I needed 'originate' but without creating a local black hole route. In my example, I sent 2 prefixes down from the spine ( 0.0.0.0/1 and 128.0.0.0/1 just to override a "real" default route at existing Linux hosts ), and I needed an export policy to pass those on at the leaves. 'originate' obviously doesn't work for that.

Many BGP implementations don't want to advertise something you have no clue about (prefix not in RIB).

> No redistribution or relation to IGP (although perhaps a broader import/export policy framework would include that too)

It's not a redistribution, it's just that BGP wants to see a RIB prefix before advertising a configured BGP prefix. Redistribution is an "uncontrolled" dump of a subset of RIB prefixes into BGP table.

ipspace:
**default-originate** has been implemented in #358, if there's anything else in this PR that needs to be implemented, please open an issue or another PR.

PR #188, opened 2022-01-26
Add basic evpn module

jbemmel:
This PR adds a basic evpn module, enabling the EVPN address family towards BGP peers (iBGP or eBGP)

See https://github.com/ipspace/netsim-tools/issues/136


ipspace:
I see no problems in merging this as a starting point for EVPN module, but I can't see how you link EVPN with L2 interfaces to get end-to-end transport.

jbemmel:
See my parallel comment: https://github.com/ipspace/netsim-tools/issues/136#issuecomment-1024371889

I agree there are missing bits, but I kind of like the idea of having a minimal evpn module that only enables EVPN AF, while the services on top are built elsewhere. Once we get some examples on different platforms, we can consolidate the common concepts into additional 'service' modules

PR #191, opened 2022-01-27
SR OS and SR Linux set custom port mtu

jbemmel:
See https://github.com/ipspace/netsim-tools/issues/171

Note that the maximum supported L2 MTU on SR OS is 9800, any larger values are currently ignored without warning.
We may want to add a sanity check and min/max supported values per device in topology-defaults.yml

### DIFF
--- a/netsim/ansible/templates/initial/srlinux.j2
+++ b/netsim/ansible/templates/initial/srlinux.j2
@@ -24,13 +24,14 @@ updates:
 {% if mtu is defined %}
 {% if (clab.type in ['ixr6','ixr10'] and mtu>9486)
    or (clab.type in ['ixrd1','ixrd2','ixrd3','ixrh1','ixrh2'] and mtu>9398) %}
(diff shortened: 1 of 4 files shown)


ipspace:
There's something wrong with the L2 "system MTU" calculation for SR Linux: either the L2 MTU takes L2 header into account, in which case 1500 should be 1514, or it doesn't, in which case you shouldn't add 14 to **node.mtu** parameter.

Also, just checking: is the **mtu** parameter on an SR Linux interface L3 MTU?

jbemmel:
I agree the minimum value of 1500 for L2 MTU is somewhat counter-intuitive, but that's what it is. If the user specifies mtu=1400, we can put that as IP MTU, but the L2 MTU cannot be less than 1500 - hence the max[mtu+14,1500] construct


SR Linux interface mtu is L2 ("Port MTU in bytes including ethernet overhead but excluding 4-bytes FCS") - that's a bug, thanks for reporting it

jbemmel:
Fixes https://github.com/ipspace/netsim-tools/issues/177

PR #192, opened 2022-01-28
Determine same iBGP RR cluster-id for redundant clusters

jbemmel:
See https://datatracker.ietf.org/doc/html/rfc4456#section-7 - a cluster of iBGP route reflectors should be configured with the same cluster-id as a loop prevention measure.

This patch calculates abgp.rr_cluster_id using the lowest id of the rr nodes


ipspace:
_cluster-id_ has never been a loop prevention measure, it was always about optimizing the updates (the loop would stop the moment a RR sees its own ID in cluster-list), but I agree, things should be done right.

Merging, will have to amend it to make cluster_id configurable (in case someone wants to experiment).

PR #195, opened 2022-01-30
Make BGP cluster ID optional

ipspace:
We might not want to enforce bgp.rr_cluster_id in all situations. Setting bgp.rr_cluster_id to False in node data does that, but that possibility has to be recognized in configuration templates.

There are two types of device configuration templates in regards to BGP RR cluster-id attribute:

* Devices that configure RR-clients on BGP sessions (IOS, EOS, NXOS, Cumulus...) -- on those devices we simply don't configure **cluster-id** if bgp.rr is not True or bgp.rr_cluster_id is not set.
* Devices that use **cluster-id** to change a BGP speaker into a route reflector -- on those devices we have to use **bgp.rr_cluster_id** or **bgp.router_id** as the cluster ID.

@jbemmel: Please check whether I made correct changes to SR Linux and SR OS templates. Thank you!

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.cli.j2
+++ b/netsim/ansible/templates/bgp/srlinux.cli.j2
@@ -43,7 +43,7 @@ send-community large {{ 'true' if 'large' in list else 'false' }}
 #}
 {% if g=='ibgp' and bgp.rr|default(0)|bool %}
  route-reflector {
-  cluster-id {{ bgp.rr_cluster_id }}
+  cluster-id {{ bgp.rr_cluster_id|default(False) or bgp.router_id }}
   client true
  }
  next-hop-self false
--- a/netsim/ansible/templates/bgp/cumulus.j2
+++ b/netsim/ansible/templates/bgp/cumulus.j2
@@ -10,7 +10,7 @@ router bgp {{ bgp.as }}
 {% if bgp.router_id|ipv4 %}
   bgp router-id {{ bgp.router_id }}
 {% endif %}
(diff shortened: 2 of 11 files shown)


netsim/ansible/templates/bgp/srlinux.cli.j2
```diff
@@ -43,7 +43,7 @@ send-community large {{ 'true' if 'large' in list else 'false' }}
 #}
 {% if g=='ibgp' and bgp.rr|default(0)|bool %}
  route-reflector {
-  cluster-id {{ bgp.rr_cluster_id }}
+  cluster-id {{ bgp.rr_cluster_id|default(False) or bgp.router_id }}
```
jbemmel, line 46: I would suggest
```
bgp.rr_cluster_id|default( bgp.router_id )
```

jbemmel (changes requested):
Suggest
```
bgp.rr_cluster_id | default( bgp.router_id )
```

netsim/ansible/templates/bgp/srlinux.cli.j2
ipspace: Thanks for the feedback! That's what I had initially ;)

However, the only way to say "don't calculate cluster ID" is to set **bgp.rr_cluster_id** to **False**, in which case it is **defined** and so the **default** filter does nothing.

I'm guessing `bgp.rr_cluster_id or bgp.router_id` would also work, but with the longer version we're also handling an (impossible?) edge case of "there is no **bgp.rr_cluster_id** attribute"

PR #199, opened 2022-02-03
Support link.node module attributes, and add local_as override on a per-node-per-link basis

jbemmel:
local_as is currently supported in a plugin under extra, but only as a node level attribute. 

This adds support for more "advanced" (some might say "insane") topologies that use different link-level eBGP local_as values

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -329,7 +329,7 @@ def augment_lan_link(link: Box, addr_pools: Box, ndict: dict, defaults: Box) -> 
     node_if['data'].neighbors = []
     for remote_if in interfaces:
       if remote_if['node'] != node_if['node'] or remote_if['data'].ifindex != node_if['data'].ifindex:
-        ngh_data = { 'ifname': remote_if['data'].ifname, 'node': remote_if['node'] }
+        ngh_data = Box({ 'ifname': remote_if['data'].ifname, 'node': remote_if['node'] })
         ifaddr_add_module(ngh_data,ngh_data,defaults.get('module'))
         for af in ('ipv4','ipv6'):
           if af in remote_if['data']:
@@ -405,10 +405,7 @@ def augment_p2p_link(link: Box, addr_pools: Box, ndict: dict, defaults: Box) -> 
     link[end_names[i]] = { 'node': link_nodes[i]['name'],'ifname': interfaces[i].get('ifname') }
 
     remote = link_nodes[1-i].name
-    neighbor = {
-        'node': remote,
-        'ifname': interfaces[1-i]['ifname']
-    }
+    neighbor = Box({ 'node': remote, 'ifname': interfaces[1-i]['ifname'] })
     ifaddr_add_module(neighbor,interfaces[1-i],defaults.get('module'))
     for af in ('ipv4','ipv6'):
       if af in interfaces[1-i]:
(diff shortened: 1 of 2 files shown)


jbemmel:
Note that tests are currently failing due to attributes like ospf.cost; those could use some refinement (as those are 'pure' link attributes, as opposed to 'link.node' attributes)

ipspace:
Ignoring the ebgp-local-as part... Did I get it right that you'd like to copy module data from remote interface into the neighbor dictionary?

That makes perfect sense, but if we're doing that change, I'd clean up the code that does that (as P2P link turns out to be nothing else but a special case of multi-access link in this part of the code), so I suggest:

* We figure out what exactly you're trying to achieve (a simple "yeah, that's it" would do ;)
* I clean up the code in links.py and make sure everything works as expected
* You resubmit this PR with just the ebgp-local-as part.

As for OSPF, you can specify ospf.cost on a link or on an interface (link.node data), and the same applies to most everything else. It's just that we rarely care how the other end is configured -- **local-as** is one of those few exceptions. 

Obviously, having neighbor interface module data available would allow us to do additional checks down the road, for example check the OSPF area mismatches.

jbemmel:
As also mentioned [here](https://github.com/ipspace/netsim-tools/pull/187#issuecomment-1024312508) I think we're missing a level in the hierarchy, link.node (after global, node, link levels)

local_as is just an example of a use case where I need it. 'underlay_as' is currently modeled as a node level attribute, but for an eBGP peering between spines (to avoid allow_own_as) I needed a node-level underlay_as AND a link.node level override for the spine-spine link (both have the same underlay_as to peer eBGP with leaves, but a different unique local_as to peer eBGP with each other - that way I can get the loopback route needed for iBGP peering, without resorting to allow_own_as)

From what I understand we could define ospf.cost as a link level attribute, while allowing it to be specified at link.node level too but not copying it from link to node.link (like the plugin effectively does for the node level underlay_as attribute)

ipspace:
> As also mentioned [here](https://github.com/ipspace/netsim-tools/pull/187#issuecomment-1024312508) I think we're missing a level in the hierarchy, link.node (after global, node, link levels)

So you'd like to have an interface-level attribute that is NOT at the same time a link-level attribute. No problem there, but that's not what your changes are doing.

> From what I understand we could define ospf.cost as a link level attribute, while allowing it to be specified at link.node level too but not copying it from link to node.link (like the plugin effectively does for the node level underlay_as attribute)

Every module attribute that is defined at the link level is copied into interface data ([line 313](https://github.com/ipspace/netsim-tools/blob/dev/netsim/augment/links.py#L313)) where it might be overwritten by node(interface) data (https://github.com/ipspace/netsim-tools/blob/dev/netsim/augment/links.py#L315). 

I do agree that https://netsim-tools.readthedocs.io/en/latest/modules.html should be more explicit though ;)

Anyway, coming back to the original question: what exactly are you trying to achieve with your changes to links.py?

jbemmel:
What I'm doing is to make this snippet work:
```
nodes:
 spine1:
  bgp.underlay_as: 65010
 spine2:
  bgp.underlay_as: 65010
links:
- spine1:
   bgp.underlay_as: 65011 # Build eBGP peering between spines
  spine2:
   bgp.underlay_as: 65012
```

For lack of a proper node.link attribute, I'm (ab)using a regular link attribute for that. We could add full support, but that would require more changes

ipspace:
> As also mentioned [here](https://github.com/ipspace/netsim-tools/pull/187#issuecomment-1024312508) I think we're missing a level in the hierarchy, link.node (after global, node, link levels)

Interface-level attributes have already been implemented, but they were "underdocumented" and hidden under weird name (**link_node**). a490d81 fixes the documentation, adds **interface** key to **attributes** dictionary and adds **local_as** to BGP interface attributes just to have a test case ;)

jbemmel:
Nice, I agree 'interface' makes more sense than link_node.

Closing this episode

PR #215, opened 2022-02-21
Always define node.interfaces

jbemmel:
Fixes https://github.com/ipspace/netsim-tools/issues/213

There are 92 cases of interfaces|default([]) in the templates that could be simplified


ipspace:
Thank you!

Will recreate the test cases and amend the commit.

PR #219, opened 2022-02-24
Doc for Management VRF in vEOS and vSRX

ssasso:
I'm a big fan of keeping the management traffic on a dedicated vrf.
Feel free to reject the PR if you don't like the approach (but I see you are doing the same on CSR, so I guess it's ok).

### DIFF
--- a/netsim/install/libvirt/eos.txt
+++ b/netsim/install/libvirt/eos.txt
@@ -5,6 +5,8 @@ Creating initial configuration for Arista EOS
 * Login as 'admin' (no password)
 * Disable zero-touch with 'zerotouch disable'
 
+NOTE: the management traffic is isolated in a dedicated management VRF (management).
+
 ===============
 *** WARNING ***
 ===============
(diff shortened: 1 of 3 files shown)


ipspace:
I totally like the idea, but I have a vague feeling that the default setup of Arista EOS includes management VRF. Give me a few days to check it.

ssasso:
As far as I can see, it does not have it activated by default.

See https://aristanetworks.force.com/AristaCommunity/s/article/introduction-to-managing-eos-devices-setting-up-management - it says:
```
In order to use the management VRF:
* the management VRF must be created
```

(I know it is an old document, but I am not able to find sth newer. Btw, it also says an RD must be set, but in my environment it works even without that).

On my test (EOS 4.27), by default there is no VRF apart from the default, and the `Management 1` interface is in it:
```
r1#sh vrf
Maximum number of VRFs allowed: 1024
   VRF           Protocols       State         Interfaces
------------- --------------- ---------------- ----------
   default       IPv4            no routing
   default       IPv6            no routing

r1#sh ip route

VRF: default
Codes: C - connected, S - static, K - kernel,
       O - OSPF, IA - OSPF inter area, E1 - OSPF external type 1,
       E2 - OSPF external type 2, N1 - OSPF NSSA external type 1,
       N2 - OSPF NSSA external type2, B - BGP, B I - iBGP, B E - eBGP,
       R - RIP, I L1 - IS-IS level 1, I L2 - IS-IS level 2,
       O3 - OSPFv3, A B - BGP Aggregate, A O - OSPF Summary,
       NG - Nexthop Group Static Route, V - VXLAN Control Service,
       DH - DHCP client installed default route, M - Martian,
       DP - Dynamic Policy Route, L - VRF Leaked,
       G  - gRIBI, RC - Route Cache Route

Gateway of last resort is not set

 C        172.23.131.0/24 is directly connected, Management1

! IP routing not enabled
```

The configuration that I put for vEOS in the PR is the one that I set on my EOS "golden image", when I created it.

ipspace:
My bad. NXOS has management interface in a VRF, not EOS, so this one is a go.

Just a question before we move forward (cc @jbemmel @petercrocker): I think we should make management VRF the default in build recipes. Any concerns/objections?

ipspace:
Thank you!

PR #231, opened 2022-03-31
Enable Cumulus 5.0.1 NVUE for Clab

barajus:
Hi everyone,
it looks like the support of NVUE for CL 5.0.1 and containerlab is missing, so this patch enables the (small) missing part.
Unless I missed something, that works, `run-tests` looks good as well:
`Success: no issues found in 53 source files`

Let me know if something is needed for that part :)

Julien

### DIFF
--- a/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
+++ b/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
@@ -1,3 +1,9 @@
+- name: "start nvue daemon"
+  command: systemctl start nvued
+
+- name: "enable nvue"
+  command: systemctl enable nvued
+
 - name: "copy the cumulus nvue yaml config file to switch for {{ config_module }} config"
   template:
     src: "{{ config_template }}"
--- a/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
+++ b/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
@@ -1,17 +1,28 @@
-- name: "start nvue daemon"
-  command: systemctl start nvued
-
-- name: "enable nvue"
-  command: systemctl enable nvued
-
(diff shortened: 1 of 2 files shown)


ipspace:
The settings part looks good to me (and if it works for you, then I guess we're all set).

However, I would be better to optimize the Ansible configuration part -- the nvue daemon has to be enabled during initial configuration, but not during every configuration activity (because it should have been enabled before).

The easiest way to get that done (speaking off-the-cuff here) would be to move your changed tasklist into `initial` directory and keep the original task list in `deploy-config` directory, mirroring what I had to do to check interfaces on Nexus OS.

barajus:
Okay, the deploy-config commands I added (enable and start NVUE) in the playbook are not necessary, everything is already in place. The problem is different, NVUE needs a couple of seconds to be up and running inside a container, and Ansible is too fast (I never thought one day I would say this) and try to apply the config while NVUE is not necessary started.

```
❯ docker exec clab-netsim-tools-leaf01 systemctl status nvued
* nvued.service - NVIDIA User Experience Daemon
   Loaded: loaded (/lib/systemd/system/nvued.service; enabled; vendor preset: enabled)
   Active: inactive (dead)
 [...] many attempts until finally:
 * nvued.service - NVIDIA User Experience Daemon
   Loaded: loaded (/lib/systemd/system/nvued.service; enabled; vendor preset: enabled)
   Active: activating (start) since Tue 2022-04-05 22:03:15 UTC; 9s ago
```
A quick and dirty solution would be to use the module `pause` in ansible/tasks/deploy-config/cumulus_nvue.yml such as:
```
- name: "testing"
  pause:
    seconds: 30
```
   
 I tried and it works, but again, this is not elegant. 
 Maybe instead I could use something with `wait_for` to verify if NVUE is ready. 
 Give me your opinion and I will fix accordingly.
 
 Julien

barajus:
This is much better, thanks for the Ansible magic :)

PR #232, opened 2022-03-31
Change python path to avoid Ansible issue

barajus:
Hi,
on some flavours, the python binary cannot be found in `/usr/local/bin/`
I experienced the same issue with some containers (for linux hosts) as well, which breaks Ansible at the task "template":
`Failed to get information on remote file (/tmp/config.sh): /bin/sh: 1: /usr/local/bin/python3: not found`

Even though, this is a packaging issue and not really a netsim issue, I wonder if we could proactively avoid some Ansible drama by using another $PATH with more better to succeed (or maybe implement a smart routine to check different $PATH).

Not sure this patch makes sense, but anyway, maybe someone will get a smarter solution :)

Julien

### DIFF
--- a/netsim/topology-defaults.yml
+++ b/netsim/topology-defaults.yml
@@ -409,7 +409,7 @@ devices:
       group_vars:
         ansible_connection: docker
         ansible_user: root
-        ansible_python_interpreter: /usr/local/bin/python3
+        ansible_python_interpreter: /usr/bin/python3
 
   vsrx:
     interface_name: ge-0/0/%d
--- a/netsim/topology-defaults.yml
+++ b/netsim/topology-defaults.yml
@@ -409,7 +409,7 @@ devices:
       group_vars:
         ansible_connection: docker
         ansible_user: root
-        ansible_python_interpreter: /usr/bin/python3
+        ansible_python_interpreter: auto
 
   vsrx:
     interface_name: ge-0/0/%d
@@ -503,7 +503,17 @@ devices:
           lla: True
       ospf:
         unnumbered: True
-
+    clab:
+      mtu: 1500
+      runtime: docker
+      node:
+        kind: cvx
+      image: networkop/cx:5.0.1
+      group_vars:
+        ansible_connection: docker
+        ansible_user: root
+        ansible_python_interpreter: auto
+        
   srlinux:
     mgmt_if: mgmt0
     interface_name: ethernet-1/%d


ipspace:
It might be better to remove the `ansible_python_interpreter` group variable -- Ansible should be able to figure out what to use, and we require Python 3.7+ during netsim-tools installation anyway.

Could you try that and add another commit if that works?

barajus:
Hi,
unfortunately removing `ansible_python_interpreter` does not work, but if I try with the following values, then it works:
```
ansible_python_interpreter: auto
ansible_python_interpreter: auto_slient
ansible_python_interpreter: auto_legacy
```

Which is strange to me, because removing `ansible_python_interpreter` like you suggested should let Ansible using `interpreter_python=auto_silent` from the default in `ansible.cfg` and handle this case.

ipspace:
What's weird is that I tested Linux-in-containers, and @petercrocker tested Cumulus NVUE.

Just to make sure we're not chasing ghosts: based on our previous discussion, are you sure you're running Ansible with Python3? Please run `ansible --version` and check `python version`.

ipspace:
OK, I can confirm that you need `ansible_python_interpreter` set when the path to `python3` on the container differs from the path to `python3` on Ansible host. Stupid, stupid, stupid...

Anyway, setting `ansible_python_interpreter` to `auto` solves the problem for me, so I'd suggest we go with that. If you feel like changing that setting and adding another commit, I'll merge your PR directly, otherwise I'll merge your PR, fix it and amend the commit, so you'll be listed as a contributor regardless of the final outcome.

barajus:
Hi, sorry I got very busy with different things (I will also answer to the other PR later today).
FWIW, I work within a venv, and the version are `Python 3.9.9` and `pip 22.0.4` but that does not really matter anymore.

Let me add another commit with the fix, but before that, what do you think about `linux` boxes? 
Do you want to keep it like it is today, with the hardcoded path or do you prefer to set `ansible_python_interpreter` on `auto` too, to make things more consistent?

ipspace:
Yes, we should do the same for Linux boxes. I hoped the behavior was caused by the `docker` connection, but I get the same behavior on Linux boxes if I move `python3` into `/usr/local/bin`.

@petercrocker: Should we do the same for Cumulus VX?

barajus:
Hi,
- Pulled the last diff from dev (please verify if that the merge is fine, sorry if that was not expected)
- Fix python interpreter path for `Linux`
- Add the config block for clab and `cumulus_nvue`
- Add python interpreter path for `cumulus_nvu`e as well as the python

Once the patch is committed, I will fix the minor issue in PR#231 as well.
If you keep track of the contributors, you can add my full name: Julien Dhaille, instead of this github non-sense nickname :)

ipspace:
You did a picture-perfect job, thank you!

PR #239, opened 2022-04-08
Fix: errors in Mikrotik ROS BGP RR handling (and VPNv4+OSPF)

ssasso:
@ipspace I **guess** a fix is also needed in:

1. https://github.com/ipspace/netsim-tools/blob/567160cc8a0e12863b9a376615e628c3c5de1a25/netsim/ansible/templates/vrf/eos.bgp.j2#L25
2. https://github.com/ipspace/netsim-tools/blob/567160cc8a0e12863b9a376615e628c3c5de1a25/netsim/ansible/templates/vrf/ios.bgp.j2#L13

to use *vdata.bgp* instead of *bgp*.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ venv/
 # Ignore everything that should be ignored ;)
 *.ignore.*
 x/
+publish-pypi.sh
 
 # Ignore Ansible retry and log files
 *.retry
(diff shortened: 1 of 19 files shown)


ipspace:
There are three parameters in **bgp** data structure that the macros look at:

* Route reflector (not relevant as long as the PE-CE sessions are EBGP)
* Ditto for next-hop-self
* Send community -- this one is a challenge

We could either decide we won't send community to VRF BGP neighbors, or we have to copy the node BGP settings into VRF BGP settings, or we have to use node BGP settings in macros.

What do you think is the best way forward?

ssasso:
I would keep the option to send communities on the VRF BGP session. I think we can keep the *eos* and *csr* code as is for now. If in the future we will identify different use cases, we can expand the VRF BGP node settings.

PR #241, opened 2022-04-08
Support output in Graphite JSON format

ssasso:
Why use Graphite (https://github.com/netreplica/graphite) only with ContainerLab? :-)

With this output module it would be possible to export a topology file to be used within Graphite.
 
Graphite container can be launched with:

```
docker run -d \
 -v "$(pwd)/graphite-default.json":/var/www/localhost/htdocs/default/default.json \
 -p 8080:80 \
 --name graphite \
 netreplica/graphite:webssh2
```
Result example: 
image

Additionally, with the *graphite:webssh2*, SSH can be launched from the browser as well ;)

image

### DIFF
--- a/docs/outputs/graphite.md
+++ b/docs/outputs/graphite.md
@@ -1,0 +1,89 @@
+# Graphite Topology Output Module
+
+Why use [Graphite](https://github.com/netreplica/graphite) only with ContainerLab? :-)
+
+*graphite* output module creates a JSON file that can be used to display the *netsim-tools* lab topology within Graphite.
+
+By default, the output is created as *graphite-default.json* file, which can be used directly from the Graphite container. Additionally, using the container *graphite:webssh2*, it is possible to launch SSH sessions towards the nodes directly from the browser.
+
+The Graphite container can be launched with:
+```
+docker run -d \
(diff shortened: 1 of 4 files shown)


ssasso:
You're totally right, it would be great to have configurable icons: I will do that improvement.
I'll also add a way to display "host bridges" like hub/cloud with multiple endpoints.

ipspace:
> I'll also add a way to display "host bridges" like hub/cloud with multiple endpoints.

You can use the same trick to display stub networks ;)

ssasso:
> > I'll also add a way to display "host bridges" like hub/cloud with multiple endpoints.
> 
> You can use the same trick to display stub networks ;)

that was the idea ;)

image

ssasso:
(wait to merge it, still doing additional stuff)

ssasso:
(I'll write the documentation page in a while)

ipspace:
Tell me when this will be ready and i'll do the merge -- can't wait to play with this stuff ;))

ssasso:
I think we are set (for now ;)).
Feel free to change the documentation if you find it unclear.
Unfortunately there's not so much documentation on graphite. Most probably in a near future we can expand the output generation having a look at the NeXt UI documentation on which graphite is based (https://developer.cisco.com/site/neXt/).

PR #242, opened 2022-04-08
Fix: Ubuntu Vagrant Box DNS stuff

ssasso:
It seems on the Vagrant box for ubuntu 20.04, DNS Servers are hardcoded as 4.2.2.1 & Co.
This is annoying on a network with filtered DNS.
DNSMasq server used for giving out DHCP addresses on the management network is able to act as a DNS Server.
Let's use that.


ipspace:
Looks good (read: I don't know enough to see why it wouldn't work, and I trust you tested it).

PR #249, opened 2022-04-13
VyOS: Support for MPLS (and MPLS L3VPN)

ssasso:
@ipspace **WARNING**: 

I had to add to the **vrf** module the following (otherwise frr on vyos does not correctly handle the service label):
```
  config_after: [ ospf, isis, bgp, mpls ]
```
It seems to me that also *logically* this ordering is fine.
A did a quick test of a L3VPN topology with *vyos* (of course ;) ), *csr*, *eos*, *chr*, and everything looked good.
Do you see any other issue with that change?


ipspace:
I don't see any problem with the order of configuration (have to write a document describing it though ;).

The tests are failing due to the changed order, will recreate the expected test results and amend the commit after merge.

PR #250, opened 2022-04-15
Update linux.j2 - fix for Alpine

barajus:
Hi,
with clab/docker as provider, Linux hosts are using the image `python:3.9-alpine` but that triggers some issues due to the absence of `netplan` and `systemd`.

This is a quick-and-dirty fix to keep things working with Alpine, but maybe we should implement a routine to deal with the flavor of images pulled (alpine/debian/ubuntu/rh). Replacing `python:3.9-alpine` with an official ubuntu images required the install of python.

Julien

### DIFF
--- a/netsim/ansible/templates/initial/linux.j2
+++ b/netsim/ansible/templates/initial/linux.j2
@@ -14,7 +14,7 @@ SCRIPT
 # This is annoying on a network with filtered DNS.
 # DNSMasq server used for giving out DHCP addresses on the management network is able to act as a DNS Server.
 # Let's use that.
-
+{% if ansible_connection != "docker" %}
 # (Overwrite netplan config to remove DNS stuff)
 cat <<SCRIPT >/etc/netplan/01-netcfg.yaml
 network:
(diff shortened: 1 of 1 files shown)


ipspace:
Thanks for the fix. I changed the test condition into `'ubuntu' in box`, so it should work with everything that is (explicitly) Ubuntu-based, amended the commit and force-pushed the change into dev branch.

Not sure I want to go into all sorts of weird Linux distro games, Linux devices were meant to be simple hosts you could use for pings. Maybe I should write a short document on building your own Linux images?

ssasso:
I apologize for breaking this stuff. :-(
However, in my defense, the *linux.j2* template also used to invoke:
```
apt-get update
apt-get install
```
Which should not work as well in an alpine container.

Maybe we should extend the check also to that lines?

barajus:
I agree, as long as Alpine/Ubuntu-based works for basic icmp test, that's fine.
Adding a line about this won't hurt. The main requirement being python required on those images.

PR #254, opened 2022-04-17
Debian compatibility and general cleanup

ghostinthenet:
Summary of changes:

- Corrected all scripts to accept an empty response as Y
- Added Debian/Ubuntu distribution detection in containerlab.sh to add the correct docker repository
- Re-referenced Docker APT GPG key and APT source definition in containerlab.sh to use the APT trusted.gpg.d directory
- Removed deprecated use of "apt-key add" and "add-apt-repository" in libvirt.sh and replaced it with current mechanism
- Added curl to missing packages in ubuntu.sh as this is not installed by default in Debian

To-do:

Deal with the fact that sudo is not installed by default on Debian. Scripts test successfully on Debian 10/11 as long as sudo is installed and configured or if run as root. Scripts test successfully on Ubuntu 20/21.

### DIFF
--- a/netsim/install/ansible.sh
+++ b/netsim/install/ansible.sh
@@ -15,11 +15,23 @@ during the installation process.
 =====================================================================
 
 EOM
+
+# Add sudo / root check - ghostinthenet 20220418
+SUDO=''
+if [ "$UID" != "0" ]; then
+ if [ -x "$(command -v sudo)" ]; then
+  SUDO=sudo
+ else
+  echo 'Script requires root privileges.'
+  exit 0
+ fi
+fi
+
 if [[ -z "$FLAG_YES" ]]; then
-  read -p "Are you sure you want to proceed [Y/n] " -n 1 -r
+  # Remove implied default of Y - ghostinthenet 20220418
+  read -p "Are you sure you want to proceed [y/n] " -n 1 -r
   echo
-  # Original script didn't properly accept an empty response as a default Y - ghostinthenet 20220417
-  if ! [[ $REPLY =~ ^$|[Yy] ]]; then
+  if ! [[ $REPLY =~ [Yy] ]]; then
    echo "Aborting..."
    exit 1
   fi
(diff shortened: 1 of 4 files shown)


ipspace:
Great job, there are just a few minor items:

* Please resubmit the PR against the dev branch (see also https://netsim-tools.readthedocs.io/en/latest/dev/guidelines.html). The only time I merge into master branch is when I'm releasing new version or when we need to fix a nasty bug.
* I'm not perfectly happy with saying "empty response = Y" -- a keyboard glitch can easily send you over the edge -- but I can be persuaded otherwise ;)
* The DISTRIBUTION logic can end with empty DISTRIBUTION variable when `/etc/lsb-release`  does not contain `ubuntu`. It would be better to test whether the DISTRIBUTION variable is set after the decision tree and abort if it's not.

It would be great if you could fix these bits and resubmit the PR, otherwise I'll fix them after the merge.

ghostinthenet:
- Will make corrections and resubmit to dev.
- I was reading the capital Y in "Y/n" as suggesting a default answer being Y, which is why I added the default. I suggest we either accept Y as a default or switch to "y/n" in the question.
- I'll add an error out if we don't have a valid debian_version / lsb-release combination.
- If I have time, I'll add the sudo check as well.

ipspace:
> Will make corrections and resubmit to dev.

Thank you.

> I was reading the capital Y in "Y/n" as suggesting a default answer being Y, which is why I added the default. I suggest we either accept Y as a default or switch to "y/n" in the question.

Thought so. Let's switch to "y/n" in the question.

ghostinthenet:
254 can be ignored. 255 was submitted against dev and includes all of the changes in 254 with the changes based on feedback. I probably should have withdrawn 254.

ghostinthenet:
254 can be ignored. 255 was submitted against dev and includes all of the changes in 254 with the changes based on feedback. I probably should have withdrawn 254.

ipspace:
Yeah, too late :( but little harm done ;) Withdrawing it would be nice, or I should have closed it when I merged the other one.

PR #255, opened 2022-04-18
Debian Compatibility

ghostinthenet:
Okay... here're the latest updates.

- Tested to see if the root user is doing the installation or if sudo is in the path before proceeding.
- Defined sudo as a variable so that the scripts can run as root in the absence of the sudo command.
- Corrected Y/n test to y/n and require an explicit [Yy] to proceed.
- Added an additional else statement in the lsb-release test to cover all of the bases. There's no need to check the DISTRIBUTION variable because the script will fail without a valid debian_version / lsb-release combination.

Hopefully these updates cover the bases.

### DIFF
--- a/netsim/install/ansible.sh
+++ b/netsim/install/ansible.sh
@@ -15,11 +15,23 @@ during the installation process.
 =====================================================================
 
 EOM
+
+# Add sudo / root check - ghostinthenet 20220418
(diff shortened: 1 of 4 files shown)


ipspace:
Thanks again! Merged.

PR #257, opened 2022-04-20
[Fix] Graphite mount point for default.js

ssasso:
According to: https://twitter.com/netreplica/status/1516803584175747080


ipspace:
Thank you! Cherry-picked it into master branch to make sure the documentation is correct.

PR #259, opened 2022-04-21
[Fix] Cumulus NVUE daemon start error on libvirt

ssasso:
At least on my deployment with libvirt, the task "Start NVUE daemon" brutally dies with the following:

```
fatal: [s1]: FAILED! => changed=true
  cmd:
  - systemctl
  - start
  - nvued
  delta: '0:00:00.014788'
  end: '2022-04-21 10:49:43.656199'
  msg: non-zero return code
  rc: 4
  start: '2022-04-21 10:49:43.641411'
  stderr: |-
    Failed to start nvued.service: Access denied
    See system logs and 'systemctl status nvued.service' for details.
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

this seems due to the fact that, on the libvirt image, the ansible login is done with the vagrant/cumulus ssh user.

Using *sudo* shall work for both the libvirt and container use case.

(I tested the patch both with libvirt and CLAB, but please @barajus give it a check as well, thanks!)

### DIFF
--- a/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
+++ b/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
@@ -6,7 +6,8 @@
 
 - block:
   - name: "Start NVUE daemon"
-    command: sudo systemctl start nvued
+    command: systemctl start nvued
+    become: true
 
   - name: "Wait for nvued to start"
     service_facts:
--- a/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
+++ b/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
@@ -6,7 +6,7 @@
 
 - block:
   - name: "Start NVUE daemon"
-    command: systemctl start nvued
+    command: sudo systemctl start nvued
 
   - name: "Wait for nvued to start"
     service_facts:


ssasso:
@ipspace It's downloadable from here: https://www.nvidia.com/en-us/networking/ethernet-switching/cumulus-vx/download/

Maybe we can just point the docs to the right URL?

ipspace:
Yeah, it's downloadable from that link which is well-protected with an annoying regwall. Will fix the documentation. They also persuaded me not to use Cumulus in any of my examples :(

ipspace:
BTW, we'll also have to change the box name, as the box downloaded from NVIDIA web site contains no metadata. I would suggest using **cumulus/vx5** similar to **juniper/vsrx3**. Would you be OK with that? /cc @petercrocker @barajus

barajus:
@ipspace I agree for the renaming :)

@ssasso I just tested with the last dev branch, and I don't see any breaks.

ipspace:
> @ipspace I agree for the renaming :)

5.0.1 is back online, so I'll leave the setup as it is. If it disappears again we know what to do.

> @ssasso I just tested with the last dev branch, and I don't see any breaks.

Because we fixed them in the meantime. I was able to reproduce the bug we introduced, and verified that "become: true" parameter fixed it on VM version of VX while not affecting the container version.

PR #264, opened 2022-04-22
VyOS: Support for VLAN Access mode

ssasso:


ipspace:
Looks great. I will fix the documentation.

Would it help if I'd add "vlan_id" (not just "vlan_name") as interface parameter for SVI/BVI interface so you wouldn't have to parse the interface name in initial.j2?

ssasso:
In my current idea (which can change based also on future needs, of course) that parsing code can be useful for future use cases - I'm thinking about plain bridges, or vxlan bridges (and so IMHO can be considered as quite "agnostic" about any specific VLAN tag, and in the future it can also support QinQ just taking the third element of the array, in example).

thanks!

PR #266, opened 2022-04-22
libvirt Vagrantfile support for node-level nic_model

ssasso:
It would be useful to support libvirt nic_model override at node level.

Use case example:
VyOS, but Linux in general, has problems in creating a *LACP LAG* with the *virtio* driver. For that reason, in that case the use of *e1000* is adviced.
On the other hand, *e1000* has problems with IPSec/GRE encapsulation...

So it would be useful to override that based on the specific test use case.

(@ipspace I might have been messing up the *topology-defaults* and *platforms.md* files between this and my previous PR. Sorry for this, but I prefer starting from the *dev* branch.)

### DIFF
--- a/docs/nodes.md
+++ b/docs/nodes.md
@@ -105,8 +105,18 @@ These node attributes are recognized and used by *netsim-tools*:
 * **loopback** -- static loopback addresses. Must be a dictionary with **ipv4** and/or **ipv6** attributes.
 * **memory** -- memory allocated to the VM lab device. Not applicable for container-based devices.
 * **cpu** -- virtual CPU cores allocated to the VM lab device. Not applicable for container-based devices.
+* **nic_model** - virtual NIC model allocated to the VM lab device. Applicable only to **libvirt**. Supported values are:
+  * virtio (*libvirt default*)
+  * e1000
+  * rtl8139
+  * pcnet
+  * ne2k_pci
+  * i82559er
+  * i82557b
+  * i82551
+  * ne2k_isa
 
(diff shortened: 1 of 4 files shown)


ipspace:
We did something similar for **clab.kind** (and a few other things), so the attribute should be **libvirt.nic_model**. We could then set that attribute in device defaults or in node. Also, you don't have to extend node attributes with provider keywords. Once I start enforcing node attribute values, that will have to be implemented by default or **clab** topologies won't work at all.

In the future, I have to extend links and interfaces with provider attributes (to implement a "do tunnel" switch for libvirt). At that time, we'll be able to specify nic_model on per-interface basis.

ipspace:
Merged the other PR. Now we have a merge conflict anyway. It would be best if you'd rebase your local branch on **dev** and force-push it to GitHub.

ssasso:
> We did something similar for clab.kind (and a few other things), so the attribute should be libvirt.nic_model. We could then set that attribute in device defaults or in node. Also, you don't have to extend node attributes with provider keywords.

Sorry I'm feeling pretty dumb on that. Is this meaning that I have to check if a default value exists, and in case use it?
(because right now if I define `defaults.devices.vyos.libvirt.nic_model: e1000` I cannot find it inside the node attributes)

ipspace:
My bad :( You have to put the attribute into `defaults.devices.vyos.libvirt.node.nic_model`. See **eos** device for a **clab** example.

Also, I'll add a tweak to send a copy of the topology to output modules so the `defined` checks don't destroy the main data structure.

ssasso:
Thanks!
I updated the *nodes.md* documentation to try to reflect this behavior.

PR #267, opened 2022-04-24
Update topology-defaults.yml

barajus:
Fix python error for FRR
`Failed to get information on remote file (/tmp/config.sh): /bin/sh: /home/ubuntu/venv2/bin/python3: not found`

### DIFF
--- a/netsim/topology-defaults.yml
+++ b/netsim/topology-defaults.yml
@@ -414,6 +414,7 @@ devices:
     group_vars:
       ansible_connection: docker
       ansible_network_os: frr
+      ansible_python_interpreter: auto_silent
     clab:
       image: frrouting/frr:v7.5.0
       mtu: 1500


ipspace:
Thank you!

This is turning into a whack-a-mole. Could you please check whether we need a similar fix for any other platform? Also, https://xkcd.com/1987/

barajus:
Hi back, I have been quiet busy lately, I couldn't follow-up before :)

I checked all the platforms supported by clab:

1. EOS requires `paramiko` 
```
fatal: [h3]: FAILED! => changed=false
  elapsed: 60
  msg: 'timed out waiting for connection port up: paramiko is not installed: No module named ''paramiko'''
```
=> I suggest add it into the `requirements.txt`

2. Nokia requires a collection from ansible-galaxy.
```
TASK [Check that required plugin exists, use 'ansible-galaxy collection install nokia.grpc' to install it] ************************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'google'
fatal: [h2]: FAILED! =>
  msg: Unexpected failure during module execution.
  stdout: ''
```
=> The documentation is already explicit, however; I suggest to automatically install this plugin. That's quite small in term of size, but most importantly, I think that give a better user experience overall = everything "works" outside the box.

3. Something is broken for Srlinux:
```
TASK [Update SRL initial node configuration (gNMI SET template=templates/initial/srlinux.j2,CA=/home/ubuntu/tutu/LOL/clab-LOL/ca/root/root-ca.pem)] *******************
fatal: [h2]: FAILED! => changed=false
  code: -32603
  msg: |-
    <_InactiveRpcError of RPC that terminated with:
            status = StatusCode.UNAVAILABLE
            details = "failed to connect to all addresses"
            debug_error_string = "{"created":"@1651742769.315096483","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3217,"referenced_errors":[{"created":"@1651742769.315095992","description":"failed to connect to all addresses","file":"src/core/lib/transport/error_utils.cc","file_line":165,"grpc_status":14}]}"
    >
```

4. I didn't test `SROS` since it needs a license
5. Another suggestion on a totally different topic, maybe we could remove those messages from Ansible:
`[DEPRECATION WARNING]: Use 'ansible.utils.ipaddr' module instead. This feature will be removed from ansible.netcommon in a release after 2024-01-01. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

Julien

ipspace:
> Hi back, I have been quiet busy lately, I couldn't follow-up before :)

No problems. Thanks a million for an extensive report!

Warning: a pile of negativity and strong language ahead, all of it directed solely toward people who can't be bothered to ship consistent software.

> 1. EOS requires `paramiko`
> => I suggest add it into the `requirements.txt`

All of us have a life, and the worst thing we could do with it would be to turn _netsim-tools_ into Ansible dependency management tool 🥴

For people who want to have out-of-the-box experience, `netlab install ansible` installs all reasonable dependencies, including paramiko and netmiko... because Ansible doesn't list either one of those as a dependency.

> 2. Nokia requires a collection from ansible-galaxy.

If that were true, I'd add it to the Ansible installation script. Unfortunately, the Nokia collection on ansible-galaxy wasn't updated in at least two years, and so the recommendation was made to install the collection straight from GitHub.

Call me paranoid or crazy, but I'm not going to put that into a generic installation script ;)

> 3. Something is broken for Srlinux:

Works for me (yeah, I know, the famous answer) after upgrading containerlab (which has another bug, but that's a different story) and pulling down the latest image, so at least our Ansible playbook, their Docker container or containerlab are not broken.

The only things left in the middle are gRPC modules and their Ansible collection, and I don't want to touch either one of those :(

> 5. Another suggestion on a totally different topic, maybe we could remove those messages from Ansible:
>    `[DEPRECATION WARNING]: Use 'ansible.utils.ipaddr' module instead. This feature will be removed from ansible.netcommon in a release after 2024-01-01. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

And this is another reason I don't want to get into the business of Ansible dependency (and version) management.

Some people can't use Ansible beyond 2.9 because RedHat does not support the newer crap. "Community" version and 2.9 are diverging with every release (hooray for consistency), and they can't get their shit together.

We're not using fully-qualified name of `ipaddr` filter anywhere because that wouldn't work with 2.9, and what's happening now is that they're still mapping `ipaddr` into `ansible.utils.ipaddr` and throwing deprecation warnings instead of fixing their mapping. I'm disgusted.

I'm guessing that since we use Ansible in a pretty closed environment we could set `deprecation_warnings=False` in `ansible.cfg`. Any objections? /cc @ssasso @petercrocker

ssasso:
No objections for me

barajus:
No worries with the strong language, I also share most of these statements 😄
 
Thanks for the feedback, the last thing I would like to comment is about Nokia SR. 
I would suggest to comment the block in `netsim/ansible/tasks/deploy-config/srlinux.yml`: 
`-- name: Update SRL {{ netsim_action }} node configuration (gNMI SET template={{config_template}},CA={{ansible_root_certificates_file}})`

to remove the RPC issue:
```
  code: -32603
  msg: |-
    <_InactiveRpcError of RPC that terminated with:
            status = StatusCode.UNAVAILABLE
            details = "failed to connect to all addresses"
            debug_error_string = "{"created":"@1651825266.523438742","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":3217,"referenced_errors":[{"created":"@1651825266.523438241","description":"failed to connect to all addresses","file":"src/core/lib/transport/error_utils.cc","file_line":165,"grpc_status":14}]}"
    >
```

Let me know if you guys agree with it and I will push a PR.
Julien

barajus:
The timing is good, the wait_for test works, and SSH is ready (also tested manually).
I have to be honest, I never touched gRPC so I didn't figure yet why this problem occurs, so hints are very welcome.

Also, to clarify, spinning Nokia SR Linux images with containerlab works indeed, this is really related to the 'initial node configuration' step.

ipspace:
> The timing is good, the wait_for test works, and SSH is ready (also tested manually). I have to be honest, I never touched gRPC so I didn't figure yet why this problem occurs, so hints are very welcome.

Does `netlab initial` executed a while later work (in which case it might be a timing issue) or does it return the same error, in which case we have a bigger problem?

@jbemmel created all configuration templates using gRPC, so I don't think we can just turn that task off.

Just to be on the safe side, here's the list of relevant Python modules from an environment where gRPC with Nokia SR Linux still works:

```
grpcio                 1.43.0
protobuf               3.19.3
```

barajus:
No later work, `netlab initial` returns the same error.

Originally, my modules were: `grpcio-1.46.0 and protobuf-3.20.1`
I reinstall the similar version as yours, and reinstalled the collection just in case, but no differences of behavior.

ipspace:
I'm totally out of ideas... :( Once I find enough time, I'll set up a brand-new Ubuntu VM and try it from scratch.

PR #269, opened 2022-04-25
Initial support for External Topology provider

ssasso:
There might be some situations where you want to use a Network Topology external to the *netsim-tools* VM (i.e., a physical lab, a cloud deployment, ...)... but you want to use *netsim-tools* to create and push the configuration using the "standard" methodology.

This PR wants to try to address that use case.

The **external** provider does the following:
* allow to use a different interface name as link->node attribute
* creates a TXT file with the physical cabling required for the Lab
* it asks for confirmation of the physical configuration before starting ansible

**Please wait to merge**, I'd like to receive any feedback about the feature itself, and about nice-to-have-features for it.


ipspace:
Great minds think alike -- I was musing about the physical provider for the last few days.

I had the idea to:

* Somehow generate the topology file from device operational data (see below)
* Use **ifname** attribute in interface (node-on-link) data, and add a check in **links.py** to skip interface name generation when **ifname** is already defined
* Skip **mgmt** setup in node setup when **mgmt** is already defined

As we don't check node variables yet, it's perfectly possible to add Ansible variables (authentication info) to the node data.

Now for the "generating topology file" bit. Starting with an Ansible inventory, it should be relatively easy to execute "gather facts" on all devices and use LLDP data to build the network topology. I did something similar in https://github.com/ipspace/ansible-examples/tree/master/LLDP-to-Graph.

Once we have the script to gather the operational data, we could hide it under **netlab** umbrella (for example, **netlab physical discovery** or similar).

ssasso:
Wonderful :-)

I forgot to mention that in my initial implementation I simply set the management address directly in the topology file:
```
---
provider: external

defaults:
  device: routeros
  bgp.as: 65000

module: [ ospf, bgp ]

nodes:
  r1:
    mgmt.ipv4: 10.20.30.50
  r2:
    bgp.originate: [ 100.64.0.0/24 ]
  r3:

links:
- r1:
    ifname: sfp-1
  r2:
- r1:
  ospf.area: 0.0.0.1
- r2:
  ospf.area: 0.0.0.2
- r1:
  r2:
  r3:
```

Btw, I'll give a look at *links* and *nodes* file to see how I can try to skip interface name generation & management address if they are already defined.

ipspace:
> Sh**t, I messed up with rebase :|

Pull down the latest dev, do `rebase -i dev` and deal with whatever crap Git is throwing at you ;) Merge conflicts inside interactive rebase are _interesting_ ;))

> Maybe we can keep this branch/PR "just for working on it", and then create a clean one and cherry pick (or simply port) the right commits only?

Rebase usually solves the problem, but of course you can always resubmit a PR.

ipspace:
BTW, I usually use `node.mgmt.pop('mac',None)` instead of the more complicated "check if a key exists, and delete it if it does". So much for Python being better than Perl ;))

ssasso:
1) Skip *mgmt* setup: I think it's safe to use https://github.com/ipspace/netsim-tools/pull/269/commits/a8b5a544a27b5fe9f58b160063042a814d7c4743

However, this is giving not-so-much benefits, it only skips "code call". The management IP addresses of the "other" nodes will be generated, in any case, based on *node id + start address*.

2) skip interface name generation when *ifname* is already defined: 

I guess the right place to do that is inside **add_node_interface** (here: https://github.com/ipspace/netsim-tools/blob/43e4c65e02ed0fd8145605f496e52420552ca8bf/netsim/augment/links.py#L128 ) .

With sth like:
```python
  if ifname_format and not 'ifname' in ifdata:
    ifdata.ifname = ifname_format % ifindex
```

Or did you have a specific idea on how/where to place that?

ipspace:
> 1. Skip _mgmt_ setup: I think it's safe to use [a8b5a54](https://github.com/ipspace/netsim-tools/commit/a8b5a544a27b5fe9f58b160063042a814d7c4743)
> 
> However, this is giving not-so-much benefits, it only skips "code call". The management IP addresses of the "other" nodes will be generated, in any case, based on _node id + start address_.

When using `external` provider I would expect all nodes to have mgmt IP already defined (or there's something wrong there). The cleanest way would be to add a provider callback early in the transformation process to check that. Let me add that after the merge.

> 2. skip interface name generation when _ifname_ is already defined:
> 
> I guess the right place to do that is inside **add_node_interface** (here:
> 
> https://github.com/ipspace/netsim-tools/blob/43e4c65e02ed0fd8145605f496e52420552ca8bf/netsim/augment/links.py#L128
> 
> ) .
> With sth like:
> 
> ```python
>   if ifname_format and not 'ifname' in ifdata:
>     ifdata.ifname = ifname_format % ifindex
> ```
> 
> Or did you have a specific idea on how/where to place that?

Yeah, that would be the easiest way to do it, so let's go with that. I might add a check whether static ifname is allowed later on (because we don't want people doing that in virtual labs).

PR #277, opened 2022-04-30
split should be used as .split for wider compat with Ansible

ddutt:
Signed-off-by: Dinesh Dutt <dd.ps4u@gmail.com>


ipspace:
Dang, I haven't noticed that you submitted the PR against master. No worries, I rebased the **dev** branch, so all's good.

PR #278, opened 2022-04-30
Support for additional loopback interfaces for VRF

ssasso:
It could be useful to have additional Loopback interfaces on the VRF scopes.

The feature uses the same tecnique of the VLAN module to define *virtual_interface* names/kind (i.e. LoopbackX), creating additional virtual links (type: *loopback*) - thus allowing "transparent" interface creation.

A new default address pool has been created for these interfaces.

Additionally, these loopback addresses are defined as *network* in the BGP sections.

A loopback interface creation was already addressed for VRF->OSPF part, so nothing was changed on OSPFv2 handling.

(In any case *redistribute connected* was already present both for BGP and OSPF - but for BGP I feel more "elegant" to announce the loopback with "network").

### DIFF
--- a/docs/module/vrf.md
+++ b/docs/module/vrf.md
@@ -14,15 +14,15 @@ This configuration module implements the VRF planning and configuration logic an
 (module-vrf-platform-support)=
 VRFs are supported on these platforms:
 
-| Operating system      | VRF<br />config | Route<br />leaking | VRF-aware<br />OSPF | VRF-aware<br />BGP |
+| Operating system      | VRF<br />config | Route<br />leaking | VRF-aware<br />OSPF | VRF-aware<br />BGP | VRF-aware<br />Loopback |
 | --------------------- | :-: | :-: | :-: | :-: |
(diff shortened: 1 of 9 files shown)


ssasso:
(if the module change looks good, I'll fix also the transformation tests)

ipspace:
Interesting idea, and I love the way you got it done just the right way. I do think it should be configurable though. Also, the Jinja2 filters look awful. If we're doing this in the module, we should also build a data structure that's easy-to-use in Jinja2.

Will add both things after merging your PR... just give me a bit more time, the coffee hasn't kicked in yet.

ipspace:
This one contained a nasty curveball :(

Making loopbacks optional was easy, but then the OSPF configuration template kept crashing. Turned out `ospf.area` was not set on the loopback interfaces because they were created too late in the transformation process (after node parameters were merged into interface parameters).

Anyway, fixed all that, cleaned up most of the templates, amended the commit and force-pushed it back to dev. Hope I got it all right; I was only able to test IOS and EOS implementations.

PR #282, opened 2022-05-07
Fix ansible install inside VENV

barajus:
Hi,
`netlab install ansible` is broken when launched inside a virtual env.
I checked the other bash script inside "installation" and they are not impacted since they don't call `pip`.

Cheers
Julien


barajus:
EDIT: First commit was a mistake. It is now okay, we test if the script is executed inside a VENV, regardless of $USER.
Also, I changed the return value of the if-statement above, to make it more standardized.

PR #286, opened 2022-05-12
Add initial VLAN and (start of) VRF lite support for SR Linux

jbemmel:
* Missing native vlan test case
* VRF only creates ip-vrf and adds interfaces; no BGP attributes yet


ipspace:
Nice job. Is this ready to be merged?

Issue #289, opened 2022-06-03
Routed VLAN subinterfaces on Vyos and OS10


ipspace:
0defd8e963a53ca6889d23e164f490199e974b41 added support for routed VLAN subinterfaces. @ssasso: It would be great to have them on VyOS and OS10 (they are already implemented on IOSv and EOS).

Any feedback on VLAN developer documentation https://netsim-tools.readthedocs.io/en/dev/dev/config/vlan.html would also be highly appreciated.

ssasso:
Working on the VyOS stuff, but I am on a blocking point:
When you work on formatting the `vlan_subif_name`, you are using `**parent_intf`, adding the attribute `subif_index`:
https://github.com/ipspace/netsim-tools/blob/0defd8e963a53ca6889d23e164f490199e974b41/netsim/modules/vlan.py#L687

However, this does not have the VLAN ID in it (or, at least, it's not easily accessible), which unfortunately is required for setting the right interface name in VyOS (subintf must match the vlan tag).

Quick and dirty workaround I'm using right now is:
```
    if 'trunk_id' in parent_intf.vlan:
      parent_intf.vlan_id = parent_intf.vlan.trunk_id[0]
```
which allows me to use:
```
vlan_subif_name: "{ifname}.{vlan_id}"
```

ssasso:
Additionally:

most probabily I am doing it in the wrong way on the topology file (maybe a full example like that could be useful on the documentation?):
```
---
module: [ vlan, vrf ]

vlans:
  red:
    vrf: red
  blue:
    vrf: blue
  green:
    vrf: green

vrfs:
  red:
  blue:
  green:

nodes:
  r1:
    device: vyos
    vlans:
      red:
        mode: route
      blue:
        mode: route
      green:
        mode: route
  s1:
    device: vyos
    vlans:
      red:
        mode: irb
      blue:
        mode: irb
      green:
        mode: irb

links:
- s1:
    vlan.trunk: [ red, blue, green ]
  r1:
    vlan.trunk: [ red, blue, green ]
```

BUT VRF data is not reported on the *r1* node (nor at the global node level, or at the interface level).

ipspace:
> However, this does not have the VLAN ID in it (or, at least, it's not easily accessible), which unfortunately is required for setting the right interface name in VyOS (subintf must match the vlan tag).

[...]

PR #298, opened 2022-06-04
VyOS: support for routed VLANs

ssasso:
Addresses #289

### DIFF
--- a/docs/dev/config/vlan.md
+++ b/docs/dev/config/vlan.md
@@ -70,7 +70,7 @@ devices:
     features:
       vlan:
         svi_interface_name: "br0.{vlan}"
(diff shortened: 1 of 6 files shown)


ipspace:
Thank you, merged.

I'm wondering if the `if vdata.mode is defined and vdata.mode != 'route'` is the correct condition to use when creating VLAN database, as it creates only VLANs for which `vdata.mode` is defined. It might be better to use `if vdata.mode|default('') != 'route'`.

You could use https://github.com/ipspace/netsim-examples/tree/master/VLAN/vlan-access-stretch and https://github.com/ipspace/netsim-examples/tree/master/VLAN/vlan-trunk to check whether the correct configurations are generated.

PR #299, opened 2022-06-04
RouterOS: Support for VLAN

ssasso:
@ipspace most probably this will have conflicts due to #298.
After that one has been merged, I'll rebase this.

### DIFF
--- a/docs/dev/config/vlan.md
+++ b/docs/dev/config/vlan.md
@@ -50,7 +50,7 @@ You have to specify VLAN-related capabilities of your device in `devices.<device
 
 [^SUBIF]: You can also use any other attribute from the parent interface, or attributes from the current interface (like `vlan.access_id`) that are not defined on the parent interface.
 
-The following VLAN features have been defined for Cisco IOSv, Arista EOS, VyOS, and Dell OS10:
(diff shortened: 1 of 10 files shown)


ipspace:
Looking at the VLAN code, it seems like it would work for trunks, but not for access VLAN interfaces, where you have to add the physical interface to a VLAN bridge.

Also, should I add code to generate a list of interfaces in a VRF so you don't have to do that in the Jinja2 template?

Alternatively, `interfaces|selectattr('vrf','equalto',vname)` might also work.

ssasso:
@ipspace you are totally right, I don't know why but in my mind I was assuming that the type *vlan_member* should have worked also for access ports.
I'll fix this.

For now I can survive also without the list of interfaces in the vrf structure.
However, it could be great if the attribute *parent_ifname* can be added to the *vlan_member* structure, in addition to *parent_ifindex*. If you are fine with that, I can add it on my own as part of this PR.

ipspace:
> However, it could be great if the attribute _parent_ifname_ can be added to the _vlan_member_ structure, in addition to _parent_ifindex_. If you are fine with that, I can add it on my own as part of this PR.

I have absolutely no problem with that. Thank you!

PR #318, opened 2022-07-09
Optional user specified custom ifindex per link node

jbemmel:
Many hardware platforms support different types of ports, with different speeds. For example, a Nokia 7220 IXR-D2 device (running SR Linux) has 48 25G ports and 8 100G ports.

The 100G ports are referenced as ethernet-1/49..56, and typical designs would use ports 1-48 for servers and 49-56 for spine links.

By default, netsim-tools would use ports 1, 2, 3, etc. This PR allows the user to pick port 48 for a spine facing link instead, by adding a custom 'ifindex' to override the auto-generated value


ipspace:
Nice one, thank you, the only thing I'm missing is a test case. I will add that after the merge and amend the commit.

PR #327, opened 2022-07-29
Initial VXLAN support on VyOS and Dell OS10

ssasso:
+ Test topology for EOS+VyOS+OS10 interworking.


ipspace:
Perfect. Thank you!

PR #329, opened 2022-08-01
Add L2 vxlan for SR Linux; requires evpn

jbemmel:
Addresses https://github.com/ipspace/netsim-tools/issues/324

On SR Linux, VXLAN requires EVPN; it might be more elegant to allow a per-platform override of the 'requires' attribute for a module, but for now I've added a 'requires_evpn' flag which is set to True on SR Linux

### DIFF
--- a/docs/module/vxlan.md
+++ b/docs/module/vxlan.md
@@ -29,6 +29,7 @@ The following table describes per-platform support of individual VXLAN features:
 | Cisco Nexus OS     | ✅  |
 | VyOS               | ✅  |
 | Dell OS10          | ✅  |
+| Nokia SR Linux     | ✅  |
 
 Notes:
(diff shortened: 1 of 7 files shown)


ipspace:
Thank you!

Per-platform 'requires' attribute sounds like an interesting addition. In the meantime, we should solve the challenge with device feature flags (`devices.srlinux.features.vxlan.requires: [ evpn ]` -- which might be just the right way to implement additional per-platform requirements), not with additional lab topology attributes. After all, there's nothing a user can do in the lab topology that would affect whether a device needs EVPN with VXLAN or not.

Will merge this (and the next) PR and make those changes when I get back home. Please give me a few more days.

ipspace:
Made the necessary adjustments directly in this branch and merged the changes. Thanks again!

PR #333, opened 2022-08-09
Model evpn.evi attribute in evpn module

jbemmel:
The EVPN instance is a 16-bit identifier used in EVPN services. This PR allows the user to define an explicit value, or leave it as default ( = vlan id )


ipspace:
Having a hierarchical key (**evpn.evi**)in **no_propagate** list won't work -- that list is checked against dictionary keys, and a hierarchical key is really a dict within a dict.

We could decide not to propagate any EVPN attributes from VLAN to interfaces, which probably makes sense. In that case, add **evpn** to **vlan_no_propagate** list.

PR #336, opened 2022-08-19
Fix VyOS new syntax for BGP System AS number

### LINKED ISSUE #334: VyOS BGP: Need to change BGP AS Number config template
... to reflect this: 
* https://phabricator.vyos.net/T4257
* https://github.com/vyos/vyos-1x/commit/967c53e2f3e4a7ae36577ba28dd60046909224dc

ssasso:
Fix #334

Tested with VyOS 1.4-rolling-202208100217

Please @ipspace consider adding the following to the next release changelog:
`This is required for working with VyOS 1.4 rolling > 20220801 (built after August, 1st 2022)`

### DIFF
--- a/netsim/ansible/templates/bgp/vyos.j2
+++ b/netsim/ansible/templates/bgp/vyos.j2
@@ -10,7 +10,7 @@ fi
 
 configure
 
-set protocols bgp local-as {{ bgp.as }}
+set protocols bgp system-as {{ bgp.as }}
 
 {% if bgp.router_id|ipv4 %}
 set protocols bgp parameters router-id {{ bgp.router_id }}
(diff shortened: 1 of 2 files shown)


ipspace:
> Please @ipspace consider adding the following to the next release changelog: `This is required for working with VyOS 1.4 rolling > 20220801 (built after August, 1st 2022)`

Will do. Just wondering: how does this fix work with old(er) versions of the code? What is the minimum VyOS version we're supporting now?

ipspace:
> Unfortunately VyOS 1.4 for now is a _rolling_ release with daily builds (or custom builds), so potentially they can introduce problems/regressions/change of syntax on any time.

Fail fast in network operating systems. Exactly what we need ;)

On a more serious note, it seems the recent change to the BGP template would only work with VyOS 1.4 or later. If that's the case, we should add that as a note to `caveats.md` (plus a link in `platforms.md`)

PR #337, opened 2022-08-19
VyOS: EVPN VXLAN initial implementation

ssasso:
VyOS: EVPN VXLAN initial implementation

Support for VLAN-based service only.

### DIFF
--- a/docs/module/evpn.md
+++ b/docs/module/evpn.md
@@ -27,6 +27,7 @@ The following table describes per-platform support of individual VXLAN features:
 | Nokia SR Linux     | ✅  | ✅  |  ❌  |  ❌  |  ❌  |
 | Nokia SR OS        |  ❌  |  ❌  |  ❌  |  ❌  |  ❌  |
 | FRR                |  ❌  |  ❌  |  ❌  |  ❌  |  ❌  |
+| VyOS               |  ✅  |  ✅  |  ❌  |  ❌  |  ❌  |
 
 EVPN module supports three design paradigms:
 
@@ -40,6 +41,7 @@ EVPN module supports three design paradigms:
 | Nokia SR Linux     |  ❌  | ✅  | ✅  | ✅  | ✅  |
 | Nokia SR OS        |  ❌  | ✅  | ✅  | ✅  | ✅  |
 | FRR                | ✅  | ✅  | ✅  | ✅  | ✅  |
+| VyOS               | ✅  | ✅  | ✅  | ✅  | ✅  |
 
 **Notes:**
 * FRR implement is a control-plane-only implementation that can be used as a route reflector. It enables EVPN over IPv4 and/or IPv6 on configured type(s) of BGP sessions. It's expected that the other end of the session won't negotiate EVPN or IPv4 AF.
--- a/netsim/ansible/templates/evpn/vyos.j2
+++ b/netsim/ansible/templates/evpn/vyos.j2
@@ -1,0 +1,32 @@
+#!/bin/vbash
+source /opt/vyatta/etc/functions/script-template
+
+if [ "$(id -g -n)" != 'vyattacfg' ] ; then
+    exec sg vyattacfg -c "/bin/vbash $(readlink -f $0) $@"
+fi
+
+# Configuration items start here
+configure
+
+# Configure Generic EVPN Stuff
+set protocols bgp address-family l2vpn-evpn advertise-svi-ip
+set protocols bgp address-family l2vpn-evpn advertise-all-vni
+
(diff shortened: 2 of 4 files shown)


ipspace:
Great job. From the documentation perspective, I don't think you could run IBGP-over-EBGP with the current VyOS implementation as you don't disable IPv4 address family over IBGP, and would thus get into some nasty routing loops... unless there's something going on that I don't understand.

Also, you're enabling EVPN on IPv4 BGP sessions, so I'm guessing IPv6 transport wouldn't work.

No worries though, we can fix that later (when you add IRB or VLAN bundles ;).

ssasso:
My bad on the v6. v6 on the control plane side should work (as for frr), I will fix that for loop.

Will also verify i-over-e.

ipspace:
> My bad on the v6. v6 on the control plane side should work (as for frr), I will fix that for loop.

Control plane should work, but would that result in a working data plane (= can VyOS do VXLAN to IPv6 NH)?

IIRC @jbemmel added v6 EVPN support to run SRv6, not VXLAN.

ipspace:
> Will also verify i-over-e.

You'll probably need your EBGP plugin to make that work (as you need **local-as** functionality on underlay EBGP session). If you make it work, please share the topology, I'd love to have an example (even though I have a "somewhat reserved" opinion about the IBGP-over-EBGP design).

ssasso:
> > My bad on the v6. v6 on the control plane side should work (as for frr), I will fix that for loop.
> 
> Control plane should work, but would that result in a working data plane (= can VyOS do VXLAN to IPv6 NH)?
> 
> IIRC @jbemmel added v6 EVPN support to run SRv6, not VXLAN.

The control plane works,
```
vagrant@s1:~$ show bgp l2vpn evpn summary
BGP router identifier 10.0.0.5, local AS number 65000 vrf-id 0
BGP table version 0
RIB entries 11, using 2112 bytes of memory
Peers 2, using 1447 KiB of memory

Neighbor        V         AS   MsgRcvd   MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd   PfxSnt Desc
10.0.0.6        4      65000        16        18        0    0    0 00:00:21            8        8 s2
2001:db8:0:6::1 4      65000        22        27        0    0    0 00:00:20            8        8 s2

Total number of neighbors 2
```

But for having it to work with an IPv6 dataplane as well, I'm trying to patch the module VXLAN to allow to use IPv6 as VTEP Address. I should be able to have a draft commit to show by today.

ssasso:
Please have a look at:
https://github.com/ssasso/netsim-tools/commit/b86a288e6da5355617f8db56817a00c191fb268d
and
https://github.com/ssasso/netsim-tools/commit/c94442e2ec955715711c39cb295ef14d38b7a959

If this is ok for you, I can submit a PR (after a rebase)

ssasso:
> > Will also verify i-over-e.
> 
> You'll probably need your EBGP plugin to make that work (as you need **local-as** functionality on underlay EBGP session). If you make it work, please share the topology, I'd love to have an example (even though I have a "somewhat reserved" opinion about the IBGP-over-EBGP design).

Here you go :-)
https://github.com/ssasso/netsim-tools/commit/d9498083bd9060d71c25252fb3035569f927168e

Will submit a PR after rebasing.

ipspace:
> If this is ok for you, I can submit a PR (after a rebase)

I will have to fix a few minor details, but it doesn't make sense to waste our time coordinating them. Please submit a PR and then I'll do a bit of editing.

> _edit:_ VyOS can't do VXLAN to IPv6 NH, the missing piece seems this one: [FRRouting/frr#5885](https://github.com/FRRouting/frr/issues/5885)
> 
> so most probably we should remove the support for IPv6-transport also from the FRR row.

As long as FRR doesn't change the next hop, it should be OK to use it as a RR, and the FRR container doesn't support VXLAN data plane anyway.

ipspace:
> > You'll probably need your EBGP plugin to make that work (as you need **local-as** functionality on underlay EBGP session). If you make it work, please share the topology, I'd love to have an example (even though I have a "somewhat reserved" opinion about the IBGP-over-EBGP design).
> 
> Here you go :-) [ssasso@d949808](https://github.com/ssasso/netsim-tools/commit/d9498083bd9060d71c25252fb3035569f927168e)

Nice job, please submit the PR.

For whatever reason I thought that your **ebgp.utils** plugin handles local-as as well. We should rename the **ebgp-local_as** plugin to **evpn.underlay_ebgp** or something similar as it really applies only to the IBGP-over-EBGP use case... or we could include it in the EVPN configuration module (cc @jbemmel)

PR #338, opened 2022-08-21
VyOS: Updated image info and caveats

ssasso:
Updated VyOS documentation for using latest image from vagrant hub, and add platform caveats based on the fact this is a rolling release.


ipspace:
Perfect. Thank you!

PR #345, opened 2022-08-22
VyOS: support for EVPN Symm IRB

ssasso:
Current limitation:
https://phabricator.vyos.net/T4636

(shoud be fixed soon, hopefully)

Will wait for merge and then rebase, before marking it as complete.


ssasso:
Issue on VyOS should be fixed, the fix should appear on the next build.
https://github.com/vyos/vyos-1x/pull/1484

PR #356, opened 2022-08-25
SR Linux vlan fixes

jbemmel:
Finally went through all the integration test cases under tests/integration/vlan and fixed all various corner cases.

Couple of notes:
* the vlan.mode attribute wasn't being propagated, this complicated the template logic
* consider not creating SVI interfaces for l2-only VLANs (could filter them out in each template, but easier to do once globally)
* trunk parent interfaces are detected when l.subif_index is defined, perhaps a type='trunk' (instead of p2p) would simplify things

### DIFF
--- a/netsim/ansible/templates/initial/srlinux.j2
+++ b/netsim/ansible/templates/initial/srlinux.j2
@@ -40,24 +40,27 @@ updates:
   val:
 {{  ip_addresses(loopback,False,True)  }}
 
-{% for l in interfaces|default([]) if (l.type=='lan' or l.vlan is not defined) %}
-- path: interface[name={{ l.ifname }}]
+{% for l in interfaces|default([]) if (l.type in ['lan','vlan_member'] or l.vlan is not defined) %}
+{% set if_name_index = l.ifname.split('.') %}
+{% set if_name = if_name_index[0] %}
+{% set if_index = if_name_index[1] if if_name_index|length > 1 else l.vlan.access_id|default(0) if l.vlan is defined else '0' %}
+{% set vlan = l.vlan.access|default(l.vlan.access_id|default('routed')|string()) if l.vlan is defined else l.ifname %}
+{% set if_desc = l.name|default( "vlan " + vlan )|replace('->','~')|regex_replace('[\\[\\]]','') %}
+- path: interface[name={{ if_name }}]
   val:
-{% if l.name is defined %}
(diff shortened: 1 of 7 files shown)


ipspace:
I understand the need for vlan.mode on VLAN interface, but calling the whole VLAN validation logic after the transformation has been done just to set an attribute might be trying to kill a fly with a cannonball.

Will merge the PR, revert vlan.py module, and fix those two bits.

ipspace:
Merged and added a fix that always sets vlan.mode (starting with system default value **irb**).

jbemmel:
The vlan-vrf-lite example now results in 2 interfaces with the same IP (see below):
* svi
* vlan_member

On SR Linux we don't need/use the svi interface in this case - only for mode==irb

```
- bridge_group: 1
  ifindex: 5
  ifname: ethernet-1/1.1000
  ipv4: 172.16.0.1/24
  name: r1 -> [h1,h2,r2,r3]
  neighbors:
  - ifname: eth1
    ipv4: 172.16.0.4/24
    node: h1
  - ifname: eth1
    ipv4: 172.16.0.5/24
    node: h2
  - ifname: ethernet-1/2.1000
    ipv4: 172.16.0.2/24
    node: r2
  - ifname: irb0.1000
    ipv4: 172.16.0.3/24
    node: r3
  parent_ifindex: 1
  parent_ifname: ethernet-1/1
  type: vlan_member
  virtual_interface: true
  vlan:
    access_id: 1000
    mode: route
  vrf: red
```

and
```
- bridge_group: 1
  ifindex: 6
  ifname: irb0.1000
  ipv4: 172.16.0.1/24
  name: VLAN red (1000) -> [h1,h2,r2,r3]
  neighbors:
  - ifname: eth1
    ipv4: 172.16.0.4/24
    node: h1
  - ifname: eth1
    ipv4: 172.16.0.5/24
    node: h2
  - ifname: ethernet-1/2.1000
    ipv4: 172.16.0.2/24
    node: r2
  - ifname: irb0.1000
    ipv4: 172.16.0.3/24
    node: r3
  type: svi
  virtual_interface: true
  vlan:
    mode: route
  vrf: red
```

ipspace:
I had a nasty feeling that last change would break something, I just didn't have the test case to catch it :( Thank you for catching it!

The breaking change was setting `topology.defaults.vlan.mode` to `irb` which means that the `vlan.mode` precedence is wrong somewhere in the code. Will fix...

ipspace:
Fixed in 41f5154f34ea925b9320e13b234c539067919277

PR #357, opened 2022-08-25
Trim svi neighbor list to include only direct neighbors

jbemmel:
Fixes https://github.com/ipspace/netsim-tools/issues/325

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -629,10 +629,10 @@ def set_svi_neighbor_list(node: Box, topology: Box) -> None:
         if len(vlan_data.neighbors) == vlan_data.host_count +1:             # ... and there's exactly one non-host attached to the VLAN
           ifdata.role = 'stub'                                              # ... then we have a stub link, mark it for IGP modules
 
-      # JvB: Build a list of neighbors for the parent_interface
-      parent_neighbors = [ n.node for n in node.interfaces[ifdata.parent_ifindex-1].neighbors ]
+      # JvB: Build a list of neighbors for the parent_interface, if available
+      parent_neighbors = [ n.node for n in node.interfaces[ifdata.parent_ifindex-1].neighbors ] if 'parent_ifindex' in ifdata else []
 
-      ifdata.neighbors = [ n for n in vlan_data.neighbors if n.node != node.name and n.node in parent_neighbors ]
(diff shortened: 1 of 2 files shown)


jbemmel:
This may be trimming too much, reading back the issue it is more about interfaces being part of different VRFs?

ipspace:
Thanks a million for trying to solve this one, but it's more complex. We need full list of neighbors on L2 VLANs and just the local neighbors if all attached VLAN interfaces are L3 interfaces (example: multi-hop VRF Lite).

Will try to solve it in the next days, but the answer will never be precise without reinventing SPF, so maybe I'll just give up and do that. It's not like we'd be dealing with huge data sets.

jbemmel:
The patch does fix the multi-hop VRF lite case:
```
- bridge_group: 2
  ifindex: 4
  ifname: ethernet-1/1.1
  ipv4: 172.16.6.1/24
  name: r1 -> [r2]
  neighbors:
  - ifname: ethernet-1/2.1
    ipv4: 172.16.8.2/24
    node: r2
  parent_ifindex: 1
  parent_ifname: ethernet-1/1
  type: vlan_member
  virtual_interface: true
  vlan:
    access_id: 1001
    mode: route
  vrf: blue
```

Is it "all attached VLAN interfaces" or only if the direct peer is routed?

ipspace:
> The patch does fix the multi-hop VRF lite case:

That is true, but it returns wrong results for the "routed VLAN interfaces attached to a bridged network" case like `[tests/topology/expected/vlan-routed-vrf.yml](https://github.com/ipspace/netsim-tools/pull/357/files#diff-e37f3b915b08d73533c79191066bdc7276a8a8bfa26c27764a1d787517f96496)`.

In that test case, S1 and S2 are bridges, so R1 and R2 are neighbors on the same VLAN -- imagine what would happen if you were running OSPF.

jbemmel:
Actually, none of the templates currently rely on the neighbors per interface. I only found
```
mpls/eos.6pe.j2: {% for l in interfaces|default([]) if l.type != 'stub' and l.neighbors is defined and not (l.ipv6|default(False)) %}
```

Perhaps an easier fix would be to remove the attribute altogether?

ipspace:
We need link neighbors to set up EBGP sessions and to find a viable next hop for static routes on Linux hosts

ipspace:
This has been fixed by [91047b6](https://github.com/ipspace/netlab/commit/91047b6eb24bad45a24aa5f32e38cc4c8bb50601)

Issue #354, opened 2022-08-24
Add advertise_default bgp attribute (global, node, interface levels)


jbemmel:
At global, node and interface levels, I'd like to add an 'advertise_default' flag that - if set - causes the node to advertise a default route.
Similar to advertise_loopback

I've considered allowing 'originate' at the interface level for this purpose; it's more generic, but the resulting export policy logic quickly gets complicated

jbemmel:
In EOS this maps to "neighbor default-originate" (for example)

ipspace:
@ssasso: This seems like a perfect addition to ebgp.utils. Will you add it or should I?

ssasso:
I will do it! Please assign the task to me,
(give me a couple of days)

ssasso:
Question: I guess we'd like to handle both the *default-originate* and the *default-originate always*, isn't it?

jbemmel:
Yes, whatever makes sense. We could do
```
default_originate: True
```
and
```
default_originate: always
```
to support both cases

ipspace:
> Question: I guess we'd like to handle both the _default-originate_ and the _default-originate always_, isn't it?

The BGP implementations I've seen had only "default-information originate" which would unconditionally advertise the default route over a BGP session regardless of what the router had in its routing table. If there's a default route in the BGP table, it's advertised no matter what.

"default-information originate always" is usually used in OSPF meaning "advertise default route in OSPF even if the router itself does not have a default route".

I'm positive (based on recent experience 😜) there must be vendors out there that are doing totally different things...

jbemmel:
Arista has both options: https://kevin-wang-xin.medium.com/arista-eos-bgp-neighbor-default-originate-always-85effce07933
```
[ no | default ] neighbor ( ip-address | peer-group-name ) default-originate [ route-map map-name ] [ always ]
```
always: Enforce the advertisement of a fabricated default route, irrespective of availability of other default routes.

ssasso:
The current initial implementation I did simply used the *default-originate*.

Then I discovered Arista also has the *always* option :-D 

See: https://github.com/ssasso/netsim-tools/commit/b99909ab83e07942fc38c71b4c33915c647f5936

ipspace:
> Yes, whatever makes sense. We could do
> 
> ```
> default_originate: True
> ```
> 
> and
> 
> ```
> default_originate: always
> ```
> 
> to support both cases

[...]

PR #358, opened 2022-08-26
Support for BGP default-originate

ssasso:
Implement BGP *default-originate [always]* in `ebgp.utils` module.

Implements #354

Please @ipspace can you test the Cisco IOS part?


ipspace:
The Cisco IOS part looks syntactically correct (will set up a lab to test it), but the sequence of the configuration commands with so many changes in address families might break Ansible's "let's minimize the config changes" logic. Will try to rewrite it.

ipspace:
Restructured the Cisco IOS template in e3688ab169b5875ccad6d74fcd8a9c9a0bcabf4f... and then figured out we did something very similar for Arista EOS. Oh well, maybe I should look around first the next time.

PR #362, opened 2022-08-27
Ebgp unnumbered enhancements

jbemmel:
This example is pushing the limits of what one might setup in a test topology, I wanted to see if I could do EBGP peering locally to leak routes between VRFs.

To make this work, I had to extend neighbor lists with module specific interface attributes ( bgp.underlay_as in this case )

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.j2
+++ b/netsim/ansible/templates/bgp/srlinux.j2
@@ -9,5 +9,4 @@ updates:
     accept: {}
 
 {% from "srlinux.macro.j2" import bgp_config with context %}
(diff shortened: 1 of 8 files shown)


jbemmel:
I tried to keep the module attribute copying generic; the 'vrf' module attribute isn't strictly needed, but it gets included as the vrf module defines an interface level attribute

ipspace:
> Alternatively, we could introduce a 'neighbor_interface' category of attributes, containing the list of attributes to include in neighbor lists

No worries, it's perfectly OK to have remote interface attributes available in the neighbor list. Might come useful in other scenarios as well (example: check OSPF areas).

Issue #369, opened 2022-09-01
Replacing ebgp-local_as plugin with BGP attributes on Nokia SR OS, Dell OS10, VyOS


ipspace:
The recent additions to BGP attributes (#366 and #365) allow us to implement IBGP-over-EBGP EVPN design without any extra evpn attributes or plugins -- the test case is here https://github.com/ipspace/netlab/blob/dev/tests/integration/evpn/vxlan-bridging-ibgp-over-ebgp.yml

It would probably make sense to migrate the current implementations on Nokia SR Linux, Dell OS10 and VyOS to the new attributes, all you have to do is to implement activation of IPv4 session based on **activate** neighbor attribute (see #366). No rush though, I can mark the platforms using the "old" method in the documentation before pushing out the next release.

jbemmel:
PR https://github.com/ipspace/netlab/pull/372 adds srlinux support

PR #371, opened 2022-09-01
srlinux bgp refactoring

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/369

* Support bgp.local_as, bgp.sessions, bgp.activate, bgp.replace_global_as
* Remove bgp.address_families from srlinux (TBD sros)
* Patch bgp module to also allow ipv4 af activation in an ipv6-only transport session (we may want to change how this is implemented/flagged)


jbemmel:
For example, we could do this:
```
BGP_DEFAULT_SESSIONS: typing.Final[dict] = {
  'ipv4': [ 'ibgp', 'ebgp', 'localas_ibgp' ],
  'ipv6': [ 'ibgp', 'ebgp', 'localas_ibgp' ],

  'ipv4-over-ipv6': [ 'ebgp' ],
}
```

PR #380, opened 2022-09-03
Add plugin to configure ipv4-over-ipv6 in ebgp sessions

jbemmel:
Use case: EVPN with IPv4 prefixes over an EBGP v6 unnumbered underlay fabric

### DIFF
--- a/netsim/extra/ebgp.ipv4-over-ipv6-lla.py
+++ b/netsim/extra/ebgp.ipv4-over-ipv6-lla.py
@@ -1,0 +1,23 @@
+#
+# Simple module to activate ipv4 address family for ipv6 unnumbered eBGP peers
+#
+# Changes:
+# * Sets bgp.activate['ipv4'] = True
+#
+import typing, netaddr
+from box import Box
+from netsim import common
+
+"""
+activate_ipv4_over_ipv6: Update ebgp neighbors with ipv6 lla to enable ipv4
+"""
+def activate_ipv4_over_ipv6(node: Box) -> None:
+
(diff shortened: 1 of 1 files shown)


ipspace:
Just to be on the safe side, please rebase this against the current dev branch. Thank you!

ipspace:
Also, I have no problem merging this, but it's another weird combo that's probably platform specific.

The "usual" way of enabling IPv4+IPv6 EBGP sessions over an unnumbered interface would be to set **unnumbered: true**, which would indicate that we want to use IPv4 _and_ IPv6 over that interface, we just don't have any addresses. If this does not trigger the activation of IPv4 and IPv6 sessions, then we have a bug to fix.

What you seem to be doing here is forcing an IPv4 session when there's no IPv4 address on the interface. You can get that configured almost anywhere, but in many cases it won't work because a properly-implemented layer-3 device refuses to process protocol X packets coming through an interface on which protocol X is not enabled. There are probably exceptions, but should that be the case, I'll try to stay polite and refrain from commenting.

jbemmel:
I guess ipv4: True might be a better way to model this. It’s not quite the same as ‘traditional’ ipv4 unnumbered as implemented by Cisco and others, there is no association with a particular loopback ip. Ipv4 forwarding is simply enabled for all interfaces in an ip-vrf, even if they don’t have an IPv4 address assigned.

Also, one could have an ipv6-only topology where ipv4 forwarding remains disabled. For some reason I thought ‘unnumbered: True’ was the ‘old’ way of specifying this, and now we have ipv4/ipv6: True to reflect that

ipspace:
This is probably not needed once we complete #381. I also suspect that everything would work today if you'd use **unnumbered** link/addressing pool attribute.

jbemmel:
No longer needed

PR #404, opened 2022-09-08
frr add vlan vxlan vrf support

jbemmel:
Note: Have not extensively tested all the various VLAN cases, this only adds a basic config for trunk ports such that VRF support can be demonstrated

* VRF module: add a features.vrf.keep_module flag to signal that we want to keep the module, even if there are no interfaces with VRFs associated. Use case is VXLAN overlay VRFs

I have this working in a setup with leaf/spine and mixed frr nodes - some on trunk vlans, some on vxlan overlay vrfs

TODO:
* Route target / Route Distinguisher not taken from data model yet

### DIFF
--- a/docs/dev/config/vlan.md
+++ b/docs/dev/config/vlan.md
@@ -73,7 +73,7 @@ devices:
     features:
       vlan:
         model: router
-        svi_interface_name: "{ifname}.{vlan}"
+        svi_interface_name: "vlan{vlan}"
         subif_name: "{ifname}.{subif_index}"
   vyos:
     features:
--- a/docs/module/evpn.md
+++ b/docs/module/evpn.md
@@ -23,7 +23,7 @@ The following table describes per-platform support of individual VXLAN features:
 | Operating system   | VXLAN<br>transport | VLAN-based<br>service | VLAN Bundle<br>service | Asymmetric<br>IRB | Symmetric<br>IRB |
 | ------------------ | :-: | :-: | :-: | :-: | :-: |
 | Arista EOS         | ✅  | ✅  | ✅  |  ❌  | ✅  |
-| Nokia SR Linux     | ✅  | ✅  |  ❌  |  ❌  |  ❌  |
+| Nokia SR Linux     | ✅  | ✅  |  ❌  |  ❌  |  ✅  |
 | Nokia SR OS        |  ❌  |  ❌  |  ❌  |  ❌  |  ❌  |
 | FRR                |  ❌  |  ❌  |  ❌  |  ❌  |  ❌  |
 | VyOS               |  ✅  |  ✅  |  ❌  |  ❌  |  ✅  |
@@ -56,7 +56,7 @@ EVPN module supports these default/global/node parameters:
 
 * **evpn.session** (global or node parameter): A list of BGP session types on which the EVPN address family is enabled (default: `ibgp`)
 * **evpn.vlan_bundle_service** (global or node parameter): Use VLAN bundle service for VLANs within a VRF (default: `False`)
-* **evpn.start_transit_vni** (system default parameter) -- the first symmetric IRB transit VNI
+* **evpn.start_transit_vni** (system default parameter) -- the first symmetric IRB transit VNI, range 4096..16777215
 
 ### VLAN-Based Service Parameters
 
@@ -85,6 +85,6 @@ The default value of VRF EVPN Instance identifier is the VLAN ID of the first VL
 IRB is configured whenever EVPN-enabled VLANs in a VRF contain IPv4 or IPv6 addresses:
 
 * Asymmetric IRB requires no extra parameters[^NS]
-* Symmetric IRB needs a transit VNI that has to be set with the **evpn.transit_vni** parameter. That parameter could be set to an integer value or to *True* in which case the EVPN configuration module assigns a VNI to the VRF.
+* Symmetric IRB needs a transit VNI that has to be set with the **evpn.transit_vni** parameter. This parameter could be set to an integer value or to *True* in which case the EVPN configuration module auto-assigns a VNI to the VRF. Note that the EVI value used in this case is currently based on the VRF ID (vrfidx)
 
 [^NS]: Asymmetric IRB is not supported at the moment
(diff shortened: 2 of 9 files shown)


ipspace (commented):
* While we could model a Linux host as a router or a bridge, your use case and the example you're using as the inspiration for your configs strongly suggest "l3-switch"
* I have no idea why you think FRR needs EVPN for VXLAN to work. Cumulus Linux doesn't.
* There probably is no need to name VRFs something-vrf, you got that from an example. All other platforms use just the VRF name, if at all possible, please do the same.
* The only change in initial/frr.j2 is a block of code that is commented out. Unless there's a really good reason for that, please remove it.
* If you're adding a new feature for a new platform, please do it properly. Use Linux bridges to implement VLANs. Having a VLAN on a single interface might build you the lab you need, but it won't make it into the dev/master branch.
* If you're adding VRF support, please do it properly. Moving VLAN interfaces but not physical interfaces into VRFs makes absolutely no sense.
* You should create all interfaces (including VRF loopbacks) in the initial configuration. Just because FRR doesn't throw a fit when you're trying to add an IP address to an unknown interface doesn't mean that it's cool. I learned the hard way how tough it is to find FRR-related problems :(
* You're unconditionally creating a BGP routing process in a VRF. It might be needed, but probably doesn't have to be created without some underlying good reason (example: BGP neighbors or route leaking)
* You're creating bridge interface together with VXLAN interface. That wouldn't be necessary if you did VLAN configuration correctly. L3VNI is obviously an exception.

On a positive side:
* The changes to vrf.py seem OK, although I still have a feeling they are not needed.

General comments:
* Network namespaces are not an alternative to VRFs. I did a podcast with the Cumulus team on that topic a long while ago if you want to know the technical details.

jbemmel:
> * I have no idea why you think FRR needs EVPN for VXLAN to work. Cumulus Linux doesn't.
Copied from srlinux and driven by the use case I was focusing on - I will remove it
> * There probably is no need to name VRFs something-vrf, you got that from an example. All other platforms use just the VRF name, if at all possible, please do the same.
Based on an example, but I also figured someone might call their VRF 'eth1' and cause issues. Plus, this adds a lot of devices on Linux, and having an easy way to tell which ones are 'vrf' devices can help. I started by using the pure vrf name in vtysh - but the name has to match with the device/table. I can change it back

Should we add a regex check for VPN names? e.g. [a-zA-Z0-9-_.]{1..16}

> * The only change in initial/frr.j2 is a block of code that is commented out. Unless there's a really good reason for that, please remove it.
Just to document something I tried and left out - will remove
> * If you're adding a new feature for a new platform, please do it properly. Use Linux bridges to implement VLANs. Having a VLAN on a single interface might build you the lab you need, but it won't make it into the dev/master branch.
ok
> * If you're adding VRF support, please do it properly. Moving VLAN interfaces but not physical interfaces into VRFs makes absolutely no sense.
ok
> * You should create all interfaces (including VRF loopbacks) in the initial configuration. Just because FRR doesn't throw a fit when you're trying to add an IP address to an unknown interface doesn't mean that it's cool. I learned the hard way how tough it is to find FRR-related problems :(
ok
> * You're unconditionally creating a BGP routing process in a VRF. It might be needed, but probably doesn't have to be created without some underlying good reason (example: BGP neighbors or route leaking)
ok
> * You're creating bridge interface together with VXLAN interface. That wouldn't be necessary if you did VLAN configuration correctly. L3VNI is obviously an exception.
ok
>

I'll admit I took a few shortcuts to get a MVL (Minimal Viable Lab) working - will polish it up

> On a positive side:
> 
> * The changes to vrf.py seem OK, although I still have a feeling they are not needed.
> 
> General comments:
> 
> * Network namespaces are not an alternative to VRFs. I did a podcast with the Cumulus team on that topic a long while ago if you want to know the technical details.
Good to know

ipspace:
> > * There probably is no need to name VRFs something-vrf, you got that from an example. All other platforms use just the VRF name, if at all possible, please do the same.
> 
> Based on an example, but I also figured someone might call their VRF 'eth1' and cause issues. Plus, this adds a lot of devices on Linux, and having an easy way to tell which ones are 'vrf' devices can help. I started by using the pure vrf name in vtysh - but the name has to match with the device/table. I can change it back

Interesting conundrum, we can't win either way. Do whatever feels more appropriate, but we have to document it. Traditional networking engineers might get confused because the VRF names are different (using various **show something vrf** commands) or because Ansible playbook throws weird errors.

BTW, do we use `set +e` at the beginning of every Bash script to make sure errors are caught and reported? Still reliving the FRR nightmare.

> Should we add a regex check for VPN names? e.g. [a-zA-Z0-9-_.]{1..16}

Absolutely. Thank you!

> I'll admit I took a few shortcuts to get a MVL (Minimal Viable Lab) working - will polish it up

Thanks a million. It would be great to have full functionality in FRR container, and it can be (probably) copied directly into Cumulus VX.

ipspace (changes requested):
Much better, still a few gotchas:

initial/frr.j2
* Please remove the commented-out block
* FRR configuration iterates over all interfaces, which includes interfaces that are not created yet. It might work now (because FRR treats errors as "polite opinions"), but I'm pretty positive we'd get bitten by that sooner or later.

vlan/frr.j2
* It might be cleaner to iterate over vlan_member and then over svi interfaces
* I can't see where you're creating a Linux bridge for individual VLANs. Also, where do you connect individual VLAN subinterfaces to the Linux bridge? What am I missing? Is there some FRR magic going on behind the scenes?
* Unnumbered IPv4 is configured by setting the interface IP address to loopback IP address (if it exists)

vrf/frr.j2
* You're still not moving physical interfaces into VRF. Is there a reason for that?

vxlan/frr.j2
* VLAN bridge must be created in VLAN module, not in VXLAN module. VLAN bridging should work without VXLAN.
* I would name VLAN bridge "vlan{vlan.id}" not "br{vlan.id}"

Finally, it looks like you should do a rebase against the dev branch. Changes from the other PR seem to be in this one as well.

ipspace:
Ah, sorry, I thought you were done with the changes. Set it to draft status, when you're done please change it back. Thank you!

ipspace:
> Got lost in the forest of pending branches - still working on this one

You are in a maze of twisty little branches, all alike 🤣

PR #405, opened 2022-09-08
vxlan auto pull vlans and vrfs

jbemmel:
This adds a 'node_pre_default' hook in the vxlan module (first such method in the codebase) to automatically pull in any vlans or vrfs used in vxlan.vlans

This allows a user to simply declare a top-level vxlan.vlans: [ red, blue ] statement, without also having to declare referenced vlans and vrfs on those nodes.

This is particularly useful in overlay scenarios, where the VRFs only exist on the nodes without physical interfaces attached


ipspace (changes requested):
Nope.

* VXLAN module requires VLAN module, so there's no way VLAN module would NOT be applied. The reason for that particular error message is that you can disable VNI auto-generation, but still list a VLAN in the node.vxlan.vlans list.
* There is no need to pull global VLANs and VRFs into a node. Creating VLANs and VRFs should be based on actual needs, which are interface-driven.

I don't think we need this PR at all.

jbemmel:
I still believe there are uses for being able to say that you want to create certain VRFs and VLANs on a bunch of nodes, without having interfaces attached. Those interfaces may come later, when servers are connected or VMs are spun up.

This effect can be achieved with loopbacks or stub interfaces, but then you have to create/define those everywhere.

OK to close this for now

ipspace:
> I still believe there are uses for being able to say that you want to create certain VRFs and VLANs on a bunch of nodes, without having interfaces attached. Those interfaces may come later, when servers are connected or VMs are spun up.

If someone wants to use _netlab_ in that way (not that it was ever intended to be used like that), then I can add **deploy_to** attribute to VLANs and VRFs.

However, I would love to see someone coming along and saying "I'd love to do this particular thingy" first, otherwise we'll get into an ever-expanding forest of cool-to-have-but-never-used nerd knobs.

PR #406, opened 2022-09-08
srlinux refactor templates and add vxlan l3 vrf support

jbemmel:
* Add l3 vxlan support using vrf module; configure vxlan after vrf
* Add support for loopbacks in vrfs
* Refactored templates: Move vlan handling to vlan module

Removed evpn.transit_evi, now using vrfidx as suggested. May change to use reserved vlan id in the future

### DIFF
--- a/docs/dev/config/vlan.md
+++ b/docs/dev/config/vlan.md
@@ -73,7 +73,7 @@ devices:
     features:
       vlan:
         model: router
-        svi_interface_name: "{ifname}.{vlan}"
+        svi_interface_name: "irb0.{vlan}"
         subif_name: "{ifname}.{subif_index}"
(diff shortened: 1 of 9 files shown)


ipspace:
I don't think the changes to EVPN module are needed. Why can't you use **vrfidx** as IRB EVI?

jbemmel:
We could, and I also believe some platforms (Nexus?) require a special VLAN id to be reserved for l3 vnis - we could use that as well.

Will revert the changes, add the bug fix/range check, and resubmit

ipspace:
> We could, and I also believe some platforms (Nexus?) require a special VLAN id to be reserved for l3 vnis

I wouldn't be surprised.

PR #409, opened 2022-09-08
Fix links to .md files with anchor

steffann:


jbemmel:
Should this be applied to dev instead?

ipspace:
That would have been better, but I do a rebase before merging anyway, so it doesn't hurt. I do wonder how @steffann managed to submit a PR against master branch though ;)

steffann:
Good to know, thanks!

ipspace:
> Good to know, thanks!

Thanks for becoming a contributor... and no worries, that thing would be impossible to know.

PR #414, opened 2022-09-09
Fix check for type(vni) != "int"

jbemmel:
In Python, bool is a subtype of int


ipspace (approved):
Reminds me of one of my "go home Python, you're drunk" tweets. This is now becoming "go home, mypy, you're drunk".

Issue #418, opened 2022-09-09
Inconsistency in vlan-routed-trunk test case


jbemmel:
https://github.com/ipspace/netlab/blob/1f7287d33fb70f7fc089bef90a4e9a7c3b0ea22f/tests/integration/vlan/vlan-routed-trunk.yml#L29

vlan.mode is set to 'route', but in the generated topology for 'ros' we get:
```
vlan:
  max_bridge_group: 2
  mode: route
vlans:
  blue:
    bridge_group: 1
    id: 1001
    mode: irb   <--- !!
```

Similarly, the svi interfaces contain ```mode: irb```

ipspace:
Yeah, that's the end result of **vlan.mode: irb** in system settings. Going for another trip into that snake pit :(

Thanks for catching this one!

jbemmel:
It happens here: https://github.com/ipspace/netlab/blob/dev/netsim/modules/vlan.py#L162

That method is first called for the global vlans list - it adds ```mode: irb```
Then when the code runs for the node with ```mode: route``` set, it looks like the user provided ```mode: irb```

PR #419, opened 2022-09-09
Fix routed trunk vlan mode

jbemmel:
* Not too early, when traversing global topology
* Fill it in when creating per-node vlan dict (if not set)

Fixes https://github.com/ipspace/netlab/issues/418

Note that the pattern of getting a parameter from node / if not from topology / if not default value seems generic, could add a helper method in data

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -885,7 +885,6 @@ class VLAN(_Module):
           node.vlans[vname] = topology.vlans[vname] + node.vlans[vname]
 
     validate_vlan_attributes(node,topology)
-    print( f"JvB after validate_vlan_attributes: {node.get('vlans',[])}" )
 
   def link_pre_transform(self, link: Box, topology: Box) -> None:
     if link.get('type','') == 'vlan_member':                                      # Skip VLAN member links, we've been there...
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -158,7 +158,7 @@ def validate_vlan_attributes(obj: Box, topology: Box) -> None:
           common.IncorrectValue,
           'vlan')
     else:
-      if default_fwd_mode:                                          # Propagate default VLAN forwarding mode if needed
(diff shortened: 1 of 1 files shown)


ipspace (approved):
This looks almost exactly what's needed. The test cases are failing (no surprise there), so I'll merge this, then figure out why the test cases are failing, and add router-on-a-stick test case (because obviously we need it ;)

jbemmel:
Actually, I think the test cases are failing because the fix is incomplete - there are some corner cases not covered. I'll try to isolate and create a smaller regression test case

ipspace:
> Note that the pattern of getting a parameter from node / if not from topology / if not default value seems generic, could add a helper method in data

That's why I already added get_global_parameter, if we stumble upon a use case where we need the full three-level inheritance, we'll add another helper method.

BTW, this is not that use case, because we have to treat global VLANs differently from node VLANs.

Also, we need the "manual" inheritance only very early on the the life of a module, after the early hooks the core transformation routine takes care of proper inheritance. We do have to run the early hooks first otherwise modules can't influence each other (example: EVPN setting VXLAN flooding mode).

ipspace:
> Actually, I think the test cases are failing because the fix is incomplete - there are some corner cases not covered. I'll try to isolate and create a smaller regression test case

Nope, it's just that **mode: irb** is missing in a number of places. I have to check whether that's only in global VLANs -- in that case we really don't care because they never get propagated to Ansible inventory.

ipspace:
Done. Force-pushed the amended commit.

Thanks a million for fixing this!!!

ipspace:
> > Note that the pattern of getting a parameter from node / if not from topology / if not default value seems generic, could add a helper method in data
> 
> That's why I already added get_global_parameter, if we stumble upon a use case where we need the full three-level inheritance, we'll add another helper method.

Just realized we also have `get_effective_module_attribute`. This thing is getting too big to handle mentally :(

PR #420, opened 2022-09-09
Use 'pop' to avoid exception when the attribute is not present

jbemmel:


ipspace (approved):
Interesting. I guess this happened when we had a VRF module on a node that had no need for a VRF. Will add that to one of the test cases.

PR #422, opened 2022-09-09
I'm trying to solve this problem with vlan.mode not being set, I was expecting it to be 'irb' - solved now

jbemmel:
In this test case, it should be 'bridge' (reflecting the value from the global vlan)

### DIFF
--- a/tests/topology/expected/vlan-bridge-trunk-router.yml
+++ b/tests/topology/expected/vlan-bridge-trunk-router.yml
@@ -253,6 +253,7 @@ nodes:
       vlan:
         access: red
         access_id: 1000
+        mode: bridge
     - bridge: input_3
       ifindex: 3
       ifname: Ethernet3
@@ -261,6 +262,7 @@ nodes:
       vlan:
         access: blue
         access_id: 1001
+        mode: bridge
     - bridge_group: 1
       ifindex: 6
       ifname: Vlan1000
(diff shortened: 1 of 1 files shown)


ipspace:
I don't get it. Why is this a PR? Are you doing test-driven development (creating the test results and then tweaking the code until they match)?

ipspace:
Also, why do you expect **vlan.mode** to be set on a switch interface? It has access VLAN, which is enough to configure the interface. Whether **vlan.mode** is **irb** or **bridge** does not matter for a switch, all **vlan.mode** does in this case is determine whether the IP addresses are set on the SVI/VLAN interface.

jbemmel:
For things that are not a switch, the logic gets complicated: https://github.com/ipspace/netlab/pull/423/files#diff-6bde79138c62989073790e0a32aa123a2c2f4612940e2b6859929e7c1101a6d9R60

I suppose in part this is due to me not following the mental model behind it. I look for consistency (like "mode" is always set) and try to build logic around that; the missing ```mode: bridge``` led to some wrong branching (now avoided)

I submitted a PR because I figured it would be easier to discuss - you can accept it, or reject it and discuss.

ipspace:
OK, I see you challenge. You would have made everyone's life easier if you'd started with that (it's always best to start with "I'm trying to solve this problem and this is the roadblock I'm facing" instead of "the tool should behave this way").

I'm pretty sure the flag you need is there (because I have the same issue on Cisco IOS), let me figure it out and update

https://netsim-tools.readthedocs.io/en/latest/dev/config/vlan.html

PR #423, opened 2022-09-09
srlinux vlan full coverage

jbemmel:
* add native vlan support
* add test case for all combinations

Greatly simplified and cleaned up vlan provisioning logic, much improved


ipspace:
There should be a vlan.native flag to avoid that convoluted logic. See

https://netsim-tools.readthedocs.io/en/latest/dev/config/vlan.html#interface-vlan-parameters
https://netsim-tools.readthedocs.io/en/latest/dev/config/vlan.html#router-like-platforms

If that flag is missing, then we need to add it -- there must be a way to find a native VLAN on an interface without too-convoluted logic.

Please confirm whether that flag is missing (open an issue and I'll fix it) or whether SR Linux does things in a different way that requires the logic above (in which case I'll just merge this PR).

jbemmel:
It's there in the trunk port (see https://github.com/ipspace/netlab/blob/dev/tests/topology/expected/vlan-native-routed.yml#L14) but not in the individual vlan_member (in case of mode=bridge/irb)

If we could mark the native vlan_member with ```native: True``` then the logic becomes much easier

ipspace:
I'm confused -- I got a dozen emails with new commits after the PR has been opened, and another one after it was marked as "ready for review". Is this ready to be merged?

Also, would it be possible to submit a PR when you're done with your work? Trying to figure out when you're done or reviewing work-in-progress (like what happened with the FRR commit) creates unnecessary friction. Thank you!

jbemmel:
I discovered some breaking issues after having submitted the PR, and since it was still pending during your night time I figured it was better to add commits, rather than wait or submit a different PR

In general I try to fix everything before submitting, sometimes that doesn’t work out

This one is ready now

Issue #424, opened 2022-09-09
Prefix parsing error; ipv4/ipv6 somehow become strings


jbemmel:
Minimal topology:
```
provider: clab

defaults:
 device: srlinux
 devices:
  srlinux.image: ghcr.io/nokia/srlinux:22.6.3

addressing:
  bgp-unnumbered:
    ipv6: True  # Unnumbered
    ipv4: True

vlans:
  vxlan:
    mode: irb
    role: bgp-unnumbered # Triggers parsing exception, prefix ipv4/ipv6 representing "true" as string

groups:
  fabric:
    members: [ leaf1a ]
    device: srlinux
    module: [ bgp,evpn,vlan,vrf,vxlan ]

nodes:
  leaf1a:
    bgp.local_as: 65001

  h1:
    bgp.local_as: 65099


links:

- leaf1a:
  h1:
  vlan.trunk: [ vxlan ]
```

Result:
```
Process LAN link {'linkindex': 2, 'parentindex': 1, 'vlan': {'access': 'vxlan'}, 'vlan_name': 'vxlan', 'type': 'vlan_member', 'interfaces': [{'node': 'leaf1a', 'vlan': {'access': 'vxlan'}}], 'prefix': {'ipv4': 'True', 'ipv6': 'True'}, 'name': {}, 'node_count': 1, 'bridge': 'Unnumbered_2'}

parse prefix: {'ipv4': 'True', 'ipv6': 'True'} type=<class 'box.box.Box'> True
JvB: process ipv4,True type=<class 'str'>
  File "/home/jeroen/srlinux/netlab/netlab", line 8, in <module>
    netsim.cli.lab_commands()
  File "/home/jeroen/srlinux/netlab/netsim/cli/__init__.py", line 153, in lab_commands
    mod.run(sys.argv[arg_start:])   # type: ignore
  File "/home/jeroen/srlinux/netlab/netsim/cli/create.py", line 76, in run
    augment.main.transform(topology)
  File "/home/jeroen/srlinux/netlab/netsim/augment/main.py", line 70, in transform
    transform_data(topology)
  File "/home/jeroen/srlinux/netlab/netsim/augment/main.py", line 54, in transform_data
    augment.links.transform(topology.links,topology.defaults,topology.nodes,topology.pools)
  File "/home/jeroen/srlinux/netlab/netsim/augment/links.py", line 593, in transform
    augment_lan_link(link,pools,nodes,defaults=defaults)
[...]

PR #426, opened 2022-09-09
Fix prefix parse error due to bool being converted to string

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/424


ipspace (approved):
Nice catch. Looks like nobody ever tested custom unnumbered address pools since the days we decided to go for **ipv4: true** instead of **unnumbered: true**.

Thanks a million!

PR #428, opened 2022-09-10
Fix vrf name

jbemmel:


ipspace (approved):
**** :( Apologize for that SNAFU :( Will add it as a fix to the previous commit.

PR #429, opened 2022-09-10
Containerlab: Add custom template feature to support specific devices/containers

jbemmel:
This PR adds a custom templating mechanism for containerlab, to generate and mount node-specific files.

The driving use case is the frr container: Versions after v7.5.0 do not support restarting daemon processes, blocking users from using newer versions. By pre-generating and bind-mounting /etc/frr/daemons, we can avoid having to restart processes

```
frr:
    ...
    clab:
      # image: frrouting/frr:v7.5.0
      image: frrouting/frr:v8.3.1
      mtu: 1500
      node:
        kind: linux
        config_templates:
          frr_daemons: /etc/frr/daemons
```

Note that this PR changes the location of 'binds' to be under 'clab', given that it is clab specific.
Any user netlab configs using 'binds' should be updated


ipspace:
Looks like amending the FRR VLAN commit created all sorts of weird stuff with this one. Please rebase (or close, rebase, resubmit). Apologize for the unnecessary work :(

jbemmel:
Resubmitting...

PR #431, opened 2022-09-11
Containerlab add custom templates per device type (cleaned up resubmit)

jbemmel:
This PR adds a custom templating mechanism for containerlab, to generate and mount node-specific files.

The driving use case is the frr container: Versions after v7.5.0 do not support restarting daemon processes, blocking users from using newer versions. By pre-generating and bind-mounting /etc/frr/daemons, we can avoid having to restart processes

```
frr:
    ...
    clab:
      # image: frrouting/frr:v7.5.0
      image: frrouting/frr:v8.3.1
      mtu: 1500
      node:
        kind: linux
        config_templates:
          frr_daemons: /etc/frr/daemons
```

Note that this PR changes the location of 'binds' to be under 'clab', given that it is clab specific config.
Any user netlab configs using 'binds' should be updated


ipspace (changes requested):
This is almost perfect -- uses templates, well-defined directory, does cleanup... There are just a few minor details worth changing:

* You should test for /etc/frr/daemons in clab.binds. Someone might use binds for something else.
* (Optional) I'm usually using `data.get_from_box(node,'clab.config_templates')` instead of more complex tests. You could even go for `data.must_be_list` to check the data type as well.
* (Optional) it would be nice to have the "generate extra file from template" functionality in a helper function somewhere close to common.template. We might decide to use it somewhere else (I already have a few ideas ;)
* (Optional) Wouldn't it be better to have the template at 'clab/frr/daemons.js', so if we do this for other platforms, we have that directory nicely structured?

PR #437, opened 2022-09-12
These need to be lists for the iteration to work correctly

jbemmel:


ipspace:
Dang, I missed that :(

I'll do something similar to your **binds** for Vagrant, and then we can have FRR running in a container or in a VM, and get rid of the dual logic.

PR #441, opened 2022-09-12
Fix check for routed native vlan support

jbemmel:
Move the check earlier, before any transformation logic removes the native vlan

With this patch:
```
netlab create -d srlinux -p clab  integration/vlan/vlan-routed-native.yml 
IncorrectValue in vlan: Node ros does not support routed native VLANs (red)
... link {'vlan': {'trunk': {'red': {}, 'blue': {}}, 'native': 'red'}, 'interfaces': [{'node': 's2', 'vlan': {'trunk': {'red': {}, 'blue': {}}, 'native': 'red'}}, {'node': 'ros', 'vlan': {'trunk': {'red': {}, 'blue': {}}, 'native': 'red'}}], 'linkindex': 2}
Fatal error in netlab: Cannot proceed beyond this point due to errors, exiting
```

Without it, no error is generated

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -362,14 +362,15 @@ whether the native vlan is configured as 'mode: route', and if so whether the de
 def check_native_vlan_routed(link: Box, native_vlan: str, topology: Box) -> None:
   for intf in link.interfaces:
     node = topology.nodes[intf.node]
-    vlan_mode = interface_vlan_mode(native_vlan,intf,node,topology)
-    if vlan_mode == "route":
-      features = devices.get_device_features(node,topology.defaults)
-      if not features.vlan.native_routed:
-        common.error(
-          f'Node {intf.node} does not support routed native VLANs ({native_vlan})\n... link {link}',
-          common.IncorrectValue,
-          'vlan')
+    if 'vlan' in node.get('module',[]):   # Is vlan module used on this node?
+      vlan_mode = interface_vlan_mode(native_vlan,intf,node,topology)
+      if vlan_mode == "route":
+        features = devices.get_device_features(node,topology.defaults)
+        if not features.vlan.native_routed:
+          common.error(
+            f'Node {intf.node} does not support routed native VLANs ({native_vlan})\n... link {link}',
+            common.IncorrectValue,
+            'vlan')
 
 """
 copy_vlan_attributes: copy prefix and link type from vlan to link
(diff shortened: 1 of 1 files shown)


netsim/modules/vlan.py
```diff
@@ -485,8 +480,8 @@ def create_vlan_links(link: Box, v_attr: Box, topology: Box) -> None:
           if 'mode' in intf.vlan and not get_from_box(intf_data,'vlan.mode'):
             intf_data.vlan.mode = intf.vlan.mode            # vlan.mode is inherited from trunk dictionary or parent interface
 
-          if interface_vlan_mode(intf_data,intf_node,topology) == 'bridge':     # Is this VLAN interface in bridge mode?
```
jbemmel, line 488: This also fixes a bug where vlan.mode was only being checked for the access vlan, instead of the vlan being processed

ipspace:
Before acting on this, I want to make sure we're talking about the same thing.

I added two test cases in 7a6e1b1881e6ff8944cd3f6b06aef9c1066646de, one testing  **vlan.mixed_trunk** feature flag, the other one testing **vlan.native_routed** feature flag. I verified that FRR and IOSv pass both test cases, and that the lack of feature flag results in an error in the expected test case.

It might be that you're trying to detect a different problem, in which case we have to go back to the drawing board.

jbemmel:
It’s the vlan-routed-native.yml test case which is mentioned in that commit, but not newly added

Using srlinux as a device type with that case should throw an error, it doesn’t (or didn’t when I tested it last)

ipspace:
Are you telling me that SR Linux cannot handle the "ros" device in `vlan-routed-native.yml` test case? That's a pure IP interface (no bridging) with tagged and untagged IP coming in. I don't think I've seen a device (yet) that wouldn't do that, even Arista EOS can do it, and it's definitely using layer-3 switch and not router mentality.

I did find a problem or two when trying to get this to work on Arista though, so I'll open separate issues for that.

jbemmel:
Yes, that is exactly what I'm saying: That test case cannot be provisioned like that on SR Linux

For each interface, there is a ```vlan-tagging``` flag that needs to be enabled to create more than 1 subinterface on the port. When using multiple subinterfaces (i.e. with tagging enabled), one subinterface can be configured with the 'untagged' vlan - but only in bridge mode

ipspace:
This has been fixed in b26e12bf3ecae37d75878124aaaf1e9957ebd4bb

PR #442, opened 2022-09-12
Fix unnumbered routed vlans

jbemmel:
Add a test case to check


ipspace:
I have to figure out what the root cause of the problem is. It looks like you're just fixing the _unnumbered interface_ case.

ipspace:
9e653704936c176e43b22570cee7b4686d081472 provides a generic fix -- copy **unnumbered** and **role** (later **pool**) into VLAN member links.

Also updated the VLAN addressing documentation -- you cannot use **prefix** on routed VLANs (plus a gazillion of other rewrites, hopefully making documentation a bit better).

PR #443, opened 2022-09-12
Proposal: Allow user to disable ebgp peer creation per vrf

jbemmel:
Currently, the code builds ebgp sessions between all neighbors with a different AS, within each VRF.
This results in several parallel peering sessions on trunk ports with multiple vlans

This patch allows the user to avoid ebgp sessions for a given vrf by setting ```bgp: False``` within that vrf, locally per node or globally

TODO (if accepted):
* Update other platform templates to handle ```bgp: False``` in vrfs

### DIFF
--- a/netsim/ansible/templates/vrf/srlinux.j2
+++ b/netsim/ansible/templates/vrf/srlinux.j2
@@ -10,7 +10,7 @@ updates:
 {% if 'ospf' in vdata %}
 {{ ospf_config(0,'ipv4' if vdata.af.ipv4|default(0) else 'ipv6',vname,vdata.ospf,vdata.ospf.interfaces)}}
 {% endif %}
-{% if 'bgp' in vdata %}
+{% if vdata.bgp|default(False) %}
 {{ bgp_config(vname,vrf.as,bgp.router_id,vdata.bgp,vdata) }}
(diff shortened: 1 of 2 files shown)


ipspace:
... if at all, we need something more generic that would (A) work on data structures not with template hacks, (B) work on VRF and interface level, and (C) work for all routing protocols.

jbemmel:
> ... and this is a completely wrong way of solving it. You want to reduce the number of EBGP sessions, which are described in a "neighbors" data structure, and instead of trimming that data structure you're introducing an attribute that has to be implemented in every platform template?

The patch to the bgp module does operate on the data model and results in a trimmed neighbors structure. Perhaps it would be easier to remove bgp from the vrf, rather than leaving it as ```bgp: False```

The same convention could work for all routing protocols - just set a flag to False, in the vrf or on the interface

ipspace:
This will be solved together with #515 -- VLANs already carry link-level attributes, and I plan to add the same to VRFs in #515, allowing you to specify OSPF area or BGP local-as on VRF.

When that works, you'll also be able to set **bgp: False** on a VRF, and that will get propagated to interfaces, and the EBGP code will say "oh, well..." and move on.

PR #448, opened 2022-09-14
frr add native vlan interfaces

jbemmel:
* Create VLAN devices with ID 0 for native vlans (eth[x].0, for design visibility purposes)
* Idempotency: Don't error out when interface is already associated or VXLAN device exists

### DIFF
--- a/netsim/ansible/templates/vlan/frr.j2
+++ b/netsim/ansible/templates/vlan/frr.j2
@@ -3,19 +3,12 @@
 set -e # Exit immediately when any command fails
 #
 {# Create VLAN subinterfaces. No need to set IP addresses, FRR can handle that #}
-{% for i in interfaces if i.type == 'vlan_member' %}
-if [ ! -e /sys/devices/virtual/net/{{ i.ifname }} ]; then
-  ip link add link {{ i.parent_ifname }} name {{ i.ifname }} type vlan id {{ i.vlan.access_id }}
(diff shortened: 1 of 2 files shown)


ipspace (commented):
You're submitting a PR for two completely independent issues:

* While the VXLAN fix sort-of makes sense, it's not the only non-idempotent operation in FRR templates (every **brctl addif** operation fails when executed twice), and effectively saying "meh, ignore that" might mask some other issues in the future.
* I can't fathom why you insist on creating an extra interface with no additional encapsulation. You don't need a _native VLAN_ subinterface on Linux unless you're using VLAN-aware Linux bridge (which we're not).

jbemmel:
I started going down the single bridge path (https://github.com/jbemmel/netlab/tree/frr-single-bridge), and you can even do a single VXLAN device - but that quickly becomes a project on its own, maybe some other time

For people who are using Netlab to test out various designs, it can be useful to have some visibility into native VLANs. I like to be able to see "eth[x].0" devices where there is a native vlan - even if it is not strictly needed for traffic purposes

ipspace:
> For people who are using Netlab to test out various designs, it can be useful to have some visibility into native VLANs. I like to be able to see "eth[x].0" devices where there is a native vlan - even if it is not strictly needed for traffic purposes

And we're back to those hypothetical users. I'm not biting. There are other more relevant resources on the 'net If someone wants to learn Linux networking with native VLANs.

Issue #450, opened 2022-09-14
VLAN trunks on self-links don't get populated


jbemmel:
Sample topology:
```
provider: clab

module: [ vrf, vlan ]

vrfs:
  global: # Global VRF

  customer1:

vlans:
  route-leak-to-global-customer1:

nodes:
  leaf1:
    device: srlinux

links:
- interfaces:
  - node: leaf1
    name: Global side
    vlan.trunk: [route-leak-to-global-customer1]
    vlan.trunk:
      route-leak-to-global-customer1:
        vrf: global
  - node: leaf1 # leaf2 for testing
    name: Customer side
    vlan.trunk: [route-leak-to-global-customer1]
    vlan.trunk:
      route-leak-to-global-customer1:
        vrf: customer1
```
The resulting model:
```
nodes:

  leaf1:
...
    interfaces:
    - clab:
        name: e1-1
      ifindex: 1
      ifname: ethernet-1/1
      linkindex: 1
      name: leaf1 -> leaf1
      neighbors:
      - ifname: ethernet-1/2
        node: leaf1
      subif_index: 2
      type: p2p
      vlan: {}   <--- empty
    - clab:
        name: e1-2
      ifindex: 2
      ifname: ethernet-1/2
      linkindex: 1
      name: leaf1 -> leaf1
      neighbors:
      - ifname: ethernet-1/1
        node: leaf1
      type: p2p
      vlan:
        trunk:
          route-leak-to-global-customer1:
            vrf: customer1
        trunk_id:
        - 1000
```

The issue is here: https://github.com/ipspace/netlab/blob/dev/netsim/modules/vlan.py#L725
There are 2 interfaces on the link that match the node; only the first one is ever returned
Consequently, 2 vlan_members are created with the same parent_ifindex=1, instead of one for each side of the link

Issue #319, opened 2022-07-10
Interface (node-to-link attachment) attributes are not checked


ipspace:
e5e96141dd92c53c3bc2231369e63088d8e5ab20 successfully added support for `ifindex` interface attribute even though there's no related attribute defined in `attributes` part of topology defaults. Clearly we need a check of interface attributes ;)

jbemmel:
vlan attributes are also not checked - 'role' is not listed as a valid attribute, but it is allowed and does affect the generated configuration

PR #451, opened 2022-09-14
Fix handling of self loops

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/450 in a way that topologies containing self-loops with vlan trunks don't generate invalid data models. Does not change IP address allocation or data models produced by current test set (other than fixing the [new test]([tests/topology/expected/vrf-leaking-loop.yml](https://github.com/ipspace/netlab/pull/451/files#diff-6048378ee97c0a0d7fa37425987711ec0bad670bc0a05c8694b09448716a302e) included in this PR)

### DIFF
--- a/docs/module/vlan.md
+++ b/docs/module/vlan.md
@@ -42,6 +42,17 @@ The VLAN configuration module assumes you're creating a sane design in which:
 
 It might be possible to build topologies that deviate from these rules, but don't be surprised when the results look weird.
 
+## Deployment Recommendations
+
+The VLAN module tries to support many different platform-specific VLAN configuration paradigms, and the resulting maze of various bits and pieces might frustrate the unwary. Following these guidelines might help you avoid the frustration:
+
+* Use **route** VLANs only in a router-on-a-stick or multi-hop VRF Lite designs.
+* While it's possible to mix **route** and **irb** VLANs on the same device, the results are not always intuitive. Don't do that -- a device should be a pure router (terminating **route** VLANs) or a layer-3 switch (forwarding within and between **bridge** or **irb** VLANs) but not both.
+* Do not use **mode: route** on access VLANs. Use native IP connectivity.
+* Avoid **mode: route** on native VLANs. Some devices might not like that approach.
+* Don't use link attributes to modify VLANs in a VLAN trunk. Use VLAN attributes.
+* Never combine multiple [forwarding modes](vlan-forwarding-modes) of a single VLAN *within a single node*. Using **route** VLAN mode on a router-on-a-stick and **bridge**/**irb** VLAN mode on attached switch is perfectly fine.
+
 ## Module Parameters
 
 VLANs are defined in the global or per-node **vlans** dictionary. Global- and node VLAN parameters are merged to get VLAN data for individual nodes.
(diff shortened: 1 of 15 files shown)


ipspace (commented):
(A) I wouldn't fix this at all, and fixing it this way makes no sense anyway due to VLAN interface issue (see #450)
(B) I wouldn't fix the IP addressing logic. If someone desperately insists on doing this, they should use static addresses.
(C) The proper way of fixing this would be to set **parentindex** (or some such) to the **ifindex** of parent interface in the **create_vlan_links** and then use it if needed.

It might make sense to implement (C) for academic purity reasons, but (C.1) we have better things to do and (C.2) it still won't give you what you want to do.

jbemmel:
Cleaned up and much simplified version

* Cannot use ```parentindex``` or ```parent_ifindex``` as those are used for other purposes - proposing "trunk_ifindex"
* Attribute only used internally, not exposed in data model
* Base case with len([candidates])==1 simply returns what the current code returns (simplified single iteration)
* If there are multiple candidates, use the 'trunk_ifindex' value (0 or 1) to disambiguate which side of the link is returned

Note that this is not an academic case for me, we have a specific customer use case where this feature is key. I appreciate that you prefer to keep the main code base focused on the typical use cases, but my alternative is to fork the project and do things my separate way. I prefer not to do that, but I can if you prefer

ipspace:
This looks surprisingly clean and efficient.

Do I get it right that because you're setting `selfloop_ifindex` only when creating VLAN member links, and because the interfaces created from those links become either VLAN interfaces or subinterfaces, the `selfloop_ifindex' is automatically stripped from those interfaces? That's cool...

Obviously we'd hit a snag when #319 gets implemented, but we're not there yet.

Just a thought (thinking about #319) -- what if we use `_something` for internal attributes that don't have to be checked. You'd have to rename `selfloop_ifindex` to `_selfloop_ifindex`, but we'd have a framework that we could eventually use to clean up all other similar stuff from `defaults.attributes`. For example, I had to add `start_transit_vlan` to evpn in the last commit just to make Linux VLAN-aware bridge happy, and there's absolutely no need to have that exposed or checked.

jbemmel:
All attributes on VLAN sub interfaces listed in phy_ifaddr are eventually removed at https://github.com/ipspace/netlab/blob/dev/netsim/modules/vlan.py#L817
except for those listed at https://github.com/ipspace/netlab/blob/1165d7e3e5b842975f5f89550661b8ccdbb2bd76/netsim/modules/vlan.py#L33

I did not write that code, I just used and extended what was there

We can avoid any snags with #319 by doing the validation check before adding additional internal attributes like this one.

I'm ok with _selfloop_index

ipspace:
> We can avoid any snags with #319 by doing the validation check before adding additional internal attributes like this one.

That would mean running link validation checks before any module hooks, which would mean that we wouldn't catch errors where modules add unexpected attributes to links. Also, modules or plugins could expand the list of valid attributes, so it's a very fine balancing act where to call the link attribute validation.

Adding a "this is not the attribute you're looking for" flag is much easier ;)

ipspace:
And now that this is merged, it's time for "big picture" discussions cosily hidden into the comments of a merged PR ;))

> Note that this is not an academic case for me, we have a specific customer use case where this feature is key.

Customers always come first (after all, they are indirectly paying your bills), and I completely understand you have to be focused on them. 

Totally off-topic: I found customers with very specific use cases relying on obscure features to be more of a long-term headache than a long-term win, but that's a discussion for another time.

> I appreciate that you prefer to keep the main code base focused on the typical use cases

Correct - the tool has a different set of customers, and our job is to keep the tool simple, versatile, and stable. Furthermore, I don't believe in undocumented nerd knobs added to solve a specific customer requirement _within a free open-source tool_.

> but my alternative is to fork the project and do things my separate way. I prefer not to do that, but I can if you prefer

It's great having you on board, and you're usually pushing me in the right direction, so I'd also prefer if you could find a solution that would not involve forking the project.

Maybe you could consider using plugins instead of modifying core behavior. For example, in this specific use case your plugin could generate extra interfaces (emulating the data structures of VLAN interfaces or subinterfaces) based on a plugin-specific link attribute (example: **recirculation.vlans**).

Issue #449, opened 2022-09-14
Missing loopback interfaces when using role: host with frr


jbemmel:
When declaring an frr node in 'host' role:
```
provider: clab

nodes:
  frr-host:
    device: frr
    role: host
```

There is no 'loopback' in the generated YAML -> template fails

ipspace:
Hosts don't run routing protocols and use static routing toward the first-hop gateway on the first interface to reach the rest of the network.

As hosts don't run routing protocols, and we don't support static routing on other nodes (yet), it makes no sense to have loopbacks on hosts.

Hosts were meant to be used for anything else but Linux VMs/containers. If you feel you simply must use FRR nodes as hosts, please fix the initial configuration template. If you think the documentation misled you into making a wrong choice, please fix it.

jbemmel:
Some scenario's involve hosts with multiple interfaces and/or vlans, with the recently added frr vlan support I figured that would be the shortest path.

Fixing the template was the first thing I did, but some users may still need their hosts to have a loopback IP, for simple ping check purposes (not routing protocols)

ipspace:
As I wrote, any sane use of a loopback interface would involve running a routing protocol with the device, and according to **netlab** definition of **role: host**, that's no longer a host. If you feel the documentation needs to be improved, please do it.

Also, I can't grasp why you try so hard to hammer square pegs into round holes. You have multiple interfaces on a Linux box, so you need routing, so choose **device: frr** and move on.

ipspace:
> ... but some users may still need their hosts to have a loopback IP, for simple ping check purposes (not routing protocols)

(A) I will no longer listen to arguments relying on some hypothetical users that we haven't seen so far. Spending time on features people are asking for (example: EVPN on Cumulus and Nexus OS) is a much better use of everyone's time.

(B) Having a loopback IP will not work without static routing (not available) or routing protocol. The proof is left as an exercise for the reader.

🖐️🎤

jbemmel:
On Containerlab, a minimal Linux container has 2 interfaces: eth0(management) and eth1. Static routing is already required even in the minimal case

PR #453, opened 2022-09-15
Handle role=host case where loopback is not set

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/449

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -62,10 +62,10 @@ ipv6 forwarding
 !
 interface lo0
  no shutdown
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
Apologize for being a bit rough, but if you decide to fix some imaginary problems, please do it right.

(A) There's no need to create the lo0 interface if **loopback** is not present
(B) There's no need to configure **interface lo0** if **loopback** is not present

Optional

(C) `loopback.ipv4 is defined` might be simpler than `if 'ipv4' in loopback|default({})`

ipspace (approved):
Much better, although you did miss the "let's create the loopback interface" bit.

PR #458, opened 2022-09-18
Beautiful vxlan vrf evpn test/show case

jbemmel:
85 lines of YAML to magically create a sophisticated topology using open source components

For those who frown upon iBGP-over-eBGP designs, I added hints on what to change to use OSPF or ISIS instead of eBGP

Note: I also cleaned up the iBGP-over-eBGP example that was already there, simplifying the modules on the spine


ipspace:
Given that we already have 'evpn' directory in 'tests/integration', I'd suggest you move this one there -- it's more about EVPN than VXLAN.

ipspace (approved):
Hope you're done with this one -- you keep adding stuff even after marking the PR 'ready for review'

PR #462, opened 2022-09-18
Fix clab mgmt network

jbemmel:
Adhere to the management network settings provided by Netlab, instead of using Containerlab defaults

Specifically, the Netlab default is ipv4-only while Containerlab provides ipv4+ipv6 by default.


ipspace (approved):
Nice one. Now we could also unify the "netlab connect" logic so it doesn't have to use clab hostnames to connect to devices.

Issue #460, opened 2022-09-18
Set 'ipv6: True' on interfaces used for unnumbered IPv6 BGP sessions


ipspace:
An interface that is used for unnumbered IPv4 EBGP session(s) needs IPv6 LLA (that's how the unnumbered sessions are implemented).

Not a big deal, until someone decides to disable IPv6 by default (I'm looking at you, Cumulus Linux 4.4 container). I "fixed" that SNAFU by enabling IPv6 on interfaces that have role 'external', but it would be much nicer to set **ipv6: True** on such interfaces within the BGP configuration module.

That might mess up the node address family though (all of a sudden, there would be IPv6 and VPNv6 address families) unless we do it at the very end of the transformation process.

Please note that we have a problem only in IPv4-only environments with interfaces configured either with **unnumbered: True** or **ipv4: True**.

The only devices currently supporting unnumbered IPv4 EBGP sessions are Cumulus Linux, FRR, and Nokia SR Linux. Any thoughts on this conundrum @jbemmel @petercrocker?

jbemmel:
I think the flag should be ```bgp.unnumbered```; if true, it would enable ipv6 on the interface (given the dependency)

I just hit a similar issue through the ```unnumbered``` flag; it currently works on frr even with ipv6 absent, because the device templates don't currently disable ipv6 correctly. I'll submit some PRs to fix that (also for Cumulus potentially)

jbemmel:
See https://github.com/ipspace/netlab/pull/463

I suspect the reasoning by Cumulus could be that they don't want their devices to do ipv6 unless explicitly configured to do so, for security reasons. In contrast, vanilla FRR enables ipv6 unless forced to disable it (outside of FRR)

ipspace:
> I think the flag should be `bgp.unnumbered`; if true, it would enable ipv6 on the interface (given the dependency)

I don't want to test yet-another flag in initial configuration (or later) templates. Setting 'ipv6: True' on the interface data gets the job done, but I wouldn't want it to trigger (for example) OSPFv3. Will check what happens if I set that flag late in BGP module.

> I just hit a similar issue through the `unnumbered` flag; it currently works on frr even with ipv6 absent, because the device templates don't currently disable ipv6 correctly. I'll submit some PRs to fix that (also for Cumulus potentially)

The default state depends on the system setting, and as the FRR container serves mostly as a control-plane container (with some of us abusing it as a pure Linux data-plane device), I don't think they would care.

[...]

PR #463, opened 2022-09-18
frr ipv6 disabled fixes

jbemmel:
* Disable ipv6 ND when ipv6 is not configured (frr bug prevents this from getting applied)
* Don't enable ipv6 forwarding when ipv6 is disabled
* Explicitly disable ipv6 on interfaces without ipv6 enabled, under Linux

### DIFF
--- a/netsim/ansible/templates/bgp/frr.bgp-config.j2
+++ b/netsim/ansible/templates/bgp/frr.bgp-config.j2
@@ -22,7 +22,7 @@ router bgp {{ bgp.as }}
     Create neighbors
 #}
 {% for af in ['ipv4','ipv6'] %}
-{%   for n in bgp.neighbors if n[af] is defined %}
+{%   for n in bgp.neighbors if n[af] is defined and (n[af] is string or af=='ipv6') %}
 {%     set peer = n[af] if n[af] is string else n.local_if|default('?') %}
   neighbor {{ peer }}{{ ' interface' if peer!=n[af] else '' }} remote-as {{ n.as }}
   neighbor {{ peer }} description {{ n.name }}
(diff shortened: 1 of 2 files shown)


ipspace (changes requested):
Good catch on the BGP conditional logic. However...

* FRR IPv6 forwarding is (IIRC) enabled by default, so...
* Changing FRR IPv6 forwarding does nothing to the "disable IPv6" system flag
* Your implementation will disable unnumbered BGP sessions in IPv4-only lab unless we implement #460 by setting 'ipv6: True'
* You copied my incomplete implementation (yeah, I know). The same stuff has to be done on VLAN interfaces and subinterfaces (that's why I hate it so much).

jbemmel:
I hit this during https://github.com/ipspace/netlab/pull/458, when I used ```unnumbered: True``` instead of ```ipv4/ipv6: True```
It was working with frr, but not with srlinux

For consistency, I figured it would be better to require the user to set ```ipv6: True``` for this use case, given that it requires ipv6 enabled on the link

jbemmel:
I'm only trying to be explicit and technically accurate/intuitive. If you think that the audience prefers/expects ```ipv4: True``` (and by association ```unnumbered: True```) to enable ipv6 lla under the hood, then we can make it work like that too

ipspace:
> I'm only trying to be explicit and technically accurate/intuitive. If you think that the audience prefers/expects `ipv4: True` (and by association `unnumbered: True`) to enable ipv6 lla under the hood, then we can make it work like that too

The audience expects 'ipv4: True' to work with unnumbered EBGP sessions just like the vendor promised ;) It turns out that the vendor promises depend on the unmentioned fact that their distro of Linux happens to have IPv6 enabled on all interfaces 🤷‍♂️

ipspace (changes requested):
'ipv6: false' is removed from the interface somewhere in link transformation code, so there's no need to check for that. 'l.ipv6 is not defined' is good enough.

PR #466, opened 2022-09-19
Add a check and error out when a node has VXLAN-enabled vlans without having the 'vlan' module enabled

jbemmel:
This leads to undefined vlan.id values, causing confusing errors about evpn.evi values that are based on those

Fixes https://github.com/ipspace/netlab/discussions/417

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -153,18 +153,24 @@ class EVPN(_Module):
     vlan_list = data.get_from_box(node,'vxlan.vlans') or []       # Get the list of VXLAN-enabled VLANs
     if not vlan_list:
       return                                                      # This could be a route reflector running EVPN
+    elif 'vlan' not in node.get('module',[]):
+      common.error( f"VXLAN-enabled vlans {vlan_list} on node {node.name} require the 'vlan' module", common.MissingValue, 'evpn' )
+      return
 
     _routing.router_id(node,'bgp',topology.pools)                 # Make sure we have a usable router ID
 
     for vname in vlan_list:
-      vlan = node.vlans[vname]
-      #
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
As there's no topology describing the challenge you're facing i'm trying to reverse-engineer what's going on. If I got it right, you have a node with VXLAN-enabled VLANs (+ EVPN) but without the VLAN module, so the VLANs have no vlan.id.

Should that be the case, the error should be reported in the VXLAN module, not in EVPN module, and if VXLAN module is not present, then we have to start discussing what to do with **vxlan.vlans** attribute which could only happen on a node, because globally that would be detected.

jbemmel:
The error happens with only a global ```vxlan.vlans``` defined, on a node with vxlan enabled but none of the listed vlans.
node.vlans = [] but node.vxlan.vlans == < global list >

Then when the evpn module iterates over vxlan.vlans, it errors out

One way to fix this, is for the vxlan module to remove ```vxlan.vlans``` when ```node.vlans``` is empty

ipspace:
Please add the topology that generates this. I'm worried what the root cause of this behavior is, and that we're not just papering over the symptoms.

jbemmel:
I still think it's better to also clear ```vxlan.vlans``` when no vlans match - even when the vlan module is correctly applied, and it seems to me that the vxlan module is the right place to do that.

Agree that there is the underlying issue of module dependency checks too

ipspace:
> I still think it's better to also clear `vxlan.vlans` when no vlans match - even when the vlan module is correctly applied, and it seems to me that the vxlan module is the right place to do that.

Agree. #476 has been fixed.

> Agree that there is the underlying issue of module dependency checks too

Could you please check whether #475 and #476 solved what this PR is addressing?

ipspace:
Based on lack-of-response I'm guessing that this is fixed.

PR #468, opened 2022-09-19
Early draft proposal: default gateway flag for vlans

jbemmel:
In the context of https://github.com/ipspace/netlab/discussions/467

I realize this is a bit hacky, just to share an idea of what this could look like:

```
vlans:
  internet:
    mode: irb
    vrf: internet
    gateway:
      default: True
```

I experimented with setting ```gateway.default: True``` in fix_vlan_gateways (on the first interface found with ipv4), but the iteration order is based on vlan id and I think that would be non-intuitive for users (too implicit)

Also note the bug fixes in links.py, triggered by copying 'gateway' from vlans


ipspace:
This has been implemented in the new **gateway** module, and I still don't believe in pointing a default route into the lab network ;)

PR #469, opened 2022-09-19
Add ipv6_lla flag for bgp unnumbered

jbemmel:
Context: https://github.com/ipspace/netlab/pull/463

Instead of assuming ```ipv6: True``` for BGP unnumbered, have the bgp module flag the interface as requiring ipv6_lla enabled.
This avoids triggering (e.g.) ospfv3 for such topologies

TODO: Documentation, will add if accepted


ipspace:
Not needed, adding 'ipv6: True' late in BGP module worked. See ae6a422bfebf87c46b3abde86155f84fa441b3cb

PR #470, opened 2022-09-20
Align VyOS and Dell OS10 with tunable AFIs and co

### LINKED ISSUE #363: Use bgp.local_as and bgp.activate functionality for EVPN IBGP-over-EBGP on VyOS and Dell OS10
We will eventually get to the point where someone will say "it would be great to have EBGP local-as functionality" outside of the "let's create job security by running IBGP over EBGP" EVPN use case.

The functionality is almost there (in the current ebgp-local_as.py plugin), but would need to be generalized:

* I would rename the plugin to **ebgp.local_as** to keep the **protocol.feature** naming convention
* It would have to be migrated into a directory within `extra` directory to enable platform-specific configuration templates
* It should use **bgp.local_as** and not **something.underlay_as** attribute as we're no longer solving just the EVPN use case.
* **local-as** configuration should be moved into the plugin custom configuration template, right now I have a feeling it's all over the place

That would result in a consistent and easy-to-extend (platform-wise) **local-as** functionality that's independent of EVPN.

Next, we could implement an EVPN plugin (for example, **evpn.ibgp_over_ebgp**) that would:

* Use the **evpn.underlay_as** attribute to signal the IBGP-over-EBGP design
* Copy **evpn.underlay_as** to **bgp.local_as** on all interfaces (ideally node-link-attachment data) with IBGP peers. We have to find the perfect plugin hook to do that -- it has to be done after the module attributes are copied into nodes, but before the interfaces are created or before **ebgp.local_as** plugin modifies the list of neighbors.
* Use custom configuration templates in **evpn.ibgp_over_ebgp** plugin to disable IPv4/IPv6 AF on IBGP sessions when **evpn.underlay_as** is present.

Have I missed something? Ready to go for it @jbemmel @ssasso?

ssasso:
Fixes #363

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -48,9 +48,10 @@ The following features are only supported on a subset of platforms:
 | Cisco IOS/IOS XE      |  ❌  |  ❌  |  ✅ |  ✅ |  ✅ |
 | Cumulus Linux 4.x     |  ✅ |  ✅ |  ✅ |  ✅ |  ✅ |
 | Cumulus Linux 5.x     |  ✅ |  ✅ |  ❌  |  ❌  |  ✅ |
-| Dell OS10             |  ❌  |  ❌  |  ✅ |  ❌  |  ❌  |
+| Dell OS10             |  ❌  |  ❌  |  ✅ |  ❌  |  ✅  |
 | FRR 7.5.0             |  ✅ |  ❌  |  ✅  |  ✅  |  ✅ |
 | Nokia SR Linux        |  ✅ |  ❌ |  ✅ |  ✅  |  ✅  |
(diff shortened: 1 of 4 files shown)


ipspace:
Thank you!

Next: check whether `tests/integration/evpn/vxlan-bridging-ibgp-over-ebgp.yml`(EVPN IBGP-over-EBGP without the plugin) works as expected.

ssasso:
The PR has already been tested with:
* `integration/evpn/vxlan-bridging-ibgp-over-ebgp.yml`
+ with a similar topology with also L3 VNI

ipspace:
> The PR has already been tested

Awesome. Thank you!

PR #471, opened 2022-09-20
Add explicit support for bgp large communities (RFC8092)

jbemmel:
On some Cisco devices 'standard' implies also 'large' (see e.g. [here](https://content.cisco.com/chapter.sjs?uri=/searchable/chapter/content/en/us/td/docs/switches/lan/catalyst9300/software/release/17-5/configuration_guide/rtng/b_175_rtng_9300_cg/configuring_bgp_large_community.html.xml)) but on other platforms (like SR OS and SR Linux) they are configured separately

* Update documentation
* Update topology defaults

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -261,7 +261,7 @@ The propagation of BGP communities over IBGP and EBGP sessions is controlled wit
 
 The value of **bgp.community** attribute could be:
 
-* A string: **standard** or **extended** -- only specified communities will be propagated to IBGP and EBGP neighbors. In the following example, R1 propagates standard communities to all its neighbors.
+* A string: **standard** or **extended** or **large** -- only specified communities will be propagated to IBGP and EBGP neighbors. In the following example, R1 propagates standard communities to all its neighbors.
 
 ```
 nodes:
@@ -270,7 +270,7 @@ nodes:
       community: standard
 ```
 
-* A list of strings (**standard** and/or **extended**) -- all communities specified in the list will be propagated to IBGP and EBGP neighbors. Most network operating systems will be configured with **both** configuration keyword if you specify `['standard','extended']` as the value. In the following example, R1 propagates standard and extended communities to all its neighbors.
+* A list of strings (**standard** and/or **extended** and/or **large**) -- all communities specified in the list will be propagated to IBGP and EBGP neighbors. Most network operating systems will be configured with **both** configuration keyword if you specify `['standard','extended']` as the value. In the following example, R1 propagates standard and extended communities to all its neighbors.
 
 ```
 nodes:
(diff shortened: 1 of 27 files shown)


ipspace (changes requested):
I understand the need to support large BGP communities. However, we should start with the question of whether we want to include "large" with "standard" or have it configurable -- after all, it's just another optional transitive attribute that should not impact the neighbor.

What I'd like to hear before doing anything else is whether you considered the impact of changing the meaning of an attribute used by a half-dozen platforms before changing the system default setting:

* Some configuration templates use the keywords from the **community** attribute verbatim. Have you verified that they accept **large** keyword?
* Other configuration templates translate two keywords in the **community** list into both. Have you considered what happens if someone sets BGP community to `[ standard, large ]`?

You might want to open an issue to discuss this with other contributors and evaluate the impact on other platforms. As things stand right now, I'm not merging this.

jbemmel:
I did consider the other platforms, but I was assuming that the templates for the other platforms could easily be updated to accommodate the new 'large' value. The PR was intended to trigger this discussion.

I noticed that the default for ebgp neighbors is minimal (standard only), while ibgp neighbors get all available types.
The proposed default setting follows the same logic ( i.e. ebgp still gets only standard communities ).
Lumping 'large' together with 'standard' is possible and easier, but I figured it would break the design intent of only sharing minimal, explicit information with (less trusted) ebgp peers

ipspace:
> I did consider the other platforms, but I was assuming that the templates for the other platforms could easily be updated to accommodate the new 'large' value. The PR was intended to trigger this discussion.

I might have missed some of the nuances of how GitHub email notification works, but how is a PR supposed to trigger a discussion? Don't we have issues or discussions for that?

> I noticed that the default for ebgp neighbors is minimal (standard only), while ibgp neighbors get all available types. The proposed default setting follows the same logic ( i.e. ebgp still gets only standard communities ). Lumping 'large' together with 'standard' is possible and easier, but I figured it would break the design intent of only sharing minimal, explicit information with (less trusted) ebgp peers

Large BGP communities are needed only when you're dealing with 4-byte ASN, in which case they are the only thing that works. If you want to pass any community information defined by your AS, and you happen to have a 4-byte ASN, you have to send large communities anyway... or am I missing something?

jbemmel:
> > I did consider the other platforms, but I was assuming that the templates for the other platforms could easily be updated to accommodate the new 'large' value. The PR was intended to trigger this discussion.
> 
> I might have missed some of the nuances of how GitHub email notification works, but how is a PR supposed to trigger a discussion? Don't we have issues or discussions for that?
> 

One could argue I "abused" a PR to trigger a discussion, as I felt it was useful to have the proposed code changes associated with the discussion.

> > I noticed that the default for ebgp neighbors is minimal (standard only), while ibgp neighbors get all available types. The proposed default setting follows the same logic ( i.e. ebgp still gets only standard communities ). Lumping 'large' together with 'standard' is possible and easier, but I figured it would break the design intent of only sharing minimal, explicit information with (less trusted) ebgp peers
> 
> Large BGP communities are needed only when you're dealing with 4-byte ASN, in which case they are the only thing that works. If you want to pass any community information defined by your AS, and you happen to have a 4-byte ASN, you have to send large communities anyway... or am I missing something?

Well - according to https://labs.ripe.net/author/emileaben/bgp-large-communities-uptake-by-the-community-at-large/ one can find uses of large communities that aren't in the AS path. Not sure if that would be intentional, could be due to router misconfiguration.

The issue came up as I was refactoring the SR OS templates, SR OS sends large communities by default but allows you to disable them upon request. I figured this could be useful to "clean up" routes being passed to peers, making sure we're not sending things we did not intend. It just made me wonder if it would be worthwhile to have as a knob

It's not a big issue for me, we can assume (document?) that 'standard' implies 'large' and leave it at that

PR #472, opened 2022-09-21
sros vlan vxlan evpn vrf ospfv3 support

jbemmel:
This PR adds support for Nokia SR OS for the following modules:
* vlan
* vrf
* vxlan (EVPN based)
* evpn
* ospfv3 (minimal but functional)
* bgp: remove 'address_families' attribute which was only used in SR OS templates
* device defaults: set MTU to ensure ospf comes up in scenarios with various port combinations

Test cases checked:
+ integration/vlan/vlan-bridge-native.yml
+ integration/vlan/vlan-mixed-native.yml
+ integration/vlan/vlan-bridge-trunk-router.yml
+ integration/evpn/vxlan-asymmetric-irb.yml
+ integration/evpn/vxlan-symmetric-irb.yml
+ integration/vlan/vlan-routed-trunk.yml
+ integration/vlan/vlan-bridge-trunk-router.yml

I've included a sample script with the start of an automated test setup - something like this could be run as part of github hooks


ipspace:
Wow!!! An enormous amount of work. Thanks a million!

PR #473, opened 2022-09-21
VyOS and Dell OS10: Support for Unnumbered eBGP

ssasso:
Tested with a modified version of `integration/evpn/vxlan-routing-ibgp-over-ebgp-unnumbered.yml`

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -48,10 +48,10 @@ The following features are only supported on a subset of platforms:
 | Cisco IOS/IOS XE      |  ❌  |  ❌  |  ✅ |  ✅ |  ✅ |
 | Cumulus Linux 4.x     |  ✅ |  ✅ |  ✅ |  ✅ |  ✅ |
 | Cumulus Linux 5.x     |  ✅ |  ✅ |  ❌  |  ❌  |  ✅ |
-| Dell OS10             |  ❌  |  ❌  |  ✅ |  ❌  |  ✅  |
+| Dell OS10             |  ✅  |  ✅  |  ✅ |  ❌  |  ✅  |
 | FRR 7.5.0             |  ✅ |  ❌  |  ✅  |  ✅  |  ✅ |
 | Nokia SR Linux        |  ✅ |  ❌ |  ✅ |  ✅  |  ✅  |
-| VyOS                  |  ❌  |  ❌  |  ✅ |  ❌  |  ✅  |
+| VyOS                  |  ✅  |  ✅  |  ✅ |  ❌  |  ✅  |
 
 
 ## Global BGP Configuration Parameters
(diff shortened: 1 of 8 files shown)


ipspace (commented):
Just a few questions before I'm merging this:

* Why would you enable RA only on LLA P2P interfaces? It's also needed on LAN IPv6 interfaces, for example to make SLAAC work for attached hosts
* Looks like OS10 requires explicit 'ipv6 enable' for unnumbered EBGP to work, which means we have to do something about #460 to make this work with 'tests/integration/bgp/ipv4-unnumbered.yml'. No worries, not your problem, I just need to know there's another platform that needs that flag.

ssasso:
> * Why would you enable RA only on LLA P2P interfaces? It's also needed on LAN IPv6 interfaces, for example to make SLAAC work for attached hosts

You are right on this, I'll fix that. I was focused only on the Unnumbered eBGP case forgetting the rest.
I'm setting the PR as WIP for now.

> * Looks like OS10 requires explicit 'ipv6 enable' for unnumbered EBGP to work, which means we have to do something about #460 to make this work with 'tests/integration/bgp/ipv4-unnumbered.yml'. No worries, not your problem, I just need to know there's another platform that needs that flag.

According to Dell's documentation, `ipv6 enable` is required *unless* you put any other param regarding ipv6, or if you previously configured the interface with "autoconfig" (basically, it "resets" the IPv6 interface state).
I.e., after setting the nd ra parameters the `ipv6 enable` became implicitly enabled (and in fact you don't see it in `show running-config`).

ipspace:
> According to Dell's documentation, `ipv6 enable` is required _unless_ you put any other param regarding ipv6, or if you previously configured the interface with "autoconfig" (basically, it "resets" the IPv6 interface state). I.e., after setting the nd ra parameters the `ipv6 enable` became implicitly enabled (and in fact you don't see it in `show running-config`).

Which still means "it's disabled by default unless you do something to enable it" (and #460 still applies) which is different from Cumulus Linux "IPv6 is enabled on all interfaces, so you get LLA without even asking for it. You're welcome!"

ssasso:
Maybe I explained myself in a wrong way.
*It is* enabled by default, but if you change the state the LLA may be deconfigured. With `ipv6 enable` you will reset the state.

https://www.dell.com/support/manuals/it-it/dell-emc-smartfabric-os10/smartfabric-os-user-guide-10-5-2-6/enable-or-disable-ipv6?guid=guid-873d7ea2-83a1-4f7b-99f0-a871ac22ef44&lang=en-us

```
By default:
IPv6 forwarding is enabled on physical Ethernet interfaces, VLANs, and port groups. IPv6 forwarding is disabled only when you enable IPv6 address autoconfiguration on an interface and set it in host mode using the ipv6 address autoconfig command.

If you disable IPv6 and configure a Layer (L2) interface in Layer (L3) mode, IPv6 is not automatically re-enabled on the interface. You must manually re-enable it.

A link-local address automatically generates when you re-enable IPv6 on an interface with the ipv6 enable command.
```

PR #474, opened 2022-09-21
Bugfix: 'value' can be None

jbemmel:
Actually, this may be hiding a deeper issue: Node vlans need to be pulled in from global, before this iteration happens


jbemmel:
Solved by https://github.com/ipspace/netlab/pull/481

Issue #480, opened 2022-09-21
Unable to set vrf attributes in group node_data


jbemmel:
### Discussed in https://github.com/ipspace/netlab/discussions/479

<div type='discussions-op-text'>

Configure a vrf property on all leaves:
```
groups:
 leaves:
    members: [ leaf1,leaf2 ]
    device: srlinux
    module: [ bgp,evpn,vlan,vrf,vxlan ]
    config: [ leaf-policy.j2 ]

    node_data:
     vrfs:  
      customer1:
       evpn.transit_vni: 1234
       
```
</div>

This currently does not work, because the data is copied into each node after the vrf node transformation hooks get called

jbemmel:
Debug trace:
```
jeroen@jvm:~/srlinux/netsim-examples/BGP/SRL-Route-Leaking$ netlab create 
Copying node_data for global modules: ['vlan', 'bgp', 'vxlan', 'vrf', 'evpn']
Copying node_data for global modules: ['vlan', 'bgp', 'vxlan', 'vrf', 'evpn']
Copying node_data for global modules: ['vlan', 'bgp', 'vxlan', 'vrf', 'evpn']
Copying node_data for global modules: ['vlan', 'bgp', 'vxlan', 'vrf', 'evpn']
vrf.node_pre_transform on leaf1 vrfs=None
vrf.node_pre_transform on leaf2 vrfs=None
vrf.node_pre_transform on spine1 vrfs=None
vrf.node_pre_transform on spine2 vrfs=None
vrf.node_pre_transform on mplsnet vrfs=None
JvB transform_data->adjust_groups
JvB copy_group_node_data vrfs -> node leaf1
JvB copy_group_node_data vrfs -> node leaf2
JvB copy_group_node_data clab -> node spine1
JvB copy_group_node_data bgp -> node spine1
JvB copy_group_node_data clab -> node spine2
JvB copy_group_node_data bgp -> node spine2
JvB transform_data->pre_node_transform
node.vrfs: {'customer1': None, 'customer2': None, 'customer3': None}
node.vrfs: {'customer1': None, 'customer2': None, 'customer3': None, 'global': {'bgp': {'neighbors': [{'ifindex': 8, 'local_if': 'ethernet-1/51.4001', 'ipv4_rfc8950': True, 'local_as': 65001, 'name': 'leaf1', 'as': 65111, 'type': 'ebgp', 'ipv4': True, 'ipv6': True}]}}}

Traceback (most recent call last):
[...]

PR #481, opened 2022-09-21
Fix transform hook calling order; remove calls to non-existent methods

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/480

Update: modified test case to catch regressions

Note that I left some original code commented out such that people can see the changes; can be removed in final version.
Also need to update method documentation if approved

This fix centers on the ```transform``` code in augment/main.py:
```
def transform_data(topology: Box) -> None:
  addressing.setup(topology,topology.defaults)
  augment.plugin.execute('pre_transform',topology)
  modules.pre_module_transform(topology)            <-- this was calling module, node AND link pre_transform hooks...

  augment.groups.adjust_groups(topology)             <-- this copies global vrf data into nodes - but too late
  augment.topology.check_global_elements(topology)

  augment.plugin.execute('pre_node_transform',topology)
  modules.pre_node_transform(topology)             <-- ... while the intent seems to be to call them here (for nodes)
  augment.nodes.transform(topology,topology.defaults,topology.pools)
  common.exit_on_error()
  augment.plugin.execute('post_node_transform',topology)

  if 'links' in topology:
    augment.plugin.execute('pre_link_transform',topology)
    modules.pre_link_transform(topology)                <-- ... and here (for links)
    augment.links.transform(topology.links,topology.defaults,topology.nodes,topology.pools)
    common.exit_on_error()
    augment.plugin.execute('post_link_transform',topology)
```

### DIFF
--- a/netsim/augment/groups.py
+++ b/netsim/augment/groups.py
@@ -203,6 +203,7 @@ def copy_group_node_data(topology: Box) -> None:
       for name,ndata in topology.nodes.items():
         if name in g_members:
           for k,v in gdata.node_data.items():   # Have to go one level deeper, changing ndata value wouldn't work
+            print( f"JvB copy_group_node_data {k} -> node {name}" )
             if not k in ndata:
               ndata[k] = v
             if isinstance(ndata[k],dict):
--- a/netsim/augment/groups.py
+++ b/netsim/augment/groups.py
@@ -203,7 +203,6 @@ def copy_group_node_data(topology: Box) -> None:
       for name,ndata in topology.nodes.items():
         if name in g_members:
           for k,v in gdata.node_data.items():   # Have to go one level deeper, changing ndata value wouldn't work
-            print( f"JvB copy_group_node_data {k} -> node {name}" )
             if not k in ndata:
               ndata[k] = v
             if isinstance(ndata[k],dict):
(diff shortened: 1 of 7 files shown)


ipspace:
You don't get it ;) There are three places where hooks can be executed:

* Before the transformation is started
* Before the node transformation is started
* Before the link transformation is started

Each one of these places can call module-level, node-level or link-level hooks depending on what the module wants to work on. Some modules rely on calling routines to do iteration over nodes or links, other modules have to process global data structures and do the iteration (if needed) internally.

As for the "VRF value is none" -- good catch, please submit that as a separate PR.

jbemmel:
Each of these places tries to call module-, node- and link-level hooks, but none of those exist:
```
jeroen@jvm:~/srlinux/netlab/tests$ grep -r module_pre_node_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ grep -r node_pre_node_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ grep -r link_pre_node_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ grep -r module_pre_link_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ grep -r node_pre_link_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ grep -r link_pre_link_transform ../netsim
jeroen@jvm:~/srlinux/netlab/tests$ 
```

The calls at https://github.com/ipspace/netlab/blob/dev/netsim/augment/main.py#L46 and https://github.com/ipspace/netlab/blob/dev/netsim/augment/main.py#L53 have no effect; you can comment them out, and still pass all test cases

PR #482, opened 2022-09-21
Pull in vrfs associated with vlans

jbemmel:
As illustrated [here](https://github.com/ipspace/netlab/blob/dev/tests/integration/evpn/vxlan-asymmetric-irb.yml#L32) there are cases where vlans get pulled into a node without being referenced from an interface.

Currently, any vrf associated with such a vlan does not get pulled in - while I believe it should

This may be another case of some underlying issue that is better fixed elsewhere - just submitting the case and a straightforward fix to start with

This PR also ensures there is a deterministic order in which local vrfidx values get assigned, by sorting the vrfs on their global id. This avoids a dependency on the order in which vlans (with associated vrfs) appear in the file

### DIFF
--- a/netsim/modules/vrf.py
+++ b/netsim/modules/vrf.py
@@ -342,10 +342,17 @@ class VRF(_Module):
     set_import_export_rt(topology,topology)
 
   def node_pre_transform(self, node: Box, topology: Box) -> None:
+    # Check if any global vrfs need to be pulled in due to being referenced by a vlan
+    vlan_vrfs = [ vdata.vrf for vname,vdata in node.vlans.items() if 'vlans' in node and 'vrf' in vdata ]
     if not 'vrfs' in node:
-      return
+      if not vlan_vrfs:  # No local vrfs and no vlan references -> exit
+        return
+      node.vrfs = {}     # Prepare to pull in global vrfs
 
-    for vname in node.vrfs.keys():
+    for vname in set(list(node.vrfs.keys()) + vlan_vrfs):  # Filter out duplicates
+      if node.vrfs[vname] is None:
+        node.vrfs[vname] = {}
+
       if 'vrfs' in topology and vname in topology.vrfs:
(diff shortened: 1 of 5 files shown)


ipspace (commented):
Why exactly do you want to have VRFs sorted? Why does it matter?

Also, how does this fix the issue you started with (assuming it's worth fixing)?

As for "is it worth fixing": VLAN is not used ==> no VLAN interfaces ==> no VRF reference. What's the problem?

ipspace:
> As you can see in the commit history, at first I did not sort the vrfs. This lead to failed test cases, because pulling in the vrfs associated with vlans changed the order in which they got added to the nodes in some cases.

That's an interesting one -- it would be nice to understand which part of Python3 is not deterministic. I also wonder why we never hit it so far.

> I didn't like that, the way vlans are associated with nodes and the order in which they are listed should not impact the vrfidx assignment imho.

And why does that matter?

> Those are the kinds of small irregularities that can cause operational issues.

What operational issues exactly? We're talking about a lab-building tool. But OK, let's say I'm persuaded that consistency is good, so let's sort VRFs by ID.

> My use case is [here](https://github.com/jbemmel/netsim-examples/blob/master/BGP/SRL-Route-Leaking/topology.yml#L140), I'm using a custom script to generate BGP policies and need to reference all the ipv4 prefixes associated with all vlans - not just the ones on the local node. I also need the export community from each customer vrf.

Plugin.

> I could instead reference each vrf and vlan on each leaf explicitly, instead of relying on the grouping mechanism or the pulling in of the customer vrf through the vlan - but that could lead to inconsistencies. I want to define things once, and derive the corresponding configs automatically

Ah, now I understand why you try to pull VLAN and VRF data into nodes that don't need them. What you're trying to do is outside of the scope of _netlab_ VLAN and VRF configuration modules, but could be easily solved with a plugin.

ipspace (approved):
OK, I'm merging this, but if it turns out it broke anything, it's the last unneeded bit of code I'm adding to netlab purely for someone's convenience.

PR #483, opened 2022-09-22
Custom plugin to collect topology data for use by scripts

jbemmel:
Context: https://github.com/ipspace/netlab/pull/482

For users that need topology information as input to per-node configuration scripts, in cases where this information is not 'naturally' included by the core modules.

One use case is generation of custom bgp policies with prefix filters that are configured for remote vlans (on other nodes), or communities for vrfs that are not locally present


ipspace:
Revisiting this old idea: what if we have a setting that would specify what needs to be copied into Ansible 'all' group under 'system' dictionary and use that setting when creating Ansible inventory?

You could then use the 'combine' filter in Jinja2 templates if you want to have a merged view of local and system VLANs (for example), or just peek into system settings if you're so inclined.

jbemmel:
Closing as no longer needed

PR #484, opened 2022-09-22
evpn/vxlan vni enhancements

jbemmel:
Allow users to specify ```vni: False``` for vlans to avoid auto-allocation of a vni.
This simplifies the implicit construction of vxlan.vlans in topologies with large numbers of vlans and/or varying vlans per node

Honor the global vlan.auto_vni setting: When ```False``` don't auto-assign vnis (unless requested using ```vni: True```)

* Fix type check on __name__
* Use 'data.get_from_box' to get evpn.transit_vni, for asymmetric case evpn: can be None (evpn enabled but no transit_vni)
* Check transit VNIs against list of L2 VNIs from vlan module, error if overlap
* Allow explicit ```vni: True``` similar to transit_vni allocation


ipspace (approved):
OK, I will merge this and then clean up the code and amend the commit. It will be faster this way.

Issue #446, opened 2022-09-14
Disable a routing protocol per VRF or interface


ipspace:
Based on #443 - we need a mechanism to disable a routing protocol (OSPF, IS-IS, BGP...) for an interface or for a whole VRF.

While netlab does a good job handling most of the designs (automatically pruning routing protocols based on adjacent node capability), there might be an edge case where you might not want to have EBGP sessions across an interface even though the attached devices are BGP speakers, or might not want an interface being part of an OSPF process (even though marking it **passive** would do the trick).

A crazy scenario that immediately comes to mind is a VRF Lite case where you'd want to run OSPF across one L3 domain and BGP across another one. Not that I would ever consider doing something along these lines, but the amount of weirdness found in the wild never stops surprising me.

jbemmel:
The vrf module has an "informal" ```ospf.activate``` parameter; it is [documented](https://github.com/ipspace/netlab/blob/dev/docs/module/vrf.md#interaction-with-routing-protocols) but not listed in topology-defaults as a valid (vrf) attribute

We could do something similar for bgp and isis, instead of ```bgp: False``` as proposed by https://github.com/ipspace/netlab/pull/443

ipspace:
Looking at some other code, it seems we can already set **some_protocol: False** on links/interfaces (see `tests/topology/input/vrf-igp.yml`). Have to check how generic that is -- maybe all we need is to a check in the BGP neighbor processing routing.

ipspace:
> The vrf module has an "informal" `ospf.activate` parameter; it is [documented](https://github.com/ipspace/netlab/blob/dev/docs/module/vrf.md#interaction-with-routing-protocols) but not listed in topology-defaults as a valid (vrf) attribute

To be pedantic -- we have `ospf.active` and `bgp.activate` ;), but yeah, maybe we should rename that to `ospf.enable`.

`ospf.active` does not have to be listed as a valid VRF attribute as we don't check VRF module attributes yet... but maybe we should.

jbemmel:
I see how this works per interface, but does it include the ability to disable it per vrf?
It also seems this feature isn't documented yet?

ipspace:
> I see how this works per interface, but does it include the ability to disable it per vrf?

What does your minimal test topology say? ;)

> It also seems this feature isn't documented yet?

https://netsim-tools.readthedocs.io/en/dev/module/routing.html#routing-disable

Guess we could add a few links to that section ;))

jbemmel:
[...]

PR #488, opened 2022-09-24
Disable routing protocols per vrf or interface - allow bgp: False

jbemmel:
Revisiting https://github.com/ipspace/netlab/issues/446, data-model only changes with no impact on templates

* Allow user to specify ```bgp: False``` at vrf, interface or link levels
* Includes ```vrf.bgp: False``` to disable bgp within any vrfs by default (unless overridden)

This PR is for bgp only, something similar could be implemented for the other routing protocols (when approved)

Adds a 'cleanup' hook for removing bgp: False flags, could also be done generically ( i.e. run code that removes any attributes set to 'False'). Alternatively, bgp module could iterate internally over nodes in module_post_transform, and cleanup afterwards (without adding a hook)

As to why this is needed: Even a simple topology with 2 leaves currently gets ebgp sessions on customer facing vxlan vlans

If you prefer ```bgp.disable: True``` instead, that's also possible

### DIFF
--- a/netsim/modules/__init__.py
+++ b/netsim/modules/__init__.py
@@ -81,6 +81,7 @@ def post_transform(topology: Box) -> None:
   module_transform("post_transform",topology)
   node_transform("post_transform",topology)
   link_transform("post_transform",topology)
+  module_transform("post_transform_cleanup",topology) # JvB: added to cleanup data model after all transformations are done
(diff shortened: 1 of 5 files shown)


ipspace:
Implemented. See #446 for details

PR #493, opened 2022-09-26
Error on attempt to modify global vrf or vlan attributes through group node_data

jbemmel:
Context: https://github.com/ipspace/netlab/issues/480

While waiting for this issue to be fixed, warn users about attempts to modify global vlans or vrfs through node_data in a group.
Due to the order of processing in various hooks, this currently does not work (and may result in obscure errors like https://github.com/ipspace/netlab/pull/474)

### DIFF
--- a/docs/groups.md
+++ b/docs/groups.md
@@ -166,6 +166,8 @@ groups:
 nodes: [ l1, l2, l3, s1, a1, a2, a3 ]
 ```
 
+Note that modifying attributes for global vlans or vrfs through node_data is currently not supported; an error is reported to the user instead.
+
 ## Setting Device Type or List of Modules in Groups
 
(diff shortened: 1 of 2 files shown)


ipspace:
Another one of those quick fixes that will NOT be accepted. Instead of figuring out what works and what doesn't, and documenting that, you went for a quick fix.

As for the need for the countless hours we wasted on this topic. This topology:

```
module: [ vrf,ospf ]

groups:
  g1:
    members: [ r1, r2 ]
    node_data:
      vrfs:
        red:

nodes:
  r1:
  r2:

links:
- r1:
  r2:
  vrf: red
```

... generates a very specific error message:

```
MissingValue in vrf: VRF red used on an interface in r1 is not defined in the node or globally
```

It would take a very focused mind to understand that as "but it has to work if it's defined in the group".

ipspace:
Implemented what I hope is a complete and working solution in de2232e9a4e926d9ecac16c056f62b36af7647d1

jbemmel:
Thanks, much appreciated!

Issue #496, opened 2022-09-26
Node local vrf import/export policies


jbemmel:
In the context of https://github.com/ipspace/netlab/commit/de2232e9a4e926d9ecac16c056f62b36af7647d1

At https://github.com/jbemmel/netsim-examples/tree/evpn-vxlan-to-hosts-v2/BGP/EVPN-VXLAN-to-hosts there is a use case for VRF leaking. The intent is to do VRF route leaking to 'internet' **on the frr nodes only**, not everywhere the vrf exists.

To achieve this, 'internet' is defined as a global vrf https://github.com/jbemmel/netsim-examples/blob/evpn-vxlan-to-hosts-v2/BGP/EVPN-VXLAN-to-hosts/topology.yml#L37 but the import/export attributes are set as 'spoke' for the frr group (https://github.com/jbemmel/netsim-examples/blob/evpn-vxlan-to-hosts-v2/BGP/EVPN-VXLAN-to-hosts/topology.yml#L84) and as 'hub' for a hub group (spine only): https://github.com/jbemmel/netsim-examples/blob/evpn-vxlan-to-hosts-v2/BGP/EVPN-VXLAN-to-hosts/topology.yml#L100

With the new code this is not allowed:
```
jeroen@jvm:~/srlinux/netsim-examples/BGP/EVPN-VXLAN-to-hosts$ netlab create
IncorrectValue in vrf: Cannot redefine vrfs attribute export for vrfs.internet from node_data in group hub
IncorrectValue in vrf: Cannot redefine vrfs attribute import for vrfs.internet from node_data in group hub
Fatal error in netlab: Cannot proceed beyond this point due to errors, exiting
```

From the documentation I get the impression that group level import/export policies get promoted to global attributes, and I'm wondering if that is always correct. The VRF id needs to be global, but import/export policies can vary per node (or group of nodes) - for example in case of a hub/spoke topology (with asymmetrical swapped import/export targets)

Sample hub/spoke topology:
```
vrfs:
 some_vrf:

nodes:
 spine:
  vrfs:
   some_vrf:
    export: [ 65000:1 ]
    import: [ 65000:0 ]

 leaf:
  vrfs:
   some_vrf:
    export: [ 65000:0 ] # Swapped
    import: [ 65000:1 ]
```
If a user wants global import/export policies, they can (and should) define them on a global vrf (imho)

jbemmel:
[...]

PR #498, opened 2022-09-27
evpn hub/spoke support for symmetric irb

jbemmel:
Addresses https://github.com/ipspace/netlab/issues/496

Noted your comment on consulting with EVPN experts, I'm ok if you prefer to wait for that before merging. Just wanted to have this available so we know the changes we're talking about

Note that I encountered an issue with the evpn module using vlan module global variables that hadn't been initialized - PR fixes that as well

### DIFF
--- a/docs/module/evpn.md
+++ b/docs/module/evpn.md
@@ -88,7 +88,12 @@ The default value of VRF EVPN Instance identifier is the VLAN ID of the first VL
 
 IRB is configured whenever EVPN-enabled VLANs in a VRF contain IPv4 or IPv6 addresses:
 
-* Asymmetric IRB requires no extra parameters[^NS]
+* Asymmetric IRB requires no extra parameters[^NS], but requires instantiating all vxlan-enabled vlans on all devices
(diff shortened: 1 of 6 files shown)


ipspace:
Not again :((

There are numerous topologies one could play with once we get this up and running, hub-and-spoke is just one of them, and what you created is NOT hub-and-spoke (because spokes can still talk to each other) but something like overlapping VPNs or Common Services VPN (I know of a great 20+-year old book describing various VPN topologies ;).

Yet again you created code that deals only with your particular use case and expects that all non-trivial VPN topologies will be using the same import/export template.

Finally, there's a reason I'm using a function to initialize global variables. Without that, the test suite wouldn't (or didn't?) work correctly, as the VLAN sets are not reinitialized between tests. I might be clumsy or stupid, but sometimes I do things for a reason.

jbemmel:
There is a reason I needed to fix the vlan global variable initialization code:
```
defaults.device: srlinux
provider: clab

vrfs:
 hub:
  evpn.transit_vni: 200000
  import: [spoke]
  export: [hub]

 spoke:
  evpn.transit_vni: 200000 
  import: [hub]
  export: [spoke]

module: [vrf,bgp,evpn]

bgp.as: 65000

nodes:
  r1:
  r2:

links:
- r1:
   vrf: hub
  r2:
   vrf: spoke
```

The above topology (also included in the PR) fails because the vlan module is not properly initialized. I will submit a separate PR focused on this issue, handling the re-initialization for the test cases as well

Issue #499, opened 2022-09-27
OSPF interfaces not removed from l2-only vrf in vxlan-vlan-bundle test case


jbemmel:
See https://github.com/ipspace/netlab/blob/dev/tests/integration/evpn/vxlan-vlan-bundle.yml

In the resulting output, ospf is enabled on the svi interfaces inside the 'bundle' vrf:
```
bundle:
    export:
    - '65000:1'
    id: 1
    import:
    - '65000:1'
    ospf:
      active: true
      area: 0.0.0.0
      interfaces:
      - bridge_group: 1
        ifindex: 4
        ifname: irb0.1000
        name: VLAN red (1000) -> [h1,h2,s2]
        neighbors:
        - ifname: eth1
          ipv4: 172.16.0.1/24
          node: h1
        - ifname: eth1
          ipv4: 172.16.0.2/24
          node: h2
        - ifname: irb0.1000
          node: s2
        ospf:
          area: 0.0.0.0
          passive: false
        type: svi
        virtual_interface: true
        vlan:
          mode: bridge
        vrf: bundle
```

However, these interfaces have no IP address

PR #500, opened 2022-09-27
Don't enable routing protocols on vrf interfaces that are l2-only

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/499

Use case: VLAN bundle service


ipspace:
Yet again you fixed the symptom, not the root cause:

* Symptom: l2-only VRF interfaces are in VRF OSPF process
* Root cause: we migrated the interfaces into VRF OSPF before doing AF-based cleanup.

The root cause fix (in 06ed6f92f53c7d61c20ddbb34a269a5a22a7bb73) swaps the order of operation -- cleanup first, then migration into VRF.

PR #501, opened 2022-09-27
Fix initialization of the vlan module for topologies in which it is not used explicitly

jbemmel:
Cherry-picked from https://github.com/ipspace/netlab/pull/498

The evpn module references the global list of VNI IDs to check for duplicates. When the vlan module isn't included, those don't get initialized


ipspace:
Here's an interesting challenge: figure out why the proposed fix is a bad idea.

ipspace:
reentry-safe implementation in 72441c6b596daf5ca91a366df1e15622a0c8d064

PR #502, opened 2022-09-27
Add global evpn.enforce_unique_transit_vni flag (default: True)

jbemmel:
Default: True enforces globally unique evpn.transit_vni values, which is the more common and sane default.
Advanced users can set this flag to False to skip this sanity check


jbemmel:
Replaced by #504

PR #503, opened 2022-09-28
srlinux vlan bundle support

jbemmel:
* Update srlinux evpn templates to support vlan aware bundle service
* Modify all templates to check evpn.rd and build individual vlans if defined, instead of evpn.evi
* Generate unique EVI per vlan, also in bundle (needed for srlinux, not used by other platforms)
* Update integration example to enable evpn and set vlan bundle service flag
* Enable asymmetric irb support (tested with integration/evpn/vxlan-asymmetric-irb.yml, though h4 lacks an svi interface)

evpn.rd is a better distinguisher than evpn.evi


ipspace:
No.

You might remember we added EVI to the code because SR Linux needed it. At that time I found out that "EVI" (which is not something that would be in BGP updates) stands for "EVPN Instance".

VLAN bundle is clearly an EVPN instance, and if someone designing SR Linux API didn't get that memo (or needs a unique ID for whatever internal reason), I couldn't care less. Please pull it out of thin air, it stays within the box anyway.

jbemmel:
Replaced by #507

PR #505, opened 2022-09-28
Support for RouterOS 7

ssasso:
Support for Mikrotik RouterOS7, with:

* Initial config (and config fetch)
* VLANs
* OSPF (v2 and v3)
* BGP
* VRF-lite
* Route-Leaking w/ BGP
* MPLS LDP and L3VPN (vpnv4)
* ebgp.utils plugin support

**NOTE**: with the current ROS7, route-leaking is working on the control plane, but not on the dataplane.

**NOTE**: Additionally, on ROS7 7.5, and 7.6beta8, the MPLS forwarding plane seems totally fucked up with *virtio* networking (while the control plane works fine) - so let's use *e1000*. Wrote it on the platform caveats.


ssasso:
ready for review ;)

ipspace (approved):
Awesome job. Thanks a million!

Obviously I can't comment on whether the templates work, the doc fixes are OK ;))

PR #507, opened 2022-09-28
srlinux vlan bundle service take2

jbemmel:
Supersedes https://github.com/ipspace/netlab/pull/503

SR Linux treats a VLAN bundle as a set of EVPN services, one per VNI. A unique RD is assigned per service, auto-generated based on the (locally assigned) EVI (== vlan ID) and the node's router ID.

The RD's assigned by Netlab may require further thought/testing, e.g. in case of multi-homed VLAN bundles with ECMP

Note the minor fixes in case user provided evpn: is ```None```


ipspace (approved):
Thank you! Will change the 'evpn is None' tests to use the new 'must_be_dict', so we'll catch other stupidities like 'evpn: yeah' ;) and amend the merge commit.

PR #517, opened 2022-09-29
Proposed evpn ebgp enhancements for RT value generation

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/513

This example manually configures the EVPN RTs in a way that works in multi-vendor cases.

On SR Linux there is an ```inter-as-vpn``` knob that needs to be set in order for the spines to keep the EVPN routes (which do not match any local VXLAN interfaces, because there are none)


ipspace:
Will add this in the EVPN refactoring process. The idea is to generate RTs for VLANs earlier in the EVPN processing, and do it globally instead of the current hack where we effectively run the same logic on every node.

RD would be generated on individual nodes to support the **router_id:vlan_id** paradigm recommended by EVPN RFCs. The global code could set evpn.rd to True, and that value would then be replaced on individual nodes.

Have you encountered any documentation on what the EVPN RD should be for a VRF? I don't want to worry about overlaps between vrf.id and vlan.id address spaces.

jbemmel:
On SR OS and SR Linux it's ```{ router_id }:{ EVI }```

We can set ```EVI``` to vlan.id, and vrf.id + 4094 for VRFs (including the ones that represent VLAN bundles for example)

ipspace:
I'll keep that in mind -- evpn.evi has to match whatever is in the low-order bits of the RD.

However, keep in mind that if we use **router_id:something** format, RD is locally significant, so nobody cares what you use. Route targets are obviously a different story.

ipspace:
I'm guessing this is solved in the new EVPN code. RDs don't matter and RTs are consistent across the lab topology. Should we close this one?

ipspace:
Looks like we solved this one with code changes in 1.4

PR #523, opened 2022-09-30
Avoid same subnet routes

jbemmel:
When users declare a 10.2.0.0/24 pool, the route add fails for obscure reasons (overlap with default vrf_loopback pool, even when not used)

In general, route add fails when the prefix subnet is the same as the host's own subnet

### DIFF
--- a/netsim/ansible/templates/initial/linux/vanilla.j2
+++ b/netsim/ansible/templates/initial/linux/vanilla.j2
@@ -51,7 +51,9 @@ ip link set {{ l.ifname }} mtu {{ l.mtu }}
 #
 {% for ifdata in interfaces|default([]) if ifdata.gateway is defined %}
 {%   for name,pool in pools.items()|default({}) %}
-{%     for af,pfx in pool.items() if af == 'ipv4' and name != 'mgmt' and name != 'router_id' %}
+{%     for af,pfx in pool.items() if af == 'ipv4' and
+         name not in ['mgmt','router_id'] and
+         pfx|ipaddr('network') != ifdata.gateway.ipv4|ipaddr('network') %}
 set +e
 ip route del {{ pfx }} 2>/dev/null
 set -e


ipspace (commented):
So you're saying that A. User used a reserved prefix (because it's an addressing pool) on a LAN link, and then said user got surprised when the static route for the reserved prefix failed to install?

I have no problem with the "let's make sure we're not installing a static route to the interface subnet" check, but shouldn't we rather stop the above-mentioned user from creating an overlap in the first place?

We should either give the users all the rope they need to hang themselves, or stop them from hanging themselves, but not surreptitiously take the rope away 😁

jbemmel:
I considered that too, but it gets tricky. In this case the vrf_loopbacks subnet isn't used, so there is no overlap in practice.
Even if vrf loopbacks were used, the user might be ok with not being able to reach them from Linux hosts, and reuse the same addressing space.

Lastly, we allow users to allocate overlapping static IPs, so why not overlapping subnets too?

A warning might be appropriate though

ipspace:
> Even if vrf loopbacks were used, the user might be ok with not being able to reach them from Linux hosts, and reuse the same addressing space.

Please note _netlab_ is a **lab** building tool, and I wouldn't worry too much about reuse of private address space _within a lab_. If someone wants to use it to replicate a real network, them maybe that someone should throw resources at this problem to make the tool more resilient ;))

> Lastly, we allow users to allocate overlapping static IPs, so why not overlapping subnets too?

We also don't try to stop users from shooting themselves in their feet (or wherever) when using overlapping static IP, so I don't understand why we should worry about the state of their appendages in this case.

> A warning might be appropriate though

Agreed. Please open an issue and when I get around to implementing warnings I'll make sure this one is added to the list.

ipspace:
BTW, you insisted on using FRR as a host a while ago. Did you fix this in FRR setup as well?

PR #524, opened 2022-09-30
Normalize IP address strings, check pool prefixes for host bits

jbemmel:
An address pool like 10.0.1.0/20 is currently accepted, but leads to errors when provisioning static routes on Linux hosts


jbemmel:
This patch introduces an issue when pfx=True - I'll submit a fix

PR #531, opened 2022-10-01
Refactor link transformation and interface addressing code

ipspace:
* Remove 'left' and 'right' elements from P2P links
* Implement multiple IPAM policies and select them based on prefix/link size or link type
* Remove P2P link augmentation and refactor/cleanup LAN link augmentation

Anyone brave enough to review the new spaghetti mess?

### DIFF
--- a/netsim/addressing.py
+++ b/netsim/addressing.py
@@ -223,10 +223,10 @@ def get_nth_subnet(n: int, subnet: netaddr.IPNetwork.subnet, cache_list: list) -
     cache_list.append(next(subnet))
   return cache_list[n-1]
 
-def get_pool_prefix(pools: typing.Dict, p: str, n: typing.Optional[int] = None) -> typing.Dict:
-  prefixes: typing.Dict = {}
+def get_pool_prefix(pools: typing.Dict, p: str, n: typing.Optional[int] = None) -> Box:
+  prefixes = Box({})
   if pools[p].get('unnumbered'):
-    return { 'unnumbered': True }
+    return Box({ 'unnumbered': True })
   for af in list(pools[p]):
     if not 'cache' in af:
       if isinstance(pools[p][af],bool):
@@ -254,14 +254,14 @@ def get_pool_prefix(pools: typing.Dict, p: str, n: typing.Optional[int] = None) 
 
   return prefixes
 
-def get(pools: Box, pool_list: typing.Optional[typing.List[str]] = None, n: typing.Optional[int] = None) -> typing.Dict:
+def get(pools: Box, pool_list: typing.Optional[typing.List[str]] = None, n: typing.Optional[int] = None) -> Box:
   if not pool_list:
     pool_list = ['lan']                   # pragma: no cover
   p = get_pool(pools,pool_list)
   if p:
     return get_pool_prefix(pools,p,n)
   else:
-    return {}                             # pragma: no cover -- can't figure out how to get here
+    return Box({})                        # pragma: no cover -- can't figure out how to get here
 
 def setup(topo: Box, defaults: Box) -> None:
   data.null_to_string(topo.addressing)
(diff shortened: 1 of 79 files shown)


netsim/addressing.py
```diff
@@ -280,33 +280,38 @@ def setup(topo: Box, defaults: Box) -> None:
...
+    if isinstance(pfx,bool):
+      prefix_list[af] = pfx
+      continue
+
+    try:
+      prefix_list[af] = netaddr.IPNetwork(pfx)
+    except Exception as ex:
+      common.error(f'Cannot parse {af} prefix: {prefix}\n... {ex}',common.IncorrectValue,'addressing')
+      return empty_box
+    if str(prefix_list[af]) != str(prefix_list[af].cidr):
```
jbemmel, line 311: Without normalization this check will erroneously detect host bits.
For example, "10.00.0.0/24" != "10.0.0.0/24" so the error gets triggered

ipspace: prefix_list is already `netaddr.IPNetwork` and so `str(prefix_list)` is normalized:

```
>>> import netaddr
>>> x = netaddr.IPNetwork('10.000.00.0/24')
>>> str(x)
'10.0.0.0/24'
```

netsim/augment/links.py
```diff
@@ -288,155 +214,290 @@ def augment_link_prefix(link: Box,pools: typing.List[str],addr_pools: Box,link_p
...
-        # List enabled modules that have interface level attributes; copy those attributes too
-        mods_with_ifattr = Box({ m : True for m in ndict[remote_if['node']].get('module',[]) if defaults[m].attributes.get('interface',None) })
-        ifaddr_add_module(ngh_data,remote_if['data'],mods_with_ifattr)
-
-        node_if['data'].neighbors.append(ngh_data)
+IPAM_dispatch: typing.Final[dict] = { 
+    'unnumbered': IPAM_unnumbered,
+    'p2p': IPAM_p2p,
+    'sequential': IPAM_sequential,
+    'id_based': IPAM_id_based
```
jbemmel, line 335: 'error' could be an entry in this list too

netsim/augment/links.py
```diff
@@ -288,155 +214,290 @@ def augment_link_prefix(link: Box,pools: typing.List[str],addr_pools: Box,link_p
...
+          f'Cannot parse {af} prefix {pfx_list[af]} on link#{link.linkindex}\n' + \
+            common.extra_data_printout(f'{ex}') + '\n' + \
+            common.extra_data_printout(f'{link}'),
+          common.IncorrectValue,
+          'links')
+        continue
+
+      allocation_policy = get_prefix_IPAM_policy(link,pfx_net,ndict)      # get IPAM policy based on prefix and link size
+
+    if allocation_policy == 'error':                                      # Something went wrong, cannot assing IP addresses
```
jbemmel, line 381: would suggest to include an 'error' policy instead of handling separately here

netsim/augment/links.py
```diff
@@ -288,155 +214,290 @@ def augment_link_prefix(link: Box,pools: typing.List[str],addr_pools: Box,link_p
...
+    print(f'... LAN link data: {link}\n')
 
-  return link
+  # Second phase: build neighbor list from list of newly-created interfaces
+  for node_if in interfaces:
+    ifdata = node_if['data']                                      # Get a pointer to interface data
+    ifdata.neighbors = []
+    node_if['__me'] = True                                        # Set 'that's me' flag so we know which interface to skip
+    for remote_if in interfaces:                                  # Iterate over all interfaces created from this link
+      if '__me' in remote_if:                                     # ... and skip the current interface
```
jbemmel, line 483: if node_if == remote_if:
  continue
why set a flag?

jbemmel (approved):
Some minor comments, no show stoppers

netsim/addressing.py
jbemmel: ok, just something I hit recently

netsim/augment/links.py
ipspace: This replaced a very long condition we had in place to deal with loopback links, but I like the suggestion. Will have to use "a is b" though, dicts (or boxes) with identical contents are equal

```
>>> a = { 'x': 1, 'y': 2 }
>>> b = a
>>> c = { 'x': 1, 'y': 2 }
>>> a == c
True
>>> a is c
False
>>> a is b
True
```

netsim/augment/links.py
ipspace: I had that same idea initially, but then it seemed cleaner to handle the error case separately instead of hiding the error message in a routine called through a lookup table.

netsim/augment/links.py
ipspace: Fixed the code based on your suggestion. Thanks again!

PR #532, opened 2022-10-01
Add anycast gateway support (dev-1.4 branch)

jbemmel:
This PR adds an 'anycast_gateway' attribute to address pools and vlan prefixes, allowing the user to specify an address by index or IP.
By default, 'True' picks the last available address in each subnet (to minimize chance at overlap with auto-assigned ips)

Use of an anycast_gateway associated with a pool or vlan can be overridden using 'anycast_gateway: False' at the interface

The anycast_gateway translates to an ipv4/ipv6 address that is passed to templates for configuration of devices. The SR Linux templates have been
updated to use these parameters. A device feature flag features.initial.anycast_gateway controls the appearance/support of these flags

For Linux hosts, the default gateway is set to the anycast_gateway_ipv4 address if available on the first non-host interface on the link

Notes:
* The resulting data model could be trimmed, leaving only anycast_gateway_ipv4/6 on node interfaces (and not in prefixes or vlans)
* Further data model changes could be made to provide better control over which IPs are primary/secondary etc., templates assumes non-anycast IP as primary to support BGP sessions to that IP
* Includes draft support for Cumulus and EOS (untested)

### DIFF
--- a/netsim/addressing.py
+++ b/netsim/addressing.py
@@ -81,6 +81,9 @@ def rebuild_prefix(pfx: typing.Union[dict,Box]) -> dict:
   for af in ('ipv4','ipv6'):
     if af in pfx:
       out_pfx[af] = str(pfx[af]) if not isinstance(pfx[af],bool) else pfx[af]
+    gw = 'anycast_gateway_' + af
+    if gw in pfx:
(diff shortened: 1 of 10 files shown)


jbemmel:
Rebased in #551

PR #533, opened 2022-10-02
sros mpls ldp support

jbemmel:
Works with https://github.com/jbemmel/netsim-examples/tree/master/EVPN/mpls-bridging

Assumes at least 1 interface has ldp enabled, can restructure to work with 0 interfaces too


ipspace (changes requested):
I think it would be better to make this work in all cases, including no LDP interfaces. Just move the beginning of the **ldp:** data structure into the **for** loop and add **if loop.first** around it.

PR #534, opened 2022-10-02
Refactor dataplane ID-allocation code (preparing for #509)

ipspace:
Dataplane ID-allocation code has been moved from VLAN module into new _dataplane module. VLAN IDs and VNIs are now allocated using that code, as are the VRF IDs and RDs.

As the next step, the VNI allocation code will be moved in the VXLAN module, and the EVPN module will register static VNIs with the VNI set.

### DIFF
--- a/netsim/modules/_dataplane.py
+++ b/netsim/modules/_dataplane.py
@@ -1,0 +1,67 @@
+#
+# Generic data plane routines
+#
+import typing
+from box import Box
+
+from .. import common
+from ..data import global_vars,get_from_box
+
+"""
+build_id_set: given an object (topology or node) and a data structure within the object,
+create a set of attributes used in that object
+
+* obj: parent object (topology or node)
+* dsname: name of the data structure (vlans, vrfs)
+* attr: name of the attribute we're interested in (vlan, vni, rd, import, export)
+* objname: name of the parent object in case we have to thrown an error message
+"""
+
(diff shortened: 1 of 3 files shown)


netsim/modules/_dataplane.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+
+def build_id_set(obj: Box, dsname: str, attr: str, objname: str) -> set:
+  if dsname in obj:
+    if not isinstance(obj[dsname],dict):    # pragma: no cover
+      common.fatal(f'Found a {dsname} setting that is not a dictionary in {objname}','dataplane')
+      return set()
+
+    return { 
+      v[attr]
```
jbemmel, line 27: We may want to use int( v[attr] ) here, to normalize ids

Otherwise, if a user does
```
vlans:
  vlan1:
   id: "123"
```

the check in is_vlan_id_used(123) below will fail

ipspace: Have to check where the `vlan.id` is checked. In any case, int('a') would crash in case the user does `id: a`, and as we use the same function to collect RDs, we can't cast to **int** anyway.

ipspace: ... not a big deal, in `validate_vlan_attributes` we check that the `vlan.id` is an **int** and report an error if not. At that point nobody cares what invalid VLAN IDs were already assigned ;)

PR #535, opened 2022-10-02
Move VNI allocation into VXLAN module (implements # 509)

ipspace:


jbemmel (approved):
Should we also move ```vni``` to ```vxlan.vni``` (and deprecate vlan.vni)?

ipspace:
Deprecating **vni** attribute would require changes in all VXLAN configuration templates. Not sure it's worth it.

However, I'm all for consistency, so I have no problems supporting (and documenting) **vxlan.vni** attribute that gets copied into **vni** attribute during the **pre_default** hook ;) as the first step in a potential migration process.

PR #537, opened 2022-10-04
Proposed evpn data model changes for mpls/vxlan transport

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/511

- build evpn.vlans list
- add evpn.mode: symmetric_irb | asymmetric_irb (instead of transit_vni)
- determine evpn.transport

Sample L2 and L3 topologies included

Note: Not entirely sure about the interaction between evpn.vlans and vxlan.vlans. For backwards compatibility the latter must be processed, and since you can have vxlan topologies without EVPN both need to exist


ipspace:
Please rebase this one so I can see the actual changes. I plan to refactor EVPN and add MPLS transport in the next few days (see commits in **evpn** branch); I had to document what's going on between the four tightly-coupled modules first to figure out the best place to insert modified EVPN logic.

PR #542, opened 2022-10-04
sros ospf interface model fix

jbemmel:
YANG model does not like multiple entries with same area-id -> split into separate paths


ipspace:
If you feel this needs to be fixed soon, please submit a PR agains **dev** branch. Once that one accumulates enough fixes, I'll push out 1.3.2-post1.

jbemmel:
Submitted against dev branch instead

PR #545, opened 2022-10-04
Avoid adding name: {} to links in debug mode

jbemmel:
When running with --debug, the reference to l.name sets l.name = {} which is not a string


ipspace:
Thank you!

PR #546, opened 2022-10-04
Adds a plugin that enables proxy ARP on all interfaces in selected VRFs

jbemmel:
Users can set ```proxy_arp: True``` for each VRF where proxy ARP is required

Platforms supported: srlinux, sros


ipspace:
This could be added as a core feature once we have the other bits and pieces in place. I'm guessing it belongs to the EVPN configuration, right?

PR #552, opened 2022-10-06
Add a feature flag for advanced RR topologies with combined IBGP/EBGP peers

jbemmel:
Context: https://github.com/ipspace/netlab/commit/be1c546aeb029960f231f3382ccf0768c918656a

On EOS this requires policy support to selectively set next-hop only for external routes
SR Linux (and possibly others) do not support such policies, or do not have them implemented in current templates

For SR Linux, do not set next-hop-self when operating as Route Reflector

TODO: Set flag for platforms that do correctly support this scenario

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -18,7 +18,7 @@
 
 - path: routing-policy/policy[name={{name}}]
   val:
-   default-action: 
+   default-action:
     reject: { }
    statement:
 {% if is_import %}
@@ -28,7 +28,7 @@
       bgp:
        community-set: "C{{ c|replace(':','_') }}"
      action:
-{{    accept() }}       
+{{    accept() }}
 {%  endfor %}
 {% else %}
    - sequence-id: 5
@@ -129,7 +129,6 @@
 {% if 'ibgp' in type %}
    import-policy: accept_all
    export-policy: accept_all
-   next-hop-self: {{ vrf_bgp.next_hop_self|default(False) }}
    peer-as: {{ neighbor.as }}
    transport:
     local-address: {{ transport_ip }}
@@ -139,6 +138,8 @@
    route-reflector:
     cluster-id: {{ vrf_bgp.rr_cluster_id|default(False) or router_id }}
     client: True
+{%  else %}
+   next-hop-self: {{ vrf_bgp.next_hop_self|default(False) }}
 {%  endif %}
 {% endif %}
 {% endmacro %}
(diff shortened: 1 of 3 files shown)


jbemmel:
I did run the BGP integration test cases for SR Linux, the one with x1/x2/x3 requires bgp policy support to selectively set next hops only on external routes, but not on reflected routes. That kind of policy action ( i.e. setting next hop ) is not supported on SR Linux (yet)

Specifically, the next hop for x3 (as sent to r1/r2) is wrong because on the rr one has to choose between next_hop_self=True or next_hop_self=False (default, i.e. follow the RFC). You cannot say next_hop_self=True for route to x3, but not for other routes

ipspace:
Most platforms based on "industry standard CLI" (😜) have a **next-hop-self** neighbor parameter that does not modify reflected routes, so the EBGP-on-RR test case works correctly for them. I know it works that way on IOSv, IOS/XE, Nexus OS, Cumulus Linux (and thus also FRR). VyOS probably works the same way as it runs FRR behind the scenes.

Arista EOS and Junos use route maps (there's no other way on Junos, and EOS is broken). I have no idea what RouterOS and OS10 are doing -- @ssasso could you check the `bgp/ibgp-ebgp` integration test case on them?

If it turns that that only a platform or two have this challenge, I'd rather solve that in the _device quirks_ 87b57c0186bde46b3beec380529cfa8e203bd5fa -- we don't have to set the flag for all platforms and modify common code just because a platform or two don't behave like the rest of them.

jbemmel:
```
A:rr# next-hop-self  
usage: next-hop-self

When set to true, the next-hop in all IPv4-unicast, IPv6-unicast and EVPN BGP routes advertised to all IBGP peers in the peer-group is set equal to the local-address used on each session (or to the router ID if the NLRI is IPv6 and there is no IPv6 local address to use). **This is independent of the route origin** (EBGP, IBGP-client, IBGP-non-client or redistributed direct/static/aggregate route).

When set to false, normal BGP rules from RFC 4271 apply.

[true|false, default false]
```
So on SR Linux next-hop-self affects reflected routes too

ipspace:
> So on SR Linux next-hop-self affects reflected routes too

I didn't doubt that ;) I'm just saying that so far we know of one platform that has this particular limitation and two unknowns, so maybe it's better to implement the safeguards as a device quirk instead of a feature flag.

ssasso:
> Arista EOS and Junos use route maps (there's no other way on Junos, and EOS is broken). I have no idea what RouterOS and OS10 are doing -- @ssasso could you check the `bgp/ibgp-ebgp` integration test case on them?

The integration test works smootly for OS10...

For RouterOS 7 I may have discovered an "implementation" bug on my side which I'm still investigating.

ssasso:
> For RouterOS 7 I may have discovered an "implementation" bug on my side which I'm still investigating.

Also ROS 7 integration test works fine after #570

ipspace:
OK, it looks like (almost) every other platform handles **next-hop-self** on route reflectors in one way or another, so it's time for device quirks ;)

PR #555, opened 2022-10-07
EVPN refactoring

ipspace:
The refactored code works with all previous test cases. It's time to merge it into dev-1.4 and add new functionality from there.


netsim/modules/evpn.py
```diff
@@ -6,6 +6,39 @@
 from .. import data
 from ..augment import devices
 
+"""
+validate_evpn_list
+
+* Create vxlan.vlans list with all known VLANs if the attribute is missing
+* validate that the vxlan.vlans list is a list with valid local or global VLAN names
```
jbemmel, line 13: evpn.vlans

netsim/modules/evpn.py
```diff
@@ -16,52 +49,62 @@ def enable_evpn_af(node: Box, topology: Box) -> None:
...
   data.must_be_int(
     evpn,'evi',epath,
     module='evpn',
     min_value=1,max_value=65535)                                    # Check EVI data type in range
-  if not 'rd' in evpn:                                              # Default RD value
-    evpn.rd = f'{node.bgp.router_id}:{vlan.id}'
   for rt in ('import','export'):                                    # Default RT value
     if not rt in evpn:                                              # ... BGP ASN:vlan ID
-      evpn[rt] = [ f"{node.bgp['as']}:{vlan.id}" ]
+      evpn[rt] = [ f"{asn}:{vlan.id}" ]
```
jbemmel, line 77: vlan.id -> evi.id

In case the user manually defines an evi value different from the vlan.id

netsim/modules/evpn.py
```diff
@@ -127,6 +181,9 @@ def vrf_transit_vni(topology: Box) -> None:
       continue  
     vni_list.append( vni )                                      # Insert it to detect duplicates elsewhere
 
+  if vni_error:                                                 # Found serious errors, makes no sense to continue
```
jbemmel, line 184: if evpn_transport != "vxlan" also return, no vnis to assign

jbemmel (approved):
Could add a mpls-bridging integration test case

I think we may still want to support an implied evpn.vlans list, containing all vlans with "evpn:" enabled. There is some inconsistency in doing this for vxlan.vlans with vnis, but not for evpn.vlans

netsim/modules/evpn.py
ipspace: Good catch! Thank you!

netsim/modules/evpn.py
ipspace: While not exactly mandatory, added code to count VRFs with evpn.transit_vni and exit early if there are none -- let's reduce the carbon footprint a bit ;)

Thank you!

ipspace:
> Could add a mpls-bridging integration test case

Coming in next commit ;)

> I think we may still want to support an implied evpn.vlans list, containing all vlans with "evpn:" enabled.

What attribute would you set on those VLANs to say "this VLAN should use EVPN"?

> There is some inconsistency in doing this for vxlan.vlans with vnis, but not for evpn.vlans

Agreed, but the difference is that with VXLAN we could say "all VLANs" which we cannot say with EVPN because it needs a VNI on a VLAN for EVPN/VXLAN to work.

jbemmel:
No attributed, just “evpn:” to mark it for inclusion in evpn.vlans. Then Netlab can assign rt/rd etc

ipspace:
> No attributes, just “evpn:” to mark it for inclusion in evpn.vlans. Then Netlab can assign rt/rd etc

OK, I understand it's nice to have three ways to configure the same thing ;)) but what would be the real advantage of having this instead of using **evpn.vlans** list?

jbemmel:
It’s redundant, but people think differently. Some people find it more natural to think of “enabling evpn on a vlan”, others prefer adding a vlan to an evpn.vlans list

PR #558, opened 2022-10-07
srlinux/sros: Cleanup address families enabled for BGP sessions:

jbemmel:
1. If both ipv4 and ipv6 peering addresses are given, activate ipv4 for v4 and ipv6 for v6 (not both for both)
2. If only ipv6 is given, allow ipv4 to be activated too (if enabled by Netlab flag)

The data model only provides
```
activate:
  ipv4: True/False
  ipv6: True/False
```

and does not specify which one(s) apply to which peering session

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -18,7 +18,7 @@
 
 - path: routing-policy/policy[name={{name}}]
   val:
-   default-action: 
+   default-action:
     reject: { }
    statement:
 {% if is_import %}
@@ -28,7 +28,7 @@
       bgp:
        community-set: "C{{ c|replace(':','_') }}"
      action:
-{{    accept() }}       
+{{    accept() }}
 {%  endfor %}
 {% else %}
    - sequence-id: 5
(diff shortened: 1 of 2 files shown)


ipspace:
> The data model only provides
> 
> ```
> activate:
>   ipv4: True/False
>   ipv6: True/False
> ```
> 
> and does not specify which one(s) apply to which peering session

Not true, the description of the parameter is pretty precise:

https://netsim-tools.readthedocs.io/en/latest/module/bgp.html#node-configuration-parameters

One could also say that with EBGP sessions and 'ipv4: True' on an interface, it's still clear which AF should be enabled. Without 'ipv4: True' on the interface, IPv4 forwarding wouldn't work on most platforms anyway.

jbemmel:
Wasn’t intended as criticism, perhaps I should speak for myself - it wasn’t clear to me

For template building, a more “fool proof” data structure would be

activate:
 ipv4: [ ipv4 ]
 ipv6: [ ipv6 ]
 
Which would also allow for
activate:
 ipv6: [ ipv4, ipv6 ]
 
The current structure assumes knowledge about what would work in practice

ipspace:
Oh, back to that discussion? ;))

Look, there's a single sane way of running BGP for unicast IPv4 and IPv6: IPv4 AF over IPv4 session, and IPv6 AF over IPv6 session. Everything else is either a struggle (trust me, I've tried to run IPv4 and IPv6 over the same transport session, and it's a mess), a behind-the-scenes duct tape (IPv4 over IPv6 LLA), or a way to implement a service that is not unicast IPv4 or IPv6 (example: 6PE or IPv4-over-SRv6).

I made a decision a long time ago that people who want to struggle should do that on their own ;), we support IPv4 AF over unnumbered interfaces because it looks like EBGP is running "over an interface", and if someone wants to implement an extra service, that needs to be handled in a configuration module or a plugin.

Makes sense?

jbemmel:
I don't disagree with your assessment of how things should be done, all I'm saying that (imho) it's currently represented in a way that leaves too much room for (mis)interpretation. I read the documentation multiple times, and I still got it wrong

If I understand correctly, your logic above translates to something like this:
```
def activate(i):
  if 'activate' not in i:
   i.activate = {}
   if 'ipv4' in i and i.ipv4 is string:
     i.activate['ipv4'] = ['ipv4']
   if 'ipv6' in i:
     i.activate['ipv6'] = ['ipv4','ipv6'] if 'ipv4_rfc8950' in i or ('ipv4' in i and i.ipv4 is not string) else ['ipv6']
```
Then the template logic simplifies to:
```
bgp_neighbor(n,af):
  ipv4: {{ af in n.activate and 'ipv4' in n.activate[af] }}
  ipv6: {{ af in n.activate and 'ipv6' in n.activate[af] }}
```
(could add 'evpn' to the list too).

This would allow users/plugins to implement custom combinations by manually setting ```activate``` (e.g. to say ipv4: ['ipv4','ipv6']) but things default to a sane, readable combination

PR #571, opened 2022-10-10
Add internal loopback interface to associate vlans and create interfaces

jbemmel:
Not to be merged as-is, just to illustrate a point about creating vlan interfaces

By defining a virtual 'loopback' interface and associating vlans with that, the existing code that processes physical interfaces will create the associated vlan interfaces, as needed (depending on vlan.mode etc.)

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -120,7 +120,7 @@ def add_node_interface(node: Box, ifdata: Box, defaults: Box) -> Box:
     ifindex_offset = 1
 
   # Allow user to select a specific interface index per link
-  ifindex = ifdata.get('ifindex',None) or (len(node.interfaces) + ifindex_offset)
+  ifindex = ifdata.get('ifindex',None) or (len(node.interfaces)-1 + ifindex_offset)
 
   ifname_format = devices.get_device_attribute(node,'interface_name',defaults)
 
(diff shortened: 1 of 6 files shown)


ipspace:
It might even work for access VLANs (would have to look at how the VLAN interfaces are created), but definitely wouldn't work for multiple VLANs per box, and the interface wouldn't get an IP address from the VLAN prefix because it's not associated with a link.

On a more personal note, I'm struggling to grasp how this _might be conceptually easier_ than saying "I need a VLAN interface" ;)

Finally, thanks for pushing me into thinking about the implications. Effectively, we'd have to create bogus "vlan_member" links if we want other bits of the code to do all the hard work, and it wouldn't work on _router_ platforms anyway.

jbemmel:
Earlier in a related discussion (https://github.com/ipspace/netlab/pull/405) I understood that some people might prefer to think in terms of (physical) interfaces driving the need for vlans. Following that way of thinking, creating a (virtual) interface to pull in vlans is "conceptually easier" as it reuses a pre-existing concept (as opposed to introducing a new attribute with specific semantics)

Or so my theory goes...

Anyway, you're correct that IP allocation would need fixing (along with anycast gateway logic). Multiple vlans could be handled as a trunk interface

The implementation would clearly still need some work - but I think in principle creating a 'dummy' interface for vlan interface creation and then removing it (and resulting bits like bogus vlan_member links) from the output could be an option

ipspace:
Implemented inm #590

PR #573, opened 2022-10-12
srlinux: fix static routes with ipv6 prefixes

jbemmel:
Ansible doesn't like unquoted strings that contain ':' characters...


ipspace:
This has nothing to do with Ansible but with semicolons in unquoted strings not being valid YAML ;)

PR #576, opened 2022-10-13
(dev-1.4) Move check for duplicate L2 VNI to register_static_transit_vni

jbemmel:
Due to register_static_transit_vni being called by module_pre_transform, the transit VNIs are already in the vni list. Any duplication should now be detected when the l2 vni is added (with a little help from my second commit)

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -177,12 +177,6 @@ def vrf_transit_vni(topology: Box) -> None:
         f'VRF {vrf_name} is using the same EVPN transit VNI as another VRF',
         common.IncorrectValue,
         'evpn')
-      continue
-    elif _dataplane.is_id_used('vni',vni):
-      common.error(
-        f'VRF {vrf_name} is using an EVPN transit VNI that is also used as L2 VNI {vni}',
-        common.IncorrectValue,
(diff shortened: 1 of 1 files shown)


ipspace (commented):
Have you tested your assumption with a sample topology or have you guessed that it should work as you expect it to work?

jbemmel:
Test case: https://github.com/jbemmel/netsim-examples/blob/master/BGP/SRL-Route-Leaking/topology.yml

It works with dev, but dev-1.4 breaks the duplicate vni detection logic. Removing that check fixes things

ipspace:
Oh, I never doubted that you made the code work with the current topology you're working on. However, once I put myself into the "I wonder if..." mood, it took me approximately 30 seconds to whip up this topology:

```
defaults.device: eos
module: [ vlan, vxlan, vrf, evpn, bgp ]

bgp.as: 65000

vlans:
  red:
    id: 10
    vni: 100000

vrfs:
  red:
    evpn.transit_vni: 100000

nodes:
  r1:
```

Pulled down your branch, ran the test case, and got no complaints.

That's why I keep telling you to look beyond the one topology you're currently working on. If you want to fix code, then fix it correctly.

PR #577, opened 2022-10-13
Validate node, VRF and VLAN attributes

ipspace:


ipspace:
Also: it would be great if you could run some of your more intricate topologies against this branch. No need to set up the lab (there should be no changes in data transformation), just run 'netlab create _topo-name_' -- I'd like to be sure I haven't missed some obscure attribute.

jbemmel:
I would like to be able to define the ```as``` to use for route targets, per vrf instance:
```
jeroen@jvm:~/srlinux/netsim-examples/BGP/SRL-Route-Leaking$ netlab create
IncorrectAttr in vrf: Invalid VRF attribute 'as' found in vrfs.customer1
```

This would require changes to the code obviously - but we actually have customers who assign separate AS numbers to each tenant

netsim/data/__init__.py
```diff
@@ -1,5 +1,5 @@
 #
-# Generic data model manipulation and validation routines
+# Generic data model manipulation outines
```
jbemmel, line 2: (r)outines ;)

jbemmel (approved):
Need to mention this in the release notes as potentially breaking existing topologies, but other than that I see no issues

ipspace:
> I would like to be able to define the `evpn.as` to use for route targets, per vrf instance:
[...]
> This would require changes to the code obviously - but we actually have customers who assign separate AS numbers to each tenant

IIRC (speaking off-the-cuff here although I should be looking at the code):

* You could set VRF RD to whatever value you prefer (including different AS number for each tenant)
* That value would then be used when translating VRF names in import/export RT.
* EVPN RD is generated from router ID and EVI (in the new code) so you could use EVPN RD in the VRF. That would require some tweaks in the templates, in particular if you define VRFs early in the configuration process, and if you can't redefine VRF RD (like on Cisco IOS). Fortunately, EVPN RD is locally significant.

I'm also (sort of) OK with hardcoded values in import/export RT.

Beyond that we're in the "tweaking the code to meet the idiosyncrasies of a customer" territory which should be solved with a plugin.

jbemmel:
Found a minor issue with the MPLS bundle integration test case, see https://github.com/ipspace/netlab/pull/579

PR #579, opened 2022-10-14
Allow for empty dict (like evpn:) in validation code, when enabled for module

jbemmel:
New validation code doesn't allow MPLS VLAN bundle integration test case


ipspace:
Avoiding None type when we're really expecting a Box will inevitably bite us. Wouldn't it be better to say "if we got this far down the validation code, and it's obvious from the attributes that we really need a Box, let's give up and turn that annoying None into a Box({})"

Dealing with this change would be a bit tricky, as we don't want to touch the attributes one level too high, so it might be best if the validation code would return fixed data structure which would be mostly ignored, but could be used to a value in parent dict.

I could fix the code if you also feel this approach makes sense.

ipspace:
Replaced by 94632864fb3bf5734cd91be5a3d9438795dd62f3

PR #584, opened 2022-10-15
Use the new attribute validation code for links and interfaces

ipspace:


ipspace:
Infrastructure:
* Move 'build valid attribute list' into a function as it's becoming more complex
* Consider 'no_propagate' and 'internal' attributes in 'build valid attribute list'
* Additional kludge: module attributes could be set to False on links or interfaces

The real deal:
* Validate link and interface attributes
* Remove module attribute validation for links and interfaces
* Clean up the `links.py` code

Issue #547, opened 2022-10-05
Including unmanaged devices in lab topology


ssasso:
### Discussed in https://github.com/ipspace/netlab/discussions/538

<div type='discussions-op-text'>

<sup>Originally posted by **ssasso** October  4, 2022</sup>
I have a setup when I want netlab to take care of some network gear equipment config (using the *external* provider), but at the same time I need to establish *peerings* with their counterparts which should not be configured by netlab.
(i.e. configure a BGP session on a device controlled by netlab)

I was thinking to use the **none** device type for that kind of stuff, with sth like:
```
vrfs:
  RED:
  BLUE:

vlans:
  site1red:
    mode: route
    id: 41
    vrf: RED
    bgp: false
  site1blue:
    mode: route
    id: 61
    vrf: BLUE

module: [ vlan,vrf,ospf,bgp,mpls ]
bgp.as: 65000

nodes:
  pe1:
  site1a:
    device: none
    module: [ vlan,vrf,bgp ]
    bgp.as: 65101
  site1b:
    device: none
    module: [ vlan,vrf,bgp ]
    bgp.as: 65101

links:
- pe1:
    ifname: eth3
    vlan.trunk:
      site1red.ipv4: 172.24.111.33/28
      site1blue.ipv4: 10.211.211.1/28
  site1a:
    vlan.trunk:
      site1red.ipv4: 172.24.111.34/28
      site1blue.ipv4: 10.211.211.11/28
  site1b:
    vlan.trunk:
      site1red.ipv4: 172.24.111.35/28
      site1blue.ipv4: 10.211.211.12/28
  vlan.trunk: [ site1red, site1blue ]
```

(for this specific example, I need *pe1* to have two configured bgp sessions towards *10.211.211.11* and *10.211.211.12*.)

But unfortunately the *none* device does not support vlan, vrf, and bgp.

Do you think it would be a good idea to allow the *none* device type to support all modules,
or should I use a *real platform* on all the devices, and the rely on a **group** of configurable devices to limit the ansible config with `-l group`?</div>

ipspace:
It looks like the best way to handle this would be to have a special **devices.features** flag (maybe `ghost: True`) and remove nodes with that device feature from final list of nodes (or some such trick), so they don't get into Ansible inventory or Vagrantfile/clab.yml.

[...]

PR #585, opened 2022-10-16
Support for unmanaged devices (aka "ghosts")

ssasso:
NOTE: I created the "input" test but not the "expected" one - not sure how to generate it in this very specific case.

Addresses #547

### DIFF
--- a/docs/nodes.md
+++ b/docs/nodes.md
@@ -69,6 +69,7 @@ nodes:
 * **loopback** -- static loopback addresses. Must be a dictionary with **ipv4** and/or **ipv6** attributes.
 * **memory** -- memory allocated to the VM lab device. Not applicable for container-based devices.
 * **cpu** -- virtual CPU cores allocated to the VM lab device. Not applicable for container-based devices.
+* **unmanaged** -- whether this node should me managed/handled by *netlab* or not (useful for integrating *netlab* topologies with additional external devices which should not be configured by *netlab* itself).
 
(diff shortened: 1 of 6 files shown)


jbemmel:
You can run ```tests/create-transformation-tests.sh```

ssasso:
> You can run `tests/create-transformation-tests.sh`

I don't think this will create a valid test for that specific case where the expected result should not have the *ghost* nodes in it.

jbemmel (changes requested):
This approach is not going to work - Netlab needs to be aware of these unmanaged devices, allocate IP addresses for them, etc. Adding them to the topology and then removing them as a first step makes no sense; it would break all configuration logic

As discussed, I think these 'unmanaged' devices should be modeled as a new type of device ('unmanaged'). This device type would support all modules and features, but configuration would be a no-op ( i.e. empty script that does nothing, or skip condition in Ansible )

ipspace:
Please note that the unmanaged devices are removed from the data model after the transformation has been complete -- in provider module(s) before generating configuration files (because they're "outside") and from Ansible inventory (because they are unmanaged).

jbemmel:
> Please note that the unmanaged devices are removed from the data model after the transformation has been complete -- in provider module(s) before generating configuration files (because they're "outside") and from Ansible inventory (because they are unmanaged).

I apologize, I mis-glanced - the diff shows "def transform()" and then right below it the 'ghost_buster' routine - but it gets called after transformation, so that's ok

ipspace:
Modified the test routines to support configurable "we don't want to see this" parameter, amended the commit and force-pushed it to dev-1.4 Thank you!

Issue #586, opened 2022-10-16
Allow to run containerlab with different runtime flag


q2dg:
Thanks!

ssasso:
Hello @q2dg, **netlab** is only a consumer of containerlab...
I think you should ask your "feature request" directly to the containerlab team @ https://github.com/srl-labs/containerlab

q2dg:
Thanks for pointing out! 
But containerlab is already compatible from 0.30 version: https://github.com/srl-labs/containerlab/releases/tag/v0.30.0

ssasso:
Ok, so can you please give us more info on what you think should be the requirement on netlab for supporting this?

ssasso:
I guess it's just a matter of running containerlab with the proper runtime flag.
(https://containerlab.dev/cmd/deploy/#runtime)

@q2dg can you verify if a netlab clab output file works just fine with that? (I'm guessing you already have an environment where both podman and netlab are installed)

q2dg:
Yes, it works!!

jbemmel:
There is already code for 'runtime' https://github.com/ipspace/netlab/blob/dev/netsim/templates/provider/clab/clab.j2#L44.
May not be documented (and currently not explicitly modeled as an allowed node attribute) but try:
```
nodes:
 r1:
  runtime: podman # ( should really be 'clab.runtime' )
```

q2dg:
Thaks. It works!
So I'll close this issue, if you don't mind

PR #587, opened 2022-10-16
Change node.runtime to node.clab.runtime, add documentation

jbemmel:
Context: https://github.com/ipspace/netlab/issues/586

Configuration did support multiple runtimes, but feature was undocumented and directly under the node while being a Containerlab specific setting

* Move node.runtime -> node.clab.runtime
* Add defaults.providers.clab.runtime = docker, always set 'runtime' config in clab.yml
* Add documentation

### DIFF
--- a/docs/labs/clab.md
+++ b/docs/labs/clab.md
@@ -28,7 +28,7 @@ You can also use [vrnetlab](https://github.com/vrnetlab/vrnetlab) to build VM-in
 
 ## LAN bridges
 
-For multi-access network topologies, **[netlab up](../netlab/up.md)** command automatically creates additional standard Linux bridges. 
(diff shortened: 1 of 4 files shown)


ipspace (changes requested):
I think I added provision for provider-specific settings when validating node attributes, so there's no need to add **clab** to the list of node attributes. Please test it, I'm down with COVID and high fever, so I might be spouting nonsense :(

ipspace:
Merged and fixed (c1ceb1ebf823f37053b8ede36f9482007721ba0e)

BTW, how many times have I told you to look further than the immediate problem you're solving? **runtime** has been used for Cumulus VX and your change broke that. To make it worse, that use of **runtime** was documented, all you'd have to do was to search docs directory for "runtime" (like I did).

PR #588, opened 2022-10-18
fix typo

jbemmel:


ipspace:
Dang. Thank you. I knew I shouldn't be writing code with fever (but I got bored so here we are). Will add this fix to the original commit.

The funny bit is that it worked because 'unmanaged' was the last value in that list ;))

ipspace:
https://stackoverflow.com/questions/3611760/scoping-in-python-for-loops

ipspace:
Also: https://github.com/python/mypy/issues/9576

PR #592, opened 2022-10-19
Configure link MTU on Cumulus NVUE

jbemmel:
Likely needed to make OSPF work


ipspace:
Thank you. Much appreciated, but not related to OSPF.

PR #604, opened 2022-10-22
Initial implementation of anycast gateway

ipspace:
The initial implementation of the **gateway** module supports anycast gateways, optionally without unicast IP addresses, and VRRPv3 for IPv4 and IPv6

### DIFF
--- a/docs/module/gateway.md
+++ b/docs/module/gateway.md
@@ -2,23 +2,14 @@
 
 First-hop Gateway configuration module implements mechanisms used to implement a shared router IPv4 address on a stub access network.
 
-The current implementation of the module supports statically configured anycast gateway IPv4 address.
+The module supports statically configured anycast gateway IPv4 address and VRRPv3 for IPv4 and IPv6.
 
 The module is supported on these platforms:
 
-| Operating system      | Anycast |
-| --------------------- | :-: |
-| Arista EOS            | ✅  |
-| Cumulus Linux         | ✅  |
-
-```{note}
-### Implementation Notes
-
-* The default *netlab* shared MAC address is 0200.cafe.00ff.
-* Vendors use different names for anycast gateways: VARP (Arista), VRR (Cumulus)
-* Arista EOS uses the same shared MAC address on all interfaces. Do not set link-level **gateway.anycast.mac** parameter in topologies using Arista EOS.
-* Arista vEOS cannot use VRRP MAC address as a shared MAC address (default recommended by Cumulus Linux)
-```
+| Operating system      | Anycast | VRRPv3 | VRRPv3<br>IPv6 |
+| --------------------- | :-: | :-: | :-: |
+| Arista EOS            | ✅  | ✅  | ✅  |
+| Cumulus Linux         | ✅  | ✅  | ✅  |
 
 ## Global Parameters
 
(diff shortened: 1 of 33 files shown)


docs/module/gateway.md
```diff
@@ -0,0 +1,63 @@
...
+* Vendors use different names for anycast gateways: VARP (Arista), VRR (Cumulus)
+* Arista EOS uses the same shared MAC address on all interfaces. Do not set link-level **gateway.anycast.mac** parameter in topologies using Arista EOS.
+* Arista vEOS cannot use VRRP MAC address as a shared MAC address (default recommended by Cumulus Linux)
+
+### Anycast Parameters
+
+Anycast implementation of shared first-hop IPv4 address supports these parameters that can be specified globally or on individual links or interfaces.
+
+* **gateway.anycast.unicast** (default: True) -- configure node-specific unicast IP addresses together with anycast IP address.
+* **gateway.anycast.mac** -- Static MAC address used for the anycast IP address
```
jbemmel, line 43: SROS by default generates a different virtual mac address for ipv4 versus ipv6 (derived from vrrp.group)

netsim/ansible/templates/initial/cumulus.j2
```diff
@@ -62,7 +62,7 @@ service lldpd restart
 #
 # Enable FRR modules
 #
-{% set modlist = {'bgp':'bgpd','ospf':['ospfd','ospf6d'],'isis':'isisd','vrf':'bgpd'} %}
+{% set modlist = {'bgp':'bgpd','ospf':['ospfd','ospf6d'],'isis':'isisd','vrf':'bgpd','gateway':'vrrpd'} %}
```
jbemmel, line 65: There is another instance in ```netsim/templates/provider/clab/frr/daemons.j2``` that needs to be updated like this

netsim/quirks/eos.py
```diff
@@ -24,11 +24,28 @@ def check_mlps_vlan_bundle(node: Box) -> None:
...
+    return
+
+  for intf in node.interfaces:
+    if get_from_box(intf,'gateway.protocol') != 'anycast':              # We hope that VRRP works (not tested yet)
+      continue
+
+    if get_from_box(intf,'vlan'):                                       # Anycast works on VLAN cEOS interfaces
+      continue
+
+    common.error(
```
jbemmel, line 44: This is also true for SR Linux - anycast only works on IRB interfaces (not bridge, not routed interface)

docs/module/gateway.md
```diff
@@ -0,0 +1,63 @@
...
+
+```{tip}
+Many implementations require unique unicast IP addresses configured on the interfaces that have anycast IP address. Set **gateway.anycast.unicast** to *False* only when absolutely necessary.
+```
+
+## Virtual Router Redundancy Protocol (VRRP)
+
+*netlab* supports a single VRRPv3 instance per subnet. The VRRPv3 instance can provide shared IPv4 and IPv6 addresses. More complex topologies like multiple VRRPv3 instances could be deployed with a judicious application of interface parameters[^VNS].
```
jbemmel, line 51: I don't think the current templates / parameter structure would support multiple instances? Wouldn't that require a dict keyed by VRRP group (for example)?

I think it's ok to state that Netlab only supports a single VRRPv3 instance per subnet - if people need more, they need to write a plugin

docs/module/gateway.md
```diff
@@ -0,0 +1,63 @@
...
+
+[^VNS]: These topologies are not supported and will not be integrated into *netlab* core. If you want to have an easier way of configuring the in a lab topology, please feel free to create a plugin.
```
jbemmel, line 53: typo here

docs/module/gateway.md
```diff
@@ -0,0 +1,63 @@
...
+
+### VRRP Parameters
+
+*netlab* supports these VRRP parameters that can be specified globally or on individual links or interfaces.
+
+* **vrrp.group** (default: 1) -- VRRP group
```
jbemmel, line 59: On SR OS, vrrp.group influences the virtual MAC address that gets generated. In a LAN segment with multiple pairs of VRRP gateways, one would have to ensure unique group ids - is that something we want to check/enforce?

jbemmel (changes requested):
See comment on modlist for frr

tests/integration/gateway/anycast/vlan.yml
```diff
@@ -0,0 +1,52 @@
...
+links:
+- s1:
+  h1:
+  vlan.access: red
+- s2:
+  h2:
+  vlan.access: red
+- s1:
+  h3:
+  vlan.access: red
```
jbemmel, line 42: At least 1 of the hosts should be on 'blue', else nobody can reach the VRRP gateway there

docs/module/gateway.md
ipspace: In most (static) anycast implementations you have to specify the MAC address and while CL recommends using a MAC address from the VRRP range, Arista refuses to use that. That's why I chose another MAC address as the default value.

Also, most static anycast implementations support only IPv4. You could configure IPv6 as well if you wish; the IPv6 address is in the data model.

docs/module/gateway.md
```diff
@@ -0,0 +1,63 @@
...
+* **vrrp.priority** (interface parameter, integer)
+* **vrrp.preempt** (interface parameter, boolean)
```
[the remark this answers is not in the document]
ipspace: Yeah, we could add that. It wouldn't hurt much. Although, keep in mind we're building a _lab BUILDING tool_ ;)

netsim/ansible/templates/initial/cumulus.j2
ipspace: True, but VRRP needs a control- and data-plane components (macvlan interface). It was pretty simple to add data plane component to Cumulus Linux with **ifupdown2**, not sure how much work it would be to add it to FRR the way it's currently implemented (and I'm not about to go exploring).

But yeah, once that's in place the list of FRR modules will have to be updated.

netsim/quirks/eos.py
ipspace: So you'll have to add a device quirk ;) I'm not adding a feature flag (I learned my lessons with VLAN module).

NXOS is even worse -- looks like anycast gateway works only with EVPN.

docs/module/gateway.md
ipspace: Absolutely correct. Rewrote to make the limitations more obvious.

docs/module/gateway.md
ipspace: Thank you, fixed

tests/integration/gateway/anycast/vlan.yml
ipspace: Ah, the beautiful copy/paste artifacts. Of course you're absolutely correct, fixed.

PR #605, opened 2022-10-23
Add the ability to disable a routing protocol per VRF

ipspace:
This PR implements "ospf: False" or "bgp: False" at a VRF level. Documentation and test cases will be added before the PR is merged.

### DIFF
--- a/netsim/modules/__init__.py
+++ b/netsim/modules/__init__.py
@@ -442,12 +442,13 @@ def copy_node_data_into_interfaces(topology: Box) -> None:
       for intf in n.get('interfaces',[]):                        # We might have some work to do, iterate over all interfaces
         if not isinstance(intf.get(m,{}),dict):                  # ... if the interface module data is not a dict, we can't merge
           continue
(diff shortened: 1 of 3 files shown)


jbemmel (changes requested):
See https://github.com/ipspace/netlab/pull/606 for a required change to cleanup 'False' flags
(else templates try to iterate over things like 'bgp.neighbors' and fail)

I think this works for bgp, ospf and isis (right?)

We may want to define the list of all supported routing protocols in _routing, and then import it from there. Also applies to eigrp for example?

ipspace:
> See #606 for a required change to cleanup 'False' flags (else templates try to iterate over things like 'bgp.neighbors' and fail)

Good catch. Thanks a million.

I always believe that whoever caused the spill should also be responsible for cleaning it up. The 'False' flags thus have to be cleaned up by individual routing modules, or we can play whack-a-mole in VRF module till cows come home.

The VRF check will go into _routing.remove_unused_igp, will do something similar for BGP.

ipspace:
Done. I could have had VRRP implemented in the time I spent on this super-important feature...

PR #606, opened 2022-10-23
Cleanup any <routing-proto>: False flags from vrfs

jbemmel:


ipspace:
I have not set up a topology to test this, but due to the way "transform_after" dependencies are set up, OSPF hook seems to be called after VRF hook, which means that your patch would clear the OSPF flag before OSPF transformation could see it.

Please try to consider the bigger picture when writing intricate patches that explore inter-module dependencies.

PR #607, opened 2022-10-24
Critical: sros re-add next-hop-self policy

jbemmel:
Somehow got lost since PR #561


ipspace:
I assume #561 got lost when I added dev-1.4 on top of dev, which applied #533.

Fortunately we're back in sane lands ;)

PR #608, opened 2022-10-24
Critical: sros - put 'routes' under bgp-evpn context

jbemmel:
YANG model mismatch

### DIFF
--- a/netsim/ansible/templates/vxlan/sros.j2
+++ b/netsim/ansible/templates/vxlan/sros.j2
@@ -6,12 +6,6 @@
    customer: '1'
    admin-state: enable
    service-id: {{ vni if type=='l3' else (vrf[4:]|int + 10000) }} # transit VNI for l3, should not overlap
-
-{% if type=='l3' %}
-   routes:
-     ip-prefix:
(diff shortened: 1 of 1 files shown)


ipspace:
One has to wonder whether a more thorough testing of all scenarios supposedly implemented by that template would have uncovered the problem before the first commit…

jbemmel:
Yes - I started creating a script to run all topology test cases with a given device:
```
DEVICE=${1:-frr}

set -e
for file in topology/input/${1:-*}.yml; do
  if [[ "${DEVICE}" != "frr" || ! `grep mpls ${file}` ]]; then
  netlab up -d ${DEVICE} -p clab $file
  netlab down
  fi
done
```
but the challenge is feature support (that, plus time - it takes a few minutes for the VM to start)

ipspace:
> Yes - I started creating a script to run all topology test cases with a given device:

While that's commendable (and takes forever), it won't do much more than test that the configurations generated by the Jinja2 templates are accepted by the devices. Definitely a step in the right direction, but you might want to test whether stuff works the way it's expected to work.

PR #609, opened 2022-10-25
sros: add gateway support

jbemmel:
PR against anycast branch, just let me know if you prefer ```dev``` instead

+ Also fixes OSPF on vlan interfaces

Known issues:
* ISIS only works in base routing context, not inside vrfs yet


ipspace:
This one was automatically closed when I merged the anycast branch :( Please resubmit it against the dev branch.

PR #610, opened 2022-10-25
Validate groups and node_data with the new validation routines

ipspace:
Also:
* Allow node attributes in group definitions (node_data is optional)
* Cleanup the remains of 'check_module_parameters' code (no longer needed, all validation is done elsewhere)

Note:
* 'node_data' is still used internally


jbemmel (approved):
no comments (saw one instance of double spaces in the documentation)

PR #612, opened 2022-10-25
Add SROS support for anycast gateway and VRRP

jbemmel:
Known issues:
* ISIS (unchanged here) only works in base routing instance, not inside VRFs yet


ipspace:
> ISIS (unchanged here) only works in base routing instance, not inside VRFs yet

I don't think that anyone runs IS-IS in VRFs (Junos?). To keep things simple, the IS-IS module removes **isis** from VRF interfaces.

Issue #619, opened 2022-10-28
[BUG] 'clab' provider attributes not accepted at group level


jbemmel:
**To Reproduce**
netlab create /tmp/test-topology.yml 

IncorrectAttr in groups: Invalid group attribute 'clab' found in topology.groups.leaves

Fatal error in netlab: Cannot proceed beyond this point due to errors, exiting

**Expected behavior**
Topology should be accepted

Minimal topology:
```
groups:
  leaves:
    members: [ leaf1,leaf2 ]
    device: srlinux
    clab.binds:
     create_customer_sh: /home/admin/create_customer.sh
     aliases_srlinuxrc: /home/admin/.srlinuxrc

nodes:
 leaf1:
 leaf2:
```
Same happens within ```node_data```

** Version
netlab version 1.4.0-dev2

PR #620, opened 2022-10-28
Allow provider attributes in groups too

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/619

(seems like the code that checks node attributes and the code that checks group level attributes could share some of the same logic?)


ipspace (approved):
Looks like a perfect fix to me ;)

PR #621, opened 2022-10-28
Request reporters to run netlab against the submitted minimal topology

jbemmel:
Else we might get overwhelmed by people who report fictional problems that only exist in their minds...


ipspace:
Good one. We might change the command into `netlab create -o none` unless we want to see the same printout in every bug report ;) Any better idea?

We could also define a 'bugreport' format in topology-defaults.yml and use that (see `integration/gateway/vrrp/vlan.yml` for an example of how that works).

jbemmel:
I struggle with the various formats myself sometimes - perhaps a top-level ```netlab bugreport``` command (which would print out things like versions, increase verbosity, etc.) would be better?

ipspace:
> Else we might get overwhelmed by people who report fictional problems that only exist in their minds...

Based on the number of bug reports we're getting I'd say we're solving imaginary scaling issues at scale ;) Will keep the **netlab bugreport** (or **netlab summary**) idea in mind and implement it if the number of bug reports justifies it.

PR #624, opened 2022-10-31
Remove OSPF in both symmetric and asymmetric cases

jbemmel:


ipspace:
We need OSPF within a VRF in asymmetric IRB case as the external prefixes (should there be any) cannot be transported as EVPN RT5... or am I missing something (again)?

Will add a note to the EVPN documentation.

jbemmel:
See https://github.com/ipspace/netlab/discussions/490 - FRR (for example) has a custom ```prefix-routes-only``` option to handle RT5 prefixes in the asymmetric case (without ospf)

I'm ok either way, it just wasn't clear to me why you'd remove ospf in one case but not the other

ipspace:
> See #490 - FRR (for example) has a custom `prefix-routes-only` option to handle RT5 prefixes in the asymmetric case (without ospf)

That's not asymmetric case. That's symmetric IRB for prefixes and bridging within directly-connected subnets.

Asymmetric IRB has no L3 VNI.

PR #630, opened 2022-11-03
Explicitly enable 'proxy_arp' attribute for vrfs

jbemmel:
Now that attribute checks have become more strict, we need to play along


ipspace:
👍

Issue #599, opened 2022-10-20
Document clab config templates


ipspace:
A message from the "No job is finished till the paperwork is done" department: the **clab.config_template** parameter needs to be documented -- someone might find a use for it when generating configurations for network management servers running in containers adjacent to netlab devices.

ipspace:
Fixed in #632

PR #632, opened 2022-11-03
Document clab config_templates

jbemmel:
To address https://github.com/ipspace/netlab/issues/599

Perhaps more of a 'dev' item


ipspace:
> Perhaps more of a 'dev' item

You're right... until we fix the "get path to template" code and check both the package directory and current directory ;)

That would give the users the ability to generate configuration files for their favorite container-based network management tools.

PR #635, opened 2022-11-07
Introduce Cisco ASA platform

willimei:
I have added Cisco ASA images to netlab and configurations for

- initial config
- IS-IS
- BGP

I look forward to feedback.

### DIFF
--- a/docs/module/isis.md
+++ b/docs/module/isis.md
@@ -38,7 +38,7 @@ The following table describes per-platform support of individual IS-IS features:
 **Notes:**
 * On Arista EOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address. Arista EOS implementation of IS-IS refuses to work on interfaces with missing address families.
 * On VyOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address.
-* On Cisco ASA you have to add `isis.network_type: false` on all point-to-point links to devices with other NOS, otherwise the ASA will not recognize the IS-IS packets
+* Cisco ASA does not support P2P IS-IS links. You could either add `isis.network_type: false` to point-to-point links connecting ASA to other devices or set `type: lan` on those links.
 
 ## Global Parameters
 
--- a/docs/module/isis.md
+++ b/docs/module/isis.md
@@ -30,6 +30,7 @@ The following table describes per-platform support of individual IS-IS features:
 | Cisco IOS          | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
 | Cisco IOS XE       | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
 | Cisco Nexus OS     | ✅  | ✅  | ✅  | ✅  | ✅  | ❌  |
+| Cisco ASAv         | ✅  | ✅  | ✅  | ❌  | ❌  | ❌ |
 | Nokia SR Linux     | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
 | Nokia SR OS        | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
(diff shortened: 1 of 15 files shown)


docs/module/isis.md
```diff
@@ -30,13 +30,15 @@ The following table describes per-platform support of individual IS-IS features:
...
 | Cisco Nexus OS     | ✅  | ✅  | ✅  | ✅  | ✅  | ❌  |
+| Cisco ASAv         | ✅  | ✅  | ✅  | ❌  | ❌  | ❌ |
 | Nokia SR Linux     | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
 | Nokia SR OS        | ✅  | ✅  | ✅  | ✅  | ✅  | ✅ | 
 | VyOS               | ✅  | ✅  | ✅  | ❌  | ✅  | ✅ |
 
 **Notes:**
 * On Arista EOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address. Arista EOS implementation of IS-IS refuses to work on interfaces with missing address families.
 * On VyOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address.
+* On Cisco ASA you have to add `isis.network_type: false` on all point-to-point links to devices with other NOS, otherwise the ASA will not recognize the IS-IS packets
```
jbemmel, line 41: This looks like something to be solved within the ASA template logic. Can't ask users to change their design to accommodate quirks on some device

willimei: The problem is that I need a different configuration on the remote side, because the ASA has only implemented broadcast IS-IS packets.
So if I want to connect an ASA with an ios or an nexus I must remove the command `isis network point-to-point` within the ios/nxos config.

jbemmel: That's what we have the 'quirks' package for, see e.g. https://github.com/ipspace/netlab/blob/dev/netsim/quirks/eos.py#L19

What we need to add, is some logic for ASA that throws an error when the user is trying to use anything besides 'broadcast' for IS-IS

In general this would be modeled as a feature flag, but since this platform is the only one that doesn't support all types a 'quirk' is the better solution

ipspace: @jbemmel: In principle, I agree with you, and it would be great to have all caveats covered in quirks. In practice, I appreciate the great work @willimei did ;)

Also, I reworded the documentation a bit to explain what the caveat is.

ipspace (approved):
@willimei: Thanks a million. Great work. Fixed a typo or two.

The only breaking change I could see is the "--iso-level 4", but I'm pretty positive vSRX should be able to handle that.

As @jbemmel pointed out, it would be great if you could eventually create a quirks module to check for IS-IS link type on ASA interfaces (or I could help you do it), but that's not a show-stopper. If you're OK with the rewrites I did I'll merge it.

docs/module/isis.md
jbemmel: Sure, I did not mean to imply a lack of appreciation - just nit-picking

willimei: I was not aware of the quirks package until now. Thanks for that hint.
I will have a look into it tomorrow.

willimei:
> The only breaking change I could see is the "--iso-level 4", but I'm pretty positive vSRX should be able to handle that.

This was my hope at the moment I have changed this.
If it breaks an other workflow, we have to omit the default config file and paste the config to the images at the time of creation, as is the case with most other images.

> As @jbemmel pointed out, it would be great if you could eventually create a quirks module to check for IS-IS link type on ASA interfaces (or I could help you do it), but that's not a show-stopper. If you're OK with the rewrites I did I'll merge it.

Yes the changes are fine for me.
I also will have a look at the quirks tomorrow.
I will add it to this PR or open another one if this is already merged until then.

PR #637, opened 2022-11-07
Default to new frr 8.4.0

jbemmel:
Tested using ```netlab up -d frr -p clab tests/integration/evpn/vxlan-routing-ibgp-over-ebgp-unnumbered.yml```


ipspace (approved):
Let's hope there are no breaking configuration syntax changes ;)

jbemmel:
That's what 'dev' is for, right? If there are, it's better if we find them before users do

PR #642, opened 2022-11-09
Add Cisco ASA quirk

willimei:
I have added a quirk as suggested in #635.

Furthermore while testing I have discovered that `type: lan` on those links has no impact on the generation of the `isis.network_type` Variable.
As far as I can see the code which adds the problematic attribute only looks for the number of neighbors but not for the link type: https://github.com/ipspace/netlab/blob/dev/netsim/modules/_routing.py#L77
Therefore I have removed the hint in the documentation.

### DIFF
--- a/docs/module/isis.md
+++ b/docs/module/isis.md
@@ -38,7 +38,7 @@ The following table describes per-platform support of individual IS-IS features:
 **Notes:**
 * On Arista EOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address. Arista EOS implementation of IS-IS refuses to work on interfaces with missing address families.
 * On VyOS, IPv6 is enabled on all interfaces as soon as one interface has an IPv6 address.
-* Cisco ASA does not support P2P IS-IS links. You could either add `isis.network_type: false` to point-to-point links connecting ASA to other devices or set `type: lan` on those links.
+* Cisco ASA does not support P2P IS-IS links. You could add `isis.network_type: false` to point-to-point links connecting ASA to other devices.
 
 ## Global Parameters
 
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Great job. Thank you!

FWIW, I usually use "get_from_box" when I'm not sure the key exists (because Box creates an empty dictionary on first reference), using dotted syntax (example: neighbor.node or topology.nodes[remote_node].interfaces) is probably easier to read if you know the key is there.

Also, I'm sorry about the `type: lan` SNAFU. I really should look at the code instead of trusting my failing memory.

ipspace:
Now that you removed **get_from_box** calls (it could return None, so you have to check for that or mypy barks), it might be better to change

```
remote_interfaces = topology[f'nodes.{remote_node}.interfaces']
```

into

```
remote_interfaces = topology.nodes[remote_node].interfaces
```

willimei:
> ```
> remote_interfaces = topology.nodes[remote_node].interfaces
> ```
Thanks for that hint!
I was unsure with the correct syntax for mixing variables and strings with Boxes. So I used the one I know well.
I will change it for better readability at the occurrences where the variable should always exist..

ipspace:
> I will change it for better readability at the occurrences where the variable should always exist..

Thank you! Just tell me when to merge ;)

willimei:
@ipspace I have made the changes. You can merge now.

ipspace:
One more thing (I should have noticed it earlier): you might want to check whether there’s IS-IS running on an ASA interface before checking the neighbor state. The code works because usually either both nodes use IS-IS or none does.

ipspace:
Merged, add the `if 'isis' in intf:` check and amended the commit. Hope you don't mind.

Thank you!

Issue #287, opened 2022-05-13
Incorrect interface neighbor list on single-node VLANs


ipspace:
The following topology incorrectly calculates neighbors on R1 and R2, resulting in no EBGP session between them.

```
nodes:
  r1:
    module: [ bgp ]
    bgp.as: 65000
  s1:
    module: [ vlan ]
    vlans:
      red:
  r2:
    module: [ bgp ]
    bgp.as: 65001

links:
- r1:
  s1:
    vlan.access: red
- r2:
  s1:
    vlan.access: red
```

We should convert single-node VLANs into fake global VLANs (example: vlan@node), and clean them up after the topology transformation.

ipspace:
Fixed by 140be8be9b6373b5218000312c82d3888410eee8 / a55a60d650915886c4d92bef0dc9fd127ef28da0

PR #645, opened 2022-11-10
Fix vlan lookup for local vlan defined on neighbor node only

jbemmel:
Addresses https://github.com/ipspace/netlab/issues/287


ipspace (approved):
Wow. Such a lovely and ingenious fix ;)) Thanks a million!

vlan-single-node test case seems to be a duplicate of vlan-access-neighbors test case. If that's true, I'll remove it and amend the commit.

PR #646, opened 2022-11-11
Handle dotted clab.binds by allowing file mappings to be specified as a list of strings too

jbemmel:
Example:
```yaml
clab.binds:
 - config.json:/etc/config.json
```

This is currently a problem because 'config.json' becomes ```config: { 'json': ... }```

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -76,6 +76,14 @@ class _Provider(Callback):
         processor_name = str(subprocess.check_output("cat /proc/cpuinfo", shell=True).splitlines()[1].split()[2])
       topology.defaults.processor = processor_name
 
+  def _get_binds(self, node: Box, key: str):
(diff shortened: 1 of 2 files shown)


ipspace (changes requested):
A) That's why we don't unroll dots for keys that contain '/'. What's wrong with using './config.json' if you insist on having dots in filenames?
B) That limitation is documented in https://netsim-tools.readthedocs.io/en/latest/labs/clab.html

If you still insist on doing the same thing in two different ways just so you won't have to use './config.json' as a filename, then you should:

C) fix your code so the tests don't fail
D) fix the documentation

I would suggest we'd drop this and use the documented workaround, but that's just me being simplistic...

jbemmel:
I tried ‘\.’ and ‘..’ but not ‘/‘. Even looked at the source code for Box to find some escape character 

Ok, never mind. Thanks

Issue #636, opened 2022-11-07
EVPN VLAN bundle data model


jbemmel:
I have some pending changes to add MPLS support for SR OS, but they depend on a change to the data model to support both VLAN-aware and non-VLAN-aware bundles

Proposal:
```
evpn.bundle:
  vrf: xyz
  vlan_aware: True # Omit to auto-select based on device supported feature(s)
```

```
features:
 evpn:
  vlan_bundle: True
  vlan_aware_bundle: False
```

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/discussions/554#discussioncomment-3897927_

ipspace:
Right now we have a global flag that means "create VLAN-aware bundles" instead of "create individual VLANs". I'm not sure anyone is using the VLAN-aware bundles (apart from the integration test case  😉) so I have no problem changing the data model (and documenting another breaking change 🤷‍♂️).

In any case, if we want to get more flexible, then the **evpn.bundle** should be moved into individual VRFs. I would prefer having a text value that would be set to **vlan** (RFC 7432 section 6.3) or **port** (section 6.2.1) or **whatever-else-you-prefer** (section 6.2). **features.evpn.bundle** would then be a list of allowed values.

Furthermore, we could add "l2only" keyword to the **features.evpn.bundle** list to indicate the VLANs in the bundle VRF should be layer-2-only VLANs.

ipspace:
@jbemmel Still waiting for that magic keyword ;)

jbemmel:
I’m ok with that

ipspace:
> I’m ok with that

Nice to hear that. So I'll implement the change in EVPN module/documentation, for the moment having **vlan** as the only valid keyword to support Arista EOS, and you'll add whatever you need for SR/OS ;)

PR #649, opened 2022-11-13
Add support for all EVPN bundle services

ipspace:
* Add 'evpn.bundle' VRF parameter
* Validate 'evpn.bundle' value against evpn.attributes.bundle list
* Check 'evpn.bundle' value against device.features.evpn.bundle list
* Refactor VLAN bundle code in EVPN configuration module
* Fix EVPN VLAN bundle integration test case
* Add 'evpn.bundle' to EVPN asymmetric IRB test case
* Update the documentation

Implements #636

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -95,15 +95,15 @@ def vlan_aware_bundle_service(vlan: Box, vname: str, topology: Box) -> None:
       'evpn')
     return
 
-  vlan.evpn.bundle = vlan.vrf                                       # Set a 'we are in a bundle' flag
   vrf = topology.vrfs[vrf_name]
   must_be_dict(
     parent=vrf,
     key='evpn',
     path=f'vrfs.{vrf_name}',
     create_empty=True)
-
+ 
   evpn = vrf.evpn
+  vlan.evpn.bundle = evpn.bundle                                    # Copy the bundle type into VLAN
   if not 'evi' in evpn:                                             # If needed, set EVI attribute for the global VRF
     evpn.evi = vrf.id                                               # ... to the VRF ID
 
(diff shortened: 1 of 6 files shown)


netsim/modules/evpn.py
```diff
@@ -114,6 +114,55 @@ def vlan_aware_bundle_service(vlan: Box, vname: str, topology: Box) -> None:
...
+"""
+Create VLAN services -- either as independent VLANs or as a VRF-based bundle
+"""
+
+def create_vlan_service(vname: str, topology: Box) -> None:
+  vlan = topology.vlans[vname]
+  if not get_vlan_bundle_flag(vlan,vname,topology):
+    vlan_based_service(vlan,vname,topology)
+  else:
+    vlan_aware_bundle_service(vlan,vname,topology)
```
jbemmel, line 148: This method is now a bit of a misnomer - ```vlan_bundle_service``` would be more accurate

netsim/modules/evpn.py
```diff
@@ -114,6 +114,55 @@ def vlan_aware_bundle_service(vlan: Box, vname: str, topology: Box) -> None:
...
+
+"""
+Validate that there's no evpn.bundle setting on node VRFs
+"""
+
+def validate_no_node_vrf_bundle(node: Box, topology: Box) -> None:
```
jbemmel, line 154: Is this common to have attributes that are only valid on global vrfs, but not local vrfs? Do we want to distinguish between the two in the data model?

jbemmel (approved):

netsim/modules/evpn.py
ipspace: In theory, you're correct. In practice, let's wait and see whether we need anything more than what this one contains (when you implement other bundle services).

netsim/modules/evpn.py
ipspace: Definitely evpn.transit_vni, maybe also RD/RT. Not sure it's worth modifying common code to cope with them (at the moment).

ipspace:
Thanks for the review and the comments! Merging...

PR #651, opened 2022-11-15
VyOS and Dell OS10 - support for Gateway module

ssasso:
* VyOS
  * VRRP 3 (v4 and v6)
* Dell OS10
  * Anycast - *partial* (see new caveats)
  * VRRP 3 (v4 and v6) - *partial* (see new caveats)

Tested with all the module "integration test" files.

**SAD NOTE**: New caveats for Dell OS10

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -116,3 +116,5 @@ Additionally, using always the latest build published on [Vagrant Hub](https://a
 ## Dell OS10
 
 Dell OS10 uses a concept of a so-called *Virtual Network* interface to try to handle transparently VLANs and VXLANs in the same way. However, it seems that right now it is **NOT** possible to activate OSPF on a *Virtual Network* (VLAN) SVI interface.
+
+Sadly, it's also **NOT** possible to use *VRRP* on a *Virtual Network* interface (but *anycast* gateway is supported). At the same time, *anycast* gateway is not supported on plain *ethernet* interfaces, so you need to use *VRRP* there.
--- a/docs/module/gateway.md
+++ b/docs/module/gateway.md
@@ -22,8 +22,10 @@ The module is supported on these platforms:
 | Cisco IOSv            |  ❌  | ✅  | ✅  |
 | Cisco Nexus OS        |  ❌  | ✅  | ✅  |
 | Cumulus Linux         | ✅  | ✅  | ✅  |
+| Dell OS10             | [❗](caveats.html#dell-os10)  | [❗](caveats.html#dell-os10)  | [❗](caveats.html#dell-os10)  |
 | Nokia SR OS           | ✅  | ✅  | ✅  |
 | Nokia SR Linux        | ✅  | ❌  | ❌  |
+| VyOS                  | ❌  | ✅  | ✅  |
 
 ## Global Parameters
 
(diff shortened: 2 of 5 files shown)


docs/caveats.md
```diff
@@ -116,3 +116,5 @@ Additionally, using always the latest build published on [Vagrant Hub](https://a
 ## Dell OS10
 
 Dell OS10 uses a concept of a so-called *Virtual Network* interface to try to handle transparently VLANs and VXLANs in the same way. However, it seems that right now it is **NOT** possible to activate OSPF on a *Virtual Network* (VLAN) SVI interface.
+
+Sadly, it's also **NOT** possible to use *VRRP* on a *Virtual Network* interface (but *anycast* gateway is supported). At the same time, *anycast* gateway is not supported on plain *ethernet* interfaces, so you need to use *VRRP* there.
```
jbemmel, line 120: Can't we have Netlab select the appropriate interface type based on requested features? To me this is more of a value that Netlab can provide, rather than a caveat

ssasso: Yes, it should.

This will require multiple *quirks* plus template change/rework. It is in "my roadmap", but for now these should stay as caveats.

ipspace: @jbemmel Please keep in mind we're not trying to build a perfect multi-vendor deployment tool that would cope with all the differences and caveats of individual implementations. If someone has the need for such a tool, they're most welcome to finance its development, and if someone feels like their platform implementation should provide that functionality, they're most welcome to implement all the fixes/quirks needed to get there.

Furthermore, I see no need to dance around ~~crappy~~ suboptimal implementations. It's up to the vendors to get their act together (or not).

Having said that, it's always nice to recognize the documented caveats in _quirks_ module and tell the user why a particular lab won't work, but I honestly don't care too much as long as a shortcoming is documented somewhere.

jbemmel: I’m only commenting from a (lazy) user perspective, documentation is good but a tool that “simply” does the correct thing is better.

I have no interest in hiding vendor flaws, on the contrary (;)). However, as a general line-of-thinking principle, to me users shouldn’t need to concern themselves with selecting the type of interface flavor to use in a certain design

ipspace: From the "opinions are great, technology is even better" perspective: it's a bit hard to "choose" between physical (Ethernet) and virtual (VLAN) interface.

From the lazy "exposing vendor flaws" perspective: if a vendor insists on having two ways to implement a VLAN interface on the same platform (which IIRC is the OS10 case), I wouldn't waste too much time trying to figure out which implementation to use, and I'd invite whoever is complaining to implement the missing bits ;)

jbemmel: So in summary: OSPF and VRRP are not supported on overlay networks 

I guess what triggered me to comment, was the implied suggestion that as a user I was choosing “virtual network” interfaces or “regular” interfaces -but I did no such thing…

I think caveats should ideally be framed in terms of data model choices (i.e. combinations of modules and/or feature flags).

ssasso: But, again, allowing to choose between "regular" VLANs or "overlay" VLANs requires lot of rework.
If someone will need for such "choice", (s)he can submit a PR or wait for me to do the implementation ;)

ipspace: 👍

PR #652, opened 2022-11-17
Check that interface has neighbors attribute

jbemmel:
Fails in some cases

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -570,7 +570,7 @@ get_vlan_data: Get VLAN data structure (node or topology or interface neighbors)
 """
 def get_vlan_data(vlan: str, node: Box, topology: Box, intf: Box) -> typing.Optional[Box]:
   vlan_data = get_from_box(topology,f'vlans.{vlan}') or get_from_box(node,f'vlans.{vlan}')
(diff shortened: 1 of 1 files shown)


ipspace:
Would love to see the topology that triggered this PR. It would be a good idea to have the "neighbors" list always present on an interface even if it's empty, but I wouldn't want to start a wild goose chase if you already have a topology triggering this issue.

jbemmel:
I thought it was, that’s why I felt ok to leave out the check - but I was wrong 

Code is peppered with this check in other places, I figured it was better to fix first and diagnose later. Will share in a bit

jbemmel:
Here you go:
```
message: |
  This lab triggers missing 'neighbors' attribute

provider: clab
defaults.device: srlinux

module: [ vlan ]

vlans:
  99: # using a string instead of a number works

nodes:
 r1:

 h1:
   vlan.mode: route # commenting this out solves it too

links:
- r1:
  h1:
  vlan.trunk: [99]
  # vlan.access: red # this works
```

It's a bug triggered by using an integer as vlan id, plus a vlan trunk with a 'route' mode on one node

ipspace:
Thanks a million for excellent detective work!

PR #653, opened 2022-11-17
Enforce string type for VLAN names; minor formatting fixes

jbemmel:
Goose chase resulting from https://github.com/ipspace/netlab/pull/652 revealed that using ints as vlan names triggers undesired consequences in corner cases. While we could convert them to strings silently, it's easier and less error prone to insist on strings

```access``` and ```native``` are forced to be str, but ```trunk``` is merely checked for 'dict' (not dict with 'str' keys)

Any vlan references I missed?


ipspace (approved):
Thank you, much appreciated.

Just a thought on VLAN IDs: wouldn't it be better to transform ints into strs and set .id attribute for int-named VLANs? I could do that after merging your PR, just wanted to hear your opinion.

jbemmel:
I started doing that, but I figured it would be more error prone; every time we add a feature that references vlans, we need to remember to do the conversion. 

I do think ints are natural vlan ids, and as such it would be nice if that could be supported

ipspace:
> I started doing that, but I figured it would be more error prone; every time we add a feature that references vlans, we need to remember to do the conversion.

Yeah, there's that...

> I do think ints are natural vlan ids, and as such it would be nice if that could be supported

Agree, but let's stick with strings for the moment.

PR #654, opened 2022-11-17
Check full range of Python versions 3.8 through 3.11.0, in parallel (upgrade minimum Python version to 3.8)

jbemmel:
Upgrade to minimum supported 3.8 version because 3.7 requires typing-extensions for typing.Final and refactoring of imports, could be supported if needed but to me a range of 4 supported Python versions is pretty good

(if anybody was using Python 3.7, Netlab wasn't working for them anyway - so merely documenting and verifying what was already the case)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,8 +8,9 @@ jobs:
   tests:
     runs-on: ubuntu-latest
     strategy:
+      max-parallel: 4
       matrix:
-        python-version: [ 3.9 ]
+        python-version: [3.8, 3.9, 3.10.0, 3.11.0]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


ipspace (approved):
Nice catch, thank you! Is there a reason you specified "3.10.0" and "3.11.0" as python version in GitHub tests?

ipspace:
... also, 3.7 needs to be changed to 3.8 in documentation (docs/install.md and docs/install/linux.md)

jbemmel:
> Nice catch, thank you! Is there a reason you specified "3.10.0" and "3.11.0" as python version in GitHub tests?

I copied that from the Napalm repo - figure they must have a reason for that, but didn't investigate

ipspace:
> > Nice catch, thank you! Is there a reason you specified "3.10.0" and "3.11.0" as python version in GitHub tests?
> 
> I copied that from the Napalm repo - figure they must have a reason for that, but didn't investigate

I found "the reason" the hard way. 3.10 is a number, and turns into "3.1" when converted into a string. The list of versions should be a list of strings. Fixed in cc626e5e354f4fb2818c4881ff564cf4e370a71e

PR #657, opened 2022-11-22
install ansible-pylibssh to fix connection issues with Cisco IOS and ASA

willimei:
After hours of debugging I figured out that paramiko causes authentication problems with Cisco IOSv and ASAv devices.
According to the IOS log, the username in the authentication request is empty.

Therefore I would recommend installing ansible-pylibssh which works fine with these devices.


jbemmel:
Arista EOS suggests installation of ansible-pylibssh too, but it does work also with paramiko

Root cause would seem to be in Cisco IOSv/ASAv devices and/or paramiko project

Which version are you testing? Looks like paramiko==2.12.0 is the latest

ipspace (approved):
As this is the "use this if you don't know any better" installation script that installs the latest version of Ansible anyway (I keep wondering whether that's a good idea, but that's another discussion), the addition makes perfect sense. Thank you!

PR #661, opened 2022-11-28
Dell OS10 works also with clab

ssasso:
Using info from https://containerlab.dev/manual/kinds/vr-ftosv/ and https://github.com/hellt/vrnetlab/tree/master/ftosv


ipspace (approved):
Nice ;)

PR #669, opened 2022-12-06
Update eos.vrf.j2

petercrocker:
For some reason jinja interprets the RD as possibly a sexagesimal number? TIL: https://en.wikipedia.org/wiki/Sexagesimal

The effect is when ansible reads the `host_vars` file value `rd: 65000:1` it interprets it as `3900001`.

```
TASK [eos_config: deploying initial from templates/initial/eos.j2] **************************************************************************************
fatal: [h1xgw1]: FAILED! => changed=false
  data: |-
    rd 3900001
    % Ambiguous command
    localhost(config-s-ansibl-vrf-blue)#
  msg: |-
    rd 3900001
    % Ambiguous command
    localhost(config-s-ansibl-vrf-blue)#
```

### DIFF
--- a/netsim/ansible/templates/initial/eos.vrf.j2
+++ b/netsim/ansible/templates/initial/eos.vrf.j2
@@ -1,6 +1,6 @@
 {% for vname,vdata in vrfs.items() %}
 vrf instance {{ vname }}
- rd {{ vdata.rd }}
+ rd {{ vdata.rd|string }}
 !
 {% if 'ipv4' in vdata.af|default({}) %}
 ip routing vrf {{ vname }}


ipspace (commented):
Looks like that could be a YAML, not a Jinja2 issue:

https://stackoverflow.com/questions/63241052/why-the-yaml-cant-load-value-as-expected

In any case, it's not just an EOS issue, the same approach is used in every platform supporting VRF, so we're dealing with a bigger issue.

Could you please post the minimum topology that generates this error so we can figure out exactly what's going on?

petercrocker:
Here's a topology.yml file that'll show the problem:

```
---
defaults.device: eos

module: [vrf]

vrfs:  
  blue:
    loopback: True

nodes:
  h1xgw1:

links:
  - h1xgw1:
    vrf: blue
```

petercrocker:
From my troubleshooting, if you quote the line in the YAML file, it doesn't solve it. It seems to be jinja templating problem, leading me to fix this in the eos.vrf.j2 template. If you find a better fix, I'd love doing that, instead of a hack in every platform j2 that supports VRFs.

ipspace:
Works for me...

```
lldp run
ip routing
!
vrf instance blue
 rd 65000:1
!
ip routing vrf blue
...
```

Also, the "host_vars" values are properly quotes:

```
vrfs:
  blue:
    af:
      ipv4: true
    export:
    - '65000:1'
    id: 1
    import:
    - '65000:1'
    loopback: true
    networks:
    - ipv4: 10.2.0.1/32
    rd: '65000:1'
    vrfidx: 100
```

Which YAML parser and what version of that parser and Jinja2 are you using?

ipspace:
Could be you're using Ruamel instead of PyYAML? I upgraded PyYAML to 6.0.0 just to be on the safe side. No change.

ipspace:
@petercrocker Have you figured out what the problem was?

jbemmel:
> sexagesimal

This doesn't reproduce the issue for me, as the host_vars file is written with quoted ```rd``` values

I have PyYaml installed, not Ruamel

Note that YAML 1.2 dropped support for sexagesimal numbers, but PyYaml only supports YAML 1.1

jbemmel:
Note that ruamel is a dependency of Box (https://github.com/cdgriffith/Box/blob/master/requirements.txt) so not sure why it's not installed by default?

jbemmel:
PR tests fail due to an unexpected kwargs parameter; not sure how to tell pytest to ignore that?

ipspace:
OK, so it looks like I managed to bork my Python environment, and reproduce the bug. When using ruamel.yaml, the YAML output does not include quotes around rd values (because they are not needed in YAML 1.2), whereas Ansible probably uses PyYAML which still understands that as sexagesimal number.

Now we have the root cause: we need quotes around RD values in YAML outputs when using ruamel.

ipspace:
> PR tests fail due to an unexpected kwargs parameter; not sure how to tell pytest to ignore that?

What if mypy/pytest happens to be right?

```
../netlab create -o yaml:nodes.r x.yml
  File "/Users/pipi/Dropbox/Workshops/Fundamentals/Networking 101/Examples/tools/X/../netlab", line 8, in <module>
    netsim.cli.lab_commands()
  File "/Users/pipi/Dropbox/Workshops/Fundamentals/Networking 101/Examples/tools/netsim/cli/__init__.py", line 157, in lab_commands
  File "/Users/pipi/Dropbox/Workshops/Fundamentals/Networking 101/Examples/tools/netsim/cli/create.py", line 82, in run
    output_module.write(topology)
  File "/Users/pipi/Dropbox/Workshops/Fundamentals/Networking 101/Examples/tools/netsim/outputs/yaml.py", line 47, in write
    output.write(cleantopo.to_yaml(ruamel_attrs={'version': (1,1)}))
  File "/usr/local/lib/python3.9/site-packages/box/box.py", line 848, in to_yaml
    return _to_yaml(
  File "/usr/local/lib/python3.9/site-packages/box/converters.py", line 121, in _to_yaml
    return yaml.dump(obj, default_flow_style=default_flow_style, **yaml_kwargs)
TypeError: dump() got an unexpected keyword argument 'ruamel_attrs'
```

box.to_yaml does not support ruamel_attrs when using python-box 5.4.1 ;)

jbemmel:
Another possible fix: ```ruamel_attrs={'default_style': '"'}```

This generates quotes around values when using ruamel - but lots of quotes, everywhere

ipspace:
We know what's wrong and as suspected it is a much bigger issue than RD value on EOS :(( Closing this one.

ipspace:
BTW, @petercrocker: a temporary workaround seems to be to uninstall ruamel.yaml

PR #670, opened 2022-12-06
Initial support for Juniper vMX plus "specific" device types

ssasso:
As per #668 

Specific device types "children" of a vendor can be supported (in this case, vmx is configurable using junos templates and tasks, apart from the initial stuff) - but this can be generic, of course.


ipspace (approved):
Looks good to me. Please let me know when it's ready to be merged.

ssasso:
@ipspace I think for now we are all set.
An eventual support for vMX using Vagrant/Libvirt will be on a separate PR, if needed.

PR #676, opened 2022-12-08
Force use of YAML version 1.2 when using ruamel

jbemmel:
Box defaults to ruamel.yang YAML package, which supports both YAML 1.2 and 1.1. 

See https://github.com/ipspace/netlab/pull/669 on an issue with YAML 1.1 and interpretation of 65000:1 as an int value; YAML 1.2 removes support for such 'sexagesimal' numbers

ruamel.yang is a dependency of Box, but may or may not be installed in the environment. Netlab writes inventory files using ruamel, and Ansible then reads back these same files again using ruamel - so this 'round trip' workflow should work

### DIFF
--- a/netsim/outputs/ansible.py
+++ b/netsim/outputs/ansible.py
@@ -113,7 +113,7 @@ def write_yaml(data: Box, fname: str, header: str) -> None:
   with open(fname,"w") as output:
     output.write(header)
     if callable(getattr(data,"to_yaml",None)):
-      output.write(data.to_yaml())
+      output.write(data.to_yaml(ruamel_attrs={'version': (1,2)}))
     else:                            # pragma: no cover -- this should never happen as we're using Box, but just in case...
       output.write(yaml.dump(data))
     output.close()


ipspace:
Great catch. Thanks a million. Have you tested it or are we at the "should work" stage?

ipspace:
> > Great catch. Thanks a million. Have you tested it or are we at the "should work" stage?
> 
> I have verified that it generates an inventory file with the YAML 1.2 tag at the top - but not full circle. I was hoping that the EOS case could be used to test this...

YAML tag 1.2 at the top will not help if the file is used with a YAML parser that only understands 1.1 (like PyYAML). Ansible still lists PyYAML as its dependency, so this is a wrong solution to the problem (see last comment in #669)

ipspace:
> Another option is `ruamel_attrs={'default_style': '"'}`
> 
> Need to update Box version for that - why was it pinned?

Because version 6.0 didn't work ;)

jbemmel:
> YAML tag 1.2 at the top will not help if the file is used with a YAML parser that only understands 1.1 (like PyYAML). Ansible still lists PyYAML as its dependency, so this is a wrong solution to the problem (see last comment in #669)

If ruamel is installed, it should used both for dumping those YAML files and for reading them back in Ansible. Ansible supports ruamel too, all we have to do is invoke it in the same Python environment (which I think we are doing?)

ipspace:
> > YAML tag 1.2 at the top will not help if the file is used with a YAML parser that only understands 1.1 (like PyYAML). 
> If ruamel is installed, it should get used both for dumping those YAML files and for reading them back in Ansible. Ansible supports ruamel too, all we have to do is invoke it in the same Python environment (which I think we are doing?)

Ansible lists PyYAML as a dependency. It might work with ruamel.yaml, but obviously not when both are present or @petercrocker wouldn't have a problem.

PR #684, opened 2022-12-12
Add support for containerized VyOS.

sysoleg:
We can't use the official 1.3-epa2 image from Docker Hub because it's outdated and doesn't work correctly with virtual network devices [1]. The thing is critical for the container environment with veth devices. Hence for now we can use the unofficial image [2] that is based on VyOS nightly ISO.

1. https://github.com/vyos/vyos-1x/commit/8cf5a4f023c5459cad4c84e93f73a9ddd69be81a
2. https://github.com/sysoleg/vyos-container

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -80,7 +80,7 @@ You cannot use all supported network devices with all virtualization providers. 
 | Mikrotik RouterOS 7                                |          ✅           |              ❌               |            ❌             |
 | Nokia SR Linux                                     |          ❌           |              ❌               |            ✅             |
 | Nokia SR OS                                        |          ❌           |              ❌               |            ✅             |
-| VyOS                                               |          ✅           |              ❌               |            ❌             |
+| VyOS                                               |          ✅           |              ❌               |            ✅             |
 
 **Note:**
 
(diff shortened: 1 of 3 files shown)


ipspace:
@sysoleg Great job!
@ssasso Would you please review this one?

Thank you!

ssasso:
> for i in shell:
>       args.append(i)

yes, I know, this can be better written as:

```
args = ['docker','exec','-it',host] + shell + ['-il']
```

ipspace:
The change to `docker_shell` handling in **netlab connect** is in fc32654 (because it's the right thing to do). In any case, the discussion is moot as `docker exec` usually connects to the container as root (**cml** is a Cumulus VX node):

```
$ netlab connect -v cml
Connecting to container clab-X-cml, starting bash
Executing: ['docker', 'exec', '-it', 'clab-X-cml', 'bash', '-il']
Use vtysh to connect to FRR daemon

cml(bash)#whoami
root
```

@ssasso @sysoleg: Are we good to go?

sysoleg:
@ipspace ``su - vyos`` will be enough. Check my latest commit.

ssasso:
thanks a lot to both of you @ipspace and @sysoleg ! :-)

PR #688, opened 2022-12-15
Multiple JunOS improvements

ssasso:
* vMX license from official public juniper.net website
* JunOS device quirks for better handling of "unit 0"
* VLANs for vMX and vSRX (only routed mode)
* VRF with OSPF and BGP for vMX and vSRX
* MPLS with LDP and L3VPN for vMX and vSRX

Adding multiple integration testing topologies as well.


ssasso:
@ipspace the device quirks for managing the "unit 0" are making the transformation tests to fail.
any idea on how to solve this?

ssasso:
I think that's enough, for now :-)

PR #692, opened 2022-12-20
FRR: don't lose interface IPv6 address after VRF enslavement

sysoleg:
VRF enslavement leads to link flap. Linux kernel doesn't keep IPv6 address by default in this case. keep_addr_on_down sysctl knob fix this behavior.


ipspace:
Thank you! As it's obvious you're much more fluent in Linux networking than some of us that are just fumbling around, would you please take a look at the `initial/frr.j2` template -- the `echo` command at the end of that template should probably also be replaced with a `sysctl`.

PR #693, opened 2022-12-20
FRR: support for MPLS, LDP, L3VPN

sysoleg:


ipspace (approved):
Thanks a million. I didn't believe we could make MPLS data plane work in containers without some extra work on the host side.

PR #695, opened 2022-12-21
Support for generic initial pre-checks ++

ssasso:
I think it would be useful to support initial pre-checks (device based), instead of having to duplicate the whole "initial" provisioning tasks.

Just to give an example, right now for vMX I had to create
https://github.com/ipspace/netlab/blob/dev/netsim/ansible/tasks/vmx/initial.yml

which basically has inside it the duplicate part for 
```
- name: "junos_config: deploying initial config from {{ config_template }}"
  junos_config:
    src: "{{ config_template }}"
  tags: [ print_action, always ]
```
("duplicate" from https://github.com/ipspace/netlab/blob/dev/netsim/ansible/tasks/deploy-config/junos.yml )

With this patch, the real initial config part is still on the "global deploy-config file", but the playbook calls a "pre-check" task file (platform specific) **if it exists**.

Examples are provided for OS10 and ROS7 (with the related support of ROS7 in clab).


ipspace (changes requested):
Like the idea (and will use it for NXOS, EOS in clab and IOS XR). However:

* I think we should use the same mechanism ('with_first_found') to check for file existence everywhere in the playbook. Use 'ignore-errors' and check for valid value afterwards. It would also allow to have the pre-check done based on OS type or device type.
* We have to be careful about playbook tags. If you want this to be executed regardless of how the playbook is invoked (and I think it should be), then the tasks need "tags: [ always ]"
* As it looks like there will be multiple tasks in sequence using the same **tags**, you might want to group them into a **block**.

ssasso:
I also updated the directory name ;)

ipspace (approved):
Perfect. Thank you!

PR #733, opened 2023-01-26
Running Multiple Labs on Linux Servers

ipspace:
Implements 'multilab' plugin that modifies numerous system defaults based on 'lab id' specified in lab topology, defaults, or CLI argument. Documentation @ docs/plugins/multilab.md

### DIFF
--- a/docs/netlab/create.md
+++ b/docs/netlab/create.md
@@ -38,8 +38,8 @@ The _[](netlab-create-output-formats)_ section describes how you can control the
 
 ```text
 usage: netlab create [-h] [--log] [-q] [-v] [--defaults DEFAULTS] [-d DEVICE]
-                     [-p PROVIDER] [-s SETTINGS] [-o OUTPUT]
-                     [topology]
+                     [-p PROVIDER] [-s SETTINGS] [--plugin PLUGIN]
+                     [-o OUTPUT] [topology]
 
 Create provider- and automation configuration files
 
@@ -56,6 +56,7 @@ optional arguments:
                         Default device type
   -p PROVIDER, --provider PROVIDER
                         Override virtualization provider
+  --plugin PLUGIN       Additional plugin(s)
   -s SETTINGS, --set SETTINGS
                         Additional parameters added to topology file
   -o OUTPUT, --output OUTPUT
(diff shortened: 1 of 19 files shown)


petercrocker:
Ok, found a gap in the implementation. If you define a switch with an access interface (veos in this case):
```
# Hexad01 links
  - h1xgw1:
    gateway.protocol: vrrp
    gateway.vrrp.priority: 120
    vlan.access: vlanblue1
  - h1xgw1:
    gateway.protocol: vrrp
    gateway.vrrp.priority: 120
    vlan.access: vlangreen1
```

It gets represented in the Vagrantfile in a non-unique way:
```
  h1xgw1.vm.network :private_network,
                  :libvirt__network_name => "ml_1_21",
                  :libvirt__forward_mode => "veryisolated",
                  :libvirt__dhcp_enabled => false,
                  :libvirt__iface_name => "vgif_h1xgw1_2",
                  :auto_config => false
```

Then when you spin up lab 1, it's no problem, when you spin up the same topology as lab 2, vagrant complains when booting the veos switch with the access interface:

```
 INFO interface: error: There was an error talking to Libvirt. The error message is shown
below:

Call to virDomainCreateWithFlags failed: Unable to create tap device vgif_h1xgw1_2: Device or resource busy
There was an error talking to Libvirt. The error message is shown
below:

Call to virDomainCreateWithFlags failed: Unable to create tap device vgif_h1xgw1_2: Device or resource busy
 INFO interface: Machine: error-exit ["VagrantPlugins::ProviderLibvirt::Errors::FogError", "There was an error talking to Libvirt. The error message is shown\nbelow:\n\nCall to virDomainCreateWithFlags failed: Unable to create tap device vgif_h1xgw1_2: Device or resource busy"]
```

Reporting this now, because I've got to do some $dayjob stuff, will try and look at code fix soon.

ipspace:
Got it: the interface names are non-unique. I will solve that with **defaults.providers.libvirt.ifname_prefix** which would be set to "vgif_" by default, and then modified by _multilab_ plugin.

Thanks a million, now I know what else I need to test... Expect a fix tomorrow or during the weekend.

ipspace:
@petercrocker The interface naming SNAFU is fixed. I managed to start two mixed libvirt/clab labs using a combination of P2P and multi-access links in parallel, so we should be good to go. Please check whether the new code works for you.

petercrocker (approved):
Went through all of it, nothing else to add, and works as promised!

ipspace:
> Went through all of it, nothing else to add, and works as promised!

Thanks a million!

PR #734, opened 2023-01-26
Fixes #731 - JunOS MTU

### LINKED ISSUE #731: [BUG] Cannot set MTU on vSRX
### Describe the bug

MTU can no longer be set on vSRX interfaces. The error is probably caused by the .0 interface name. Here's the relevant config:

```
interfaces {
  ge-0/0/0.0 {
    description "r1 -> r2";
    mtu 1514;

      family inet {
        address 172.16.0.1/24;
      }
  }
}
```

And here's the error message generated during initial device configuration:

```
fatal: [r1]: FAILED! => changed=false
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python3
  msg: |-
    <?xml version="1.0" encoding="UTF-8"?><rpc-error xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:junos="http://xml.juniper.net/junos/21.4R0/junos" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0">
    <error-type>protocol</error-type>
    <error-tag>operation-failed</error-tag>
    <error-severity>error</error-severity>
    <error-message>syntax error</error-message>
    <error-info>
    <bad-element>mtu</bad-element>
    </error-info>
    </rpc-error>
```

### To Reproduce

Execute **netlab up** with the attached topology

### Lab topology

```
provider: libvirt
nodes:
  r1:
    device: vsrx
    mtu: 1500
  r2:
    device: cumulus
    mtu: 1500

links:
- r1-r2
```

### Version

* netlab version 1.4.3/1.5.0

ssasso:
fixes ipspace/netlab#731

### DIFF
--- a/netsim/ansible/templates/initial/junos.j2
+++ b/netsim/ansible/templates/initial/junos.j2
@@ -33,15 +33,22 @@ interfaces {
 {% endif %}
     }
   }
+
+{# Handle MTU and outside the main interface loop, since it needs to be applied only to the master interface #}
+{% for l in interfaces|default([]) %}
+{%   if l.mtu is defined %}
+  {{ l.junos_interface }} {
+    mtu {{ l.mtu + 14 }};
+  }
+{%   endif %}
+{% endfor %}
+
 {% for l in interfaces|default([]) %}
   {{ l.ifname }} {
 {% if l.name is defined %}
     description "{{ l.name }}{{ " ["+l.role+"]" if l.role is defined else "" }}";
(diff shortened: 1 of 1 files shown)


ipspace:
Wow, that was ultra-fast. Thanks a million!

Do you think this is critical enough to push it out as 1.4.3-post2?

ssasso:
Unless there are specific requirements from users, I don't think it's critical enough for a post-release IMHO

PR #745, opened 2023-02-09
Change default password for ASAv

willimei:
In newer version ASAv enforces a minimum password length of 8.
Therefore I have changed the default password to match this criteria.

Users have to rebuild old vagrant boxes to be able to work with them after this change.


ipspace:
Thank you! Will merge it in a few days when I'll start working on 1.5.1. At the moment I'm still fixing a few things that should go into 1.5.0-post2 and I don't want to add a breaking change to it.

ipspace:
"few days" turned into "a month and a half", but finally I got there ;) Thanks again!

PR #749, opened 2023-02-28
Correct clab directory name for labs in sub-directories

jbemmel:
inventory_dir | basename happens to work when the lab is in the same directory, but...


ipspace:
Yeah, working with Ansible paths is a ***** :(

PR #754, opened 2023-03-11
Missing pip install step on Ubuntu 20.04

modonovan:
As per https://stackoverflow.com/questions/71851185/unable-to-install-uninstall-with-pip-due-to-cryptographydeprecationwarning it is required to install cryptography version 3.3.2 on Ubuntu 20.04 before running the netlab install command

Tested on multipass Ubuntu 20.04

### DIFF
--- a/docs/install/ubuntu-vm.md
+++ b/docs/install/ubuntu-vm.md
@@ -78,6 +78,7 @@ end
 sudo apt-get update
 sudo apt-get install -y python3-pip
 sudo pip3 install --ignore-installed networklab
+sudo pip3 install cryptography==3.3.2
 netlab install -y ubuntu ansible libvirt containerlab
 ```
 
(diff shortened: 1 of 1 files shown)


ipspace:
Thanks a million for pointing out this can of worms :( While pip3 worked for me, latest Ansible version required cryptography >= 3.3 and that totally borked everything. What worked for me was

```
sudo pip3 install --upgrade pyopenssl cryptography
```

I would prefer to go with this solution as it doesn't nail cryptography module to a specific version that happens to be over two years old. Could you please check whether it works in your environment?

Also, the same command needs to be added to sample Vagrantfile in docs/install/ubuntu-vm.md. I could do that before merging, but I'd really appreciate if you could test the above.

Thanks again! Ivan

ipspace:
Also, it looks like the first time someone barfs about cryptography is in the ansible.sh script where ansible-pylibssh wants to have a newer version than what's included in LTS 20.04. Maybe we could fix `netsim/install/ansible.sh` instead of adding an extra manual step?

modonovan:
Perfect Ivan, that works better and agreed on not nailing a particular package to a specific release - this makes good sense. I've updated my pull request with the suggested changes. This has now been tested and works on multipass Ubuntu 20.04

PR #763, opened 2023-04-05
Remove dependency on bgp module

jbemmel:
Template implicitly assumed bgp module was being used, it may not be

Note that we may want to simplify the router_id modeling to a single, global attribute per node - instead of per protocol, vrf, etc. There is bgp.router_id, ospf.router_id, etc. but in the typical case each router has a single unique ID

### DIFF
--- a/netsim/ansible/templates/vrf/sros.j2
+++ b/netsim/ansible/templates/vrf/sros.j2
@@ -7,8 +7,8 @@ updates:
   val:
    service-id: {{ vdata.vrfidx }}
    customer: "1"
-   autonomous-system: {{ bgp.as }}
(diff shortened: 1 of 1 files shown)


ipspace:
> Note that we may want to simplify the router_id modeling to a single, global attribute per node - instead of per protocol, vrf, etc. There is bgp.router_id, ospf.router_id, etc. but in the typical case each router has a single unique ID

I agree with that, but I also remember someone who wanted to have multiple router IDs per device for back-to-back connections between VRFs (or whatever it was).

jbemmel:
Sounds like a crazy person exploring exotic corners of the networking universe - must be braver than me...

PR #764, opened 2023-04-05
Remove dependency on Python being installed in Linux containers

jbemmel:
The 'template' action gets executed inside the target container, and hence requires Python to be installed. This change processes the template on the controller, and copies the result

### DIFF
--- a/netsim/ansible/tasks/deploy-config/linux.yml
+++ b/netsim/ansible/tasks/deploy-config/linux.yml
@@ -1,7 +1,13 @@
 - template:
     src: "{{ config_template }}"
-    dest: /tmp/config.sh
+    dest: /tmp/{{inventory_hostname}}_config.sh
+  delegate_to: localhost # Run on controller, to avoid dependency on Python
+
+- name: Copy template file into running container
+  shell: docker cp /tmp/{{inventory_hostname}}_config.sh clab-{{ netlab_name }}-{{inventory_hostname}}:/tmp/config.sh
+  delegate_to: localhost
+
 - name: "Execute /tmp/config.sh to deploy {{ netsim_action }} config from {{ config_template }}"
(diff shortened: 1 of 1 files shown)


jbemmel:
Note that I'm not sure if 'linux' nodes are also used on Vagrant - if so, this needs to be conditional on provider being 'clab'

ipspace:
> Note that I'm not sure if 'linux' nodes are also used on Vagrant

Yes they are.

> if so, this needs to be conditional on provider being 'clab'

Yes. Also, when would `deploy-config/linux.yml` be executed in a container considering we have `linux/initial-clab.yml`? Could you share the actual use case?

jbemmel:
> > Note that I'm not sure if 'linux' nodes are also used on Vagrant
> 
> Yes they are.
> 
> > if so, this needs to be conditional on provider being 'clab'
> 
> Yes. Also, when would `deploy-config/linux.yml` be executed in a container considering we have `linux/initial-clab.yml`? Could you share the actual use case?

I am playing around with your DHCP relay example: https://github.com/jbemmel/netsim-examples/blob/srlinux_dhcp_relay/DHCP/evpn-relay/topology.yml#L23

When replacing the IOS DHCP clients with Alpine containers, the linux.j2 script https://github.com/jbemmel/netsim-examples/blob/srlinux_dhcp_relay/DHCP/evpn-relay/dhcp-client/linux.j2 gets executed

The patch is similar to what you're doing in initial-clab.yml

ipspace:
What if we'd add the provider-specific deployment script to the search path, so the playbook would look for `deploy-config/linux-clab.yml` before trying `deploy-config/linux.yml`? That might produce less clutter than the interesting way Ansible implements if-then-else block ;)

jbemmel:
See https://github.com/ipspace/netlab/pull/766

PR #765, opened 2023-04-06
Use passive flag from interface

jbemmel:


ipspace:
FWIW, it would be nice if you could add the platform name to the PR title. I like to have easy-to-identify platform template fixes, it makes writing the release notes a no-brainer.

PR #766, opened 2023-04-07
For clab provider, avoid dependency on Python for custom configuration

jbemmel:


ipspace:
Merged. Thank you!

The really funny part of this whole discussion is that the same solution was already used to deploy modules and custom configs specified in the lab topology, but it's nice to have consistent implementation no matter how it's invoked.

PR #775, opened 2023-04-10
Deal with Python Box expanding dots in keys read from YAML files

ipspace:
The clab.binds and clab.config_templates dictionaries could contain dots in keys (filenames) which results in weird values in clab.yml configuration file. This patch:

* Checks the validity of clab.binds and clab.config_templates parameters
* Converts them into lists of host:target strings
* Converts those strings into dicts and back as needed

### DIFF
--- a/docs/labs/clab.md
+++ b/docs/labs/clab.md
@@ -68,7 +68,7 @@ nodes:
 
 ## Using File Binds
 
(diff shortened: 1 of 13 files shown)


ipspace:
@jbemmel: It would be great if you could try this one out on some of your more complex containerlab topologies. It works for me, but then I never used custom binds or custom templates (beyond what we need for Linux and FRR)

PR #782, opened 2023-04-13
srlinux: ipv4 unnumbered is not supported

jbemmel:


jbemmel:
Figures - will submit a more robust PR

PR #794, opened 2023-04-26
Support multi-line embedded startup-config for clab

jbemmel:


ipspace:
I think you're looking for `startup_config | indent(6)` or something similar

https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.indent

jbemmel:
See https://github.com/ipspace/netlab/pull/804

PR #803, opened 2023-05-03
Support srlinux r23.3.1

jbemmel:
SR Linux 23.3.1 introduces some breaking YANG model changes:
* ipv4/ipv6 config now has an explicit 'admin-state: enable'
* ACL entries have names instead of sequence IDs
* BGP address families are reorganized

As it would become too complex to maintain backwards compatibility, this PR sets the default SR Linux image to 23.3.1


ipspace:
Please rebuild the test results (the node `image` attribute has changed in every test that uses SR Linux), then I'll merge the PR. Thank you!

PR #804, opened 2023-05-03
Support multi-line startup-config for clab

jbemmel:


ipspace:
I'll try to find an easier way of doing it, this seems a bit too convoluted.

PR #805, opened 2023-05-15
Update clab.j2 for ipv6 mgmt

jbemmel:
Not used much, but...


ipspace:
Thanks a million. Totally missed this one.

PR #806, opened 2023-05-15
Update containerlab.sh - pick specific Containerlab version

jbemmel:
I've spoken to the maintainers and they agree the recent version upgrade could have been handled more smoothly - but going forward, they suggested picking a specific version

### DIFF
--- a/netsim/install/containerlab.sh
+++ b/netsim/install/containerlab.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+
+# Install a specific version of Containerlab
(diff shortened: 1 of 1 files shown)


ipspace:
> I've spoken to the maintainers and they agree the recent version upgrade could have been handled more smoothly

Let's call that the understatement of the year ;)

> but going forward, they suggested picking a specific version

AKA RFC 1925 Rule 6 ;) I have no problem doing that in containerlab.sh, but then we also have to document that somewhere in the installation guide for people doing installs on other Linux platforms.

PR #807, opened 2023-05-15
Update clab.yml - allow srl-agents attribute

jbemmel:
Points at a missing test case - I'll see if I can add that


ipspace:
Dang, I missed this one. Apologies, and thanks for the fix!

PR #808, opened 2023-05-15
Add clab provider attribute test case

jbemmel:


ipspace:
Thanks a million for adding the test case -- the test cases are the only thing that keeps me semi-sane with all the complexity we have to deal with ;)

jbemmel:
Yeah, I figured as much ;)

PR #813, opened 2023-06-01
Initial integration of vJunos EVO (vPTX)

ssasso:
Initial integration/testing of vJunos EVO (vPTX).


ipspace:
Wow. Thanks a million!

ssasso:
Image creation is still WIP, btw, but will arrive soon

netsim/devices/vptx.yml
```diff
@@ -0,0 +1,51 @@
...
+libvirt:
+  image: juniper/vptx
+  # New "pre-install hooks" to be created... in a similar way to the ISO building method.
+  pre_install: vptx
+  create:
+    virt-install --connect=qemu:///system --name=vm_box --arch=x86_64 --vcpus 4 --ram 8192
+      --disk path=vm.qcow2,size=10,bus=virtio --disk path=/tmp/vptx.bootstrap.qcow2,bus=usb --boot hd
+      --os-variant generic --import --nographics --serial pty
+      --cpu IvyBridge,+vmx
+      --qemu-commandline="-smbios type=0,vendor=Bochs,version=Bochs -smbios type=3,manufacturer=Bochs -smbios type=1,manufacturer=Bochs,product=Bochs,serial=chassis_no=0:slot=0:type=1:assembly_id=0x0D20:platform=251:master=0:channelized=no"
```
ssasso, line 46: @ipspace it seems the libvirt script is not able to expand this commandline params, as you see from the args array... tried both with `"` and `'`.
any idea before I get deeper into this ocean?

```
usage: virt-install --name NAME --memory MB STORAGE INSTALL [options]
virt-install: error: argument -s/--file-size: invalid float value: 'mbios'
Error executing virt-install --connect=qemu:///system --name=vm_box --arch=x86_64 --vcpus 4 --ram 8192 --disk path=vm.qcow2,size=10,bus=virtio --disk path=/tmp/vptx.bootstrap.qcow2,bus=usb --boot hd --os-variant generic --import --nographics --serial pty --cpu IvyBridge,+vmx --qemu-commandline='-smbios type=0,vendor=Bochs,version=Bochs -smbios type=3,manufacturer=Bochs -smbios type=1,manufacturer=Bochs,product=Bochs,serial=chassis_no=0:slot=0:type=1:assembly_id=0x0D20:platform=251:master=0:channelized=no' --network=network:vagrant-libvirt,model=virtio --network network=vptx_PFE_LINK,model=virtio --network network=vptx_RPIO_LINK,model=virtio --network network=vptx_RPIO_LINK,model=virtio --network network=vptx_PFE_LINK,model=virtio:
  Command '['virt-install', '--connect=qemu:///system', '--name=vm_box', '--arch=x86_64', '--vcpus', '4', '--ram', '8192', '--disk', 'path=vm.qcow2,size=10,bus=virtio', '--disk', 'path=/tmp/vptx.bootstrap.qcow2,bus=usb', '--boot', 'hd', '--os-variant', 'generic', '--import', '--nographics', '--serial', 'pty', '--cpu', 'IvyBridge,+vmx', "--qemu-commandline='-smbios", 'type=0,vendor=Bochs,version=Bochs', '-smbios', 'type=3,manufacturer=Bochs', '-smbios', "type=1,manufacturer=Bochs,product=Bochs,serial=chassis_no=0:slot=0:type=1:assembly_id=0x0D20:platform=251:master=0:channelized=no'", '--network=network:vagrant-libvirt,model=virtio', '--network', 'network=vptx_PFE_LINK,model=virtio', '--network', 'network=vptx_RPIO_LINK,model=virtio', '--network', 'network=vptx_RPIO_LINK,model=virtio', '--network', 'network=vptx_PFE_LINK,model=virtio']' returned non-zero exit status 2.
Fatal error in netlab: Aborting
```

ipspace: `lp_create_vm` expects `libvirt.create` to be a string which is then passed to `external_commands.run_command`. That function splits a string into list assuming space separator. It does not handle quoted strings.

That same function does accept command in list format, so we could fix `lp_create_vm` to handle strings and lists (minor annoyance), or you could create an XML VM definition like I did for EOS or IOSv -- you need `_device_.xml.j2` file in `netsim/install/libvirt` and `create_template` parameter (instead of `create` command) in device definition.

ipspace: If you decide to go for "list of arguments" approach, then this line has to be executed only if the command is a string:

https://github.com/ipspace/netlab/blob/dev/netsim/cli/libvirt.py#L125

Also, it would be nice to change typing hints in `abort_on_failure` from `str` to `typing.Union[str,list]` (**mypy** wouldn't notice the difference, but it's good form to keep things in order)

PR #830, opened 2023-06-21
Merge global VRF data with node VRF data in post-transform hook

ipspace:
Some module (like EVPN) modify global VRF data and expect the modified global VRF data to be used for node VRF data. Code introduced in #482 pulls global VRF data into node VRF data very early in the transformation process -- in those cases, the global changes are never copied into the node VRF data.

This PR fixes the #482 logic to create stub node VRF structures, and carefully populates them with auto-generated or global ID/RD/RT values. The full global VRF data is then merged with node VRF data in post_transform hook.


jbemmel (approved):
Looks ok to me

PR #832, opened 2023-06-21
Evpn ebgp test/example

jbemmel:
An example topology using only eBGP sessions for IPv4 and EVPN, and no IGP. The spine is configured as a route reflector; leaf1 is doing IRB, leaf2 is purely L2 with no IP or interface participating in the service

For some reason it doesn't yet work when using FRR as hosts; to be investigated

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -142,12 +142,14 @@
 {%  if vrf_bgp.rr|default(0) %}
     passive-mode: True
     _annotate_passive-mode: "Improve robustness during startup"
-   route-reflector:
-    cluster-id: {{ vrf_bgp.rr_cluster_id|default(False) or router_id }}
-    client: True
 {%  else %}
    next-hop-self: {{ vrf_bgp.next_hop_self|default(False) }}
 {%  endif %}
+{% endif %}
+{% if vrf_bgp.rr|default(0) %}
+   route-reflector:
+    cluster-id: {{ vrf_bgp.rr_cluster_id|default(False) or router_id }}
+    client: True
 {% endif %}
 {% endmacro %}
 
--- a/netsim/ansible/templates/evpn/srlinux.j2
+++ b/netsim/ansible/templates/evpn/srlinux.j2
@@ -10,12 +10,15 @@ updates:
   val:
    group:
    - group-name: {{ 'ibgp-ipv4' if type=='ibgp' else 'ebgp' }} # Could create a dedicated group for EVPN only?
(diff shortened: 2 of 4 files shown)


ipspace:
I have no problem with fixes to config templates, but the topology makes no sense. You're running EVPN/VXLAN on what you call hosts (but what are effectively full-blown routers), so the fabric is not involved in overlay packet forwarding at all.

Also, I see no value in having this as another integration test. We already have IBGP-over-EBGP, and if you wish you could create a scenario with pure EBGP, but even that is questionable as you already added vxlan-routing-ibgp-over-ebgp-unnumbered.yml.

The purpose of the integration tests is to have a small number of easy-to-understand scenarios that stress-test all the capabilities we want to test, not to have convoluted stuff that nobody (but its author) understands ;)

Please resubmit the PR without the integration test topology -- the template changes are valuable.

jbemmel:
Believe it or not, but this comes close to a topology that people are deploying. The hosts are servers running FRR with eBGP unnumbered; the issue is that for whatever reason it doesn't work with FRR (but it does with SR Linux, minus EVPN over eBGP unnumbered which is not supported)

The fabric gets involved as border leaves towards the Internet; in the lab leaf1 has an IRB interface that can be pinged from the hosts.

Convoluted, yes, useful as integration test - maybe, I was thinking perhaps we could figure out why FRR doesn't work in this scenario

ipspace:
> Believe it or not, but this comes close to a topology that people are deploying.

I can easily believe that ;)

> The fabric gets involved as border leaves towards the Internet; in the lab leaf1 has an IRB interface that can be pinged from the hosts.

Then it would be nice to have another VRF in the scenario to test the packet forwarding functionality. Pinging a directly-connected IP address seems trivial.

> Convoluted, yes, useful as integration test - maybe

The integration tests bundled with _netlab_ are supposed to help developers working on new devices or new functionality of existing devices. They have to be as simple as possible and focused on a single device behavior (as in "make sure your configuration works in this scenario").

> I was thinking perhaps we could figure out why FRR doesn't work in this scenario

Oh, that's a totally different discussion that's probably best solved in some other way, not with a PR.

jbemmel:
See #833 - allowas-in flag was missing under l2vpn evpn address family

May also apply to IOS and other platforms

ipspace:
> See #833 - allowas-in flag was missing under l2vpn evpn address family

Nice catch.

> May also apply to IOS and other platforms

I'm pretty sure nobody thought about using that. Now this is an interesting scenario that should be turned into an integration test. Will you do it or should I?

jbemmel:
I'll do it

PR #840, opened 2023-06-26
Replace gNMI plugin with srlinux ansible collection

jbemmel:
Due to lack of maintenance and support for the gnmi plugin, the Ansible galaxy collection is a better way forward

Obviously needs some further testing - will do that as I go forward

Could consider adding the installation of the collection to the 'Ansible' install script

### DIFF
--- a/netsim/ansible/tasks/deploy-config/srlinux.yml
+++ b/netsim/ansible/tasks/deploy-config/srlinux.yml
@@ -30,6 +30,7 @@
 - name: Update SRL {{ netsim_action }} node configuration (template={{config_template}})
   when: d!=[] or (u!=[] and u!="") or r!=[]
   vars:
+    ansible_network_os: nokia.srlinux.srlinux
     cfg: "{{ lookup('file', tempfile_1.path ) | from_yaml }}"
     d: "{{ cfg.delete | default([]) }}"
     u: "{{ cfg.updates | default([]) }}"
--- a/netsim/ansible/tasks/deploy-config/srlinux.yml
+++ b/netsim/ansible/tasks/deploy-config/srlinux.yml
@@ -1,12 +1,14 @@
 ---
-
-- name: Check that required plugin exists, use 'netlab install grpc' to install it
+- name: Check version of nokia.srlinux
   vars:
-    ansible_connection: nokia.grpc.gnmi
+    nokia_srlinux_version: "{{ lookup('community.general.collection_version', 'nokia.srlinux') }}"
   register: plugin_checked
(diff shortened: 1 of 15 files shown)


ipspace:
Based on "Obviously needs some further testing" I converted this PR to draft. When you're ready to merge it (= when the further testing proved it's safe to merge) please let me know and I'll merge it.

ipspace:
According to @sdktr it looks like new versions of Ansible no longer work with the software installed with `netlab install grpc` (could be related to https://github.com/ansible/ansible/pull/79372). If that's the case, then it might be worthwhile getting this PR tested and merged.

ipspace:
@jbemmel: Any progress on this or should we forget this whole idea?

hellt:
I can help with the switch. I don't know the netlab parts, but know srlinux and its collection parts.
If there is a certain target state for each of the tasks I can assist in transition

ipspace:
OK, I'm biting ;) As we have integration tests that should check most of the device configuration features, here's the plan:

* I managed to turn this PR into a local branch in this repository https://github.com/ipspace/netlab/tree/srlinux-json (@jbemmel's work is safe) 
* I did a rebase onto the current dev branch, added the device dependency on Ansible collection (a functionality we didn't have when this PR was created) and fixed a few YAML issues.
* I'll run the integration tests to see what breaks, and then we can start fixing stuff.

I can fix the obvious staff (looks like 'path' needs to start with a / and 'val' should become 'value'), but there are numerous other things that have to be fixed from the config correctness perspective, and I won't be able to fix those. FWIW, SR Linux is the most broken platform we're testing at the moment (see https://release.netlab.tools/ for details), so there will be a lot of things to fix.

Finally, what about SR OS?

hellt:
> I can fix the obvious staff (looks like 'path' needs to start with a / and 'val' should become 'value'), but there are numerous other things that have to be fixed from the config correctness perspective, and I won't be able to fix those.

great plan, feel free to ping me when I can have a look at the test harness and the tests that require more involved attention.



> FWIW, SR Linux is the most broken platform we're testing at the moment

_SR Linux implementation within netlab_ might be the most broken, not SR Linux itself. fwiw it is the best NOS out there, objectively 😋

ipspace:
> _SR Linux integration with netlab_ might be the most broken

Correct. Stupid me, didn't want to imply anything else 😎

> not SR Linux itself. fwiw it is the best NOS out there, objectively 😋

If you say so yourself 😜 But yeah, I have no problem with that claim ;)

ipspace:
Replaced by #1291

PR #843, opened 2023-06-27
Add support for RFC9355 OSPF BFD Strict-Mode

jbemmel:
Could also be added as a plug-in - let me know if that's preferred


ipspace:
Picture perfect ;) and yes, I'm perfectly OK with having this in the OSPF module. Let's see when we get implementations from other vendors ;))

PR #853, opened 2023-07-04
Generate module support lists from device features (implements #689)

ipspace:


jbemmel:
Would suggest to leave devices removed from the "supported_on" list in a comment, during the transition.
An unaware reader of the code now might think that a given feature is not supported on (say) eos

PR #857, opened 2023-07-29
Align device features/inheritance

ssasso:
Aligned device feature "auto-generation" for

- dellos10
- arubacx
- vyos
- routeros
- routeros7

Additionally, created a "meta" device junos, with no specific provider support (apart from external), to be used for other juniper devices attributes inheritance.


ipspace:
Great. Thank you!

PR #865, opened 2023-08-22
Fix EVPN VLAN attributes

jbemmel:
Due to a typo, vlan attributes were broken

Added a regression test case


ipspace:
Thank you!

We already have a few fixes of pretty annoying bugs, so I'll push out 1.6.2 in a few days (easier than cherry-picking those bugs into 1.6.1-post1)

PR #870, opened 2023-09-08
Add containerlab example to nodes docs

sdktr:
Document a working containerlab + sros example. Adds defaults and node specific overrides.

### DIFF
--- a/docs/nodes.md
+++ b/docs/nodes.md
@@ -106,10 +106,30 @@ nodes:
 
 ### Containerlab Attributes
 
-* **clab.kind** -- [containerlab device kind](https://containerlab.dev/manual/kinds/kinds/). Set in the system defaults for all supported devices.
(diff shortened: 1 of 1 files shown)


ipspace:
Glad to see you joined netlab contributors. Welcome!

jbemmel:
While I appreciate the contribution intent, I'm afraid this may confuse users - the way to select SR OS support, is to use a node of kind 'sros' - ~not~ clab.type: vr-sros

ipspace:
@jbemmel Agreed, thanks for the feedback. I planned to rewrite the commit anyway and move the text to clab.md. Will use some strong language to point out one should NOT change clab.kind.

ipspace:
@jbemmel Hope you're OK with changes in eff1e39

jbemmel:
Yes, much better thanks

Issue #874, opened 2023-09-21
[BUG] vlan.mode set to 'irb' when set as 'bridge' at node level


jbemmel:
### Describe the bug

In a topology where the intent is to pass a trunk of vlans in L2 mode by setting `vlan.mode=bridge` at the node level, IPs are correctly not assigned but the individual vlan.mode remains as 'irb'

### To Reproduce

`netlab create test.yml`

### Expected behavior
The host_vars for r1 should say `vlan.mode: bridge`, as there is no IP to do `irb`

### Lab topology

```
provider: clab

defaults.device: frr

module: [vlan]

vlans:
 v1:
  mode: irb
 v2:
  mode: irb

nodes:
  r1:
   vlan.mode: bridge # Override to pass VLANs as L2
  r2:

links:
- r1:
  r2:
  vlan.trunk: [v1,v2]
```

### Output

host_vars/r1:
```
- bridge_group: 1
  ifindex: 4
  ifname: vlan1000
  name: VLAN v1 (1000) -> [r2]
  neighbors:
  - ifname: vlan1000
    ipv4: 172.16.0.2/24
    node: r2
  type: svi
  virtual_interface: true
  vlan:
    mode: irb   <!-- should be 'bridge'
```

host_vars/r2:
```
- bridge_group: 1
  ifindex: 4
  ifname: vlan1000
  ipv4: 172.16.0.2/24  <!-- correctly assigns ip here, but not on r1
  name: VLAN v1 (1000) -> [r1]
  neighbors:
  - ifname: vlan1000
    node: r1
  type: svi
  virtual_interface: true
  vlan:
    mode: irb
```
### Version

Latest dev branch

### Additional context

Another edge case, but I thought I'd document it anyway ;)

ipspace:
Agreed, looks like the code does not conform to the documented behavior https://netlab.tools/module/vlan/#vlan-forwarding-modes

Thank you, will fix.

PR #876, opened 2023-09-22
Fix node vlan.mode override, update test case

jbemmel:
Fix for https://github.com/ipspace/netlab/issues/874

Note how explicitly adding the default vlan.mode setting changes things - it could be argued that the code should consistently spell out attribute values for each object (and hence the 'blue' vlan would also have a 'mode' attribute)


ipspace:
This fix is incorrect:

* node-level **vlan.mode** overwrites VLAN-level **vlan.mode** for VLANs defined in the node.
* global VLAN definitions are copied/merged with node-level VLAN definitions in several places, this fix addresses only one of those scenarios.

#879 addresses all scenarios I was able to identify.

PR #879, opened 2023-09-23
Copy node-level vlan.mode into global VLANs copied into node VLANs

### LINKED ISSUE #874: [BUG] vlan.mode set to 'irb' when set as 'bridge' at node level

ipspace:
Fixes #874

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -614,6 +614,9 @@ def create_node_vlan(node: Box, vlan: str, topology: Box) -> typing.Optional[Box
     for m in list(node.vlans[vlan].keys()):                         # Remove irrelevant module parameters
       if not m in node.module and m in topology.module:             # ... it's safe to use direct references, everyone is using VLAN module
         node.vlans[vlan].pop(m,None)
+    node_mode = node.get('vlan.mode',None)                          # When copying global VLANs into node VLANs
+    if not node_mode is None:                                       # ... make sure node vlan.mode takes precedence
+      node.vlans[vlan].mode = node_mode                             # ... over global VLAN vlan.mode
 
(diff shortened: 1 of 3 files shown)


ipspace:
This is an alternate fix for #874. It copies **vlan.mode** into a VLAN only if the VLAN is copied from the global VLAN definition.

Issue #881, opened 2023-09-27
[BUG] Global VLAN mode still leaks into node VLAN data


ipspace:
### Describe the bug

Even after fixing #874 (or not 🤷‍♂️) there's still a scenario where the VLAN forwarding mode set in global VLAN data is taking precedence over node `vlan.mode`.

### Lab topology

```
defaults.device: frr

module: [ vlan ]
vlans:
  red:
  green:
    mode: route

nodes:
  s1:
  s2:
    vlan.mode: bridge

links:
- s1:
  s2:
  vlan.trunk: [ red, green ]
```

### Expected behavior

VLAN data on S2 is correct:

```
netlab create vm.yml -o yaml:nodes.s2.vlans
green:
  bridge_group: 1
  id: 1001
  mode: bridge
  prefix:
    allocation: id_based
    ipv4: 172.16.1.0/24
red:
  bridge_group: 2
  id: 1000
  mode: bridge
  neighbors:
  - ifname: vlan1000
    node: s2
  - ifname: vlan1000
    ipv4: 172.16.0.1/24
    node: s1
  prefix:
    allocation: id_based
    ipv4: 172.16.0.0/24
```

However, the S2 subinterface for the green VLAN is routed. Even worse, it has no IP address, but also no corresponding VLAN interface.

```
- ifindex: 1
  ifname: eth1
  linkindex: 1
  name: s2 -> s1
  neighbors:
  - ifname: eth1
    node: s1
  subif_index: 2
  type: p2p
- bridge_group: 1
  ifindex: 2
  ifname: eth1.1001
  name: s2 -> s1
  neighbors:
  - ifname: eth1.1001
    ipv4: 172.16.2.1/24
    node: s1
  parent_ifindex: 1
  parent_ifname: eth1
  type: vlan_member
  virtual_interface: true
  vlan:
    access_id: 1001
    mode: route
    routed_link: true
```

jbemmel:
The issue is in `routed_access_vlan`: https://github.com/ipspace/netlab/blob/dev/netsim/modules/vlan.py#L37
```
def routed_access_vlan(link: Box, topology: Box, vlan: str) -> bool:
  def_link  = link.get('vlan.mode',None)
  def_vlan  = topology.get(f'vlans.{vlan}.mode',None)
  def_global = topology.get('vlan.mode','irb')

  if log.debug_active('vlan'):
    print(f'routed_access_vlan: {link}')
[...]

PR #882, opened 2023-09-27
Extend vlan mode test case to include node local override to 'bridge'

jbemmel:
See https://github.com/ipspace/netlab/issues/881

Bug is triggered by `mode: route`, using `mode: vl_route` does not trigger the logic that generates the routed subinterface

Note: One could argue that `neighbors` should not contain an IP for s3:
```
vlans:
      black:
        bridge_group: 2
        id: 1004
        mode: n_irb
        neighbors:
        - ifname: vlan1004
          ipv4: 172.16.4.2/24
          node: s2
        - ifname: vlan1004
          ipv4: 172.16.4.1/24
          node: s1
        - ifname: vlan1004
          ipv4: 172.16.4.3/24  <-- not allocated on s3 due to 'bridge' mode
          node: s3
        prefix:
          allocation: id_based
          ipv4: 172.16.4.0/24
```

### DIFF
--- a/netsim/cli/__init__.py
+++ b/netsim/cli/__init__.py
@@ -27,7 +27,7 @@ def parser_add_debug(parser: argparse.ArgumentParser) -> None:
                     'vlan','vrf','quirks','validate','addressing','groups','quirks','status',
                     'external','defaults'
                   ],
-                  help=argparse.SUPPRESS)
+                  help='Debug specific transformation logic')
 
 def common_parse_args(debugging: bool = False) -> argparse.ArgumentParser:
   parser = argparse.ArgumentParser(description='Common argument parsing',add_help=False)
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -43,16 +43,16 @@ def routed_access_vlan(link: Box, topology: Box, vlan: str) -> bool:
     print(f'routed_access_vlan: {link}')
     print(f'... vlan {vlan} def_mode {def_vlan}')
   for intf in link.interfaces:
+    node_vlan_mode = topology.nodes[intf.node].get('vlan.mode',def_vlan)
     mode = (
       intf.get('vlan.mode',def_link) or
-      topology.nodes[intf.node].get(f'vlans.{vlan}.mode',def_vlan) or
-      topology.nodes[intf.node].get('vlan.mode',def_global) or
-      'irb')
+      topology.nodes[intf.node].get(f'vlans.{vlan}.mode',node_vlan_mode) or
+      def_global)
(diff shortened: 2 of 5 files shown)


ipspace:
Thanks for the PR, but I'll create a separate simpler regression test for #881. Admittedly the tests will run marginally slower, but it's really hard figuring out what's going on once a complex test case (like the vlan-routed-vrf) blows up.

As for the changes to `parser_add_debug` -- I don't want to show that to the regular users. It would just clutter the help messages.

Finally, thanks for pointing out I probably messed up the setting precedence in `routed_access_vlan`. Will fix that, but along the lines of how it's done now. When looking at your code I was pretty sure it was wrong, and it took me a while to realize it's probably correct, which is a bad symptom to have in an already-overcomplex bit of code.

jbemmel:
I looked for `--debug` documentation and couldn't find it anywhere. If not in the help, at least the developer docs should mention it

`routed_access_vlan` could call `interface_vlan_mode` instead of duplicating the logic. 

I do see the value of having a 'computed' `_vlan_mode` attribute, separate from what the user provided. Need to clean it up from the final output

Lastly, if we make sure `_vlan_mode` is set on all interfaces of a link (in all cases), then we could also compute `_routed_access_vlan` once. I think that's a nice way to reduce complexity: Make computations explicit by setting internal attributes, then avoid complex derivation logic in the code

ipspace:
> I looked for `--debug` documentation and couldn't find it anywhere. If not in the help, at least the developer docs should mention it

Agreed. Will add.

> `routed_access_vlan` could call `interface_vlan_mode` instead of duplicating the logic.

Correct. Will fix.

> I do see the value of having a 'computed' `_vlan_mode` attribute, separate from what the user provided. Need to clean it up from the final output

It is already done that way, but inconsistently.

> Lastly, if we make sure `_vlan_mode` is set on all interfaces of a link (in all cases), then we could also compute `_routed_access_vlan` once.

We could, but I don't think we need that. Have to check how often _routed_access_vlan is called.

> I think that's a nice way to reduce complexity: Make computations explicit by setting internal attributes, then avoid complex derivation logic in the code

Agreed.

Thanks a million for the suggestions. I will implement the minimum amount of changes necessary to fix #881 in #883 and then when I have well-defined results of test cases go for a cleanup in another PR.

ipspace:
Just FYI:

> routed_access_vlan could call interface_vlan_mode instead of duplicating the logic.

Wanted to do that, but unfortunately `routed_access_vlan` also needs to consider link `vlan.mode` and it's somewhere in the middle of the precedence chain.

PR #883, opened 2023-09-27
Propagate node VLAN mode into link 'interfaces' early enough to matter

ipspace:


jbemmel (approved):

ipspace:
Almost there. Have to verify the vlan-routed-vrf test case.

PR #888, opened 2023-09-29
Add ebgp multihop

jbemmel:
This adds bgp.multihop and bgp.vpn.multihop dictionaries that enable users to declare multi-hop sessions at the node level.
Including documentation and updated scripts for SR Linux / SR OS

One use case for this is peering between route reflectors in different AS, see https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9600/software/release/16-11/configuration_guide/mpls/b_1611_mpls_9600_cg/configuring_mpls_vpn_inter_as_ipv4_bgp_label_distribution.html

### DIFF
--- a/docs/plugins/ebgp.utils.md
+++ b/docs/plugins/ebgp.utils.md
@@ -3,9 +3,12 @@
 The **ebgp.utils** plugin (contributed by Stefano Sasso) implements several EBGP nerd knobs, including:
 
 * **bgp.allowas_in** is an interface (node-to-link attachment) attribute that controls whether a BGP router accepts one or more copies of its own AS number in incoming update. It takes an integer value between 1 and 10. A *true* value sets it to 1.
+* **bgp.tcp_ao** is a string attribute that sets the algorithm to use for TCP Authentication Option (RFC5925)
 * **bgp.as_override** is an interface (node-to-link attachment) boolean attribute that controls whether a BGP router replaces peer AS with its own AS.
 * **bgp.default_originate** is a boolean attribute that controls whether a BGP router advertises a default route to its neighbor(s). It can be specified on global, node, link or interface level.
 * **bgp.gtsm** is an integer attribute that enables the Generic TTL Security Mechanism (GTSM). It can be specified on global, node, link or interface level and takes an integer value between 1 and 254. A *true* value sets it to 1 (remote router can be up to one hop away).
+* **bgp.multihop** and **bgp.vpn.multihop** is a dictionary of neighbor=hopcount key/values to create multihop EBGP sessions for ipv4/6
+and VPN-v4 address families
 * **bgp.password** is a string attribute that can be specified on global, node, link or interface level. It specifies the MD5 password used on EBGP sessions.
 * **bgp.timers** is a dictionary of BGP timers that can be specified on global, node, link or interface level. It has three elements:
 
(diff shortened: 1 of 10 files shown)


ipspace:
I have nothing against EBGP multihop added to the attributes supported by **ebgp.utils** However, I see limited functionality if you want it applied to EBGP neighbors created by the BGP module as that one won't create the BGP session between route reflectors you want to have.

With that in mind, I think we should stick to the "configure **ebgp.utils** attributes at link level" approach. That will add **bgp.multihop** to the BGP neighbor and you can use that in the templates if you so desire. However, for the "true" EBGP multihop session you will have to generate the neighbor data somewhere else, and in that case it doesn't matter where you apply **bgp.multihop** attribute.

Finally, having per-AF **bgp.multihop** makes absolutely no sense. It's a session-level (not AF-level) attribute and can only be meaningfully applied to a neighbor, not an AF with that neighbor.

jbemmel:
fair enough, back to the drawing board

The first issue is creating the multihop ebgp sessions, as the bgp module only creates 1-hop ebgp and multihop ibgp to route reflectors. In this PR I added https://github.com/ipspace/netlab/pull/888/files#diff-1ac5feff8c23f6b8f6739aeba3a7e8754e6f12d63e5e03a18d4add3b6870b1f8R81 but I agree a 'plugin' may not be the best place to be creating these. Would you prefer modifying the regular bgp module to create multihop ebgp sessions, given some flag or session structure?

For example, at top level similar to 'links'
```
bgp.multihop:
- n1-n2
- n3-n4
```

ipspace:
Wouldn't it be better to have a separate `ebgp.multihop` plugin and use it to implement the `bgp.multihop` structure like you suggested? The only consideration is that it would have to create neighbors soon enough (before BGP module does).

You could also use the `augment.links` routines to parse the data structure, and then validate the attributes (add a separate attribute namespace to `defaults.bgp.attributes`).

jbemmel:
I'm thinking that one would want things like password protection and TCP-AO on those multi-hop sessions, given that increasing the TTL hop count expands the attack surface. For that, I think it would be better to add the multi-hop sessions in the `ebgp.utils` plugin, instead of a separate plug-in which would then require the user to include both

jbemmel:
Note: Does not handle VPNv4/v6 address family cases yet - only ipv4/v6

ipspace:
> I'm thinking that one would want things like password protection and TCP-AO on those multi-hop sessions, given that increasing the TTL hop count expands the attack surface.

You're yet again trying to boil the ocean.

> For that, I think it would be better to add the multi-hop sessions in the `ebgp.utils` plugin, instead of a separate plug-in which would then require the user to include both

Where's the harm in that? You want two features, you add two plugins.

> Specifically, the `bgp` module creates bgp sessions in `node_post_transform`. The plugin could add a virtual interface (of type 'multihop' with a single neighbor) in `post_transform`, and (optionally) remove it after the bgp module has processed the interfaces to create the sessions

I like the generic idea:

* Generate a link-like structure defining application-level sessions (not just for multihop BGP, we could also use it for MPLS-TE or directed LDP)
* Add virtual links to the `links` list early in the process. For multihop BGP sessions it would have to be after `node_transform` to ensure we have loopback interfaces already set up.
* Cleanup resulting links and interfaces in `post_transform` phase or even add the `session` link type as core feature

We do have to make sure that `ebgp.multihop` is executed after `ebgp.utils`, so we might need `execute_after` attribute for plugins. Don't tell me that doing everything in a single plugin would solve it ;) 

> We would add `defaults.bgp.attributes.session` and include `password` and other attributes there, keeping the convention where 1-hop ebgp session attributes can be defined on the links

No need to do that if you create `session` links out of BGP sessions -- you can just check against BGP links attributes

jbemmel:
> We do have to make sure that `ebgp.multihop` is executed after `ebgp.utils`, so we might need `execute_after` attribute for plugins. Don't tell me that doing everything in a single plugin would solve it ;)
> 

How about we put the logic in a new `session` module instead, such that we can use the existing module dependency controls:

The `bgp`, `mpls` and other modules that need it would automatically bring in the `session` module (early, e.g. in `module_pre_default`) if not already included

PR #890, opened 2023-09-30
Consistently use _vlan_mode attribute throughout the VLAN module

### LINKED ISSUE #885: Rewrite the 'vlan.mode' handling to use '_vlan_mode'
Current handling of VLAN forwarding mode is a mess with a ton of somewhat-overlapping functions doing more-or-less the same thing.

Fixing #881 will involve consistently setting `_vlan_mode` on interfaces (first on link.interfaces and then on node.intefaces). We have to use that as the sole source of information.

ipspace:
Fixes #885

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1150,12 +1150,16 @@ def cleanup_vlan_flags(topology: Box) -> None:
 
   for n in topology.nodes.values():
     for intf in n.interfaces:
-      for int_attr in ['vlan_name','_global_merge']:
-        intf.pop(int_attr,None)
+      for int_attr in ['vlan_name','_global_merge']:        # vlan_name is available in other attributes
+        intf.pop(int_attr,None)                             # ... and _global_merge was propagated from VLAN data
+      if '_vlan_mode' in intf:
+        intf.vlan.mode = intf._vlan_mode                    # Copy _vlan_mode into vlan.mode
+        intf.pop('_vlan_mode',None)                         # ... and remove _vlan_mode
+
     if 'vlans' in n:
       for vdata in n.vlans.values():
-        for int_attr in ['_global_merge','neighbors']:
-          vdata.pop(int_attr,None)
+        for int_attr in ['_global_merge','neighbors']:      # Remove merge flag and neighbors from node VLANs
+          vdata.pop(int_attr,None)                          # ... see discussion #887 for details
 
 class VLAN(_Module):
 
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -49,23 +49,6 @@ def routed_access_vlan(link: Box, topology: Box, vlan: str) -> bool:
     print(f'... Routed VLAN {vlan} on link {link}')
   return routed_vlan
 
-#
-# interface_vlan_mode: Given an interface, a node, and topology, find interface VLAN mode
-#
-# FIX!!! streamlining interface_vlan_mode failed, we must be doing something wrong
-#
-def interface_vlan_mode(intf: Box, node: Box, topology: Box) -> str:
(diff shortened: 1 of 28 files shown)


netsim/modules/vlan.py
```diff
@@ -35,42 +35,40 @@ def populate_vlan_id_set(topology: Box) -> None:
...
-    # Test rewritten based on #882 by jbemmel
-    mode = (
-      intf.get('_vlan_mode',None) or
-      intf.get('vlan.mode',def_link) or
-      topology.nodes[intf.node].get(f'vlans.{vlan}.mode',None) or
-      topology.nodes[intf.node].get('vlan.mode',None) or 
-      def_vlan or def_global or 'irb')
-    if mode != 'route':
+    intf_mode = intf.get('_vlan_mode',None)
+    if intf_mode is None:                           # Not a VLAN-enabled device
```
jbemmel, line 41:
```
if intf_mode is not None:   # VLAN-enabled device?
  if intf_mode != 'route':
    return False
  routed_vlan = True
```

netsim/modules/vlan.py
```diff
@@ -391,6 +389,77 @@ def set_node_vlan_mode(n_vlan: Box, node: Box, g_vlan: Box, topology: Box) -> No
...
+
+* We're assuming the VLAN needs to be created, so we're not tiptoeing around.
+* Node VLAN mode is evaluated using the precedence principles, everything else
+  is merged
+* _global_merge flag is set once we're done so we're not doing the same thing
+  too many times
+"""
+def populate_node_vlan(vlan: str, node: Box, topology: Box) -> Box:
+  if '_global_merge' in node.vlans[vlan]:
+    return node.vlans[vlan]
```
jbemmel, line 403: Does this happen?

netsim/modules/vlan.py
```diff
@@ -391,6 +389,77 @@ def set_node_vlan_mode(n_vlan: Box, node: Box, g_vlan: Box, topology: Box) -> No
...
+
+"""
+compute_interface_vlan_mode: Given an interface, a node, and topology, find interface VLAN mode
+"""
+def compute_interface_vlan_mode(intf: Box, node: Box, topology: Box) -> str:
+  vlan = intf.get('vlan.access',None) or intf.get('vlan.native',None)
+  if not vlan:
+    return 'irb'
+  return (
+    intf.get('_vlan_mode',None) or
```
jbemmel, line 430: Curious if this happens - should only be called once?

jbemmel (approved):
I think the code could benefit from a top-level comment in vlan.py about the high level call graph, i.e. order in which things happen.

For example:
1. module_pre_transform: This is where A,B,C is done
2. node_pre_transform: For each node, do D,E,F
3. link_pre_transform: For each link, ...
4. module_post_link_transform: etc...

especially highlighting where the _vlan_mode and other internals get computed

netsim/modules/vlan.py
ipspace: A non-VLAN device (host) might be attached to an access link. It should not participate in the "routed vlan" computation.

netsim/modules/vlan.py
ipspace: Yes, we call that function several times, first in `pre_transform` hook to build the node VLANs configured on nodes or groups, and later when processing access/trunk links to ensure the VLAN is also defined on the node.

netsim/modules/vlan.py
ipspace: Didn't check, but it could be that (similar to above) this function is called several times. The "do we already have a result" **get** is just a bit of optimization, the results shouldn't change anyway.

ipspace:
> I think the code could benefit from a top-level comment in vlan.py about the high level call graph, i.e. order in which things happen.

Agree.

PR #903, opened 2023-10-08
Add ebgp.multihop support for SR Linux and SR OS

jbemmel:
* Docs says VRFs not supported, but templates do include support?
* No support yet for address family activation like the EVPN example
* Templates assume multihop is an integer between 1 and 255


ipspace:
> Docs says VRFs not supported, but templates do include support?

Because the plugin code doesn't handle VRFs. I left the VRFs in the templates so I won't have to figure it out again once the plugin builds the VRF multihop sessions. Claiming you can run VRF multihop on SR Linux and SR OS is thus a bit of a stretch.

> No support yet for address family activation like the EVPN example

The AF activation should work out-of-the-box. Once you have the EBGP neighbors in the neighbor table, other templates configure their stuff.

> Templates assume multihop is an integer between 1 and 255

Exactly.

PR #920, opened 2023-10-25
Fix typo breaking virtualbox support

jbemmel:


ipspace:
Thanks a million. Will push out another -post release in a few days.

PR #921, opened 2023-10-25
Typo 'feature' instead of 'features'

jbemmel:
Implies missing regression test...


ipspace:
> Implies missing regression test...

Absolutely agree, but it looks like that was a safeguard in case the module data claimed a device supports some feature but the device data says otherwise.

Have to go through that code now that we migrated everything (AFAIK) to the new system and clean it up (probably for release 1.7.0). Thanks for the poke!

PR #928, opened 2023-10-26
Extend BGP session support for SR Linux

jbemmel:
* BGP timers
* BFD
* remove_private_as ('on', 'all' or 'replace')
* passive

Note: `replace` implies `all` for consistency with other devices, could distinguish `remove_private_as: [ replace, all ]`


ipspace:
Thank you! Now I'm guessing I can add SR Linux to the list of "potential external routers" for BGP labs.

Issue #942, opened 2023-11-02
[BUG] Nokia SR-OS stub interfaces fail the initial config rendering


sdktr:
### Describe the bug

Using SR-OS, the initial config template can not be rendered when a 'stub' interface is in the topology.

### To Reproduce

 Use the template from https://blog.ipspace.net/2023/09/ecmp-ospf-unnumbered.html with SR-OS

### Expected behavior

Eighter:
1) the config should render with a stub interface (in Cisco IOS I assume this will produce a loopback?)
2) Render config without the interface at all or
3) some validation error 'a stub interface is not supported by SR-OS'

### Lab topology

Use the template from https://blog.ipspace.net/2023/09/ecmp-ospf-unnumbered.html or this minimalist version:
```yaml
---
provider: clab
defaults:
  device: sros

defaults.devices.sros.clab:
  image: vrnetlab/vr-sros:23.7.R2
  node:
    kind: vr-sros
    type: ixr-ec
    license: ../../sros/license-sros23.3.R3-202310.txt

addressing.p2p.ipv4: True
module: [ ospf ]

nodes: [ r1 ]
links: [ r1 ]
```

### Output

Running `netlab initial -vvv`

```
TASK [debug] *********************************************************************************************************************************************************************************************
task path: /home/sdktr/.local/lib/python3.10/site-packages/netsim/ansible/tasks/deploy-module.yml:40
redirecting (type: filter) ansible.builtin.ipaddr to ansible.netcommon.ipaddr
redirecting (type: filter) ansible.netcommon.ipaddr to ansible.utils.ipaddr
redirecting (type: filter) ansible.builtin.ipv4 to ansible.netcommon.ipv4
redirecting (type: filter) ansible.netcommon.ipv4 to ansible.utils.ipv4
redirecting (type: filter) ansible.builtin.ipv6 to ansible.netcommon.ipv6
redirecting (type: filter) ansible.netcommon.ipv6 to ansible.utils.ipv6
fatal: [r1]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: 'null' is undefined. 'null' is undefined. 'null' is undefined. 'null' is undefined
  
    The error appears to be in '/home/sdktr/.local/lib/python3.10/site-packages/netsim/ansible/tasks/deploy-module.yml': line 40, column 5, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
      - debug:
        ^ here
```

### Version

[...]

PR #943, opened 2023-11-02
Fix quoting of "[null]" value

jbemmel:


sdktr:
this fixes the 1st problem of #942 for me

PR #944, opened 2023-11-02
Fix SROS OSPF Stub interface naming

sdktr:
Reuse the interface naming logic of the inital config in SROS OSPF module. This results in consistent interface naming for stub interfaces
Fix the OSPF config deployment as described in #942 

Is this the intended way to fix this @jbemmel ?

### DIFF
--- a/netsim/ansible/templates/ospf/sros.j2
+++ b/netsim/ansible/templates/ospf/sros.j2
@@ -14,7 +14,11 @@
    area:
    - area-id: "{{ l.ospf.area }}"
      interface:
-     - interface-name: "{{ if_name(l.ifname) }}"
+{%  if l.type|default("") == "stub" %}
+     - interface-name: "{{ if_name("stub-"+l.ifname) }}"
+{%  else %}
+     - interface-name: "{{ if_name(l.ifname) }}" 
+{%  endif %}
 {%  if l.ospf.passive|default(False) %}
        passive: True
 {%  endif %}


jbemmel:
Thanks for chiming in

I'm thinking that we'd actually want to include the 'stub-' convention in the if_name macro, to avoid duplicating logic. You're welcome to prepare a PR along those lines

```
{% macro if_name(i) -%}
{% set _pre = "stub-" if i.type=='stub' else 'i' if i.ifname[0].isdigit() else "" %}
{{- (('oc_%s%s/1_0' if sros_use_openconfig|default(true) else '%s%s')|format(_pre,i.ifname)).replace(':','_') -}}
{%- endmacro %}
```

ipspace:
@sdktr @jbemmel Please let me know when this is ready for merge. Thank you!

sdktr:
Agreed we shouldn't be duplicating code, bleh.

Is it the intended netlab -way to build the interface naming in the initial.j2? It seems there is a place within netlab to assign interface names as well, shouldn't we move the logic there?
https://github.com/ipspace/netlab/blob/7213fb39d6bd663cf6e6459896eacdfeadc3d8e6/netsim/devices/sros.yml#L3

sdktr:
new pr in #947

ipspace:
> Is it the intended netlab -way to build the interface naming in the initial.j2?

Nope. SR OS is the only device (I'm aware of... as of a few days ago ;) that does that.

> It seems there is a place within netlab to assign interface names as well, shouldn't we move the logic there?

Yes: https://github.com/ipspace/netlab/blob/dev/netsim/augment/links.py#L218

And it uses eval (= f-string) so you could make the interface naming scheme as complex as you want.

https://github.com/ipspace/netlab/blob/dev/netsim/utils/strings.py#L48

Alternatively, you could use "device quirks" to fix interface names in the device data model. Junos implementation is using that to figure out what device to configure based on whether it uses VLANs or not:

https://github.com/ipspace/netlab/blob/dev/netsim/devices/junos.py

Jinja2 template is the worst place to do it because then you have to do it every time you configure something.

ipspace:
> new pr in #947

As in 'this one is obsolete'? If so, then please close it.

sdktr:
Based on your comment I think both PRs are not merge worthy yet. We should move it to 'somewhere else' so interface naming is set in stone before hitting ansible. Agreed @jbemmel ?

ipspace:
> Based on your comment I think both PRs are not merge worthy yet. We should move it to 'somewhere else' so interface naming is set in stone before hitting ansible. Agreed @jbemmel ?

It would be best for that "somewhere else" to be either interface name parameter, or device quirks (similar to Junos) if that's too complex to implement. If you go for device quirks, I'd set _sros_ifname interface parameter or something similar -- things that start with _ are never type-checked.

You could also decide to use _sros_ifname only for exceptions. Of course you'd have to use `_sros_ifname|default(ifname)` or something similar in Jinja2 templates in that case.

PR #946, opened 2023-11-06
Add SROS to OSPF module testing suite

sdktr:
Nokia SR-OS was not actually connected to the OSPF topology, it had no links configured.

### DIFF
--- a/tests/module/ospf-feature-test.yml
+++ b/tests/module/ospf-feature-test.yml
@@ -43,6 +43,9 @@ links:
   j_vsrx:
     ospf:
       cost: 40
+  n_sros:
+    ospf:
+      cost: 50
   ospf:
     area: 0.0.0.1
 
@@ -51,6 +54,7 @@ links:
 - c_csr:
 - a_eos:
 - j_vsrx:
+- n_sros:
 
 # Stub link (by role) - should be passive
 # Testing OSPF cost specified on the link
@@ -61,6 +65,7 @@ links:
   c_csr:
   a_eos:
   j_vsrx:
+  n_sros:
 
 # External link - should no be in OSPF process
 - role: external
@@ -68,6 +73,7 @@ links:
   c_csr:
   a_eos:
   j_vsrx:
+  n_sros:
 
 # P2P links - test unnumbered interfaces
 # NX-OS has a different default area, the area number has to be specified on the link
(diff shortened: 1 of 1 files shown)


ipspace:
Of course we can merge this one (as soon as it leaves the "draft" status ;). However, there's a whole range of OSPFv2/OSPFv3 test cases in `tests/integration` if you want to stress-test the OSPF implementation.

sdktr:
My naive -self was hoping for some automated CI/CD that would be now 'more fixed' with this patch. That howver seems not to be the case. While I can confirm that SROS now works for the different OSPF types, I can't test end-to-end because I don't have all the other images setup-to-go.
So what's the expected quality of these tests?

ipspace:
> My naive -self was hoping for some automated CI/CD that would be now 'more fixed' with this patch.

Nope 🤷‍♂️

> That howver seems not to be the case. While I can confirm that SROS now works for the different OSPF types, I can't test end-to-end because I don't have all the other images setup-to-go. So what's the expected quality of these tests?

I should remove this one. There are integration tests a template developer can use to test their code, and we might add an easy-to-get device (= Cumulus VX or Nokia SR Linux) to multi-vendor tests to stress-test things work (like, for example, unnumbered OSPF on Arista EOS required a nerd knob), but I don't think we'll get to the CI/CD phase anytime soon.

For the starters, we wouldn't verify more than "templates don't break". We can do that now, and maybe we should be doing it. Anything more requires device images, which brings us into all sorts of copyright issues I don't want to deal with.

sdktr:
OK; I'll leave it up to you what to do with this PR then.

That CI/CD for aalllll the tests would be awesome though:)

jbemmel:
FRR and SR Linux images could easily be used in some extensive CI/CD tests, similar to https://github.com/jbemmel/napalm-srlinux/blob/main/.github/workflows/tests.yml

See https://github.com/jbemmel/napalm-srlinux/tree/main/test/ci for the Python tests that get run against this trivial one-node topology: https://github.com/jbemmel/napalm-srlinux/blob/main/.clab/ci-topology.yml

ipspace:
Agreed. We could do the same with Cumulus container. However, that would cover 1/6th of netlab functionality, and require some fancy testing framework if we want to check whether the devices do what they're supposed to be doing, not just whether the configuration commands can be parsed by the device.

PR #947, opened 2023-11-06
Move Nokia SROS interface naming logic to initial.j2 if_name macro

sdktr:
Alternative fix for #942 . Moved the naming logic of SROS interfaces to the macro in intial.j2


netsim/ansible/templates/initial/sros.j2
```diff
@@ -21,7 +22,7 @@
 {%- endmacro %}
 
 {% macro declare_interface(i,name=None) %}
```
jbemmel, line 24: Not sure if the `name` parameter is really used anywhere, could probably remove it and use `i.ifname` instead

jbemmel (approved):
Could be improved further, but it's better than what we have right now

PR #948, opened 2023-11-08
Add srlinux ldp support

jbemmel:


ipspace:
Thank you. Will fix the integration test when I decide to add LDP to yet-another platform (Nexus OS?). Integration tests should be completely device-agnostic so it can be used with any device, or we could trip up someone trying to add the feature for another platform.

PR #949, opened 2023-11-08
Add python 3.12

jbemmel:


ipspace:
Thank you! So nice to have someone who's tracking these things!

PR #955, opened 2023-11-09
Bug fix: installation scripts have to work with root user

### LINKED ISSUE #951: [BUG] install broken if user is root?
### Describe the bug

Don't know if this is happening only to me,
I tried to perform a fresh install on a new virtual machine w/ Ubuntu 22.04, but when launching the installation **directly as root**, the install scripts are "broken".

### To Reproduce

In example, looking at https://github.com/ipspace/netlab/blob/2a38f37309e4592072289c6ec019980bca587528/netsim/install/ubuntu.sh#L16 

when the user is already root, the *SUDO* var stays untouched, and this results in:
```
root@clab-playground:~# SUDO='DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a'
root@clab-playground:~# $SUDO apt-get update -y
DEBIAN_FRONTEND=noninteractive: command not found
root@clab-playground:~#
```

### Solution/Workaround

Quick & Dirty workaround is to always use *sudo*, even if you are root:
```
root@clab-playground:~# SUDO='DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a'
root@clab-playground:~# SUDO="sudo $SUDO"
root@clab-playground:~# $SUDO apt-get update -y
Get:1 https://download.docker.com/linux/ubuntu jammy InRelease [48.8 kB]
Get:2 https://apt.releases.hashicorp.com jammy InRelease [12.9 kB]
Get:3 http://security.ubuntu.com/ubuntu jammy-security InRelease [110 kB]
```

ipspace:
Fixes #951

### DIFF
--- a/netsim/install/containerlab.sh
+++ b/netsim/install/containerlab.sh
@@ -16,15 +16,17 @@ during the installation process.
 
 EOM
 
-# Add sudo / root check - ghostinthenet 20220418
-SUDO='DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a'
-if [ "$UID" != "0" ]; then
- if [ -x "$(command -v sudo)" ]; then
-  SUDO="sudo $SUDO"
- else
-  echo 'Script requires root privileges.'
-  exit 0
- fi
+# If we have sudo command, then we use it to set environment variables
+if [ -x "$(command -v sudo)" ]; then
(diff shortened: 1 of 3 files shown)


ipspace:
I thought about installing sudo in ubuntu.sh and then expecting sudo to be there, but someone could decide to run just the installation script for libvirt or containers on top of their existing installation, so it's safer to do the same thing three times, and not install anything extra.

ssasso (approved):
looks good to me!

PR #960, opened 2023-11-14
Fabric-generating plugin

ipspace:


ipspace:
Added reviewers just in case someone wants to look at that code. No worries if you don't have time, there's a reason it's a plugin ;)

PR #971, opened 2023-12-05
Fix release notes referral to validation feature

sdktr:
Release notes for 1.7.0 has broken referral to Validation feature


ipspace:
Thank you!

PR #980, opened 2023-12-22
Proposal: VXLAN configurable VTEP address/interface

ssasso:
@ipspace This is a proposal to make the VXLAN VTEP source configurable, and not "hardcoded" to loopback 0.

It is based on a custom flag that can be set on an additional loopback interface. In that case, the first loopback interface with the flag will be used as vtep source (interface/address)

If this is ok for you, I will try to add the proper doc.

/cc @jbemmel

### DIFF
--- a/netsim/modules/vxlan.py
+++ b/netsim/modules/vxlan.py
@@ -98,6 +98,13 @@ def node_set_vtep(node: Box, topology: Box) -> bool:
   vtep_interface = node.loopback
   loopback_name = devices.get_device_attribute(node,'loopback_interface_name',topology.defaults).format(ifindex=0)
 
+  # Search for additional loopback interfaces with vxlan.vtep' flag, and use the first one
+  for intf in node.interfaces:
+    if intf.get('type', '') == 'loopback' and 'vxlan' in intf and intf.vxlan.get('vtep', False):
+      vtep_interface = intf
+      loopback_name = intf.ifname
+      break
+
   if topology.defaults.vxlan.use_v6_vtep and not 'ipv6' in vtep_interface:
     log.error(
       f'You want to use IPv6 VTEP -- VXLAN module needs an IPv6 address on loopback interface of {node.name}',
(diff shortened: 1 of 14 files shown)


netsim/modules/vxlan.py
```diff
@@ -94,24 +94,36 @@ def assign_vni(toponode: Box, obj_path: str, topology: Box) -> None:
...
+      vtep_interface = intf
+      loopback_name = intf.ifname
+      break
 
   vtep_ip = ""
   vtep_af = 'ipv6' if topology.vxlan.use_v6_vtep else 'ipv4'
-  vtep_ip = node.loopback[vtep_af]
+  vtep_ip = vtep_interface[vtep_af]
   node.vxlan.vtep = str(netaddr.IPNetwork(vtep_ip).ip)              # ... and convert IPv4(v6) prefix into an IPv4(v6) address
+  node.vxlan.vtep_interface = loopback_name
```
ssasso, line 126: A change that can be done is to create the 'vtep_interface' ONLY if it's different than the default one.
Default interface name can be then delegated to the template itself, like it's today (i.e. for Cisco or Arista)

ipspace:
Sure, why not ;) However:

* Whatever new stuff we're adding should be within the VXLAN module. For example, 'use current interface as VTEP' should be (for example) **vxlan.vtep** link/interface attribute, not **use_as_vtep** link/interface attribute.
* I would prefer to have the default interface name computed in Python so we don't have to worry about that in the configuration templates. It's easier to fix the Python code once than 10+ configuration templates ;)

Also, I would search for the VTEP interface first in the `node_set_vtep` and then do all the checks, otherwise you might report an irrelevant error (example: no IPv6 on loopback which is not used as VTEP anyway).

ssasso:
> * Whatever new stuff we're adding should be within the VXLAN module. For example, 'use current interface as VTEP' should be (for example) **vxlan.vtep** link/interface attribute, not **use_as_vtep** link/interface attribute.

Ok, so also the link attribute **vxlan.vtep** should be added on the validation list by the vxlan module itself, right?

With
```
_top: # Modification of global defaults
  attributes:
    link:
      vxlan.vtep: bool
```

is my understanding correct?

> Also, I would search for the VTEP interface first in the `node_set_vtep` and then do all the checks, otherwise you might report an irrelevant error (example: no IPv6 on loopback which is not used as VTEP anyway).

You are absolutely right.

ipspace:
Changed link attributes definition in `vxlan.yml`. Hope I got it right.

The module attributes should be defined in the module **attributes** dictionary. If you define them as nested attributes (**vxlan.vtep**) in the global **attributes** dictionary they can be used on all objects (= links in our case). I also suspect that would cause problems if you'd have other VXLAN link attributes.

ssasso:
@jbemmel up to you if you want to aling also SRLinux+SROS ?

jbemmel:
> @jbemmel up to you if you want to aling also SRLinux+SROS ?

Feel free to update, looks trivial

Issue #1058, opened 2024-03-21
[BUG] netlab status -i X - unknown lab instance


sdargoeuves:
### Describe the bug

When performing the command `netlab status -i 2` I have the following errors:

```
~/c/n/labs/arista-test ❯ /home/seb/code/netlab/netlab/netlab status -i 3    seb@linux-lab 17:26:53
[FATAL]   Unknown lab instance 3.
[HINT]    Use 'netlab status --all' to display the list of lab instances
~/c/n/labs/arista-test ❯ /home/seb/code/netlab/netlab/netlab status -i 2    seb@linux-lab 17:27:03
[FATAL]   Unknown lab instance 2.
[HINT]    Use 'netlab status --all' to display the list of lab instances
~/c/n/labs/arista-test ❯ /home/seb/code/netlab/netlab/netlab status --all   seb@linux-lab 17:27:07
Active lab instance(s)

┏━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━┓
┃ id ┃ directory                              ┃ status  ┃ providers ┃
┡━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━┩
│ 2  │ /home/seb/code/netlab/labs/arista-ospf │ started │ libvirt   │
├────┼────────────────────────────────────────┼─────────┼───────────┤
│ 3  │ /home/seb/code/netlab/labs/arista-test │ started │ libvirt   │
└────┴────────────────────────────────────────┴─────────┴───────────┘
~/c/n/labs/arista-test ❯
```

Do I understand correctly the notion of lab instance, is it indeed the lab id?

### To Reproduce

Have multiple labs active, perform the `netlab status -i X`

### Expected behavior

Display a list of the active devices for this lab.

### Lab topology

```
##Lab3
---
plugin: [ multilab ]
defaults:
  multilab.id: 3
  device: eos
  links.stub_loopback: false
....removed the rest of the config.....

##Lab2
---
plugin: [ multilab ]
defaults:
    multilab.id: 2
    device: eos
    links.stub_loopback: false
....removed the rest of the config.....
```

### Version

* netlab version (as displayed by `netlab version`)
```
[...]

PR #1059, opened 2024-03-21
#1058 add type int for netlab status instance

sdargoeuves:
My first Pull Request via fork, it's not about the fix (it's the simplest fix I could think of!) it's about learning how to do more git things, for hopefully more things to come in the future!

### DIFF
--- a/netsim/cli/status.py
+++ b/netsim/cli/status.py
@@ -26,7 +26,6 @@ def status_parse(args: typing.List[str]) -> argparse.Namespace:
     '-i','--instance',
     dest='instance',
     action='store',
-    type=int,
     help='Display or cleanup specific lab instance(s)')
   parser.add_argument(
     '-l','--log',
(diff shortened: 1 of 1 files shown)


ipspace:
Welcome to the club ;) So nice to see another contributor. Unfortunately, the "instance ID" could be a string (example: "default") or an integer, the integer case arising only because of peculiarities of YAML- and `--set` option parsing.

To fix this one, we have to modify https://github.com/ipspace/netlab/blob/dev/netsim/cli/status.py#L53 to return str or int (`Union[str,int]`), and to try str and int value during the `lab_states` lookup process. Would you like to solve that or shall I implement the fix?

sdargoeuves:
I knew that fix would be too simple! thanks for the comment and the help. I've added the suggested fix, where the instance_id can now be a `str` or a `int`

ipspace (commented):
Reading the documentation for "isdigit()" I wonder whether everything that is recognized as having digits (ex: superscripts) would go work with int. Personally, I would go for something along the lines of "try: int(args.instance) except: args.instance"

Also, we're now expecting that a string parseable as integer will always appear as an integer key in lab_states. That is _probably_ true, but I can easily generate a counterexample (but then people shouldn't use too-smart tricks, right).

Anyway, let's go with this one until someone complains ;)

ipspace:
Ouch, the changes made mypy decidedly grumpy. Never mind, I'll merge the PR, make the necessary adjustments, and amend the commit (yeah, I know I'm probably violating a half-dozen best practices).

sdargoeuves:
That's a good point, I don't have any strong reason why I went with the `isdigit` rather than a `try... except...`, it seemed like a good enough way of doing that check, without thinking of the odd cases, like superscript.
If someone complains I'd be happy to change this to a try/except method.
Thank you for the updates and the merge!

PR #1100, opened 2024-04-18
Fix bgp communities inside VRFs

jbemmel:
The config has a 'community' array at the global level, not per VRF


jbemmel:
We might want to add a regression test for this one, if possible

PR #1103, opened 2024-04-18
Fixes to accommodate IXR-X devices

jbemmel:


ipspace:
I merged your PRs in sequence, but even so, it looks like this one still has a merge conflict. Could you please resolve the conflict, then I'll merge the PR (I don't know enough about SROS to risk fixing it).

Issue #1104, opened 2024-04-19
[BUG] netlab initial -o is not considering specific host modules


ssasso:
When running `netlab initial -o` it seems that all the modules defined on the topology "global" level are considered, even if a device is using only a subset (or different modules).

I.e., given this topology snippet:
```
module: [ vlan, vrf, vxlan, ospf, bgp, evpn, gateway ]

groups:
  routers:
    members: [ r1, r2 ]
    device: vyos
    module: [ bgp ]
```

`netlab initial -o -l r1` reports:

```
TASK [Create device configurations for ['vlan', 'bgp', 'ospf', 'vrf', 'vxlan', 'gateway', 'evpn']] *******************************************************************************************************************************
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=vlan)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=bgp)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=ospf)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=vrf)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=vxlan)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=gateway)
included: /root/GIT_H/netlab/netsim/ansible/tasks/create-config.yml for r1 => (item=evpn)
[cut]
TASK [Compute final module name for vlan] ****************************************************************************************************************************************************************************************
ok: [r1 -> localhost]

TASK [Find vlan configuration template] ******************************************************************************************************************************************************************************************
ok: [r1 -> localhost]

TASK [Throw an error message if we can't find the configuration template] ********************************************************************************************************************************************************
skipping: [r1]

TASK [Create vlan configuration from /root/GIT_H/netlab/netsim/ansible/templates/vlan/vyos.j2] ***********************************************************************************************************************************
ok: [r1 -> localhost]
```

while `netlab up` still "computes" all the modules, but then it skips the unused ones:
```
TASK [Deploy individual configuration modules] ***********************************************************************************************************************************************************************************
included: /root/GIT_H/netlab/netsim/ansible/tasks/deploy-module.yml for r1, r2, sw1, sw2 => (item=vlan)
[...]

PR #1105, opened 2024-04-19
OS10: Improvements/Fixes on EVPN+BGP+IPv6

ssasso:
++ bonus point: do not fail on vyos template if `vrfs` is not defined (see #1104 )


ipspace:
Thank you!

PR #1118, opened 2024-04-27
Doc fix: Aruba CX box-building recipe

### LINKED ISSUE #1117: [BUG] Aruba box building instructions do not enable SSH server
Using Aruba CX 10.13 box built according to https://netlab.tools/labs/arubacx/, the SSH server is not enabled in the management VRF and the box is never started (and could not be configured).

The missing command is **ssh server vrf mgmt**

@ssasso: I don't know enough about Aruba software to have an opinion about this. Is this needed in all ArubaCX versions, or is it a new thing? Should we add it to the initial device configuration?

ipspace:
It seems that Aruba CX 10.13 consistently messes up startup configuration on first reboot. Changed the box-building recipe to reboot the device first, then apply the desired configuraton. Fixes #1117

### DIFF
--- a/docs/labs/arubacx.md
+++ b/docs/labs/arubacx.md
@@ -3,29 +3,30 @@
 Aruba AOS-CX 10 is supported by the **netlab libvirt package** command. To build it:
 
 * Create an empty directory on a Ubuntu machine with *libvirt* and *Vagrant*.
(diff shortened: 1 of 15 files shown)


ipspace:
Aborting the pull request, should have pushed **dev** branch first :(

PR #1124, opened 2024-04-27
Fix Dell OS10 Ansible-related information

ipspace:
* It looks like OS10 requires dellemc.os10 Ansible collection to work. Added that information to the box-building instructions
* The 'dellos10.yml' configuration deployment task failed for me. It looks like the newer versions of the dellemc.os10 Ansible collection expect a filename, not the actual commands, as the parameter to dellemc.os10.os10_config.

### DIFF
--- a/docs/labs/dellos10.md
+++ b/docs/labs/dellos10.md
@@ -34,5 +34,8 @@ During the box-building process you'll have to copy-paste initial device configu
    :literal:
 ```
 
+## Ansible Galaxy for Dell OS10
+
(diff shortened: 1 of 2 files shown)


docs/labs/dellos10.md
```diff
@@ -34,5 +34,8 @@ During the box-building process you'll have to copy-paste initial device configu
    :literal:
 ```
 
+## Ansible Galaxy for Dell OS10
+
+You must install the Ansible modules for Dell OS10 with the  `ansible-galaxy collection install arubanetworks.aoscx` command.
```
ssasso, line 39: I guess this should be `dellemc.os10` and not aruba ;)

Issue #1073, opened 2024-04-09
[BUG] SR Linux: disable scenarios with inter-VRF route leaking without EVPN


ipspace:
The SR Linux caveats clearly state that:

> Inter-VRF route leaking is supported only in combination with BGP EVPN

We have to add a device quirk that will identify this scenario and generate an error.

PR #1138, opened 2024-05-02
Add SR Linux quirk to catch inter-VRF leaking scenarios without EVPN

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1073

### DIFF
--- a/netsim/devices/srlinux.py
+++ b/netsim/devices/srlinux.py
@@ -1,0 +1,25 @@
+#
+# SR Linux quirks
+# # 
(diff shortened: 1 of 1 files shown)


ipspace:
In principle, one could build hub-and-spoke topologies where a VRF does not import its own routes (yeah, don't ask, people were doing crazy things before the IBGP-over-EBGP craze), so both import and export lists could have a single entry, but I guess this is good enough ;)

jbemmel:
I did consider that, this covers the 31- through 34- test cases in the integration test directories so I figured this would do for now

PR #1143, opened 2024-05-03
New module for link aggregation support (mvp)

jbemmel:
Example:
```
links:
- s1:
  s2:
  vlan.trunk: [ irb, route ]
  vlan.native: irb

- s1:
   lag.id: 1
   lag.peer: s2
  h1:

- s2:
   lag.id: 1
   lag.peer: s1
  h1:
```

For discussion - syntax could also work like vlans, declaring the links

### DIFF
--- a/netsim/ansible/tasks/frr/mpls-clab.yml
+++ b/netsim/ansible/tasks/frr/mpls-clab.yml
@@ -6,5 +6,14 @@
   run_once: true
   tags: [ print_action, always ]
 
+- name: "Load bonding kernel module"
+  when: "'lag' in module"
+  shell: |
+    modprobe bonding
+  become: true
+  delegate_to: localhost
+  run_once: true
+  tags: [ print_action, always ]
+
 - include_tasks: deploy-config.yml
   tags: [ print_action, always ]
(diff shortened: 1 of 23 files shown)


ipspace:
I like it the generic idea, and having lag.id on an interface is probably the way to go. However, the whole thing is way more complex than what can be done with your approach... unless we limit ourselves to LAG-with-VLAN-trunks, but then I know we'll pay a heavy price sometime in the future (see also: VLAN module).

For example, here's a simplest possible topology modeling a L3 LAG link between two adjacent nodes:

```
module: [ lag ]
nodes: [ r1, r2 ]
links:
- r1:
  r2:
  lag.id: 1
- r1:
  r2:
  lag.id: 1
```

You approach (fixing the interface names afterwards) would fail miserably, as we have to assign a single subnet to the LAG link. I think we have to do something similar to how we handle VLAN trunks and create extra virtual links for LAGs, turning the physical link into a L2-only link.

jbemmel:
Some changes - highlights:
* Create virtual interfaces of type 'lag', move IP parameters there
* Added frr support (no mc_lag, just basic bonding) -> made if_name configurable
* Working MC_LAG interop example between sros and frr -> made LACP default enabled (link down detection doesn't work in clab)
* Cleanup and basic test cases

TODO:
* Test VLAN interop
* Test gateway interop

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/1335

PR #1144, opened 2024-05-03
Doc fix: nokia.grpc collection works with Ansible release 9.5.1

ipspace:


docs/caveats.md
```diff
@@ -269,10 +269,10 @@ ansible-galaxy collection install git+https://github.com/nokia/ansible-networkin
 python3 -m pip install grpcio protobuf==3.20.1
 ```
 
-* As of September 2023, the Nokia Ansible Galaxy collection crashes with Ansible versions 5.x and later. To downgrade Ansible to 4.10.0, use the following commands (also included as part of **netlab install grpc** script):
+* As of May 2024, the Nokia Ansible Galaxy collection crashes with Ansible versions between 4.10.0 and 9.5.1. We recommend upgrading to Ansible release 9.5.1 (also included as part of **netlab install grpc** script):
```
jbemmel, line 272: Suggest to clarify as 'nokia.grpc Ansible Galaxy collection'

jbemmel (changes requested):

docs/caveats.md
ipspace: Reworded. Thank you!

PR #1149, opened 2024-05-03
Speed up BGP on Aruba CX

### LINKED ISSUE #1148: [BUG] ArubaCX does not resend BGP routes after configuring 'remove-private-as'
After configuring **neighbor remove-private-as**, Aruba CX does not resend BGP routes. If possible, it would be great to do **clear bgp <neighbor>** after configuring neighbors.

* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp.session/08-remove-private-as.yml
* Validation results: https://tests.netlab.tools/arubacx/libvirt/bgp.session/08-remove-private-as.yml-validate.log

ipspace:
Fixes #1148

### DIFF
--- a/netsim/ansible/templates/bgp/arubacx.j2
+++ b/netsim/ansible/templates/bgp/arubacx.j2
@@ -3,6 +3,7 @@
 !
 router bgp {{ bgp.as }}
   bgp log-neighbor-changes
+  timers bgp connect-retry 10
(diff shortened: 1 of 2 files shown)


ipspace:
Wrong title (now fixed) and order of PRs, this one has to be applied after #1150

ssasso (approved):
Super approved!! thank you!

Issue #1169, opened 2024-05-09
[BUG] isis/02-ipv6.yml validation fails due to FRR requiring multi-topology


jbemmel:
SR Linux fails to exchange IPv6 routes with FRR in an ipv6-only scenario isis/02-ipv6.yml

The reason is that FRR sends a multi-topology ipv4/ipv6, with its management address as ipv4 router ID; SR Linux does not support multi-topology
```
tlv ipv4-interface-addresses {
                            ipv4-interface-addresses {
                                address [
                                    192.168.121.103
                                ]
                            }
                        }
                        tlv ipv4-te-router-id {
                            ipv4-te-router-id {
                                router-id [
                                    192.168.121.103
                                ]
                            }
                        }
                        tlv mt-ipv6-reachability {
                            mt-ipv6-reachability {
                                prefixes {
                                    prefix ::/0 mt-id 2 {
                                        up-down false
                                        s-bit false
                                        metric 3087008512
                                    }
                                    prefix 2001:db8:1:3::/64 mt-id 2 {
                                        up-down false
                                        s-bit true
                                        metric 10
                                    }
                                    prefix 4020:10d::/32 mt-id 2 {
                                        up-down false
                                        s-bit false
                                        metric 0
                                    }
                                }
                            }
                        }
                        tlv multi-topology {
                            multi-topology {
                                topologies {
                                    topology 0 {
                                    }
                                    topology 2 {
                                    }
                                }
                            }
                        }
                        tlv nlpid {
                            nlpid {
                                nlpid [
                                    ipv6
                                    ipv4
                                ]
                            }
                        }
                        tlv router-capability {
                            router-capabilities {
                                capability 0 {
                                    router-id 192.168.121.103
                                }
                            }
                        }
```

Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/isis/02-ipv6.yml

This is incorrect - there should not be any ipv4 configuration in this scenario

ipspace:
> SR Linux fails to exchange IPv6 routes with FRR in an ipv6-only scenario isis/02-ipv6.yml

It also fails to exchange IPv6 routes with FRR in dual-stack scenario.

> The reason is that FRR sends a multi-topology ipv4/ipv6, with its management address as ipv4 router ID; SR Linux does not support multi-topology

In which case there's nothing we can do. The IS-IS module assumes "multi-topology" (see https://netlab.tools/module/isis/) and that's how most other devices are configured. I will document the caveat.

[...]

PR #1171, opened 2024-05-10
Add an explicit feature flag for IS-IS multi-topology support

jbemmel:
And don't require it for ipv6-only scenarios

SR Linux removed support for IS-IS multi-topology in 23.10, but can still support ipv6-only deployments

Fixes https://github.com/ipspace/netlab/issues/1169

### DIFF
--- a/docs/module/isis.md
+++ b/docs/module/isis.md
@@ -6,7 +6,7 @@ The module supports the following IS-IS features:
 
 * IPv4 and IPv6
 * IS type (L1 and/or L2)
-* Multi-topology IPv6 (enabled by default as soon as the node has at least one IPv6 address, cannot be disabled)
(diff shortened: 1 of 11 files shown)


ipspace:
No.

You're adding a feature flag and modifying 5 device definitions (missing at least two others) just because you have a problem with a single device. Solve it with a device quirk.

Also, saying "we'll use multi-topology only in dual-stack deployments" doesn't help. What if part of the network is dual-stack and part of it is IPv6-only?

Finally, you fixed FRR template (because you care about that), but not all the others that also configure multi-topology. That's not how things work.

If you care about running IPv6 with IS-IS on SR Linux, please create a plugin that would implement all the IS-IS deployment scenarios you'd like to see, and add configuration templates for whichever devices you think should be supported by that plugin.

PR #1173, opened 2024-05-10
Implement the management VRF for FRR containers (fixes #1170)

### LINKED ISSUE #1170: Management VRF on FRR containers
FRR uses its management IP address in routing protocols, which causes some confusion in IPv6-only networks (and some default route issues). 

We could implement the management VRF in FRR, but that could disable outbound communication with the outside world (example: installation of additional software). Have to check what could be done.

ipspace:


jbemmel (approved):
Do we need to add a section to the running config to model this vrf, or does that work automatically?

ipspace:
> Do we need to add a section to the running config to model this vrf, or does that work automatically?

Done. Can't hurt ;)

PR #1175, opened 2024-05-10
Fix the checking for stub_loopback feature

jbemmel:
Enable it by default for SR Linux devices

Fixes isis/12-passive.yml test case

### DIFF
--- a/netsim/devices/srlinux.yml
+++ b/netsim/devices/srlinux.yml
@@ -32,6 +32,7 @@ features:
       unnumbered: True
     ipv6:
       lla: True
+  stub_loopback: True
   vlan:
     model: router
     svi_interface_name: "irb0.{vlan}"
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -749,10 +749,10 @@ def set_link_loopback_type(link: Box, nodes: Box, defaults: Box) -> None:
   if not lb_name:
     return
 
-  if features.stub_loopback is False:
+  if 'stub_loopback' not in features:
     return
-
-  if not defaults.links.stub_loopback and not features.stub_loopback:
+  
+  if not defaults.links.stub_loopback and 'stub_loopback' not in features:
     return
 
   link.type = 'loopback'


netsim/devices/srlinux.yml
```diff
@@ -32,6 +32,7 @@ features:
       unnumbered: True
     ipv6:
       lla: True
+  stub_loopback: True
```
jbemmel, line 35: This feature flag becomes ```stub_loopback: {}```, the 'True' value is lost in translation

ipspace: I think you made a wrong conclusion. If features.stub_loopback would have been set, referencing it wouldn't have reset it to `{}`.

ipspace:
I think you misunderstood what's going on here. This feature **enables** the user to turn stub links into loopbacks **if they want**. The code you 'fixed' **disables that features** on per-device basis, that's why it checks for **False**.

See https://netlab.tools/links/#loopback-links for details.

netsim/devices/srlinux.yml
jbemmel: What I mean is: If you print out the 'features' dict at the place where it checks for stub_loopback, it shows as '{}' regardless whether I put ```stub_loopback: True``` or ```stub_loopback: False``` in the device yaml. Same is true for 'bfd' for example

There seems to be a difference in the levels - because the 'lla' flag does get retained as 'True' for example

This is confusing to me - in your example you have ```devices.iosv.features.stub_loopback: False```, so somehow that does get retained?

ipspace: That looks like a bug. Will check what's going on, and change the logic to consider device flag before the system default, so you'll be able to set the device flag to **True**.

ipspace:
This was a tricky one -- a **True** feature is converted into a **Box({})** to support things like **ospf: True**.

Fixed in e2ca48d333d7cbf547a6798f06ac76730f67d3d7 -- you can set **stub_loopback: True** in device features if you wish.

PR #1176, opened 2024-05-10
SR Linux: Don't provision MTU on loopbacks or stubs

jbemmel:
Fixes osfp test 04-passive.yml


ipspace:
I can merge that, but I think it would be better if you stop pretending that the stub links are loopbacks? You're just increasing the complexity here.

The 'lets emulate stub links with loopbacks' mattered ages ago until you figured out how to connect a container to a Linux bridge, which completely solved that problem.

jbemmel:
'mtu' would be an example of a property that can be configured on an ethernet link, but not on a loopback interface

I'll look into fixing this in a different way

PR #1180, opened 2024-05-11
VyOS: refactor of bgp.session template

### LINKED ISSUE #1159: [BUG] VyOS does not apply bgp.session parameters to unnumbered EBGP sessions
It looks like VyOS applies bgp.session parameters only to neighbors with IPv4/IPv6 addresses. The unnumbered session is established, but VyOS does not advertise the default route over it.

* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp.session/24-default-unnumbered.yml
* Validation results: https://tests.netlab.tools/vyos/clab/bgp.session/24-default-unnumbered.yml-validate.log

ssasso:
Fix #1159

### DIFF
--- a/netsim/extra/bgp.session/defaults.yml
+++ b/netsim/extra/bgp.session/defaults.yml
@@ -113,6 +113,7 @@ devices:
     allowas_in: True
     as_override: True
     password: True
+    remove_private_as: True
   arubacx.features.bgp:
     allowas_in: True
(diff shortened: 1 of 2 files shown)


ipspace:
The new test results: https://tests.netlab.tools/_html/vyos-clab-bgp.session

Great job ;)

PR #1183, opened 2024-05-11
SR Linux: Remove the hardcoded notion that stubs get implemented as loopbacks

jbemmel:
Fixes isis/12-passive.yml


ipspace:
Merged. Hope this change won't impact other templates.

PR #1191, opened 2024-05-16
Add 'bgp.link_bandwidth' knob to the bgp.session plugin

jbemmel:
Enables weighted ECMP based on link capacity

Inspired by https://bgplabs.net/lb/2-dmz-bw/


ipspace:
Love the idea. However, I think this functionality should be part of the **bgp.policy** plugin:

* It modifies existing routing updates, and we're doing that in **bgp.policy** for weight, LocPref, MED, and communities
* On some platforms you have to set Link Bandwidth with a route map, and the **bgp.policy** plugin already has the framework to deal with platforms that can set attribute X either directly or with a route map.

Also, in theory you could set Link Bandwidth on incoming or outgoing updates, so we might need **bgp.bandwidth.in** and **bgp.bandwidth.out** (with a number instead of a dictionary meaning **in**).

I could write the framework Python code in **bgp.policy** plugin, and then you'd just add the SR OS template.

jbemmel:
Sure, we can put it there. There are indeed additional knobs to send or accept the extended community to/from ebgp peers.

Let me know when you're ready for the template and I'll add it

ipspace:
Alternate implementation in release 1.8.3

PR #1192, opened 2024-05-17
Implement BGP Link Bandwidth community in bgp.policy plugin

ipspace:


ipspace:
@jbemmel You can probably just commit your template(s) and corresponding settings to this branch, and then I'll merge it all.

jbemmel:
Will do as soon as we have power back here in Houston - could take a few more daysJeroenSent from my iPhoneOn May 17, 2024, at 5:52 AM, Ivan Pepelnjak ***@***.***> wrote:﻿
@jbemmel You can probably just commit your template(s) and corresponding settings to this branch, and then I'll merge it all.

ipspace:
> Will do as soon as we have power back here in Houston - could take a few more days

No rush. Hang in there and focus on important things, software will wait ;) All the best!

docs/plugins/bgp.policy.md
```diff
@@ -16,17 +16,21 @@ The **bgp.policy** plugin implements simple BGP routing policies :
 
 The plugin adds the following BGP attributes:
 
+* **bgp.bandwidth** link attribute sets the BGP Link Bandwidth extended community. It can be an integer (in megabytes), in which case the Link Bandwidth community is attached to inbound EBGP updates[^BCP], or a dictionary with **in** and **out** integer values if you want to set the Link Bandwidth community in both directions (or just on the outbound updates).
```
jbemmel, line 19: My use case was actually having the router append the bandwidth value corresponding to the interface it was received on, such that it can do UCMP on the routes

In a sample deployment, a customer would have 100G links and then add a new 400G link. Rather than getting ECMP with at most 100G over the new link, we'd want to send 4x the amount across the new path

ipspace: Got it. That's how IOS does things. OTOH, some other devices can't even change the bandwidth on an interface :(

Anyhow, we could make the attribute take an integer value (where supported) or **auto** for "use interface bandwidth", like **bgp.bandwidth: auto** or **bgp.bandwidth.in: auto**

If that works for you I'll fix the parser and the check for device support.

ipspace:
@jbemmel: Added 'auto' keyword to bgp.bandwidth attribute, plus means of specifying what a device supports. Does this meet your needs/expectations?

jbemmel:
I think so - I'll take a look

PR #1201, opened 2024-05-28
srlinux: Add quirk to warn about extended communities not being filtered

jbemmel:
Test case 05 filters extended communities from EBGP - SR Linux only supports filtering standard or large communities


ipspace:
Of course I can merge this, but wouldn't it be better to print a warning (`category=Warning`) and then document the caveat in test results?

ipspace:
Added caveat to the integration test results. Hope the wording is OK.

https://tests.netlab.tools/_html/srlinux-clab-bgp

Issue #1209, opened 2024-06-04
Warning: don't use jammy cloud 2204 no VRF


benbgg:
Just finished getting an (apparently viable) jammy kvm cloud image going on Proxmox.... lots of learning about cloudinit .... and then while running "netlab test clab" ...
   + ip link add mgmt type vrf table 42
    Error: Unknown device type.
  stderr_lines: <omitted>
  stdout: '# Created by netlab (containerlab Linux host)'
and no there isn't a VRF module waiting to be loaded!!!

sharing so no one else has to go through this....

benbgg:
Image name: for above: jammy-server-cloudimg-amd64-disk-kvm.img

ssasso:
What is `lsmod | grep vrf` returning?
Can you please try to manually load the module with `modprobe vrf`?

ssasso:
Maybe I misunderstood your issue.

Are you saying that images from https://cloud-images.ubuntu.com/jammy/current/ are shipped with a reduced kernel with no vrf support?

jbemmel:
Could it be that we are missing a ```modprobe vrf``` since we changed the management vrf on frr?
Similar to https://github.com/ipspace/netlab/blob/dev/netsim/ansible/tasks/frr/mpls-clab.yml#L3

benbgg:
1) yes, no VRF support in that cloud image. (ie lsmod | grep vrf gives nothing, mod probe comes back "not found")

2) OK, done a bit of reading on Ubuntu forums and it appears that the cloud images ship with kernels optimised for different clouds/ platforms. Canonical appears to have decided that VRFs was superfluous for this use case.

I tried to install linux-5.15.105 generic and its modules but that seems to have made the machine go into a boot loop. Should  have snapshotted :-(

jbemmel:
The use of a vrf is conditional on ```netlab_mgmt_vrf|default(False)``` https://github.com/ipspace/netlab/blob/dev/netsim/ansible/templates/initial/frr.j2#L52

Is ```netlab_mgmt_vrf``` set in your environment?

benbgg:
Couldn't say what was set or unset - I'd just completed install and was running netlab test clab as described in docs.
I managed to blow up the vm by trying to move doing an apt install of the generic kernel (see above).  Cloud images not proving a very happy experience., but I'll have another crack seeding from a generic image if anyone wants me to.
[...]

PR #1210, opened 2024-06-04
FRR: Load VRF kernel module upon initial startup, disable use of vrfs when not available

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1209

### DIFF
--- a/netsim/ansible/tasks/frr/initial-clab.yml
+++ b/netsim/ansible/tasks/frr/initial-clab.yml
@@ -1,0 +1,17 @@
+- name: Attempt to load VRF kernel module 
+  block:
+    - name: "Load VRF kernel module"
+      when: netlab_mgmt_vrf|default(False) or 'vrf' in module
+      shell: |
+        modprobe vrf
+      become: true
+      delegate_to: localhost
+      run_once: true
+      tags: [ print_action, always ]
+  rescue:
+    - name: Unable to load vrf kernel module - disable use of VRFs
+      set_fact: 
+        netlab_mgmt_vrf: False
+
+- include_tasks: deploy-config.yml
+  tags: [ print_action, always ]
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -124,7 +124,7 @@ ip link set {{ l.ifname }} mtu {{ l.mtu }}
 cat >/tmp/config <<CONFIG
 hostname {{ inventory_hostname }}
 !
-{% if clab is defined %}
+{% if clab is defined and netlab_mgmt_vrf|default(False) %}
 vrf mgmt
  exit-vrf
 !


netsim/ansible/tasks/frr/initial-clab.yml
```diff
@@ -0,0 +1,17 @@
...
+    - name: "Load VRF kernel module"
+      when: netlab_mgmt_vrf|default(False) or 'vrf' in module
+      shell: |
+        modprobe vrf
+      become: true
+      delegate_to: localhost
+      run_once: true
+      tags: [ print_action, always ]
+  rescue:
+    - name: Unable to load vrf kernel module - disable use of VRFs
```
jbemmel, line 12: Could report an error if 'vrf' module is being used

ipspace:
Love the idea. However:

* Ideally, we'd fix this in the FRR quirks. 
* **modprobe** can be executed as non-root but will fail if it needs to load the driver. As the **vrf** driver is loaded if installed, this shouldn't be a big dea.
* I would deal with the VRF module separately (in the VRF configuration task list). 

Finally:

* **run_once** is implemented by executing the task on a single node, which means that the **netlab_mgmt_vrf** would be set only for that node.

So, I would propose to add the "modprobe" call to FRR quirks and act on that. It would also be easier to cache the results there (ensuring we only call **modprobe** once). I could add that if you wish (but not before tomorrow or so).

ipspace:
... also, if we just want to test whether the VRF module is loaded, looking for 'vrf' in **lsmod** printout is a better alternative.

jbemmel:
> Finally:
> 
> * **run_once** is implemented by executing the task on a single node, which means that the **netlab_mgmt_vrf** would be set only for that node.
> 

I tested for that, actually. The ```modprobe``` runs once, but the failure handler runs for every frr node :)

image

jbemmel:
> ... also, if we just want to test whether the VRF module is loaded, looking for 'vrf' in **lsmod** printout is a better alternative.

I checked that the vrf kernel module is loaded automatically on demand when one issues the ```ip link ... type vrf``` command. So it not being loaded doesn't mean that it won't work

ipspace:
> > * **run_once** is implemented by executing the task on a single node, which means that the **netlab_mgmt_vrf** would be set only for that node.
> 
> I tested for that, actually. The `modprobe` runs once, but the failure handler runs for every frr node :)

That's REALLY NICE to know (thanks a million), I'm just worried that the behavior might change in another Ansible release. 

Anyway, should we go down this path or implement a quirk? Using a quirk would give us more control earlier in the process, and the ability to generate a warning.

jbemmel:
> Anyway, should we go down this path or implement a quirk? Using a quirk would give us more control earlier in the process, and the ability to generate a warning.

I modeled it after what we're doing for the mpls kernel modules, I think this is similar and consistent. I do agree that the vrf module case is different and should be checked for and aborted earlier in the process, with a decent warning

The 'initial' config can work with or without management vrf

ipspace:
> > Anyway, should we go down this path or implement a quirk? Using a quirk would give us more control earlier in the process, and the ability to generate a warning.
> 
> I modeled it after what we're doing for the mpls kernel modules, I think this is similar and consistent.

To be fair, we probably added the MPLS hack before implementing device quirks ;)

> I do agree that the vrf module case is different and should be checked for and aborted earlier in the process, with a decent warning

We have to be consistent and do one or the other for the VRF and the MPLS modules.

> The 'initial' config can work with or without management vrf

... but it would be nice to know that we turned the management VRF off ;)

Anyway, will merge this one as-is and then work on a more generic solution. Thank you!

PR #1212, opened 2024-06-06
Install libssh-dev dependency

jbemmel:
Install fails due to missing dependency

### DIFF
--- a/netsim/install/ansible.sh
+++ b/netsim/install/ansible.sh
@@ -57,6 +57,10 @@ echo "Install Ansible Python dependencies"
 echo ".. pynacl lxml"
 $SUDO pip3 install $REPLACE $IGNORE $FLAG_PIP pynacl lxml
 echo ".. paramiko netmiko ansible-pylibssh"
+#
+# Install ansible-pylibssh build dependency
+#
+sudo apt install -y libssh-dev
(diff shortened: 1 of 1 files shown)


ipspace:
I don't get it. The `libssh-dev` library is not listed as a dependency of `ansible-pylibssh` and nobody complained (so far) that they cannot install Ansible.

Just to be on the safe side, I created a brand new Ubuntu 22.04 VM, started the full installation process and it completed without a glitch. What exactly is the problem?

jbemmel:
On Ubuntu 2024.4 server, the install (inside a venv) fails because
libssh-dev is not installed. Installing libssh-dev fixed it

ipspace:
> On Ubuntu 2024.4 server, the install (inside a venv) fails because libssh-dev is not installed. Installing libssh-dev fixed it

Ah, so we're facing the first 2404 SNAFUs :( Thank you.

The only remaining question: we're installing all other system library dependencies in **install.sh**, which makes **ansible.sh** platform-independent (it has no **apt** calls, just **pip3** and **sudo**). Wouldn't it be better to add `libssh-dev` to https://github.com/ipspace/netlab/blob/dev/netsim/install/ubuntu.sh#L66?

ipspace:
Replaced by d0bd5814bc2ce33b6beb9cb326482fd989b29e0c

PR #1215, opened 2024-06-10
RIPv2/RIPng configuration module

ipspace:
* Simple configuration module to add RIPv2/RIPng support in global routing table
* Configuration template for Arista EOS, Cisco IOS, Cumulus Linux, and FRR


jbemmel:
```netlab validate``` fails for me using ```frr``` devices

PR #1235, opened 2024-07-18
Implement named prefixes for use in validation and prefix lists

ipspace:
This PR implements 'named prefixes' -- a dictionary of prefixes that can be used in 'links.prefix', 'vlans.prefix', validation code, and (future) prefix lists.

The named prefixes can be static (ipv4 and/or ipv6 subnets, optionally with allocation mechanism) or allocation from one of the addressing pools, in which case the named prefix serves as a mechanism to prevent data duplication

### DIFF
--- a/docs/prefix.md
+++ b/docs/prefix.md
@@ -1,0 +1,80 @@
+(named-prefixes)=
+# Named IP Prefixes
+
+In most lab topologies, you probably don't care about the exact IP addresses and subnets. Defining IPv4 and IPv6 prefixes in [addressing pools](address pools) is good enough.
+
+If you want to tighten control over IP address allocation, use the **prefix** attribute on links or VLANS or the **ipv4**/**ipv6** attributes on node interfaces.
+
+However, there are scenarios in which you have to use the same prefix in multiple places, for example:
+
+* Using a link prefix in the lab validation code
+* Using a link prefix in a [prefix list](module-routing)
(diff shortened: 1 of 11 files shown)


ipspace:
Optional idea (in case I get bored): validate the mutually-exclusive attributes. For example, you cannot use 'ipv4' or 'ipv6' in a prefix that has the 'pool' attribute.

docs/prefix.md
```diff
@@ -0,0 +1,80 @@
...
+
+If you want to tighten control over IP address allocation, use the **prefix** attribute on links or VLANS or the **ipv4**/**ipv6** attributes on node interfaces.
+
+However, there are scenarios in which you have to use the same prefix in multiple places, for example:
+
+* Using a link prefix in the lab validation code
+* Using a link prefix in a [prefix list](module-routing)
+* Using the same prefix on multiple links, for example, to implement a stretched subnet using a technology not supported by _netlab_.
+
+You can use *named IP prefixes* in all three scenarios. The named prefixes are defined in the top-level **prefix** dictionary. The dictionary keys are prefix names, the values are dictionaries defining individual prefixes. The prefix values can include **ipv4**, **ipv6**, **pool** and **[allocation](addr-allocation-sequential)** attributes.
```
ssasso, line 14: I don't know if it's me (maybe yes), but cannot find the `allocation` specific documentation.

ssasso (approved):

PR #1234, opened 2024-07-15
Generic routing module (routing policies)

ipspace:
The generic routing module implements route maps (routing.policy) which can be used in bgp.policy plugin with bgp.policy.in and bgp.policy.out interface attributes.

bgp.policy plugin has been modified to use the shared route-map infrastructure and activate 'routing' module whenever
necessary.

The configuration templates for all devices supported by the bgp.policy plugin have been split into route-map and BGP parts, and the route-map part moved into the 'routing' module.

PR #1238, opened 2024-07-20
Add "generic routing" module and implement set-only routing policies

ipspace:


ipspace:
This is the first part of #1234

PR #1242, opened 2024-07-20
Implement prefix-lists in the generic routing module

ipspace:


ipspace:
@ssasso -- just in case you'd have some time to review this. The next step will be 'match' conditions in routing policies, I'll probably do them in a separate PR

ssasso (approved):
Looks good! as soon as it enters in `dev` I will add the support for vyos and arubacx

netsim/ansible/templates/routing/eos.j2
```diff
@@ -8,3 +9,26 @@
 {{     routemap.common_route_map_entry(p_entry) }}
 {%-   endcall %}
 {% endif %}
+{% if routing._prefix.ipv4|default(False) %}
+!
+! IPv4 prefix lists
+!
+{%   for p_name,p_value in routing._prefix.ipv4.items() %}
```
ssasso, line 16: Isn't this a duplication, if the for loop is already performed by the included file `_prefix_list_ios.j2`?

(same for ipv6?)

ipspace: EOS has a different syntax for IPv6 prefix lists than for IPv4 ones, and once I figured that out, I decided to inline them both for consistency (and who knows what else might pop up later)

ipspace:
> Looks good! as soon as it enters in `dev` I will add the support for vyos and arubacx

Thanks for the review. Merged. However, you might want to review the next PR first and then implement both.

PR #1244, opened 2024-07-22
Implement 'match.prefix' in routing policies

ipspace:


ssasso (approved):
The code looks good... 

but it seems I have to deal with this on Aruba side :-( :

```
dut(config)# ipv6 prefix-list lb_only seq 10 permit 2001:db8:1::/48 ge 64 le 64
IPv4 and IPv6 prefix-lists can't have the same name.
```

I don't know if it could be worth having at the "general level" the same logic you implemented for different route maps also for prefix lists.
Otherwise I will implement it only for Aruba templates

ipspace:
> I don't know if it could be worth having at the "general level" the same logic you implemented for different route maps also for prefix lists. Otherwise I will implement it only for Aruba templates

I guess there might be other crap out there with similar limitations, so let's go for the same logic applied everywhere. Will you fix that in all related places or shall I?

ipspace:
> I guess there might be other crap out there with similar limitations, so let's go for the same logic applied everywhere. Will you fix that in all related places or shall I?

You know what, let me fix that, so it will be a separate commit, and I can immediately run the integration tests.

ipspace:
> I don't know if it could be worth having at the "general level" the same logic you implemented for different route maps also for prefix lists.

Implemented in 686789904a5662310dc6c7da67b909a606345f74. Go ahead ;)

ssasso:
ready for aruba ;)
vyos will require some more time.

PR #1246, opened 2024-07-22
VyOS: support for route map prefix list match

ssasso:


ipspace:
Wow. That was fast ;) Thank you!

PR #1247, opened 2024-07-22
Implement as-path access lists and match.aspath in routing policies

ipspace:


ssasso (commented):
Two notes:

1) you forgot to add the `devices/xxxxxx.yml` files

```
[TYPE]    routing: Device frr (node dut) does not support routing policy 'match' keyword 'aspath' used in in_x1 entry #10
[TYPE]    routing: Device frr (node dut) does not support routing policy 'match' keyword 'aspath' used in out_x2 entry #10
[ATTR]    routing: Device frr (node dut) does not support 'aspath' objects (in_x1)
[ATTR]    routing: Device frr (node dut) does not support 'aspath' objects (out_x2)
```

2) while most probably this works with FRR 9, on version 7.5 (which - if I recall correctly - is also the one used on bgplabs) `bgp as-path` cannot have `seq`.

```
  stderr: |-
    line 11: % Unknown command[56]: bgp as-path access-list in_x1 seq 10 deny ^65100$
    line 12: % Unknown command[56]: bgp as-path access-list in_x1 seq 20 permit .*
    line 14: % Unknown command[56]: bgp as-path access-list out_x2 seq 10 permit 65203$
```

```
# netlab connect dut
Connecting to container clab-routing-dut, starting bash
Use vtysh to connect to FRR daemon

dut(bash)#vtysh

Hello, this is FRRouting (version 7.5_git).
Copyright 1996-2005 Kunihiro Ishiguro, et al.

dut# conf
dut(config)# bgp as-path
  access-list  Specify an access list name
dut(config)# bgp as-path access-list
  WORD  Regular expression access list name
dut(config)# bgp as-path access-list aaa
  deny    Specify packets to reject
  permit  Specify packets to forward
dut(config)# bgp as-path access-list aaa permit
  LINE  A regular-expression (1234567890_^|[, {}() ]$*+.?-\) to match the BGP AS paths
dut(config)# bgp as-path access-list aaa permit
```

ipspace:
> 1. you forgot to add the `devices/xxxxxx.yml` files

Even worse, I managed to pass the integration tests 🤦‍♂️

The test topology used a very specific setup (node AS-path filter and routing policy) that did not trigger "does this work here" check. I fixed that check to make Aruba integration test fail (because AS-path ACL is not implemented yet) but forgot the side effects 🤷‍♂️

> 2. while most probably this works with FRR 9, on version 7.5 (which - if I recall correctly - is also the one used on bgplabs) `bgp as-path` cannot have `seq`.

Thank you, fixed (and yeah, FRR 9 has AS-path ACL sequence numbers, but it's the only device I work with that does)

PR #1249, opened 2024-07-23
Implement set.community routing policy option (EOS, FRR, IOSv)

ipspace:


ipspace:
The code includes community-list stuff (object definition, transformations). However, I figured out I need 'set.community' first to run integration tests ;) so I decided to pivot to that (how's that for MBA blabber 😜) and submit it as a separate PR.

Next on the list: community-lists

ssasso (approved):

PR #1251, opened 2024-07-23
BGP community-list implementation on EOS, FRR, and IOS

ipspace:


ipspace:
I will wait for users to start asking for Large and Extended community lists before implementing them. It should be a relatively easy extension, but let's see what the interest is first.

ssasso (approved):

ipspace:
Thank you!

PR #1271, opened 2024-08-12
ArubaCX: OSPFv3 in VRF, and ebgp multihop

ssasso:
Tested with:

* https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/21-multi-vrf-ospfv3.yml
* https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp.multihop/02-vrf.yml


ipspace:
The `23-multi-vrf-mixed-ipv6.yml` test case is failing, probably it's a missing redistribution.

The latest test results are @ https://tests.netlab.tools/

ssasso:
checking, thanks!

PR #1272, opened 2024-08-12
ArubaCX Fix: redistribute ospfv3 as well if present

ssasso:
tested with https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/23-multi-vrf-mixed-ipv6.yml

and also with https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/15-multi-vrf-mixed.yml for backward compatibility


ipspace:
Now it all works ;) Thank you!

PR #1281, opened 2024-08-15
Controlled import of IGP/connected routes into VRF BGP

ipspace:


ssasso (approved):

ipspace:
Thank you! Now for the default routes ;)

PR #1286, opened 2024-08-17
ArubaCX + VyOS OSPF/v3 default originate + bonus

ssasso:
ArubaCX + VyOS OSPF/v3 default originate

++ bonus: Support for OSPFv3 in VRF for VyOS

**Testing caveats for ospfv2/20-default.yml**

- ArubaCX allows to generate the default only as `E2` - the test is failing for the wrong metric received from `dut_c` --> created device quirk for removing `metric-type` from ospf struct
- VyOS does not have a management VRF; a default route is always present on the main routing table - hence, test is failing because the probe always receive the E1 route announced by `dut_c` (which shall announce the default only if present) --> documented in `caveats.md`

### DIFF
--- a/netsim/devices/arubacx.py
+++ b/netsim/devices/arubacx.py
@@ -16,6 +16,7 @@ class ARUBACX(_Quirks):
   @classmethod
   def device_quirks(self, node: Box, topology: Box) -> None:
     mods = node.get('module',[])
+    # Checks for OSPF Process ID (index based)
     if 'ospf' in mods and 'vrf' in mods:
         ospfidx = 2
         for vrf in node.get('vrfs', {}).keys():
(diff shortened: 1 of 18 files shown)


netsim/devices/arubacx.py
```diff
@@ -27,6 +28,18 @@ def device_quirks(self, node: Box, topology: Box) -> None:
                 return
             node.vrfs[vrf]['ospfidx'] = ospfidx
             ospfidx = ospfidx + 1
+    # Remove OSPF default originate route-type (not supported, yet)
+    if 'ospf' in mods:
```
ipspace, line 32: FWIW, you could have used `if node.get('ospf.default.type',None)` -- the get functionality on a box survives hierarchical stuff that's not there, and does not modify the box if you supply the default value.

Issue #1293, opened 2024-08-26
[BUG] BGP MD5 passwords on SR Linux


ipspace:
The https://github.com/ipspace/netlab/blob/dev/netsim/extra/bgp.session/srlinux.j2 template configures per-peer keychain and applies it to BGP neighbors, but the whole thing doesn't work -- the session with FRRouting is never established.

Integration test: https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp.session/07-password.yml

hellt:
@ipspace I know I need to read the contrib docs, but maybe there is a ready-made command that I can paste to run this specific test and validate things?

jbemmel:
```
source netlab/setup.sh
~/Projects/netlab/tests/integration/bgp.session$ netlab up 07-password.yml -p clab -d srlinux
```

ipspace:
Thank you @jbemmel!

I would usually work in a scratch directory (X is an ignored directory within the netlab repo), so assuming your local copy of the netlab Git repo is in ~/netlab:

* Execute `source setup.sh` in top directory (to get netlab in the path)
* Create directory ~/netlab/X
* Within ~/netlab/X execute `netlab up -p clab -d srlinux ../tests/integration/bgp.session/07-password.yml`
* Execute `netlab validate` to test stuff
* I create symlinks for tests I'm using a lot to make things simpler. For whatever a stupid reason I haven't figured out yet, dots in symlinks didn't work for me, so I'd symlink bgp_session to ../tests/integration/bgp.session
* When you're fixing a template, execute `netlab initial -l nodename` to deploy the change
* Execute `netlab down` when done.

... and now I just have to copy-paste this into a Markdown file ;))

PR #1294, opened 2024-08-26
Update srlinux.j2 - fix MD5 auth

jbemmel:
Adds admin-state: enable

Fixes https://github.com/ipspace/netlab/issues/1293


ipspace:
\*\*\*\* -- now I'm embarrassed :(

jbemmel:
Now you can get back to important stuff, like SRv6 support...

PR #1300, opened 2024-08-30
SR Linux configuration template changes needed to support release 24.7.1

ipspace:


hellt:
thnx, it looks good from the diff perspective
I haven't tried it. Will have to find time over the weekedn

ipspace:
> thnx, it looks good from the diff perspective

Thank you.

> I haven't tried it. Will have to find time over the weekend

All the integration tests (for whatever they're worth ;) passed, so I'm merging it, and moving on to the next interesting challenge ;)

PR #1301, opened 2024-08-31
SR Linux: route import (redistribution) into OSPF and BGP

ipspace:
Adds SR Linux support for unconditional route import into BGP and OSPF. Also, creates chains for BGP import/export policies so the per-neighbor routing policy works correctly (as opposed to dumping the whole RIB to the neighbor)


hellt:
LGTM, the next policy was key

hellt (approved):

PR #1304, opened 2024-09-03
SR Linux: Implement OSPF default route

ipspace:
Another feature not yet supported in SR Linux netlab templates: OSPF default route. Validation tests pass, although with warnings as it's impossible to set OSPF metric or OSPF metric type on SR Linux routes exported into OSPFv2/OSPFv3


hellt:
I will circle back re export/policy features, but it might be true that they are not supported as of now

ipspace:
According to the YANG data model explorer, you can only change the BGP parameters in the action part of a routing policy statement, and changing MED does not change the OSPF cost (I checked ;)

hellt:
ok, confirmed that currently you can't set OSPF attributes in the policy engine when importing routes to OSPF. The warnings are legit

hellt (approved):

PR #1307, opened 2024-09-07
SR Linux: Simplify the IBGP export policy with the new 'next-hop: self' action

ipspace:
SR Linux needs an export to set next-hop to "self" on EBGP routes but not on reflected IBGP routes.
The original template created two per-AF export policies and set the BGP next hop to the IP address
of the loopback interface.

With the new "action.bgp.next-hop: self" option, we can simplify that into a single route map that does
not depend on IP addresses or address families.

Last but not least, we're removing the internal BGP communities used to mark IBGP routes.

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -6,7 +6,7 @@
       mask-length-range: exact
 {% endmacro %}
 
-{% macro bgp_export_policy(vrf,type,nh=None,import=[]) %}
+{% macro bgp_export_policy(vrf,type,nhs=False,import=[]) %}
 {%   if vrf == 'default' and type == 'bgp' %}
 - path: /routing-policy/prefix-set[name={{ vrf }}_bgp_advertise]
   value:
@@ -25,7 +25,7 @@
       action:
         policy-result: next-policy
 {%   endif %}
-{%   if 'ibgp' in type and nh %}
+{%   if 'ibgp' in type and nhs %}
     - name: rr-next-hop-unchanged
       match:
         protocol: bgp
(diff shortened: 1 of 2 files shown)


netsim/ansible/templates/bgp/srlinux.macro.j2
```diff
@@ -25,28 +25,36 @@
...
+              ibgp-mark
     - name: ebgp-next-hop-self
       match:
         protocol: bgp
       action:
         policy-result: next-policy
         bgp:
           next-hop:
-            set: "{{ nh }}"
+            set: self
```
hellt, line 47: @ipspace I am curious, what is the use case here where the nhs is needed on the ebgp session? I believe 4271 rules should be applied by default on eBGP sessions.

Or is it some redistribution use case?

ipspace: It's not on an EBGP session. The policy is used on IBGP sessions, and we cannot use the per-neighbor/AF setting because that would also impact the reflected IBGP routes.

The statement is called "ebgp-next-hop-self" because we have to implement the next hop change on EBGP routes. Any better ideas to rename the statement?

hellt: purely up to you, maybe I'd name it `bgp-next-hop-self`, as the policy is universally matching on the bgp protocol

hellt (approved):

netsim/ansible/templates/bgp/srlinux.macro.j2
jbemmel: I seem to recall having tested this on iBGP EVPN sessions, and I believe I saw that next hop self wasn't being applied even if the policy asked for it. Could be different in recent releases, but we may want to test for that

ipspace: If you're saying "I don't know whether the action.bgp.next-hop.set: self would work", it does (I'm so glad I wasted so much time writing integration tests), or were you thinking about applying next-hop-self to the peer group? That doesn't work either, the next hop is still changed on reflected routes.

PR #1311, opened 2024-09-11
SR Linux: prefix filters and match.prefix in routing policies

ipspace:


ipspace:
I think I hit a bug in SR Linux :( or I can't figure out how it works :(( 

Here's what happens when running the `tests/integration/routing/10-match-prefix.yml` topology:

* The **in_x1** policy is applied as **import-policy** to X1. Please note I have to use the **replace** not **update** action because updating a list does not remove the existing list elements.
* The import policy does not seem to work: it does not filter out the 172.42.42.0/24 prefix advertised by X1. Clearing the BGP session on the other end (and thus resending the BGP prefixes) does not change the results.
* When I delete the import policy from the neighbor, commit the changes, reapply the import policy, and commit the changes, the route is filtered out.

FWIW, after starting the lab, **netlab initial -o -l dut** will generate the API requests sent to SR Linux in the `config` directory.

ipspace:
@hellt: I think I found what the problem was: bogus activation of IPv4 AF on IPv6 neighbor and vice versa. Fixed that, running integration tests.

It will be "fun" to figure out why that matters though :(

ipspace:
9c1814ae4ca626fc95620c41af2b685f52eddaa3 fixed the "weird lack of import policy impact" issue. I think this is good to go.

hellt (approved):

PR #1320, opened 2024-09-19
SR Linux: IS-IS VRF instances

ipspace:


ipspace:
Integration test results in a4851663

hellt (approved):
SR Linux should do ISIS better than it does OSPF, I believe
SP roots calling.

ipspace:
> SR Linux should do ISIS better than it does OSPF, I believe

I don't think anyone will notice a significant difference within a lab environment ;)

PR #1335, opened 2024-10-03
Add 'lag' module for basic link aggregation

jbemmel:
Simplified and more robust version of the earlier attempt, with support for FRR and Cumulus to start

This module focuses on simple bonds of 1 or more P2P links between 2 devices, with or without LACP. 
MC-LAG is left FFS


jbemmel:
I put the 'lag' module interface logic in 'initial', rather than its own lag/device.j2 script. In theory it could be put there if run before 'initial', at least in case of FRR, but this may be confusing as 'initial' wouldn't be the initial script to run

jbemmel:
Put back to draft as the scenario with VLANs should work too

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/1337

Issue #1334, opened 2024-10-02
Check uses 'sshpass' without checking for its existence


jbemmel:
https://github.com/ipspace/netlab/blob/eb60af9f0691ff35ba3cd238717170cf21e2651f/netsim/ansible/tasks/readiness-check/dellos10.yml#L10

The password used is 'vagrant' even when using clab provider (as defined in group vars)

ipspace:
Are you saying "I don't have sshpass installed" or "we're using the wrong password"?

jbemmel:
First problem was that I did not run 'netlab install ubuntu' and hence sshpass wasn't installed; the check was failing silently because of that. Beyond that, I could already see the next problem once I did install sshpass, more like a 'note to self' when peeling the onion

ipspace:
@ssasso It looks like we should:

* Rename dellos10.yml into dellos10-clab.yml and remove the "are we using clab" checks
* Replace the ssh call with "include_tasks: vm-clab-ssh-check.yml" in dellos10.yml
* Add the "do we have sshpass" check into vm-clab-ssh-check.yml

I can do the last bit in a day or two. Can you test the OS10-specific parts (I won't waste my time trying to build an OS10 container)?

@jbemmel: And what might he "next problem" be?

ssasso:
> @ssasso It looks like we should:
> 
> * Rename dellos10.yml into dellos10-clab.yml and remove the "are we using clab" checks
> * Replace the ssh call with "include_tasks: vm-clab-ssh-check.yml" in dellos10.yml
> * Add the "do we have sshpass" check into vm-clab-ssh-check.yml
> 
> I can do the last bit in a day or two. Can you test the OS10-specific parts (I won't waste my time trying to build an OS10 container)?

Yes, I will try to have a new working container image, there are also some "new" caveats on that which I documented on vrnetlab page here: https://github.com/hellt/vrnetlab/tree/master/ftosv#caveats--working-vs-non-working-images

The last sentence needs to be converted in: "it seems, but sometimes it seems not" :(

jbemmel:
> Yes, I will try to have a new working container image, there are also some "new" caveats on that which I documented on vrnetlab page here: https://github.com/hellt/vrnetlab/tree/master/ftosv#caveats--working-vs-non-working-images
> 
> The last sentence needs to be converted in: "it seems, but sometimes it seems not" :(

Yep, saw that. Have been trying to make the prompt scraping logic more robust, but it's finicky

[...]

PR #1336, opened 2024-10-03
OS10: Fix CLAB credentials

ssasso:
Related to #1334


jbemmel (approved):
Minor comment that it might be nice to sort properties alphabetically, but it's not done consistently anywhere else either

PR #1337, opened 2024-10-03
Updated lag module changes

jbemmel:
(reworking to remove manual ```lag.id``` assignment by users)

This PR introduces a new virtual link type 'lag' which represents a collection of one or more physical links to be bonded.
Support is included for FRR and Cumulus, with both plain and VLAN use cases validated

Sample config:
```
links:
- r1:
  r2:
  lag.members: 2
```
If desired, users can also define member links explicitly:
```
lag.members: [ r1-r2, r1-r2 ]
```
or
```
lag.members:
- r1:
    ifindex: 49 # To use a particular pair of ports
  r2:
    ifindex: 49
- r1:
    ifindex: 50
  r2:
    ifindex: 50
```

Unlike previous attempts, in this module the ```lag.id``` parameter is auto-generated internally, like VLAN ids. The naming convention for corresponding bond interfaces uses the existing logic for loopbacks and other virtual interfaces; note that a LAG with id 1 may or may not become "bond1", for example.

Future extensions towards MC-LAG may be considered at some point in the future - for example:
```
links:
- r1:
  r2:
  r3:
  lag.members: [ r1-r2, r1-r3 ]
```

### DIFF
--- a/netsim/ansible/tasks/frr/initial-clab.yml
+++ b/netsim/ansible/tasks/frr/initial-clab.yml
@@ -5,6 +5,7 @@
     modprobe vrf || echo "FAILED"
   become: true
   delegate_to: localhost
+  run_once: true
   tags: [ print_action, always ]
   register: modprobe_result
   ignore_errors: True
@@ -13,6 +14,15 @@
   when: modprobe_result.stdout|default('') is search("FAILED")
   set_fact:
     netlab_mgmt_vrf: False
+
+- name: Load bonding kernel module
+  when: "'lag' in module|default([])"
+  shell: |
+    modprobe bonding
+  become: true
+  delegate_to: localhost
+  run_once: true
+  tags: [ print_action, always ]
 
 - include_tasks: deploy-config.yml
   tags: [ print_action, always ]
--- a/netsim/ansible/tasks/frr/initial-clab.yml
+++ b/netsim/ansible/tasks/frr/initial-clab.yml
@@ -15,14 +15,5 @@
   set_fact:
     netlab_mgmt_vrf: False
 
-- name: Load bonding kernel module
-  when: "'lag' in module|default([])"
-  shell: |
-    modprobe bonding
-  become: true
-  delegate_to: localhost
-  run_once: true
-  tags: [ print_action, always ]
-
 - include_tasks: deploy-config.yml
   tags: [ print_action, always ]
--- a/netsim/ansible/templates/initial/cumulus.j2
+++ b/netsim/ansible/templates/initial/cumulus.j2
@@ -63,7 +63,7 @@ done
 #
 echo "INIT: creating other interface"
 cat >/etc/network/interfaces.d/11-physical.intf <<CONFIG
-{% for l in interfaces if l.type in ['lan','p2p','stub'] %}
+{% for l in interfaces if l.type in ['lan','p2p','stub','lag'] %}
 auto {{ l.ifname }}
 {%   if l.ipv4 is defined %}
 
@@ -99,6 +99,7 @@ iface {{ l.ifname }} inet6 static
 iface {{ l.ifname }}
   mtu {{ l.mtu }}
 {%     endif %}
+
 {% endfor %}
 CONFIG
 #
@@ -109,7 +110,7 @@ done
 #
 # For whatever crazy reason, I had to enable IPv6 in containers
 #
-{% for l in interfaces if l.type in ['lan','p2p','stub'] and (l.ipv6 is defined or 'external' in l.role|default('')) %}
+{% for l in interfaces if l.type in ['lan','p2p','stub','lag'] and (l.ipv6 is defined or 'external' in l.role|default('')) %}
 sysctl -qw net.ipv6.conf.{{ l.ifname }}.disable_ipv6=0
 {% endfor %}
 #
--- a/netsim/ansible/templates/lag/frr.j2
+++ b/netsim/ansible/templates/lag/frr.j2
@@ -1,0 +1,33 @@
+#!/bin/bash
+#
+set -e # Exit immediately when any command fails
+#
+
+#
+# Create bonds for LAGs, if any. Requires kernel bonding module loaded
+#
+{% for l in interfaces if 'lag' in l %}
+{%  if l.type=='lag' %}
+{%   set _lacp = l.lag.lacp|default(lag.lacp) %}
+{%   set lacp_act = 'off' if _lacp=='off' else 'on' %}
+{%   set lacp_rate = ('lacp_rate ' + _lacp) if _lacp!='off' else '' %}
+{%   set _p = "balance-xor" if _lacp=='off' else "802.3ad" %}
+{%   set mode = "mode " + _p + " xmit_hash_policy encap3+4" %}
+ip link add dev {{l.ifname}} type bond {{mode}} lacp_active {{lacp_act}} {{lacp_rate}} || true
+ip link set dev {{l.ifname}} down
+{%  else %}
+ip link set dev {{ l.ifname }} down
+{%  endif %}
+{% endfor %}
+
+{% for l in interfaces if 'lag' in l %}
+{%  if l.type=='lag' %}
+ip link set dev {{l.ifname}} up
+{%  else %}
+ip link set dev {{ l.ifname }} master {% for i in interfaces if i.type=='lag' and i.lag.id==l.lag.id %}{{ i.ifname }}
+{% endfor %}
+ip link set dev {{ l.ifname }} up
+{%  endif %}
+{% endfor %}
+
+exit 0
--- a/netsim/ansible/templates/vlan/cumulus.j2
+++ b/netsim/ansible/templates/vlan/cumulus.j2
@@ -18,11 +18,10 @@ auto bridge
 iface bridge
     bridge-vlan-aware yes
 {% if vlans is defined %}
-{%   for vdata in vlans.values() %}
-    bridge-vids {{ vdata.id }}
-{%   endfor %}
+{%  set vids = vlans.values() | map(attribute='id') | sort | map('string') %}
+    bridge-vids {{ ",".join(vids) }}
 {%  endif %}
-{% for ifdata in interfaces if ifdata.virtual_interface is not defined and ifdata.vlan is defined %}
+{% for ifdata in interfaces if ifdata.vlan is defined and ifdata.type!="svi" %}
     bridge-ports {{ ifdata.ifname }}
 {% endfor %}
 CONFIG
@@ -35,7 +34,7 @@ cat >/etc/network/interfaces.d/51-bridge-interfaces.intf <<CONFIG
 auto {{ ifdata.ifname }}
 {%   endif %}
 iface {{ ifdata.ifname }}
-{%   if ifdata.vlan.trunk_id is defined %}
+{%   if ifdata.vlan.trunk_id is defined and ifdata.type == "svi" %}
     bridge-vids {{ ifdata.vlan.trunk_id|sort|join(",") }}
 {%     if ifdata.vlan.native is defined %}
     bridge-pvid {{ ifdata.vlan.access_id }}
--- a/netsim/ansible/templates/vlan/cumulus.j2
+++ b/netsim/ansible/templates/vlan/cumulus.j2
@@ -34,7 +34,7 @@ cat >/etc/network/interfaces.d/51-bridge-interfaces.intf <<CONFIG
 auto {{ ifdata.ifname }}
 {%   endif %}
 iface {{ ifdata.ifname }}
-{%   if ifdata.vlan.trunk_id is defined and ifdata.type == "svi" %}
+{%   if ifdata.vlan.trunk_id is defined and ifdata.type != "lag" %}
     bridge-vids {{ ifdata.vlan.trunk_id|sort|join(",") }}
 {%     if ifdata.vlan.native is defined %}
     bridge-pvid {{ ifdata.vlan.access_id }}
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -15,7 +15,7 @@ from ..data.types import must_be_string,must_be_list,must_be_dict,must_be_id
 from . import devices,addressing
 
 VIRTUAL_INTERFACE_TYPES: typing.Final[typing.List[str]] = [
-  'loopback', 'tunnel' ]
+  'loopback', 'tunnel', 'lag' ]
 
 def adjust_interface_list(iflist: list, link: Box, nodes: Box) -> list:
   link_intf = []
@@ -298,8 +298,8 @@ def add_node_interface(node: Box, ifdata: Box, defaults: Box) -> Box:
       if sys_mtu and node.mtu == ifdata.mtu:    # .. is it equal to node MTU?
         ifdata.pop('mtu',None)                  # .... remove interface MTU on devices that support system MTU
     else:                                       # Node MTU is defined, interface MTU is not
-      if not sys_mtu:                           # .. does the device support system MTU?
-        ifdata.mtu = node.mtu                   # .... no, copy node MTU to interface MTU
+      if not sys_mtu and ifdata.get('type',None)!='lag':  # .. does the device support system MTU?
+        ifdata.mtu = node.mtu                   # .... no, copy node MTU to interface MTU unless it's a LAG
 
   node.interfaces.append(ifdata)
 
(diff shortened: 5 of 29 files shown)


jbemmel:
Note that the patch removes this bit of code from initial/frr.j2 which I believe was generating invalid ipv6 config:
```
{% if 'ipv6' not in l.dhcp.client|default({}) %}
 ipv6 nd ra-interval 5
 no ipv6 nd suppress-ra
{% endif %}
```

netsim/ansible/tasks/frr/initial-clab.yml
```diff
@@ -14,5 +15,14 @@
   set_fact:
     netlab_mgmt_vrf: False
 
+- name: Load bonding kernel module
```
ipspace, line 18: We have a generic "load kernel module X if module Y is active" infrastructure. Use that.

netsim/ansible/templates/initial/cumulus.j2
```diff
@@ -135,10 +136,6 @@ interface {{ l.ifname }}
 {% elif l.type|default("") == "stub" %}
  description Stub interface
 {% endif %}
-{% if 'ipv6' not in l.dhcp.client|default({}) %}
```
ipspace, line 138: Without 'ipv6 nd ra' the EBGP LLA sessions won't start. It must be enabled.

netsim/ansible/templates/vlan/cumulus.j2
```diff
@@ -18,11 +18,10 @@ auto bridge
...
     bridge-vlan-aware yes
 {% if vlans is defined %}
-{%   for vdata in vlans.values() %}
-    bridge-vids {{ vdata.id }}
-{%   endfor %}
+{%  set vids = vlans.values() | map(attribute='id') | sort | map('string') %}
+    bridge-vids {{ ",".join(vids) }}
 {%  endif %}
-{% for ifdata in interfaces if ifdata.virtual_interface is not defined and ifdata.vlan is defined %}
+{% for ifdata in interfaces if ifdata.vlan is defined and ifdata.type!="svi" %}
```
ipspace, line 24: Did you run the integration tests? Are you sure this is safe?

netsim/ansible/templates/vlan/cumulus.j2
```diff
@@ -35,7 +34,7 @@ cat >/etc/network/interfaces.d/51-bridge-interfaces.intf <<CONFIG
 auto {{ ifdata.ifname }}
 {%   endif %}
 iface {{ ifdata.ifname }}
-{%   if ifdata.vlan.trunk_id is defined %}
+{%   if ifdata.vlan.trunk_id is defined and ifdata.type == "svi" %}
```
ipspace, line 37: This looks fishy. Why would an SVI interface have VLAN trunk ID defined?

netsim/augment/links.py
```diff
@@ -298,8 +298,8 @@ def add_node_interface(node: Box, ifdata: Box, defaults: Box) -> Box:
       if sys_mtu and node.mtu == ifdata.mtu:    # .. is it equal to node MTU?
         ifdata.pop('mtu',None)                  # .... remove interface MTU on devices that support system MTU
     else:                                       # Node MTU is defined, interface MTU is not
-      if not sys_mtu:                           # .. does the device support system MTU?
-        ifdata.mtu = node.mtu                   # .... no, copy node MTU to interface MTU
+      if not sys_mtu and ifdata.get('type',None)!='lag':  # .. does the device support system MTU?
```
ipspace, line 301: And why would we want to ignore system MTU on LAG interfaces?

netsim/augment/links.py
```diff
@@ -387,6 +387,11 @@ def assign_link_prefix(
       addr_pools: Box,
       nodes: Box,
       link_path: str = 'links') -> Box:
+  
+  # Don't assign a prefix to physical links that are part of a lag
+  if 'lag' in link and link.get("type","")!="lag":
```
ipspace, line 392: This should be handled with 'ipv4: False' and 'ipv6: False' in the LAG module. There's no need to tweak generic link augmentation if we have other mechanisms.

netsim/modules/lag.py
```diff
@@ -0,0 +1,75 @@
...
+      return False
+  return len(l1)==len(l2)
+
+class LAG(_Module):
+
+  """
+  link_pre_transform: Create virtual LAG links
+  """
+  def link_pre_transform(self, link: Box, topology: Box) -> None:
+    if log.debug_active('vlan'):
```
ipspace, line 25: Why is LAG debugging tied to VLAN flag?

netsim/modules/lag.py
```diff
@@ -0,0 +1,75 @@
...
+            and same_list(l.interfaces,link.interfaces):
+          vlag = l
+          break
+      
+      if vlag is None:
+        vlag = data.get_box(link)
+        vlag.type = "lag"
+        vlag.linkindex = len(topology.links) + 1
+        vlag._linkname = f"links[{vlag.linkindex}]"
+        vlag.interfaces = [ i for i in link.interfaces ] # Make a deep copy, could use a set?
```
ipspace, line 42: BoxList(link.interfaces) might also work. Or link.interfaces + []

netsim/modules/lag.py
```diff
@@ -0,0 +1,75 @@
...
+        vlag.pop('mtu',None)                             # Remove any MTU attribute
```
ipspace, line 43: I'm pretty sure this will break multi-vendor OSPF

netsim/modules/lag.py
```diff
@@ -0,0 +1,75 @@
...
+  """
+  node_post_transform: Check for correct supported configuration of LAG
+  """
+  def node_post_transform(self, node: Box, topology: Box) -> None:
+    features = devices.get_device_features(node,topology.defaults)
+    for i in node.interfaces:
+      # 1. Check if the interface is part of a LAG
+      if 'lag' in i:
+        if 'lag' not in features:
+          log.error(
```
ipspace, line 61: You could report this earlier in the pre_transform module

netsim/modules/lag.py
```diff
@@ -0,0 +1,75 @@
...
+              f'Node {node.name} does not support LAG configured on {i.ifname}',
+              category=log.IncorrectAttr,
+              module='lag',
+              hint='lag')
+
+        _type = i.get("type")
+        if _type=="lag":
+          continue
+        elif _type!="p2p":
```
ipspace, line 70: You should do this check earlier when creating the virtual LAG link

ipspace (changes requested):
Commented individual snippets I didn't like. The bigger problem is the generic LAG data structure. More about that in another comment.

ipspace:
I hate the data model. It's an ugly hack. Why don't you use a link data structure like:

```
links:
- lag.id: 1
  vlan.trunk: [ a,b ]
  lag.members:
  - r1-r2
  - r1-r2
```

Obviously, the LAG module would have to transform that very early on into additional links and create keys for nodes, but it would be very clear what the LAG members are, the "is this sane" logic would be easy, and we would have a clean way toward MLAG implementation.

The LAG module should add nodes mentioned in the 'lag.members' to the parent link to get the rest of the transformation code to work. That would also allow you to specify 'r1' and 'r2' attributes on the LAG link if needed.

The "only" drawback of this approach is that we would have to make the LAG module a plugin (so it's 'topology_expand' hook is called before the 'links.link_init' routing). Unfortunately we can't implement a module 'topology_expand' hook because the list of modules is not fully evaluated until the groups are processed.

netsim/ansible/templates/initial/cumulus.j2
jbemmel: Could we at least make it conditional on ```if l.ipv6|default(False)```?

netsim/augment/links.py
jbemmel: On Linux the bond mtu is derived from the member ports, it won't accept setting mtu on the bond device. But that condition should be applied there, not here

netsim/modules/lag.py
jbemmel: Need to add a 'lag' debug flag

netsim/modules/lag.py
jbemmel: I wanted a set for easier order independent comparison - but that would require link.interfaces to become a set too

netsim/modules/lag.py
jbemmel: I'll keep the mtu attribute, and skip setting it on bond interfaces in the platform specific scripts

jbemmel:
> I hate the data model. It's an ugly hack. Why don't you use a link data structure like:
> 
> ```
> links:
> - lag.id: 1
>   vlan.trunk: [ a,b ]
>   lag.members:
>   - r1-r2
>   - r1-r2
> ```
> 
> Obviously, the LAG module would have to transform that very early on into additional links and create keys for nodes, but it would be very clear what the LAG members are, the "is this sane" logic would be easy, and we would have a clean way toward MLAG implementation.
> 
> The LAG module should add nodes mentioned in the 'lag.members' to the parent link to get the rest of the transformation code to work. That would also allow you to specify 'r1' and 'r2' attributes on the LAG link if needed.
> 
> The "only" drawback of this approach is that we would have to make the LAG module a plugin (so it's 'topology_expand' hook is called before the 'links.link_init' routine). Unfortunately we can't implement a module 'topology_expand' hook because the list of modules is not fully evaluated until the groups are processed.

I did consider something like this, we discussed in the past and I thought we agreed but we can change it like this, no problem.

I think it could remain a module and apply this transformation in ```link_pre_transform```, mimicking the relevant parts of ```links_init``` for the newly added links

netsim/ansible/templates/initial/cumulus.j2
ipspace: No. That would break the "unnumbered IPv4 EBGP sessions" that really rely on IPv6 LLA behind the scenes.

Why is it bothering you?

netsim/modules/lag.py
ipspace: link.interfaces could have been a set if there wouldn't have been an individual who insisted on having two interfaces of the same box connected to the same link 🤷‍♂️ We have to do quite some code modification to accommodate him.

ipspace:
> I think it could remain a module and apply this transformation in `link_pre_transform`, calling `links_init` for the newly added links:

* You can try that. https://netlab.tools/dev/transform/ might help if it doesn't work.
* No need to create a fake topology and call 'links_init'. There's another lower-level routine that normalizes a single link data structure. It's called by links_init, may be two levels down.

netsim/ansible/templates/initial/cumulus.j2
jbemmel: I'm working on an ipv4-only network, and there is no ipv6 anywhere in the configs. 

There is no BGP either, so if this setting is intended to fix an issue with BGP, perhaps we could do 
```if 'bgp' in modules { ...``` ?

File: netsim/modules/lag.py (hunk 8 above)
Comment (jbemmel) [reply netsim/modules/lag.py:42]: We could easily keep accommodating that individual by including a counter, as follows:
```
link.interfaces: { { "node": "n1" }, { "node": "n1", "_j": 2 } }
```

File: netsim/ansible/templates/initial/cumulus.j2 (hunk 2 above)
Comment (jbemmel) [reply netsim/ansible/templates/initial/cumulus.j2:138]: Added ```if 'bgp' in module``` condition

File: netsim/ansible/templates/vlan/cumulus.j2 (hunk 4 above)
Comment (jbemmel) [reply netsim/ansible/templates/vlan/cumulus.j2]: ```if ifdata.vlan.trunk_id is defined and ifdata.type != "lag"```

File: netsim/ansible/templates/vlan/cumulus.j2 (hunk 3 above)
Comment (jbemmel) [reply netsim/ansible/templates/vlan/cumulus.j2:24]: Struggle to run integration tests, the check is needed to include 'lag' type links but exclude 'svi' (like the virtual_interface check does)

File: netsim/modules/lag.py (hunk 7 above)
Comment (jbemmel) [reply netsim/modules/lag.py]: flag added

File: netsim/modules/lag.py (hunk 11 above)
Comment (jbemmel) [reply netsim/modules/lag.py]: moved up

File: netsim/augment/links.py (hunk 6 above)
Comment (jbemmel) [reply netsim/augment/links.py]: I added ```prefix: False```

File: netsim/ansible/templates/lag/frr.j2 [hunk 12]
```diff
@@ -0,0 +1,33 @@
...
+{%  else %}
+ip link set dev {{ l.ifname }} down
+{%  endif %}
+{% endfor %}
+
+{% for l in interfaces if 'lag' in l %}
+{%  if l.type=='lag' %}
+ip link set dev {{l.ifname}} up
+{%  else %}
+ip link set dev {{ l.ifname }} master {% for i in interfaces if i.type=='lag' and i.lag.id==l.lag.id %}{{ i.ifname }}
```
jbemmel, line 27: Could set ```parent_ifindex``` on each member link

netsim/ansible/templates/initial/cumulus.j2
ipspace: It looks like you're throwing spaghetti at the wall. I learned the hard way that doesn't end well.

If you have a problem with that particular bit of template, please open a separate issue, explain why it bothers you, and we might decide to fix it. Changing some random stuff in an unrelated PR usually results in hard-to-figure-out problems further down the line.

jbemmel: ok, left the comment explaining why it's there but restored the condition to the way it was

netsim/ansible/templates/lag/frr.j2
jbemmel: Also missing a lag.id uniqueness check

jbemmel:
Actually we could even remove the ```lag.id``` attribute, and auto-generate a unique value instead

netsim/ansible/templates/vlan/cumulus.j2
ipspace: There might be other virtual interface types (we already have "tunnel" and "loopback"). Not sure it's a good idea to test if the interface type is not SVI.

netsim/ansible/templates/lag/frr.j2
ipspace: > Could set `parent_ifindex` on each member link

We have "parentindex" for links and "parent_ifindex" for interfaces. However, the "parent_ifindex" is not a valid attribute, so it has to be set pretty late in the transformation process.

ipspace: > Also missing a lag.id uniqueness check

Correct. That should be done in the lag module/plugin.

ipspace:
> Actually we could even remove the `lag.id` attribute, and auto-generate a unique value instead

Correct. Although I would leave the "lag.id" as an attribute like we can set the VLAN ID or VXLAN VNI manually.

jbemmel:
What’s the use case for setting it to any particular value? The interface id is derived per device locally (unlike vlan id which appears on the wire)Sent from my iPhoneOn Oct 5, 2024, at 8:20 AM, Ivan Pepelnjak ***@***.***> wrote:﻿

Actually we could even remove the lag.id attribute, and auto-generate a unique value instead

Correct. Although I would leave the "lag.id" as an attribute like we can set the VLAN ID or VXLAN VNI manually.

ipspace:
> What’s the use case for setting it to any particular value? The interface id is derived per device locally (unlike vlan id which appears on the wire)

Vendor limitations? Probably not applicable if you plan to set the lag.id per device (= not globally unique), but then I don't know what has to match across MLAG members (different vendors might have different ideas), so maybe lag.id cannot be completely local.

ipspace:
> Could consider to base lags on link groups:

Yes, that's another option. Still think we'd have to get the LAG code involved before we expand the link groups, which probably means it has to be a plugin (too late to look at the code right now), or we'd have to modify the link group expansion code.

However, 'lag: true' block the ability to have any LAG-specific attributes, so it might be better to go with 'type: lag'.

Finally, you need some attribute that's propagated from the LAG link to the member links, yet again pointing to the need to do expansion of this construct very early in the transformation process.

jbemmel:
@ipspace feel free to choose, but my suggestion is to use https://github.com/ipspace/netlab/pull/1350 instead of this one

ipspace:
> @ipspace feel free to choose, but my suggestion is to use #1350 instead of this one

Well, once you have what you feel is a final (working) version, please let me know and I'll start looking at it. Right now this one is in draft, and the other one isn't, but there's comment saying "but maybe we should do things differently", so... ;)

jbemmel:
Superseded by link group based implementation

PR #1341, opened 2024-10-04
Fix Cumulus mgmt ip prefix

jbemmel:
ssh to Cumulus nodes is currently broken due to the implied /32 prefix

It might be more elegant to fix this by normalizing ```mgmt.ipv4``` to include the prefix, and then removing it everywhere else.

It could also be considered to condition the use of the mgmt vrf on ```netlab_mgmt_vrf``` like done for FRR

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -163,7 +163,9 @@ class _Provider(Callback):
       file_name = file.replace(out_folder+"/","")
       template_name = self.find_extra_template(node,file_name,topology)
       if template_name:
-        node_data = node + { 'hostvars': topology.nodes, 'hosts': get_host_addresses(topology) }
+        node_data = node + { 'hostvars': topology.nodes, 
+                             'hosts': get_host_addresses(topology),
+                             'mgmtpool': topology.addressing.mgmt }  # Needed for subnet prefix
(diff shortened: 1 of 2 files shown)


netsim/providers/__init__.py
```diff
@@ -163,7 +163,9 @@ def create_extra_files(
       file_name = file.replace(out_folder+"/","")
       template_name = self.find_extra_template(node,file_name,topology)
       if template_name:
-        node_data = node + { 'hostvars': topology.nodes, 'hosts': get_host_addresses(topology) }
+        node_data = node + { 'hostvars': topology.nodes, 
+                             'hosts': get_host_addresses(topology),
+                             'mgmtpool': topology.addressing.mgmt }  # Needed for subnet prefix
```
ipspace, line 168: Is there a reason you're not copying 'topology.addressing' into 'addressing'? That would give us more information in case we ever need it.

ipspace (changes requested):
Nice catch. Thank you! Just make it a bit more generic ;)

netsim/providers/__init__.py
jbemmel: done

ipspace (approved):

PR #1344, opened 2024-10-07
Split set_fhrp_gateway to avoid accessing link.interfaces before it is populated

jbemmel:
Possibly points at wrong calling order for set_fhrp_gateway - called too early in process?

Hit this with following lab snippet:
```
vlans:
  v1:
    gateway:
      id: 1
      protocol: vrrp
```

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -338,14 +338,12 @@ def set_fhrp_gateway(link: Box, pfx_list: Box, nodes: Box, link_path: str) -> No
   if not gwid:                                                        # No usable gateway ID, nothing to do
     return
 
-  fhrp_assigned = False
   for af in log.AF_LIST:
     if not af in pfx_list or isinstance(pfx_list[af],bool):           # No usable IPv4/IPv6 prefix, nothing to do
       continue
 
     try:                                                              # Now try to get N-th IP address on that link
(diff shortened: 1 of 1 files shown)


jbemmel:
```
provider: clab

module: [vlan,gateway]

nodes: [n1,n2]

vlans:
 v1:
  gateway:
   id: 1  # Without this bug is not triggered
links:
- n1:
  n2:
  vlan.trunk: [v1]
```

jbemmel:
Wondering if ```copy_fhrp_gw_to_interfaces``` wouldn't be better placed in modules/gateway.py

ipspace:
I'll start working on this in a day or two. The root cause is the set_fhrp_gateway being called (indirectly) from the VLAN module when the VLAN module is creating virtual links for the VLANs in the VLAN trunk.

It could be that your fix is OK and that the set_fhrp_gateway gets called twice (so the nodes get the gateway IP address anyway), or it could be something nasty that has to be fixed within the VLAN module.

Anyway, thanks a million for a nice catch!

ipspace:
Replaced by #1352. I hoped I'd be able to find a better solution, but it seems like yours was close to the best that could be done considering all the other dependencies (and the order in which things have to be done).

Anyway, it turns out that specifying low gateway ID on a link triggers another bug (#1351) that is very hard to fix in an optimal way. I will do my best there and document the caveats.

PR #1350, opened 2024-10-10
lag based on lag.members syntax

jbemmel:
A variation on the earlier PR:
```
links:
- lag.members: [ r1-r2, r1-r2 ]
```

This forces a check for the ```lag``` module when using this functionality, it is more concise

Replaces https://github.com/ipspace/netlab/pull/1337


netsim/modules/lag.py
```diff
@@ -0,0 +1,99 @@
...
+  """
+  link_pre_transform: Process LAG link groups and add virtual parent links to the topology
+  """
+  def link_pre_transform(self, link: Box, topology: Box) -> None:
+    if log.debug_active('lag'):
+      print(f'LAG link_pre_transform for {link}')
+
+    # Iterate over links with type lag, created for link group(s)
+    if link.get('type',"")=="lag" and not ('lag' in link and '_parent' in link.lag):
+      _match = GROUPNAME.search(link._linkname)
```
jbemmel, line 35: We could also have the link group code add an internal ```group``` attribute to the links it creates, instead of reverse engineering the link names

ipspace: That would make much more sense than reverse-engineering. Just set '_link_group' attribute in copy_group_data somewhere around https://github.com/ipspace/netlab/blob/dev/netsim/augment/links.py#L1057. It would also be nice to check if the parent already has '_link_group' attribute and make hierarchical group names if necessary.

ipspace:
You keep adding stuff, so I put the PR back into draft. Please let me know when you're done.

ipspace:
@jbemmel Is this final and ready for review? Thank you!

jbemmel:
> How would you like to proceed?

Thanks for reviewing, I will rework it to use ```lag.members``` instead (implying link.type='lag')

ipspace:
Wow, this was super-fast. Will review it tomorrow, it's getting late over here.

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+  for l in list(topology.links):
+      if 'lag' in l:
+        if not 'members' in l.lag:
+          log.error(
+            f'must define "lag.members" on link {l._linkname}',
+            category=log.IncorrectAttr,
+            module='lag')
+        l.type = 'lag'
+
+        if 'ifindex' not in l.lag:                   # Use user provided lag.ifindex, if any
```
jbemmel, line 69: Else could check it's a valid int

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+        copy_link_data = data.get_box(l)             # We'll copy all link data into member links
+        copy_link_data.type = "p2p"                  # Mark as p2p
+        copy_link_data.prefix = False                # Don't allow IP assignment for these links
+        for k in ['vlan','lag.members']:             # Remove lag.members and any VLAN parameters 
+          copy_link_data.pop(k,None)
+
+        lag_members = l.lag.members
+        l.lag.pop("members",None)                    # Remove explicit list of members
+        for idx,member in enumerate(lag_members):
+          member = links.adjust_link_object(member,f'lag{l.lag.ifindex}[{idx+1}]',topology.nodes)
```
ipspace, line 81: I would use the parent link name as part of the member link name for easier troubleshooting, then you can also skip the l.lag.ifindex, so the link name would be something.lag[n]

ipspace (changes requested):

ipspace:
Great job, just a ton of minor suggestions ;)

docs/module/lag.md
```diff
@@ -0,0 +1,57 @@
...
+* **mode**: lag mode, one of "802.3ad" (default), "balance-xor" or "active-backup"
+* **lacp**: LACP protocol interval: "fast", "slow" or "off"
+
+  Note that 'link down' is not easily detectable in a virtual environment with veth pairs, therefore it is strongly recommended
+  to enable LACP whenever possible
+
+* **lacp_mode**: "active" (default) or "passive"; note that at most 1 node can be passive
+
+The following parameters can be set per link:
+* **members**: List or dict of links that form the LAG, mandatory and formatted like **topology.links**
```
ipspace, line 26: Let's limit ourselves to 'list'. The code deals only with lists anyway.

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+
+        copy_link_data = data.get_box(l)             # We'll copy all link data into member links
+        copy_link_data.type = "p2p"                  # Mark as p2p
+        copy_link_data.prefix = False                # Don't allow IP assignment for these links
+        for k in ['vlan','lag.members']:             # Remove lag.members and any VLAN parameters 
+          copy_link_data.pop(k,None)
+
+        lag_members = l.lag.members
+        l.lag.pop("members",None)                    # Remove explicit list of members
+        for idx,member in enumerate(lag_members):
```
ipspace, line 80: You don't know that lag.members is a list at this point (we're before links.validate), so you should call **must_be_list** to check that.

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+from ..augment import devices, links
+
+ID_SET = 'lag_id'
+
+"""
+populate_lag_id_set -- Collect any user defined lag.ifindex values globally and initialize ID generator
+"""
+def populate_lag_id_set(topology: Box) -> None:
+  _dataplane.create_id_set(ID_SET)
+  LAG_IDS = { l.lag.ifindex for l in topology.links if 'lag' in l and 'ifindex' in l.lag }
```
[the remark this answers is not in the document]
jbemmel: ```l.lag.ifindex``` can be 0 in this case (I realize that's inconsistent with the way ```ifindex``` is used elsewhere, should we rename it to something else? ```devindex``` perhaps?

netsim/modules/lag.py
jbemmel: added

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+    features = devices.get_device_features(n,topology.defaults)
+    if 'lag' not in features:
+      log.error(
+          f'Node {n.name} does not support LAG configured on link {memberlink._linkname}',
+          category=log.IncorrectAttr,
+          module='lag',
+          hint='lag')
+      ok = False
+    ATT = 'lag.lacp_mode'
+    lacp_mode = i.get(ATT) or memberlink.get(ATT) or n.get(ATT) or topology.defaults.get(ATT)
```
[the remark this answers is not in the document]
jbemmel: moved check to post processing hook

netsim/modules/lag.py
```diff
@@ -0,0 +1,103 @@
...
+def populate_lag_id_set(topology: Box) -> None:
+  _dataplane.create_id_set(ID_SET)
+  LAG_IDS = { l.lag.ifindex for l in topology.links if 'lag' in l and 'ifindex' in l.lag }
+  _dataplane.extend_id_set(ID_SET,LAG_IDS)
+  _dataplane.set_id_counter(ID_SET,topology.defaults.lag.start_lag_id,100)
+
+"""
+check_lag_feature -- Verify that all nodes involved in a LAG link support the feature
+"""
+def check_lag_feature(memberlink: Box,topology: Box) -> bool:
```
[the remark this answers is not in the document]
jbemmel: done

netsim/modules/lag.py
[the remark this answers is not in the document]
ipspace: No, that's fine. We can probably assume that an interface has lag.ifindex if it has any lag-related data.

Issue #1351, opened 2024-10-10
[BUG] Node interface IP/gateway IP overlap on VLAN links


ipspace:
### Describe the bug

The gateway IP can overlap the node IP on the VLAN links when the gateway.id is not set globally (so it cannot be taken out
of the ID pool).

### Lab topology

```
provider: clab
defaults.device: eos

module: [vlan,gateway]

nodes: [n1,n2]

vlans:
 v1:
  gateway:
   id: 1
links:
- n1:
  n2:
  vlan.trunk: [v1]
```

### Solutions

In theory, we could scan through all link and VLAN data to find gateway.id values and remove them from the ID pool.

In practice, we'll just check for an overlap and report an error explaining that the gateway.id has to be set globally if
you want to set it to a low value.

ipspace:
Left to do: create a warning if we ever manage to create node address equal to gateway address.

PR #1355, opened 2024-10-10
Fix: Reserve gateway.id parameters specified on links and global VLANs

ipspace:
Fixes the easier part of #1351

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -80,7 +80,7 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
     import sphinx_rtd_theme
 
     html_theme = "sphinx_rtd_theme"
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -80,7 +80,7 @@
     import sphinx_rtd_theme
 
     html_theme = "sphinx_rtd_theme"
-    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+##    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
```
jbemmel, line 83: Someone once said to me that including unrelated "fixes" in a PR is bad practice...

ipspace: And you're absolutely correct 🤦‍♂️

PR #1356, opened 2024-10-10
Install 'curl' used in script below

jbemmel:


ipspace:
* curl is installed with 'netlab install ubuntu' which is recommended on Ubuntu.
* I don't understand what the problem with lsb-release is. Which version of Ubuntu is giving you a headache?

jbemmel:
> * curl is installed with 'netlab install ubuntu' which is recommended on Ubuntu.
> * I don't understand what the problem with lsb-release is. Which version of Ubuntu is giving you a headache?

Docker container image ubuntu:24.4 - I didn't call the install target for ubuntu yet (though the libvirt script should install its own dependencies imho)

Issue #1343, opened 2024-10-07
DHCP relay for VLANs


jbemmel:
One can enable DHCP relay for a given interface, but not for all interfaces created for a given VLAN

Example:
```
vlans:
  red:
    dhcp.server: [ dhcp-s1, dhcp-s2 ]
```

This should result in DHCP relay being enabled for all interfaces attached to that VLAN. It also implies a check for DHCP module support for every node using that VLAN

ipspace:
Agreed, that should work. It should be applied to the VLAN interface. Which platform(s) were you testing?

jbemmel:
┌──────────────────────────────────────────────────────────────────────────────────┐
│ CREATING configuration files                                                     │
└──────────────────────────────────────────────────────────────────────────────────┘
[ERRORS]  Errors found in topology.yml
[ATTR]    dhcp: Invalid dhcp vlan attribute 'server' found in topology.vlans.V1234.dhcp
[HINT]    use 'netlab show attributes --module dhcp' to display valid attributes

I'm using Cumulus, but the error happens irrespective of that

ipspace:
The problem is that the **dhcp.server** attribute is an interface, not a link attribute. After 30c8a204c74572f0564429d0ee050d29f7e13b7e you'll be able to apply it to node VLAN definition, but not to global VLAN definition.

Have to think about the implications of making DHCP attributes link attributes. It might make sense, but should come with a heavy documentation warning (imagine someone configuring **dhcp.client** and **dhcp.server** on the same VLAN)

ipspace:
@jbemmel: I can't figure out what the best way to deal with this would be. Any suggestions are much appreciated ;)

Imagine a VLAN that has a bunch of DHCP clients and a few DHCP relays. Configuring **dhcp.client** on the VLAN sort-of makes sense (as in "we're using DHCP on this VLAN") but if we also configure **dhcp.server** on the same VLAN, what does that mean? How are we going to figure out who is a relay and who isn't?

I could implement a rule saying "if someone is a DHCP relay, they cannot also be a client" but that would still require that the **dhcp.server** is configured within node (or group) VLAN definition.

PR #1357, opened 2024-10-10
Use dhcp.client link attribute to enable DHCP on all non-relay nodes

ipspace:
Addresses the configuration part of #1343. The DHCP pools for the VLAN segments are not correct (duplicate pools, missing 'excluded' list)


ipspace:
See #1358 for more details on the DHCP pools issue

docs/module/dhcp.md
```diff
@@ -56,17 +56,21 @@ DHCP relays are supported on these platforms:
...
+
+You can enable the DHCP client on all non-relaying devices attached to a link with the boolean **dhcp.client.ipv4** link attribute and the DHCPv6 client with the boolean **dhcp.client.ipv6** attribute.
+
 ## Interface Parameters
 
-* To start a DHCP client on an interface, set **ipv4** or **ipv6** address to **dhcp**.
-* The DHCP client usually installs a default route to an IPv4 default gateway[^DGv6] sent in a DHCP reply. To disable this functionality, set **dhcp.client.default** interface parameter to `False`.
-* To start a DHCP relay on an interface, set **dhcp.server** interface parameter to a list of DHCP servers.
+* To start a DHCP client on an interface, set the **ipv4** or **ipv6** address to **dhcp**. You can use the **dhcp.client.ipv4** or the **dhcp.client.ipv6** interface attribute only when the node already uses the DHCP module. 
+* The DHCP client usually installs a default route to an IPv4 default gateway[^DGv6] sent in a DHCP reply. To turn off this functionality, set **dhcp.client.default** interface parameter to `False`.
```
jbemmel, line 72: Shouldn't this be ```dhcp.client.ipv4.default```?

netsim/modules/dhcp.py
```diff
@@ -35,11 +35,14 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
   # DHCP clients
   #
   for intf in node.interfaces:
-    if not intf.get('dhcp.client',False):
+    if not intf.get('dhcp.client',False):         # Not a DHCP client? OK
       continue
+    if intf.get('dhcp.server',False):             # Also a DHCP relay? That can't be
+      intf.dhcp.pop('client',None)                # ... remove the client flag
```
jbemmel, line 41: Log a warning?

netsim/modules/dhcp.py
```diff
@@ -35,11 +35,14 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
+      continue                                    # ... and move on
     for af in ('ipv4','ipv6'):
       if not af in intf.dhcp.client:
-        continue
+        continue
```
jbemmel, line 45: white space not needed

jbemmel (commented):
Some minor comments, no show stoppers

docs/module/dhcp.md
ipspace: If we set **dhcp.client.ipv4** to **True**, how can we then set **dhcp.client.ipv4.default** to something?

jbemmel: Various options:
1. dhcp.client.ipv4:
      request: True
      default: False
2. dhcp.client.ipv4_default: False

It's just to emphasize that this applies to ipv4 only

ipspace: DHCPv6 does not have a "default gateway" option. IPv6 uses RA to get the default gateway information.

Issue #1359, opened 2024-10-11
[BUG] DHCP 'default router' option is missing for VLAN DHCP pools


ipspace:
The default gateway used by hosts attached to a VLAN is not saved anywhere and thus not accessible to the DHCP module building DHCP pools. The DHCP pools for the VLAN prefixes thus lack the correct 'default router' option.

A workaround is to use the **gateway** module.

### Lab topology

```
provider: libvirt

vlans:
  red:
    dhcp.client.ipv4: True
    links: [ s1-h1, s1-h2 ]

nodes:
  s1:
    module: [ dhcp, vlan ]
    device: eos
    vlans:
      red:
        dhcp.server: dhs
  h1:
    device: linux
  h2:
    device: linux
  dhs:
    device: dnsmasq
    provider: clab

links:
- s1-dhs
```

### Current behavior

The `vlan_red` DHCP pool does not have a default gateway:

```
$ netlab inspect dhcp
pools:
- clean_name: vlan_red
  excluded:
    ipv4:
    - 172.16.0.1
  ipv4: 172.16.0.0/24
  name: vlan_red
```

PR #1360, opened 2024-10-11
Build VLAN-aware DHCP pools (fixes #1358, #1343)

### LINKED ISSUE #1358: [BUG] VLAN DHCP pools are incorrect
### Describe the bug

The DHCP pools built for a DHCP-on-VLAN topology contain duplicate subnets and do not have excluded IP addresses

### Expected behavior

The IP subnets in the **dhcp.pools** list should be unique. Each entry should have the **excluded** IP addresses (at the very minimum, the DHCP relays). However, this is what the following topology generates:

```
pools:
- clean_name: vlans_red_links_1_
  ipv4: 172.16.0.0/24
  name: vlans.red.links[1]
- clean_name: vlans_red_links_2_
  ipv4: 172.16.0.0/24
  name: vlans.red.links[2]
```

### Lab topology

```
provider: libvirt

vlans:
  red:
    dhcp.client.ipv4: True
    links: [ s1-h1, s1-h2 ]

nodes:
  s1:
    module: [ dhcp, vlan ]
    device: eos
    vlans:
      red:
        dhcp.server: dhs
  h1:
    device: linux
  h2:
    device: linux
  dhs:
    device: dnsmasq
    provider: clab

links:
- s1-dhs
```

### Proposed solution

* Build excluded IP lists for all VLANs (iterate over SVI/subinterfaces or use the VLAN neighbors list)
* Iterate over all links, add link data if the link is not a VLAN link, or adjust the VLAN data if the link is a VLAN access link
* Use the combined data structure to generate the DHCP pools

ipspace:
The simplistic 'build DHCP pools from links' algorithm has been heavily modified to cope with VLANs where a single subnet can be stretched across multiple links:

* In the first phase, the algorithm collects VLAN prefixes, using the VLAN name as the pool ID
* In the second phase, it scans all node interfaces, extracts IP addresses from SVI/subinterfaces and adds them to the VLAN 'excluded addresses' lists
* Finally, it scans the links for DHCP clients, creates new pools for classic links, and marks a pool/prefix active.

The collected data is then transformed into the topology.dhcp.pools list compatible with existing configuration templates

There's another outstanding bug (#1359), but that one might be applicable to non-VLAN segments as well.

### DIFF
--- a/netsim/modules/dhcp.py
+++ b/netsim/modules/dhcp.py
@@ -118,38 +118,93 @@ Each DHCP pool could have:
 '''
 
 def build_topology_dhcp_pools(topology: Box) -> None:
-  topology.dhcp.pools = []
+  pools = data.get_empty_box()
+
+  # Phase 1: Build a list of potential pools from VLANs
+  #
+  for vname,vdata in topology.get('vlans',{}).items():
+    if vdata.get('mode') == 'route':                        # Won't deal with routed VLANs
+      continue
+
+    pid = strings.make_id(f'vlan_{vname}')
+    for af in log.AF_LIST:
+      af_pfx = vdata.get(f'prefix.{af}',None)
+      if isinstance(af_pfx,str):
+        pools[pid][af] = af_pfx
+
+      if vdata.get(f'gateway.{af}'):
+        pools[pid].gateway[af] = vdata.gateway[af]
+
+    if pid in pools and 'vrf' in vdata:                     # Copy VLAN VRF if we got some usable prefixes
+      pools[pid].vrf = vdata.vrf
+
+  # Phase 2: Iterate over node interfaces, find SVIs, and build excluded IP list
+  #
+  for node,ndata in topology.get('nodes',{}).items():
+    for intf in ndata.get('interfaces',[]):
+      vname = intf.get('vlan.name')
+      if not vname:
+        continue
+      pid = strings.make_id(f'vlan_{vname}')
+
+      if pid not in pools:
+        continue
+
+      # Iterate over all address families, finding true interface addresses,
+      # converting them from CIDR format to IPv4/IPv6 address format, and
+      # appending them to the pool.excluded.af list
+      #
+      for af in log.AF_LIST:
+        if af in intf and isinstance(intf[af],str):
+          data.append_to_list(pools[pid].excluded,af,str(netaddr.IPNetwork(intf[af]).ip))
 
   for link in topology.get('links',[]):                     # Iterate over lab topology links
     if not link.get('dhcp.subnet'):                         # dhcp.subnet is set if there's at least one DHCP client on the link
       continue
 
-    subnet = data.get_empty_box()                           # Create a DHCP pool data structure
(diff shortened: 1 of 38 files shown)


netsim/modules/dhcp.py
```diff
@@ -118,38 +118,93 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
+    if vdata.get('mode') == 'route':                        # Won't deal with routed VLANs
+      continue
+
+    pid = strings.make_id(f'vlan_{vname}')
+    for af in log.AF_LIST:
+      af_pfx = vdata.get(f'prefix.{af}',None)
+      if isinstance(af_pfx,str):
+        pools[pid][af] = af_pfx
+
+      if vdata.get(f'gateway.{af}'):
```
jbemmel, line 135: Wondering why you're using ``` vdata.get(f'prefix.{af}',None)``` but ```vdata.get(f'gateway.{af}')``` (without None)

netsim/modules/dhcp.py
```diff
@@ -118,38 +118,93 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
+    pools[pid].active = True
+    for af in log.AF_LIST:                                  # Iterate over link address families
       if af not in link.dhcp.subnet or af not in link.prefix:
         continue                                            # No AF prefix or no clients within this AF
 
-      subnet[af] = link.prefix[af]                          # Copy link prefix into the pool
-      subnet.name = link.get('name','') or link.get('_linkname','')
-      subnet.clean_name = strings.make_id(subnet.name)      # Get pool name from link and clean it up
+      if af in pools[pid]:                                  # Pool already has a prefix
+        if pools[pid][af] != link.prefix[af]:               # Check for mismatch between link and VLAN prefix
```
jbemmel, line 181: What happens if ```link.prefix[af]``` is not defined?

netsim/modules/dhcp.py
```diff
@@ -118,38 +118,93 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
+          log.error(
+            f'Mismatch in DHCP pool {pid} prefix {pools[pid][af]}, link {lname} claims the prefix should be {link.prefix[af]}',
+            category=log.IncorrectValue,
+            module='dhcp')
+      else:
+        pools[pid][af] = link.prefix[af]                    # New pool, add prefix
```
jbemmel, line 187: Could throw an error if link.prefix[af] is False

netsim/modules/dhcp.py
ipspace: Good catch. I don't think anything bad happens (because we're using Box not dict), but it's good to be consistent. Fixed.

netsim/modules/dhcp.py
ipspace: Another good one. It would definitely mess up the DHCP configuration template (or confuse the DHCP daemon). I added a test earlier on whether the prefix is a string (because "True" doesn't help either).

ipspace:
> Missing a topology/input test case that uses the dhcp module (let alone dhcp+vlan combination)

Will add.

PR #1369, opened 2024-10-14
Alternate loopback pools (fixes #1331)

### LINKED ISSUE #1331: Custom loopback pools
Suppose I have a topology with 2 groups, one representing nodes in data center 1 and a second for data center 2

I'd like to assign loopback IPs that are specific per data center: e.g. 10.1.x.y for DC1, and 10.2.x.y for DC2. How could I do this?

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/discussions/1328_

Solution: use **loopback.pool** attribute

ipspace:


docs/nodes.md
```diff
@@ -83,6 +83,11 @@ nodes:
 The [Supported Virtualization Providers](platforms.md#supported-virtualization-providers) section of [Supported Platforms](platforms.md) lists the default **memory** and **cpu** values for all devices that can be run as virtual machines.
 ```
 
+(node-loopback)=
+You can use the **loopback** node attribute to change the [default allocation of loopback addresses](addressing-loopback). It's a dictionary that can contain static loopback prefixes (**ipv4** and/or **ipv6** attributes) or an alternate addressing pool (**pool** attribute[^LBIN]).
+
+[^LBIN]: The alternate pool you use for IPv4 loopback addresses should have **loopback** in its name (to tell _netlab_ to set the allocated prefix length to /32) or a [**prefix** attribute](address-pool-specs), preferably set to 32.
```
jbemmel, line 89: Isn't the pool also used for ipv6 addresses?

jbemmel (changes requested):
Should clarify behavior for ipv6, I would expect the alternative pool would work for both?

PR #1375, opened 2024-10-15
Implement node-level 'gateway' parameters

### LINKED ISSUE #1370: [BUG] Node level VRRP gateway priority not propagated to interfaces
### Describe the bug

In the sample topology provided the ```gateway.vrrp.priority``` value configured at the node level is not propagated to the interfaces

### To Reproduce
```netlab create test.yml```

### Expected behavior

Correct VRRP priority value applied to interfaces

### Lab topology

```
module: [gateway,vlan]

vlans:
 v1:
  gateway.protocol: vrrp

nodes:
 r1:
 r2:
  gateway.vrrp.priority: 220

links:
- r1:
  r2:
  vlan.trunk: [v1]
```

### Output
In host_vars/r2/topology.yml
```
gateway:
  vrrp:
    group: 1
    priority: 220
...
- bridge_group: 1
  gateway:
    id: -2
    ipv4: 172.16.0.254/24
    protocol: vrrp
    vrrp:
      group: 1   <--- 'priority' not included here, device templates don't pick it up
  ifindex: 3
  ifname: vlan1000
  ipv4: 172.16.0.2/24
  name: VLAN v1 (1000) -> [r1]
  neighbors:
  - ifname: vlan1000
    ipv4: 172.16.0.1/24
    node: r1
  type: svi
  virtual_interface: true
  vlan:
    mode: irb
```

### Version

netlab version 1.9.1

### PR DESCRIPTION
**Pull request by ipspace**
Fixes #1370

### DIFF
--- a/tests/topology/expected/vlan-vrrp.yml
+++ b/tests/topology/expected/vlan-vrrp.yml
@@ -28,12 +28,14 @@ links:
     node: s1
     vlan:
       trunk:
+        blue: {}
         red: {}
   - ifindex: 1
     ifname: Ethernet1
     node: s2
     vlan:
       trunk:
+        blue: {}
         red: {}
   linkindex: 1
   node_count: 2
@@ -41,6 +43,7 @@ links:
   type: p2p
   vlan:
     trunk:
+      blue: {}
       red: {}
 - bridge: input_2
   gateway:
@@ -227,9 +230,11 @@ nodes:
       type: p2p
       vlan:
         trunk:
+          blue: {}
           red: {}
         trunk_id:
         - 1000
+        - 1001
     - bridge: input_2
       ifindex: 2
       ifname: Ethernet2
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/topology/expected/vlan-vrrp.yml [hunk 1]
```diff
@@ -274,6 +279,35 @@ nodes:
...
+        protocol: vrrp
+        vrrp:
+          group: 1
+      ifindex: 6
+      ifname: Vlan1001
+      ipv4: 172.16.1.2/24
+      name: VLAN blue (1001) -> [s2]
+      neighbors:
+      - gateway:
+          anycast:
```
jbemmel, line 295: Interfaces get cleaned up, but not neighbors

jbemmel (approved):

tests/topology/expected/vlan-vrrp.yml
ipspace: Correct. Have to mention that "neighbors" are not a reliable data source, they are more of a hint of who's around. Changes to interface data after the link transformation are usually not reflected in _neighbors_ list (and that applies to a bunch of modules and plugins)

PR #1396, opened 2024-10-19
Fix iol* images in wake of 3517253

DanPartelly:


ipspace:
Ouch. Yeah, I should have thought about that... 🤦‍♂️

PR #1398, opened 2024-10-19
Netlab exec

DanPartelly:
Synposis:
	netlab exec [-h] [-v] [-q] [--dry-run] [--snapshot [SNAPSHOT]]
	node

The rest of the arguments are passed to SSH or docker exec command.
Let my people go execing!

This is a draft pull request. It misses cosmetics and documentation.
It is also the first line of python I wrote in my life. So more merciless your review is it,  the better for me to learn this language.
I choose to duplicate ~24 lines of code. In the end is 0 maintenance burden.  It could have been refactored, but it works this way.
Error reporting is no better or worse than other part of netlab. It may be improved. 
Please use it. Break it.

### DIFF
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -242,7 +242,7 @@ def run(cli_args: typing.List[str]) -> None:
   host = args.host
 
   if host in topology.nodes:
-    connect_to_node(args,rest,topology,host,log_level)
+    connect_to_node(node=host,args=args,rest=rest,topology=topology,log_level=log_level)
   elif host in topology.tools:
     connect_to_tool(host,rest,topology,log_level)
   else:
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -164,17 +164,17 @@ def create_command_list(host: Box, args: argparse.Namespace, rest: list) -> list
     return rest
 
 def connect_to_node(      
+      node: str, 
       args: argparse.Namespace,
       rest: list,
       topology: Box,
-      node: typing.Optional[str] = None, 
       log_level: LogLevel = LogLevel.INFO) -> typing.Union[bool,int,str]:
   
   host_data = outputs_common.adjust_inventory_host(
                 node=topology.nodes[node],
                 defaults=topology.defaults,
                 group_vars=True)
-  host_data.host = node or args.host
+  host_data.host = node
   connection = host_data.netlab_console_connection or host_data.ansible_connection
 
   rest = create_command_list(host_data,args,rest)
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -167,8 +167,9 @@ def connect_to_node(
       args: argparse.Namespace,
       rest: list,
       topology: Box,
+      node: str,
       log_level: LogLevel = LogLevel.INFO) -> typing.Union[bool,int,str]:
-  node = args.host
+  
   host_data = outputs_common.adjust_inventory_host(
                 node=topology.nodes[node],
                 defaults=topology.defaults,
@@ -241,7 +242,7 @@ def run(cli_args: typing.List[str]) -> None:
   host = args.host
 
   if host in topology.nodes:
-    connect_to_node(args,rest,topology,log_level)
+    connect_to_node(args,rest,topology,host,log_level)
   elif host in topology.tools:
     connect_to_tool(host,rest,topology,log_level)
   else:
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -163,18 +163,18 @@ def create_command_list(host: Box, args: argparse.Namespace, rest: list) -> list
   else:
     return rest
 
-def connect_to_node(
+def connect_to_node(      
       args: argparse.Namespace,
       rest: list,
       topology: Box,
-      node: str,
+      node: typing.Optional[str] = None, 
       log_level: LogLevel = LogLevel.INFO) -> typing.Union[bool,int,str]:
   
   host_data = outputs_common.adjust_inventory_host(
                 node=topology.nodes[node],
                 defaults=topology.defaults,
                 group_vars=True)
-  host_data.host = node
+  host_data.host = node or args.host
   connection = host_data.netlab_console_connection or host_data.ansible_connection
 
   rest = create_command_list(host_data,args,rest)
--- a/netsim/cli/exec.py
+++ b/netsim/cli/exec.py
@@ -24,7 +24,7 @@ from ..augment.groups import group_members
 #
 # CLI parser for 'netlab ' command
 #
-def exec_parse(args: typing.List[str]) -> typing.Tuple[argparse.Namespace, typing.List[str]]:
+def exec_parse(args: typing.List[str]) -> typing.Tuple[argparse.Namespace, typing.List[str]]::
   parser = argparse.ArgumentParser(
       prog="netlab exec",
       description='Run a command on one or more network devices',
@@ -61,16 +61,16 @@ def run(cli_args: typing.List[str]) -> None:
   topology = load_snapshot(args)
   selector = args.node
   args = argparse.Namespace(show=None,verbose=False, quiet=True,Output=True) 
-  if selector in topology.nodes:   
-    connect_to_node(args,rest,topology,selector,log_level)
+  if selector in topology.nodes:
+    connect_to_node(node=selector,args=args,rest=rest,topology=topology,log_level=log_level)
   elif selector in topology.groups:
     node_list = group_members(topology,selector)
     for node in node_list:           
-        connect_to_node(args,rest,topology,node,log_level)   
+        connect_to_node(node=node,args=args,rest=rest,topology=topology,log_level=log_level)   
   else:  
     node_list = _nodeset.parse_nodeset(selector,topology)
     for node in node_list:
-        connect_to_node(args,rest,topology,node,log_level)
+        connect_to_node(node=node,args=args,rest=rest,topology=topology,log_level=log_level)
   
  
 
--- a/netsim/cli/exec.py
+++ b/netsim/cli/exec.py
@@ -24,7 +24,7 @@ from ..augment.groups import group_members
 #
 # CLI parser for 'netlab ' command
 #
-def exec_parse(args: typing.List[str]) -> typing.Tuple[argparse.Namespace, typing.List[str]]::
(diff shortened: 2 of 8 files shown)


netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+from box import Box
+
+from . import external_commands, set_dry_run
+from . import load_snapshot, _nodeset, parser_add_verbose
+from ..outputs import common as outputs_common
+from ..utils import strings, log
+
+from .connect import quote_list, docker_connect, ssh_connect, LogLevel, get_log_level
+
+
```
ipspace, line 22: I would not have more than a single blank line between sections, but that's just my preference.

netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+      default='netlab.snapshot.yml',
+      const='netlab.snapshot.yml',
+      help='Transformed topology snapshot file')
+  parser.add_argument(
+      dest='node', action='store',
+      help='Node(s) to run command on')
+  return parser.parse_known_args(args)
+
+       
+def exec_on_node(
```
ipspace, line 52: It should be possible to use the function from connect.py. What am I missing?

netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+  elif connection in ['paramiko','ssh','network_cli','netconf','httpapi'] or not connection:
+    if connection in ['netconf','httpapi']:
+      print(f"Using SSH to connect to a device configured with {connection} connection")
+    return ssh_connect(host_data,args,rest,log_level)
+  else:
+    log.fatal(f'Unknown connection method {connection} for host {node}',module='connect')
+
+
+def run(cli_args: typing.List[str]) -> None:
+  (args, rest) = connect_parse(cli_args)
```
ipspace, line 77: You should test that there are some extra arguments, otherwise the command will interactively connect to the every box in the nodeset.

netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+  set_dry_run(args)
+
+  if not rest:
+    log.fatal("No command to execute on node(s) was specified. Aborting.")
+
+  rest = quote_list(rest)    
+  topology = load_snapshot(args)
+  selector = args.node
+
+  if selector in topology.nodes:
```
ipspace, line 89: Alternatively, we could make the parse_nodeset deal with groups (which would not be a bad idea anyway) and then you could just call the parse_nodeset and loop over the results

ipspace:
> This is a draft pull request. It misses cosmetics and documentation. 

Also: the command has to be added in numerous places. I searched for "netlab connect"; these would be the obvious places:

```
netsim/cli/alias.txt
docs/cli-overview.md
docs/netlab/cli.md
docs/index.md
README.md
```

> It is also the first line of python I wrote in my life.

Wow. You did a great job ;)

> I choose to duplicate ~24 lines of code

I always prefer having functionality in one place, even if it involves refactoring. In this case, it seems like the only difference is that you need **node** as an argument whereas **connect** gets it from **args.host**. The gotcha is that we don't know who else is using that function (although **mypy** will be very quick to point that out)

The seamless change would add optional node parameter with default value set to None, and set it to args.host if it has no usable value.

netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+    exec_on_node(args,rest,topology,selector,log_level)
+  elif selector in topology.groups:
+    node_list= topology.groups.get(selector, {}).get('members', [])
```
ipspace, line 92: You forgot recursive groups. There must be a function in `groups.py` that returns a node list from a group name.

On a more generic note, we use **python-box** (Box) because it makes dict syntax palatable. Also, we configure it in a way that returns empty dict when referring to undefined values (that's why you have to be careful when referencing stuff; we don't want to have spurious empty dicts in the data model).

In this case, you don't have to use **get** because you already know **selector** is in groups. `topology.groups[selector]` works just fine. Also, we know the group must have members (or the validation would bark), so we can go with `topology.groups[selector].members`

netsim/cli/exec.py
```diff
@@ -0,0 +1,104 @@
...
+    for node in node_list:
```
ipspace, line 93: Finally, even if the group did not have members, the returned value would be an empty dict. Iterating over a dict means iterating over its keys, so this line would work no matter what.

ipspace:
To recap: I would add the **node** parameter to `connect_to_node` and add group functionality to `parse_nodeset`. 

Also: maybe we should make the `parse_nodeset` change a separate PR. It doesn't matter if that happens after you're done with this one.

netsim/cli/exec.py
DanPartelly: I  assumed that all groups are toplevel, after looking at may topologies. Ill Ffx it.

DanPartelly:
> To recap: I would add the **node** parameter to `connect_to_node` 

The key change was removal of  call create_command_list() in exec_on _node vs   I did it to solve name-spacing issues.  Worse is better, have this now. Ill refactor as you want. 


>> and add group functionality to `parse_nodeset`.

We can do that of course.  But I would also prefer to finish this one, then refactor in separate Pr.

netsim/cli/exec.py
DanPartelly: your project, your rules. Ill obey it.

DanPartelly:
Related to this:  from an UI standpoint, this change will allow globing over both hosts and groups at the same time. Sure we wanna allow user to do that ?  

> Also: maybe we should make the `parse_nodeset` change a separate PR. It doesn't matter if that happens after you're done with this one.

netsim/cli/connect.py
```diff
@@ -167,8 +167,9 @@ def connect_to_node(
       args: argparse.Namespace,
       rest: list,
       topology: Box,
+      node: str,
```
ipspace, line 170: This could have been `node: typing.Optional[str] = None`

netsim/cli/connect.py
```diff
@@ -167,8 +167,9 @@ def connect_to_node(
...
       log_level: LogLevel = LogLevel.INFO) -> typing.Union[bool,int,str]:
-  node = args.host
+
```
ipspace, line 172: ... and then this would be `node = node or args.host`

Anyway, separating **node** from the other arguments makes sense, but then we should put it somewhere closer to the top of the list, maybe even as the first argument (which would require fixing all calls to `connect_to_node`)

netsim/cli/connect.py
```diff
@@ -241,7 +242,7 @@ def run(cli_args: typing.List[str]) -> None:
   host = args.host
 
   if host in topology.nodes:
-    connect_to_node(args,rest,topology,log_level)
+    connect_to_node(args,rest,topology,host,log_level)
```
ipspace, line 245: Moving **host** to the first place (see above) would make this lovely similar to the next call ;)

netsim/cli/validate.py
```diff
@@ -498,7 +498,7 @@ def get_parsed_result(v_entry: Box, n_name: str, topology: Box, verbosity: int)
   # Set up arguments for the 'netlab connect' command and execute it
   #
   args = argparse.Namespace(quiet=True,host=n_name,output=True,show=v_cmd,verbose=False)
-  result = connect_to_node(args=args,rest=[],topology=topology,log_level=LogLevel.NONE)
+  result = connect_to_node(args=args,rest=[],topology=topology,node=n_name,log_level=LogLevel.NONE)
```
ipspace, line 501: Just put node to the first place here because it's the most important argument ;)

netsim/cli/validate.py
```diff
@@ -498,7 +498,7 @@ def get_parsed_result(v_entry: Box, n_name: str, topology: Box, verbosity: int)
   # Set up arguments for the 'netlab connect' command and execute it
   #
   args = argparse.Namespace(quiet=True,host=n_name,output=True,show=v_cmd,verbose=False)
```
ipspace, line 500: We don't need **host** argument, do we?

netsim/cli/validate.py
```diff
@@ -547,7 +547,7 @@ def get_result_string(
   # Set up arguments for the 'netlab connect' command and execute it
   #
   args = argparse.Namespace(quiet=True,host=n_name,output=True,show=None,verbose=False)
```
ipspace, line 549: Same as above, we don't need the **host** argument

netsim/cli/validate.py
```diff
@@ -547,7 +547,7 @@ def get_result_string(
...
-  result = connect_to_node(args=args,rest=v_cmd,topology=topology,log_level=LogLevel.NONE)
+  result = connect_to_node(args=args,rest=v_cmd,topology=topology,node=n_name,log_level=LogLevel.NONE)
```
ipspace, line 550: ... and **node** should be the first argument

ipspace:
Now we're down to the nit-picking phase ;)

Ignoring those (but please don't, future readability is important), the code is good to go. All we need now is the documentation.

DanPartelly:
Ive added the initial documentation. Please note that I couldn't install Sphinx today on the development machine, so  I couldn't complete a compilation of documentation.  It misses from repos. If anyone could do that I would be forever grateful.

netsim/cli/connect.py
```diff
@@ -163,17 +163,18 @@ def create_command_list(host: Box, args: argparse.Namespace, rest: list) -> list
...
-def connect_to_node(
+def connect_to_node(      
       args: argparse.Namespace,
       rest: list,
       topology: Box,
+      node: typing.Optional[str] = None, 
       log_level: LogLevel = LogLevel.INFO) -> typing.Union[bool,int,str]:
-  node = args.host
+  
   host_data = outputs_common.adjust_inventory_host(
```
ipspace, line 173: This would now crash if the node would not be specified (i.e. it would be set to None)

ipspace (approved):
I think we're done. I can polish the documentation before publishing the release.

Congratulation, your first Python module is now part of netlab ;)

DanPartelly:
thank you. Next one will be better. Learned a lot those two days from doing it and from your comments. 
> 
> Congratulation, your first Python module is now part of netlab ;)

ipspace:
> I couldn't install Sphinx today on the development machine, so I couldn't complete a compilation of documentation.

The documentation is built automatically as part of the CI/CD pipeline when the master branch is pushed. Having a local copy of Sphinx just helps you check the documentation looks as intended.

PR #1400, opened 2024-10-19
STP module

jbemmel:
* Brings enable/disable STP under Netlab control
* ```stp.protocol```: One of ```stp```(default),```mstp```,```rstp```,```pvrst```, global flavor to run on all supporting nodes
* ```stp.priority``` for setting bridge priority
* ```stp.port_priority``` for selecting between multiple local interfaces

PR modifies FRR vlan template to enable STP before the bridge is enabled, to prevent loops also during startup. A slight violation of modularity, but warranted imho

Cumulus container was found to start sending duplicate BPDUs with both old and new port priority values, requiring a mstpd daemon restart to restore sanity

```
## Platform Support

The following table describes per-platform support of individual STP features:

| Operating system   | STP | MSTP | RSTP | Per-VLAN (R)STP | Enable per port
| ------------------ | :-: | :--: | :--: | :-------------: | :--------------:
| Arista EOS         | ✅  |  ✅  |  ✅  |       ✅        |       ✅      ! Note: STP is enabled by default, using MSTP
| Cumulus Linux      | ✅  |  ❌  |  ✅  |       ❌        |       ✅      ! Note: STP is enabled by default, unless disabled through this module
| FRR                | ✅  |  ❌  |  ❌  |       ✅        |       ❌      ! Note: STP is disabled by default
```

### DIFF
--- a/netsim/modules/stp.py
+++ b/netsim/modules/stp.py
@@ -1,0 +1,35 @@
+#
+# STP transformation module
+#
+from box import Box
+
+from ..augment import devices
+from ..utils import log
+
+from . import _Module
+
+class STP(_Module):
+
+  # Check for device support when pvrst is required
+  def node_pre_transform(self, node: Box, topology: Box) -> None:
+
+    if not topology.stp.get('pvrst',False):
+      return
+
+    if not 'stp' in node.get('module',[]):
+      return
+    
+    features = devices.get_device_features(node,topology.defaults)
+
+    if not 'stp' in features:
+      log.error(
+        f'node {node.name} (device {node.device}) does not support STP module',
+        log.IncorrectValue,
+        'stp')
+      return
+
+    if not features.stp.get('pvrst',False):
+      log.error(
+        f'node {node.name} (device {node.device}) does not support per-VLAN STP (PVRST)',
+        log.IncorrectValue,
+        'stp')
--- a/netsim/modules/stp.py
+++ b/netsim/modules/stp.py
@@ -33,3 +33,31 @@ class STP(_Module):
         f'node {node.name} (device {node.device}) does not support per-VLAN STP (PVRST)',
         log.IncorrectValue,
         'stp')
+
+  # Check max port_priority values
+  def node_post_transform(self, node: Box, topology: Box) -> None:
+    if not 'stp' in node.get('module',[]):
+      return
+    features = devices.get_device_features(node,topology.defaults)
+
+    priority = node.get('stp.priority',0)
+    if priority and (priority % 4096):
+        log.error(
+            f'node {node.name} (device {node.device}) stp.priority: {priority} must be a multiple of 4096',
+            log.IncorrectValue,
+            'stp')
+
+    port_priority = features.get('stp.port_priority', { 'max': 255 } )
+    for intf in node.get('interfaces',[]):
+      if 'stp' in intf:
+        val = intf.get('stp.port_priority',0)
+        if val > port_priority.max:
+          log.error(
+            f'node {node.name} (device {node.device}) only supports stp.port_priority up to {port_priority.max}, found {val}',
+            log.IncorrectValue,
+            'stp')
+        elif 'multiple' in port_priority and (val % port_priority.multiple):
+          log.error(
+            f'node {node.name} (device {node.device}) stp.port_priority {val} must be a multiple of {port_priority.multiple}',
+            log.IncorrectValue,
+            'stp')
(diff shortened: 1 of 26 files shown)


netsim/modules/stp.py
```diff
@@ -0,0 +1,71 @@
...
+from ..utils import log
+
+from . import _Module
+
+class STP(_Module):
+
+  # Check for device support when pvrst is required
+  def node_pre_transform(self, node: Box, topology: Box) -> None:
+
+    if not topology.stp.get('pvrst',False):
```
ipspace, line 16: This will create topology.stp if it doesn't exist. Are you counting on global stp.enable being there?

netsim/modules/stp.py
```diff
@@ -0,0 +1,71 @@
...
+      return
+
+    if not 'stp' in node.get('module',[]):
```
ipspace, line 19: I don't think the hook gets called if the node does not use the module

netsim/modules/stp.py
```diff
@@ -0,0 +1,71 @@
...
+      return
+    
+    features = devices.get_device_features(node,topology.defaults)
+
+    if not 'stp' in features:
```
ipspace, line 24: This should be detected in the generic node/module compatibility check

netsim/modules/stp.py
```diff
@@ -0,0 +1,71 @@
...
+      log.error(
+        f'node {node.name} (device {node.device}) does not support STP module',
+        log.IncorrectValue,
+        'stp')
+      return
+
+    if not features.stp.get('pvrst',False):
```
ipspace, line 31: I would combine this with the first test into a single "can we do it" test. We will need other tests, for example checking **features.stp.mst**.

ipspace:
Looks pretty good. The "how do we specify which protocol to use" it the show-stopper, everything else are probably just nits.

jbemmel:
> Looks pretty good. The "how do we specify which protocol to use" it the show-stopper, everything else are probably just nits.

Would you prefer a ```stp.protocol``` attribute with the various flavors ```[ "STP", "PVST", "RSTP", "RPVST", "MSTP" ]``` defaulting to "RPVST"?

ipspace:
> Would you prefer a `stp.protocol` attribute

Yes

> with the various flavors `[ "STP", "PVST", "RSTP", "RPVST", "MSTP" ]`

Lowercase? We could then use **stp.protocol** to check device features.

> defaulting to "RPVST"?

I try to stay away from STP. Is this a good default?

ipspace:
I never claimed to be an STP expert (I tried to stay as far away from it as possible), but MSTP (according to the little I know, and as I understand https://en.wikipedia.org/wiki/Multiple_Spanning_Tree_Protocol) is not per-VLAN STP.

Should that be the case, FRR (implementing VLANs with Linux bridges) cannot run MSTP because each VLAN bridge runs a separate instance of STP and BPDUs are sent tagged with VLAN tags (this https://serverfault.com/questions/824621/linux-bridges-vlans-and-rstp explains it pretty well). It probably does PVST on VLAN interfaces and traditional STP on non-VLAN interfaces.

Also, according to Cumulus Linux 4.4 documentation, it cannot run MSTP but (because we used VLAN-aware bridge) only RSTP: https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-44/Layer-2/Spanning-Tree-and-Rapid-Spanning-Tree/. Also, it effectively tunnels PVST BPDUs.

I would suggest you take a device we can trust to have a decent STP implementation (Arista cEOS, maybe SR Linux) and test interoperability with FRR and CL in VLAN environment. Also, we will not get total interoperability between CL and FRR because our configuration templates used different types of Linux bridges.

jbemmel:
I may have been looking at more recent documentation. Removed MSTP from supported protocols, and fixed propagation of ```protocol``` attribute

SRL does not support STP at all, and I don't have easy access to any of the others

jbemmel:
* Added EOS support (all flavors)
* Tested interop with FRR on per-VLAN (R)STP - that seems to work fine
* Updated docs to reflect current features per platform

ipspace (approved):

netsim/modules/stp.yml
```diff
@@ -0,0 +1,30 @@
...
+transform_after: [ vlan ]
+config_after: [ vlan ]
+no_propagate: [ pvrst ]
+
+enable: True        # By default, enable STP everywhere (on all bridges created by Netlab) when this module is used
+
+attributes:
+  global:
+    enable: bool
+    pvrst: bool      # Topology requires Per-VLAN Rapid Spanning Tree (802.1w)
```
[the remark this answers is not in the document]
jbemmel: Implemented

Issue #1402, opened 2024-10-19
[BUG] ':' character in clab binds causes processing errors


aceplean:
### Describe the bug

cEOS has a habit of creating interface Ma0 when a Ma1 should exist; this can have some subtle issues with Cloudvision. Arista recommends mapping the interface names.
In containerlab this is done with a json file bind-mounted in flash: of the router (/mnt/flash)
Netlab appears to allow containerlab binds but they are not properly created. I thought that my netlab may be dated (1.7) so I moved to latest one 1.9.1.
Same situation.

### To Reproduce

Use minimal topology example below

### Expected behavior

Create binds that containerlab understands; or more accurately: correctly create yaml sequence for binds.

### Lab topology

```
provider: clab

defaults.devices.eos.clab.image: ceos:4.30.5M

nodes:
  temgm1:
    device: eos
    clab:
      binds:
        mymapping_json: '/mnt/flash/EosIntfMapping_json:ro'
```
It produces this clab.yml output:
```
name: topos

mgmt:
  network: netlab_mgmt
  ipv4-subnet: 192.168.121.0/24
  # Note: 'start' not validated
topology:
  nodes:
    temgm1:
      mgmt-ipv4: 192.168.121.101
      kind: ceos
      env: {'INTFTYPE': 'et', 'CLAB_MGMT_VRF': 'management'}
      image: ceos:4.30.5M
      runtime: docker
      binds:
      - mymapping_json:/mnt/flash/EosIntfMapping_json:ro
```
The listed binds - just one in the example- not correctly indented, cause errors further:

```
[CREATED] provider configuration file: clab.yml
Traceback (most recent call last):
  File "/home/aceplean/netlab/venv/bin/netlab", line 12, in <module>
    netsim.cli.lab_commands(__file__)
  File "/home/aceplean/netlab/venv/lib64/python3.8/site-packages/netsim/cli/__init__.py", line 333, in lab_commands
    mod.run(sys.argv[arg_start:])   # type: ignore
  File "/home/aceplean/netlab/venv/lib64/python3.8/site-packages/netsim/cli/up.py", line 371, in run
    run_up(cli_args)
  File "/home/aceplean/netlab/venv/lib64/python3.8/site-packages/netsim/cli/up.py", line 306, in run_up
    topology = get_topology(args,cli_args)
  File "/home/aceplean/netlab/venv/lib64/python3.8/site-packages/netsim/cli/up.py", line 91, in get_topology
[...]

PR #1403, opened 2024-10-20
Allow for ':' in values of binds dict

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1402 by limiting split on ':' to at most 2 parts


ipspace:
Lovely fix, thank you!

Issue #1405, opened 2024-10-20
Ability to specify eos device identification numbers in topology


aceplean:
EOS has a device serial number and a system MAC address identifying the instance. Netlab creates random values at spin-up.
It would be convenient in certain cases being able to pre-set one or both of these values by entering them in topology under the node.

Example of the values:
```
gw2#sh ver
Arista cEOSLab
Hardware version:
Serial number: 49CCB95CF768B841CC791ACB9A3CB2FF
Hardware MAC address: 001c.7324.e592
System MAC address: 001c.7324.e592
```

ipspace:
netlab does not do anything. EOS takes the system MAC address from somewhere, and if you could figure out where it gets it, we might be able to do something. I have zero idea what they do for the serial number.

However, keep in mind that netlab uses Vagrant or containerlab to start the workload, and has no influence (apart from what they allow us to configure) on how they set things up. We can tweak things, but only after the VMs/containers are started, which might be a bit late.

sdargoeuves:
There are some values you can override with Arista. I have used [this workaround](https://github.com/ipspace/netlab/discussions/910#discussioncomment-8871926) to set the SN on vEOS after reboot of the netlab server. I'd like to think it works the same on cEOS, but I haven't tested.

Related to cEOS, I came across this page recently: [Populate Arista cEOS Serial Number with Containerlab](https://campusnetworkengineering.com/posts/containerlab_populate_serial/)

Side note, not related to clab, but libvirt: I am currently using a pretty ugly workaround with virsh, to reuse the uuid from a backup of a dumpxml, so I can restart the vm and have it using the previous SN, as it seems the SN of some of the virsh/vagrant boxes are linked on the uuid (and the mac address), or at least this is the case for F5!

ipspace:
> Related to cEOS, I came across this page recently: [Populate Arista cEOS Serial Number with Containerlab](https://campusnetworkengineering.com/posts/containerlab_populate_serial/)

Thank you!

> Side note, not related to clab, but libvirt: I am currently using a pretty ugly workaround with virsh, to reuse the uuid from a backup of a dumpxml, so I can restart the vm and have it using the previous SN

vagrant-libvirt plugin can specify the uuid, and the management MAC addresses are fixed any (or DHCP wouldn't work correctly), so we could add **libvirt.uuid** node option. Will you open an issue or should I? 😎

sdargoeuves:
[...]

PR #1406, opened 2024-10-21
Add 'eos.serialnumber' and 'eos.systemmacaddr' device properties

jbemmel:
This PR allows for device-specific topology attributes that get added dynamically when a node uses a given device type.
Concretely, here it is used to define a ```eos.serialnumber``` node property (str) to configure the serial number, and ```eos.systemmacaddr``` for the system MAC (type checked and converted to the right format)

Fixes https://github.com/ipspace/netlab/issues/1405

Note: ```ceos-copy-to-flash``` was suggested, but I tested that regular bind-mounting also works. Code checks for conflicting user provided bind mounts

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -187,6 +187,10 @@ def find_node_device(n: Box, topology: Box) -> bool:
   if dev_def.get('node.module') and 'module' not in n:      # Have to copy default device module into node data
     n.module = dev_def.node.module                          # ... before modules are initialized
 
+  # JvB: Add device specific attributes to the data model, could auto-prefix with provider name
+  if 'attributes' in dev_def:
+    topology.defaults.attributes = topology.defaults.attributes + dev_def.attributes
+
   return True
 
 """
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -187,8 +187,7 @@ def find_node_device(n: Box, topology: Box) -> bool:
   if dev_def.get('node.module') and 'module' not in n:      # Have to copy default device module into node data
     n.module = dev_def.node.module                          # ... before modules are initialized
 
-  # JvB: Add device specific attributes to the data model, could auto-prefix with provider name
-  if 'attributes' in dev_def:
+  if 'attributes' in dev_def:                               # Add any device specific attributes to the data model
     topology.defaults.attributes = topology.defaults.attributes + dev_def.attributes
 
   return True
--- a/netsim/devices/eos.py
+++ b/netsim/devices/eos.py
@@ -5,6 +5,7 @@ from box import Box
 
 from . import _Quirks
 from ..utils import log
+from .. import data
 from ..augment import devices
 
 def check_mlps_vlan_bundle(node: Box) -> None:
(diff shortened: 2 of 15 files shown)


jbemmel:
Could make this more fancy and have the added attributes only be valid for the nodes in question, as opposed to globally, but I could also see the benefit of being able to switch node device types without having to comment out a ton of stuff every time. It doesn't really hurt that one can now define a 'ceos.serialnumber' for a FRR node in the same topology

netsim/augment/nodes.py
```diff
@@ -187,6 +187,10 @@ def find_node_device(n: Box, topology: Box) -> bool:
   if dev_def.get('node.module') and 'module' not in n:      # Have to copy default device module into node data
     n.module = dev_def.node.module                          # ... before modules are initialized
 
+  # JvB: Add device specific attributes to the data model, could auto-prefix with provider name
+  if 'attributes' in dev_def:
+    topology.defaults.attributes = topology.defaults.attributes + dev_def.attributes
```
ipspace, line 192: Good one. Like it.

netsim/devices/eos.yml
```diff
@@ -112,3 +115,8 @@ external:
   image: none
 graphite.icon: switch
 bgp._cprop_order: [ standard, extended, large, link-bandwidth ]
+
+# Define custom device-specific attributes
+attributes:
+ node:
+  ceos.serialnumber: int  # Referenced in ceos-config.j2 template
```
ipspace, line 122: It would be nice to have attributes that match device types to minimize confusion. From that perspective, it could be either `eos.serialnumber` (or is it `serial_number` 😉) or `clab.eos.serialnumber` if we want to emphasize that it only works on containers.

netsim/templates/provider/clab/eos/ceos-config.j2
```diff
@@ -0,0 +1 @@
+SERIALNUMBER={{ ceos.serialnumber if ceos is defined and 'serialnumber' in ceos else 123456789 }}
```
ipspace, line 1: This would set the same serial number on all cEOS nodes in the lab. We learned the hard way with Cisco IOL that could be a bad idea. Just make the whole line depend on the existence of the serial number attribute

ipspace (changes requested):

tests/topology/expected/isis-bfd-test.yml
```diff
@@ -233,6 +233,10 @@ nodes:
       ipv6: true
     box: ceos:4.32.1F
     clab:
+      binds:
+      - clab_files/n4/ceos-config:/mnt/flash/ceos-config
```
jbemmel, line 237: Just noticing this now - this is going to create conflicts...

jbemmel:
Note to self: Need to add a note to ceos documentation to tell users not to craft their own ceos-config mounts

ipspace:
> Note to self: Need to add a note to ceos documentation to tell users not to craft their own ceos-config mounts

Alternatively, we could be careful when adding binds from templates and check whether the user already created that bind (which would not be a bad idea anyway)

netsim/devices/eos.yml
jbemmel: Opting for ```eos.serialnumber```, we might one day be able to apply the same setting for other providers too. The modeling of the property and the user intent remains the same

netsim/templates/provider/clab/eos/ceos-config.j2
jbemmel: Made it depend and insert a comment on how to configure it if desired. Also added systemmacaddr

ipspace:
It was easier to fix stuff than to explain what's wrong. Merging...

netsim/devices/eos.py
```diff
@@ -65,6 +66,32 @@ def check_dhcp_clients(node: Box, topology: Box) -> None:
...
+  mnt_config = '/mnt/flash/ceos-config'
+  for ct in node.get('clab.binds',[]):
+    if mnt_config in ct:
+      log.error(
+        f"{mnt_config} file is already mapped, unable to configure eos.serialnumber (node {node.name}).",
+        category=log.Skipped,
+        module='quirks')
+      return
+
+  data.append_to_list(node.clab,'binds',f'clab_files/{node.name}/ceos-config:{mnt_config}')
```
jbemmel, line 92: I think this happens automatically after the config_template is rendered successfully

jbemmel (commented):
Let's double check if it works also when commenting out that line appending to 'binds'

jbemmel:
> It was easier to fix stuff than to explain what's wrong. Merging...

I did test the code and modifying the defaults does work, even if it may be late in the process

netsim/devices/eos.py
ipspace: I took the pragmatic approach of creating a fake template, dumping the node data in the quirks, and recreating the relevant data structures.

ipspace:
> > It was easier to fix stuff than to explain what's wrong. Merging...
> 
> I did test the code and modifying the defaults does work, even if it may be late in the process

I might have done something wrong, but I also ran your code and it did not work as expected. Also, why would we modify the settings for all EOS nodes if someone wants to set a serial number on one of them?

PR #1412, opened 2024-10-21
Be extra pedantic about IP address values

jbemmel:
Errors in transformation logic can lead to ```int``` type values being passed to templates.

As it turns out, ```ipv4``` believes ```1``` is a perfectly valid IP address -  this PR correctly marks it as "invalid"

### DIFF
--- a/netsim/ansible/templates/initial/eos.j2
+++ b/netsim/ansible/templates/initial/eos.j2
@@ -52,9 +52,9 @@ interface {{ l.ifname }}
     Set interface IPv4 addresses
 #}
 {% if 'ipv4' in l %}
-{%   if l.ipv4 == True %}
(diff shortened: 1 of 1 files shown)


ipspace:
> Errors in transformation logic can lead to `int` type values being passed to templates.

Working on it ;) That SHOULD NOT happen.

> As it turns out, `ipv4` believes `1` is a perfectly valid IP address - this PR correctly marks it as "invalid"

It's worse than that. The particular **ipv4** value you had was `1` and `True` is just another name for `1`. The expression `1 == True` is True (or 1)

Thanks for the fix though, it's definitely safer than what we had before. Do we have a volunteer to fix and test the other 20 templates? 😜

Issue #1413, opened 2024-10-21
[BUG] Global and node level 'gateway' attributes not propagated to interfaces


jbemmel:
### Describe the bug
In the given topology, the intent is to make ```n1``` a gateway on VLAN v1

### To Reproduce
```netlab create test.yml```

### Expected behavior
n1 should be configured as VRRP gateway

### Lab topology

```
module: [gateway,vlan]
gateway.protocol: vrrp

vlans:
 v1:

nodes:
 n1:
  vlans:
   v1:
    gateway: True
```

### Output
```
  File "/home/jvanbemmel/netlab/netsim/modules/gateway.py", line 39, in check_protocol_support
    gw_proto = intf.gateway.protocol                                  # We're asuming someone else did a sanity check on this value
               ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'bool' object has no attribute 'protocol'

```

### Version

dev branch

### Additional context
Suggested fix, add in gateway.py ```check_protocol_support``` line 36
```
    if intf.gateway is True:                                          # Expand global and node specific gateway parameters
      intf.gateway = topology.gateway + node.get('gateway',{})
```

ipspace:
It should be noted we never promised that you can enable a gateway protocol on a single node. It was designed to be enabled on a link, and all nodes on the link would run that protocol: https://netlab.tools/module/gateway/#link-parameters

One could argue this is a FAD, but even then the whole thing should not crash. One could also "fix" the crash with stricter requirements on interface **gateway** attribute (instead of https://github.com/ipspace/netlab/blob/dev/netsim/modules/gateway.yml#L28).

However, let's say I tentatively agree there might be a crazy niche case (that I can't figure out at the moment) where you would want to run VRRP (or whatever) on one node (with **gateway** module) but not on another one (also using **gateway** module for some other link).

Will look into how this could get fixed without too many hacks.

jbemmel:
The topology was minimized to simplify the discussion, but the full use case is having a bunch of switches forwarding that vlan across the network, with only a subset of them acting as gateway. The gateway properties would be defined once on the VLAN - for consistency - and then selected nodes would have ```gateway: True```

ipspace:
> The full use case is having a bunch of switches forwarding that vlan across the network, with only a subset of them acting as gateway.

Can't you use the **gateway** module just on that subset? It was not designed to be used selectively (per-node).

jbemmel:
[...]

Issue #1411, opened 2024-10-21
[BUG] Relative IPs on node VLANs are unresolved


jbemmel:
### Describe the bug

Assigning a relative IP to a node on a VLAN does not work as (I) expected

### To Reproduce

```netlab create test-topo.yml```

### Expected behavior

n1 should get IP 10.0.0.2/24

### Lab topology

```
---
module: [ vlan ]

vlans:
 v1:
  prefix: 10.0.0.0/24

nodes:
 n1:
  vlans:
   v1:
    ipv4: 2  # Assign relative IP 2 -> this does not work
 
 n2:

links:
- n1:
  n2:
   vlan.trunk:
    v1:
     ipv4: 3   # Assign relative IP 3 -> works
  vlan.trunk: [ v1 ]
```

### Output

```
# Ansible inventory created from ['/home/jvanbemmel/test/vlan-relative-ip.yml', '/home/jvanbemmel/.netlab.yml', 'package:topology-defaults.yml']
#

af:
  ipv4: true
box: generic/ubuntu2004
interfaces:
- ifindex: 1
  ifname: eth1
  libvirt:
    type: tunnel
  linkindex: 1
  name: n1 -> n2
  neighbors:
  - ifname: eth1
    node: n2
  remote_id: 2
  remote_ifindex: 1
  subif_index: 1
  type: p2p
- ifindex: 2
  ifname: eth1.1000
  parent_ifindex: 1
  parent_ifname: eth1
  type: vlan_member
  virtual_interface: true
  vlan:
    access: v1
    access_id: 1000
- bridge_group: 1
  ifindex: 3
  ifname: vlan1000
  ipv4: 2                                         <-- problem is here
  name: VLAN v1 (1000) -> [n2]
  neighbors:
  - ifname: vlan1000
    ipv4: 10.0.0.3/24
    node: n2
  type: svi
  virtual_interface: true
  vlan:
    mode: irb
    name: v1
loopback:
  ifindex: 0
  ifname: lo
  ipv4: 10.0.0.1/32
  neighbors: []
  type: loopback
  virtual_interface: true
mgmt:
  ifname: eth0
  ipv4: 192.168.121.101
  mac: 08:4f:a9:00:00:01
module:
- vlan
vlan:
  max_bridge_group: 1
vlans:
  v1:
    bridge_group: 1
    id: 1000
    ipv4: 2
    mode: irb
    prefix:
      allocation: id_based
      ipv4: 10.0.0.0/24
```

### Version

[...]

PR #1414, opened 2024-10-21
Tentative fix for gateway attribute propagation

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1413

Not 100% sure if this is complete, since link level gateway attributes don't get propagated - "gateway" is included in link_no_propagate


ipspace:
This is most probably not the correct way to go.

First I have to figure out whether this behavior is VLAN-specific (tied to #1411) or generic, then it will be clear what needs to be fixed.

ipspace:
As I suspected, this would just mask the true problem and make our life worse when we would have to troubleshoot why the adjacent hosts don't have the correct default gateway (because it wouldn't be calculated for the link)

Issue #1410, opened 2024-10-21
[BUG] Topology containing multiple 'nodes', 'links' or 'vlans' sections results in confusing errors


jbemmel:
A topology like:
```
nodes:
  n1:
  n2:

....
nodes:
  n3:
  n4:
```
does not generate an error message about having multiple ```nodes``` sections. Instead, you get confusing errors about members in groups being invalid nodes (because only the last section seems to be retained)

Triggered by https://github.com/ipspace/netlab/discussions/1409#discussioncomment-11005163

Perhaps it would be useful to allow for such composite topologies, by merging the sections?

ipspace:
There's not much we can do. This is a YAML parser issue, and it looks like the YAML parser does not get bothered by duplicate keys 🥲 

I know we had PyYAML versus Ruamel discussion a long while ago. In those days, I grudgingly added support for Ruamel because someone might use it, but since Ansible uses PyYAML, I decided to stick with that.

However, based on https://stackoverflow.com/questions/44904290/getting-duplicate-keys-in-yaml-using-python it looks like Ruamel (when used with the new API) triggers the duplicate key error. 

It looks like Box can take arguments that are then passed to PyYAML, so do you want to go down the path of custom safe loader described in the above URL? I wouldn't, but that's just me ;)

jbemmel:
Easy fix in netsim/utils/read.py:
```
# special loader with duplicate key checking
class UniqueKeyLoader(yaml.SafeLoader):
  def construct_mapping(self, node, deep=False):
      mapping = []
      for key_node, value_node in node.value:
          key = self.construct_object(key_node, deep=deep)
          if key in mapping:
            log.fatal(f"Duplicate section in YAML file: {key}") 
          mapping.append(key)
      return super().construct_mapping(node, deep)

...
yaml_data = Box().from_yaml(filename=filename,default_box=True,box_dots=True,default_box_none_transform=False,Loader=UniqueKeyLoader) (2x)
```

Box defaults to Ruamel

ipspace:
> Easy fix in netsim/utils/read.py:

Perfect. Check it with both YAML libraries and submit a PR

> Box defaults to Ruamel

... but we have PyYAML in the requirements so people using Ansible (= eveyone) don't end with two YAML parsers on their system (https://xkcd.com/927/ comes to mind)

jbemmel:
> > Easy fix in netsim/utils/read.py:
> 
> Perfect. Check it with both YAML libraries and submit a PR

[...]

PR #1416, opened 2024-10-21
Check for duplicate keys when loading YAML through PyYAML

jbemmel:
Also add an error test case

Fixes https://github.com/ipspace/netlab/issues/1410


ipspace:
That was fast ;) Thank you!

ipspace:
@jbemmel: FWIW, [bbffd0b](https://github.com/ipspace/netlab/pull/1416/commits/bbffd0b98dbf6d110cbbd55997c4f61eae6e306d) was the absolutely worst stunt I've seen in ages -- when the CI/CD tests fail, let's fake the expected results without ever thinking about the root cause.

If you call log.fatal from something that's under "try" block, sys.exit() is caught and returned to the caller. That's why you got the "1" back. Under the CI/CD testing condition, **fatal** raises an error with the error message as its value instead of calling sys.exit() (or we wouldn't be able to do a sequence of error tests), so you got back the error.

I made the code work identically in both environments and amended the commit. I also think this was an absolute time waste -- if you edit YAML in any decent code editor, you'll get warnings with duplicate keys.

Issue #1418, opened 2024-10-22
[BUG] VRRP on cEOS - no preempt needs to be explicit


sdargoeuves:
### Describe the bug

Despite setting the `vrrp.preempt` to false on cEOS, I can see on the `show vrrp` output that Preemption is enabled.

### To Reproduce

topology:

```yaml
vlans:
  vl_blue:
    id: 3032
    prefix: 172.16.32.0/24
    ospf.passive: true
    gateway:
      protocol: vrrp
      vrrp.preempt: false
  vl_red:
    id: 3033
    prefix: 172.16.33.0/24
    ospf.passive: true
    gateway:
      protocol: vrrp
      vrrp.preempt: false
```

Output:

```bash
Connecting to clab-ml_2-gw1 using SSH port 22, executing show vrrp
Vlan3032 - Group 1
  VRF is default
  VRRP Version 3
  State is Master
  Last state transition was 00:01:37 ago
  Virtual IPv4 address is 172.16.32.254
  Virtual MAC address is 0000.5e00.0101
  Mac Address Advertisement interval is 30s
  VRRP Advertisement interval is 1s
  Preemption is enabled
  Preemption delay is 0s
  Preemption reload delay is 0s
  Priority is 120
  Master Router is 172.16.32.1 (local), priority is 120
  Master Advertisement interval is 1s
  Skew time is 0.530s
  Master Down interval is 3.530s

Vlan3033 - Group 1
  VRF is default
  VRRP Version 3
  State is Backup
  Last state transition was 00:01:40 ago
  Virtual IPv4 address is 172.16.33.254
  Virtual MAC address is 0000.5e00.0101
  Mac Address Advertisement interval is 30s
  VRRP Advertisement interval is 1s
  Preemption is enabled
  Preemption delay is 0s
  Preemption reload delay is 0s
  Priority is 100
  Master Router is 172.16.33.2, priority is 120
  Master Advertisement interval is 1s
  Skew time is 0.600s
  Master Down interval is 3.600s
```

### Expected behavior

I would expect to see ` Preemption is disabled` in the output

### Lab topology

```yaml
---
plugin: [ multilab ]
defaults.multilab.id: 2
defaults.addressing.mgmt.start: 10
provider: clab

defaults.devices.eos.clab.image: "ceos:4.32.2F.1"

groups:
  vl_gws:
    device: eos
    module: [vlan, gateway, ospf]
    members: [ gw1, gw2 ]

vlans:
  vl_blue:
[...]

PR #1419, opened 2024-10-22
#1418 - vrrp no preempt needs to be explicit on cEOS

sdargoeuves:
#1418 - simple tweak to force the `no vrrp <group> preempt`

### DIFF
--- a/docs/module/gateway.md
+++ b/docs/module/gateway.md
@@ -92,6 +92,6 @@ More complex topologies, such as multiple VRRPv3 instances *on the same subnet*,
 
 * **vrrp.group** (default: 1) -- VRRP group
 * **vrrp.priority** (interface parameter, integer)
-* **vrrp.preempt** (interface parameter, boolean)
+* **vrrp.preempt** (interface parameter, boolean) -- VRRP preempting is enabled by default; you can disable it with **vrrp.preempt: False**.
 
 No other aspect of VRRP (VRRP version, timers...) is manageable through *netlab* parameters; if you want to configure them, create a plugin to implement additional VRRP functionality.
--- a/netsim/ansible/templates/gateway/arubacx.j2
+++ b/netsim/ansible/templates/gateway/arubacx.j2
@@ -30,7 +30,7 @@ interface {{ intf.ifname }}
   address fe80::1 primary
   address {{ intf.gateway[af]|ipaddr('address') }} secondary
 {%       endif %}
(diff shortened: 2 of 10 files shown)


sdargoeuves:
I've just been thinking about it, do you think we should change the default to be True, so if it's not specified, preemption will be enabled?

ipspace:
There are two independent issues here:

* What is the default value of **gateway.vrrp.preempt** parameter
* How to configure that on Arista EOS

Will open another issue on the first one, once that's settled, you might have to change the `|default` bit. Other than that, the code looks good.

ipspace:
> Edit: It seems to be how cEOS behaves, if you don't specify `no vrrp <group> preempt` the default behaviour is to enable preemption. So I feel like changing the default to True follows the same principle.

I don't want different devices to have different configuration template behavior or different default values. We have to agree on what the default is (please comment in #1420, "do nothing" is an option) and then ensure all configuration templates are consistent with that.

sdargoeuves:
## Tests Results

| Platform | Preempt False | Preempt True | Preempt Default | Status |
|:---------|:--------------|:-------------|:----------------|:-------|
| ArubaCX | Disabled | Enabled | Enabled | ✅ |
| Cumulus | Disabled | Enabled | Enabled | ✅ |
| DellOS | not tested |-|-|❔|
| EOS | Disabled | Enabled | Enabled | ✅ |
| FRR | Disabled | Enabled | Enabled | ✅ |
| IOS | Disabled | Enabled | Enabled | ✅ |
| NXOS | not tested |-|-|❔|
| SRLinux | N/A | N/A |
| SROS | not tested |-|-|❔|
| VyOS | Disabled | Enabled | Enabled | ✅ |

I haven't got SROS not DellOS currently, but looking at the code, both should work as expected

sdargoeuves:
@ipspace thank you for the doc update and the merge #1419 
I was wondering if you already have a (draft) page for the next release notes. We should mention the change of behaviour, as topology using the default, could result in a different state.
Maybe you already have some magic happening to capture all those changes, and update this part just before the new version is released.

ipspace:
> I was wondering if you already have a (draft) page for the next release notes.

`docs/release/1.9.md`, but I haven't started the release 1.9.2 yet.

> Maybe you already have some magic happening to capture all those changes, and update this part just before the new version is released.

**git changelog** from https://github.com/tj/git-extras followed by some copy-paste-sorting. Don't worry about that (but thank you!), I do it before pushing out the release.

PR #1422, opened 2024-10-23
Enable gateway protocols with interface-level granularity

### LINKED ISSUE #1413: [BUG] Global and node level 'gateway' attributes not propagated to interfaces

ipspace:
Fixes #1413, replaces #1414

### DIFF
--- a/netsim/modules/gateway.py
+++ b/netsim/modules/gateway.py
@@ -172,10 +172,11 @@ class FHRP(_Module):
       if 'gateway' in link:                                 # Do we already have 'gateway' on link?
         if link.gateway is False:                           # If we have 'gateway:false' we're dealing with a stupidity
           log.error(
-            'Cannot combine gateway:false on link with gateway:true on interface',
+            'Interface gateway attribute ignored due to gateway:false configured on the link',
             more_data=f'Link {link._linkname} node {intf.node}',
-            category=log.IncorrectValue,
+            category=Warning,
             module='gateway')
+          intf.pop('gateway',None)
       else:
         link.gateway = True                                 # Found interface gateway attribute but no link GW. Fix that
         copy_to_link = True                                 # And remember that someone gave us extra work to do :(
--- a/netsim/modules/gateway.py
+++ b/netsim/modules/gateway.py
@@ -153,8 +153,50 @@ class FHRP(_Module):
         process_gw_id(vdata.get('gateway.id',None),required=False,path=f'VLAN {vname}')
 
   def link_pre_link_transform(self, link: Box, topology: Box) -> None:
-    if not 'gateway' in link:
-      return
+    #
+    # Step#1: deal with 'gateway' interface attributes
+    #
+    # If an interface has 'gateway' parameter (and it's not False), we must enable
+    # gateway on the link and remember that we have to turn 'gateway' off on other
+    # interfaces if needed
+    #
+    # Also: combining 'gateway:False' on link with 'gateway:True' on interface is
+    # a stupidity and will be rejected
+    #
+    copy_to_link = False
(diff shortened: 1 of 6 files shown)


ipspace:
Documentation is coming in a separate commit ;)

netsim/modules/gateway.py
```diff
@@ -153,8 +153,50 @@ def module_init(self, topology: Box) -> None:
...
+    copy_to_link = False
+    for intf in link.interfaces:
+      if 'gateway' not in intf:                             # No 'extra special' stuff, move on
+        continue
+      if intf.get('gateway',None) is False:                 # The 'special' stuff is don't do it?
+        continue                                            # ... sure, I can do that ;)
+      if 'gateway' in link:                                 # Do we already have 'gateway' on link?
+        if link.gateway is False:                           # If we have 'gateway:false' we're dealing with a stupidity
+          log.error(
+            'Cannot combine gateway:false on link with gateway:true on interface',
```
jbemmel, line 175: ```gateway``` on interface could also be a dict setting the vrrp.priority for example

"Cannot combine gateway:false on link with gateway attribute on interface" (although I could come up with scenarios where that could make sense to me, I'll spare you the eye-rolling exercise this time)

jbemmel:
```
---
module: [gateway,vlan]

gateway:
 protocol: vrrp
 id: 1

vlans:
 core:
 access:
  # gateway: True -> This enables the gateway

groups:
 _auto_create: True
 
 core:
  members: [ c1, c2, c3 ]
  vlans:
   core:
    gateway: True # -> This does work
 
 access:
  members: [ a1, a2 ]
  vlans:
   access:
    gateway: True  # -> This does not

 hosts:
  members: [ h1 ]
  device: linux

links:
- c1:
  c2:
  c3:
  a1:
  a2:
  vlan.trunk: [ core ]

- a1:
    vlan.access: access
  a2:
    vlan.access: access
  h1:
```

Possibly not an issue introduced by this PR, but enabling the gateway in a trunk vlan works but not in an access vlan

ipspace:
> Possibly not an issue introduced by this PR, but enabling the gateway in a trunk vlan works but not in an access vlan

Nice one, thank you. Will figure out what's going on. Merge postponed.

netsim/modules/gateway.py
jbemmel: ```Warning: interface gateway attributes ignored due to link.gateway=False```

ipspace:
> Possibly not an issue introduced by this PR, but enabling the gateway in a trunk vlan works but not in an access vlan

That issue is related to #1424. It looks like I copied node vlan data into interface data only on trunks but not on access links. Will open a separate issue for that and fix the code.

The "this should be a warning" part has been implemented, merging...

PR #1424, opened 2024-10-23
Fix: copy select node VLAN data into intf data before link transform

### LINKED ISSUE #1411: [BUG] Relative IPs on node VLANs are unresolved

ipspace:
Majority of node VLAN data is copied into SVI interface data after the link transformation. However, that means we could not influence the SVI IP address or selectively enable gateway functionality on a node VLAN.

This fix copies a subset of node vlan data into link interface data for all (virtual or real) access VLAN links, and ensures those attributes do not overwrite the results of link transformation.

Fixes #1411, probably also VLAN part of #1413 (need a regression test after #1422 is merged)

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -789,7 +789,7 @@ def create_svi_interfaces(node: Box, topology: Box) -> dict:
                                                                             # Overwrite interface settings with VLAN settings
       vlan_ifdata = vlan_ifdata + {
                       k:v for k,v in vlan_data.items()
-                        if k not in svi_skipattr and
+                        if k not in svi_skipattr and k not in copy_attr and
                            (v is not True or k not in vlan_ifdata) }
       fix_vlan_mode_attribute(vlan_ifdata)
       node.interfaces.append(vlan_ifdata)                                   # ... and add SVI interface to list of node interfaces
@@ -1279,6 +1279,29 @@ class VLAN(_Module):
 
   def module_pre_link_transform(self, topology: Box) -> None:
     create_loopback_vlan_links(topology)
+
+  # We need to copy a few attributes from node VLAN data into access link interface data, for example
+  # the node IPv4/IPv6 address (to support static addressing) and "gateway" attribute to enable
+  # selective VLAN gateway module activation
+  #
+  # This operation needs to be done when the link data structures have already been transformed into
+  # the canonical "interfaces" list format, so we cannot do it any sooner than this point.
+  #
+  def link_pre_link_transform(self, link: Box, topology: Box) -> None:
+    if 'vlan' not in link:
+      return
+
+    for intf in link.interfaces:
+      vname = intf.get('vlan.access',None)
+      if not vname:
+        continue
+      if vname not in topology.nodes[intf.node].get('vlans',{}):
+        continue
+      vdata = topology.nodes[intf.node].vlans[vname]
+      for attr in topology.defaults.vlan.attributes.copy_vlan_to_intf:
+        if attr not in vdata:
+          continue
+        intf[attr] = vdata[attr]
 
   def module_post_link_transform(self, topology: Box) -> None:
     for n in topology.nodes.values():
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1299,9 +1299,8 @@ class VLAN(_Module):
         continue
       vdata = topology.nodes[intf.node].vlans[vname]
       for attr in topology.defaults.vlan.attributes.copy_vlan_to_intf:
-        if attr not in vdata:
-          continue
-        intf[attr] = vdata[attr]
+        if attr in vdata:
+          intf[attr] = vdata[attr]
 
   def module_post_link_transform(self, topology: Box) -> None:
     for n in topology.nodes.values():
--- a/netsim/modules/vlan.yml
+++ b/netsim/modules/vlan.yml
@@ -28,6 +28,13 @@ attributes:
     mode:
     prefix:
     evpn:
+
+  # Copy these attributes from node VLAN data into interface-on-link data
+  copy_vlan_to_intf:
+    ipv4:
+    ipv6:
+    gateway:
+
   #
   # Do not copy these VLAN attributes into SVI interfaces
   #vlan_svi_no_propagate:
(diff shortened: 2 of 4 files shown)


jbemmel:
May have found an issue:
```
module: [ vlan ]

vlans:
 v1:
  prefix: 10.0.0.0/24

nodes:
 n2:
  vlans:
   v1:
    ipv4: 2  # Assign relative IP 2

links:
- n1:
  n2:
   vlan.trunk:
    v1:
     ipv4: 3   # Assign relative IP 3, overriding '2'
  vlan.trunk: [ v1 ]
```
Result:
```
[ERRORS]  Errors found in vlan-relative-ip.yml
[VALUE]   vlan: VLAN v1 used by node n2 on links[1] is not used by any other node on that link
```

ipspace:
It does look like two bugs:

* The VLAN module should not complain about the mismatch in VLAN trunk
* It should not allow to configure (for example) **ipv4** on the VLAN trunk if the VLAN mode is not IRB.

We attached too many nerd knobs to this thing 🤷‍♂️, and some people keep asking for more 🤨

ipspace:
> May have found an issue:

Indeed (thank you!) but it's unrelated to this PR. The same thing happens in the main dev branch.

jbemmel:
> It does look like two bugs:
> 
> * The VLAN module should not complain about the mismatch in VLAN trunk
> * It should not allow to configure (for example) **ipv4** on the VLAN trunk if the VLAN mode is not IRB.
> 
> We attached too many nerd knobs to this thing 🤷‍♂️, and some people keep asking for more 🤨

As nerd knobs go, putting IPs on VLAN interfaces is pretty fundamental to networking. Some people are correct for expecting that to work properly

ipspace:
> > We attached too many nerd knobs to this thing 🤷‍♂️, and some people keep asking for more 🤨
> 
> As nerd knobs go, putting IPs on VLAN interfaces is pretty fundamental to networking.

I had other nerd knobs in mind that are now causing headaches with fundamental stuff. Now that we got the exchange of opinions out of the way, how does the code look to you?

netsim/modules/vlan.py
```diff
@@ -1280,6 +1280,29 @@ def link_pre_transform(self, link: Box, topology: Box) -> None:
...
+      return
+
+    for intf in link.interfaces:
+      vname = intf.get('vlan.access',None)
+      if not vname:
+        continue
+      if vname not in topology.nodes[intf.node].get('vlans',{}):
+        continue
+      vdata = topology.nodes[intf.node].vlans[vname]
+      for attr in topology.defaults.vlan.attributes.copy_vlan_to_intf:
```
jbemmel, line 1301: Do you anticipate that other modules might need to extend this list, and that's why it's defined in ```vlan.yml```?
If not, I would just list the attributes here:
```
for attr in ["ipv4","ipv6","gateway"]
```

In particular for the ```gateway``` attribute, it would be "better" (more elegant imho) to have the ```gateway``` module code append "gateway" to the list of attributes to be copied. As is, the ```vlan``` module is making assumptions about another module

netsim/modules/vlan.py
```diff
@@ -1280,6 +1280,29 @@ def link_pre_transform(self, link: Box, topology: Box) -> None:
...
+        if attr not in vdata:
```
jbemmel, line 1302: if attr in vdata:
   ... copy it ...

jbemmel (approved):
Unrelated to this PR, but I've always found ```phy_ifattr``` a bit of a misnomer.

For all the lists of attributes in vlan.yml: if they don't need to be modified externally, I would keep them internally in vlan.py for improved readability

netsim/modules/vlan.py
ipspace: I love having stuff like that list defined in the defaults (as opposed to hardcoded inline or defined in the module header) to make it easier to change if we hit a snag, similar to how I could write the instructions for fixing containerlab startup parameters (blog post coming in an hour or two)

netsim/modules/vlan.py
ipspace: Yeah, I was in an auto-pattern I guess ;) Thank you

Issue #1428, opened 2024-10-24
Check for unicast MACs


jbemmel:
https://github.com/ipspace/netlab/blob/1b3218cae21b1132321adae850db11dc1871baca/netsim/data/types.py#L761

In many (all?) cases I believe the MAC should be a unicast MAC address - we could validate that

ipspace:
> In many (all?) cases I believe the MAC should be a unicast MAC address - we could validate that

Agree, makes perfect sense. Will you implement that?

For future reference: if we ever need to recognize broadcast addresses, we can add an extra parameter to the validation function (similar to min_value / max_value on ints)

PR #1431, opened 2024-10-25
Check for the multicast bit in MAC addresses, and report an error if set

jbemmel:
Adds error test case to check

Fixes https://github.com/ipspace/netlab/issues/1428

### DIFF
--- a/netsim/data/types.py
+++ b/netsim/data/types.py
@@ -764,6 +764,9 @@ def must_be_mac(value: typing.Any) -> dict:
 
   try:
     parse = netaddr.EUI(value)                                        # now let's check if we have a MAC address
+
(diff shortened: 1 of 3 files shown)


netsim/data/types.py
```diff
@@ -764,6 +764,9 @@ def must_be_mac(value: typing.Any) -> dict:
 
   try:
     parse = netaddr.EUI(value)                                        # now let's check if we have a MAC address
+
+    if int(parse) & 0x010000000000:                                   # Check if the multicast bit is set
```
ipspace, line 768: One could use `bool(parse.words[0] & 0b01)` but we get the same result anyway ;)

ipspace (approved):

ipspace:
Thank you!

PR #1432, opened 2024-10-25
Jvb fix access vlan gateway

jbemmel:
1. In ```check_link_vlan_attributes```, reverse the order: First check if the vlan exists on the node, and if so add the node to the node_set. Only after that check the global vlans
2. ```copy_vlan_attributes```: Also copy gateway attributes, if any
3. ```set_link_vlan_prefix```, after copying global vlan attributes also copy any node level vlan attributes, overriding the global ones if any
4. ```link_pre_transform``` populate the node_set for access and native vlans, line 175 was doing this only when a node vlan was present

Fixes https://github.com/ipspace/netlab/pull/1422

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -171,10 +171,10 @@ def check_link_vlan_attributes(obj: Box, link: Box, v_attr: Box, topology: Box) 
     v_attr[attr].use_count = v_attr[attr].use_count + 1
 
     for vname in vlan_list:                                       # Check that VLANs exist
-      if vname in topology.get('vlans',{}):
-        continue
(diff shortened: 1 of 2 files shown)


jbemmel:
The method ```check_link_vlan_attributes``` confused me a little, because in addition to checking it also builds up the node_set in v_attr (partly). "check" implies read-only (and report errors) in my mind

ipspace:
Thank you! Have to find a few hours of quiet time to review these changes. Unfortunately, there's often a good reason behind the way we do things (often learned the hard way), and I have to figure out what those reasons were and what the impact of these changes might be.

ipspace:
This should be fixed by 4ae26b61abf0ce7e0d00e4a0536ab0630579bffd based on the clues from #1442

PR #1439, opened 2024-10-26
Further IOLL2 enablement

DanPartelly:
Ioll2 was promoted to multilayer switch
Routed sub-interfaces where enabled and the vlan template was taught how to configure them
Routed Native vlan was enabled

This change set does not promotes iosvl2 to a multilayer switch.  The platform was not tested. When enabled and tested, depending of the results of the tests, we may have to do away with the symlinks which are still present. 

It also does not contain any attempt to mixed trunks.

Changes where tested on vlan integration test suite.  At this time all tests up and including 61-vlan-routed-native are a success.  test success is dependent of running the correct version of iouyap.

### DIFF
--- a/netsim/ansible/templates/vlan/iosvl2.j2
+++ b/netsim/ansible/templates/vlan/iosvl2.j2
@@ -5,9 +5,10 @@ vlan {{ vdata.id }}
 !
 {%   endfor +%}
 {% endif %}
-{% for ifdata in interfaces if ifdata.vlan is defined %}
+{% for ifdata in interfaces if ifdata.vlan is defined and ifdata.virtual_interface is not defined %}
 !
 interface {{ ifdata.ifname }}
+
 {%   if ifdata.vlan.trunk_id is defined %}
  switchport
  switchport trunk encapsulation dot1q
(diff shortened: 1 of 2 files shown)


ipspace:
> Ioll2 was promoted to multilayer switch Routed sub-interfaces where enabled and the vlan template was taught how to configure them Routed Native vlan was enabled

Thank you!

> This change set does not promotes iosvl2 to a multilayer switch. The platform was not tested. When enabled and tested, depending of the results of the tests, we may have to do away with the symlinks which are still present.

Will run the integration tests.

> Changes were tested on vlan integration test suite.

Perfect.

> At this time all tests up and including 61-vlan-routed-native are a success. test success is dependent of running the correct version of iouyap.

However, as I cannot reproduce your result (nor have a burning desire to run a highly customized container ;), I would suggest to (temporarily, until we have an official solution) remove IOLL2 from the integration test results.

DanPartelly:
Roman offered to host the packages for the next iouyap version on clab package site. Once I'm happy with it's state ( I estimate next weekend) , I'll  a release it, and it will be easy for everybody to build the containers. 

> However, as I cannot reproduce your result (nor have a burning desire to run a highly customized container ;), I would suggest to (temporarily, until we have an official solution) remove IOLL2 from the integration test results.

Issue #1436, opened 2024-10-26
[BUG] Node VLAN data is not copied into interface data when the access VLAN is defined on the interface


ipspace:
When you define an access VLAN on an interface but not on the parent link, the node VLAN data is not copied into the interface data. It looks like the access VLAN information is copied into the link data too late to matter.

The bug was initially observed by @jbemmel when reviewing #1422, a proposed fix is in #1432

### Lab topology

```
---
module: [ vlan ]

defaults.device: eos

vlans:
  access:

nodes:
  a1:
    vlans.access.ipv4: 42
  a2:

links:
- a1:
    vlan.access: access
  a2:
    vlan.access: access
```

Adding `vlan.access` to the link definition results in the correct behavior, so the bug must be somewhere in the VLAN attribute initialization code.

PR #1442, opened 2024-10-27
Issue error when user tries to set access vlan attributes via an interface

jbemmel:
Suggestion for https://github.com/ipspace/netlab/discussions/1440

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1288,17 +1288,22 @@ class VLAN(_Module):
   # the canonical "interfaces" list format, so we cannot do it any sooner than this point.
   #
   def link_pre_link_transform(self, link: Box, topology: Box) -> None:
-    if 'vlan' not in link:
-      return
+    
+    # Process both link level and interface level access vlans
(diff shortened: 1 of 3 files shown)


ipspace:
Thanks for the suggestion. I think we need to be stricter than that -- everything that is not whitelisted should be banned. Is there anything (apart from the **vlan** attribute itself) that should be whitelisted?

ipspace:
Replaced by #1445 (and thanks a million for finding the root cause of #1436)

jbemmel:
> Thanks for the suggestion. I think we need to be stricter than that -- everything that is not whitelisted should be banned. Is there anything (apart from the **vlan** attribute itself) that should be whitelisted?

The code itself puts ‘’’ipv4: False’’’ on access vlan interfaces at times (and then removes it before producing the topology files)

ipspace:
> The code itself puts ‘’’ipv4: False’’’ on access vlan interfaces at times (and then removes it before producing the topology files)

Thank you. Yeah, I found that out the hard way ;) It's included in #1445

PR #1445, opened 2024-10-27
Bug fix: reject per-link irb/bridge VLAN interface attributes

### LINKED ISSUE #1359: [BUG] DHCP 'default router' option is missing for VLAN DHCP pools

ipspace:
This fix introduces a strict check of link interface attributes for VLAN access connections where the VLAN forwarding mode is 'bridge' or 'irb'. The only exceptions are the physical interface attributes, STP parameters, and 'no IPv4/IPv6' parameters.

The fix resulted in 'dhcp-vlan' transformation test failure, and after fixing that transformation test, #1359 magically disappeared.

Replaces #1442, accidentally fixes #1359

### DIFF
--- a/netsim/modules/dhcp.py
+++ b/netsim/modules/dhcp.py
@@ -329,6 +329,9 @@ class DHCP(_Module):
 
         # Copy link DHCP client status to interface in case the node does not have DHCP module enabled
         if link.get(f'dhcp.client.{af}',False) and intf.get(f'dhcp.client.{af}',None) is not False:
+          vlan_intf = intf.get('_vlan_mode',None)
+          if vlan_intf in ['bridge','irb']:         # No automatic DHCP clients on bridging/IRB interfaces
+            continue
           intf.dhcp.client[af] = True
 
         if intf.get(f'dhcp.client.{af}',False):     # Do we have DHCP client on this interface?
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -296,6 +296,34 @@ def validate_link_vlan_attributes(link: Box,v_attr: Box,topology: Box) -> bool:
                 intf[af] = False                                  # ... make sure it's not connected to the native VLAN subnet
   return link_ok
 
+"""
+check_link_interface_attributes -- validate that the interfaces in bridge/IRB mode do not
+carry any extra attributes
+"""
+def check_link_interface_attributes(link: Box, intf: Box, topology: Box) -> None:
+  valid_attr = list(topology.defaults.vlan.attributes.phy_ifattr) + [ 'node','_vlan_mode' ]
+  vlan_mode = get_vlan_interface_mode(intf)                 # Get interface VLAN forwarding mode
+  if vlan_mode is None or vlan_mode == 'route':             # No access VLAN or routed subinterface
+    return                                                  # ... interface attributes are OK
+
+  invalid = []                                              # This is where we collect invalid attributes
+  for attr in intf.keys():                                  # Now iterate over interface attributes
+    if attr in valid_attr:                                  # L2 attribute below VLANs?
+      continue                                              # ... OK, move on
+    if attr in log.AF_LIST and intf[attr] is False:         # Address family set to False?
+      continue                                              # ... OK, we need that to enforce bridge forwarding mode
+    invalid.append(attr)                                    # Everything else is invalid
+
+  if not invalid:
+    return
+
+  log.error(
+    f"You must use 'nodes.{intf.node}.vlans.{intf.vlan.access}' dictionary to set VLAN interface attributes,",
+    more_data=[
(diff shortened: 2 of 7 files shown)


netsim/modules/dhcp.py
```diff
@@ -329,6 +329,9 @@ def link_pre_transform(self, link: Box, topology: Box) -> None:
 
         # Copy link DHCP client status to interface in case the node does not have DHCP module enabled
         if link.get(f'dhcp.client.{af}',False) and intf.get(f'dhcp.client.{af}',None) is not False:
+          vlan_intf = intf.get('_vlan_mode',None)
+          if vlan_intf in ['bridge','irb']:         # No automatic DHCP clients on bridging/IRB interfaces
+            continue
```
jbemmel, line 334: add a warning: Ignoring user provided dhcp.client.ipv[x] flag on interface

netsim/modules/vlan.py
```diff
@@ -296,6 +296,34 @@ def validate_link_vlan_attributes(link: Box,v_attr: Box,topology: Box) -> bool:
                 intf[af] = False                                  # ... make sure it's not connected to the native VLAN subnet
   return link_ok
 
+"""
+check_link_interface_attributes -- validate that the interfaces in bridge/IRB mode do not
+carry any extra attributes
+"""
+def check_link_interface_attributes(link: Box, intf: Box, topology: Box) -> None:
+  valid_attr = list(topology.defaults.vlan.attributes.phy_ifattr) + [ 'node','_vlan_mode' ]
```
jbemmel, line 304: Could move this down a few lines

netsim/modules/vlan.yml
```diff
@@ -50,6 +50,7 @@ attributes:
     type:
     vlan:
     mtu:
+    stp:
```
jbemmel, line 53: What about ```bandwidth```?

netsim/modules/dhcp.py
ipspace: I understand what you're saying, but there a whole saga behind the "dhcp.client" -- it is set everywhere (including the DHCP relay) and then cleaned up as needed. If we were to set it here, we'd have to modify the cleanup code as well, tightly coupling DHCP and VLAN module.

From a more realistic perspective: having a DHCP client on a VLAN interface on a VLAN-enabled LAN is less usual than not having it. Usually, the VLAN interface would be the DHCP relay.

netsim/modules/vlan.py
```diff
@@ -296,6 +296,34 @@ def validate_link_vlan_attributes(link: Box,v_attr: Box,topology: Box) -> bool:
...
+  vlan_mode = get_vlan_interface_mode(intf)                 # Get interface VLAN forwarding mode
+  if vlan_mode is None or vlan_mode == 'route':             # No access VLAN or routed subinterface
+    return                                                  # ... interface attributes are OK
+
+  invalid = []                                              # This is where we collect invalid attributes
+  for attr in intf.keys():                                  # Now iterate over interface attributes
+    if attr in valid_attr:                                  # L2 attribute below VLANs?
+      continue                                              # ... OK, move on
+    if attr in log.AF_LIST and intf[attr] is False:         # Address family set to False?
```
jbemmel, line 313: I am working on a proposal for this: Instead of setting ipv4/6: False flags on interfaces, set it in the node vlan(s) and then have the code in ```link_pre_link_transform``` copy them to the interfaces

jbemmel (approved):
Couple of minor comments, I have some suggestions for improvements but those can be added after merging this one

netsim/modules/vlan.py
ipspace: OK, let's see how that turns out. I'm not sure it's worth investing the time though.

netsim/modules/vlan.py
jbemmel: ‘’’valid_attr’’’ initialization, it may not be needed

PR #1448, opened 2024-10-27
Copy link attribute 'bandwidth' to LAG members if present

jbemmel:


ipspace:
Now you see why it's good to have lists of attributes defined in system defaults as a dictionary instead of using lists or hardcoded values -- anyone could have added that attribute (or any other attribute they wish to propagate) just by modifying user or even topology defaults.

PR #1451, opened 2024-10-28
Found issue with missing 'vlan_name' for access vlans

jbemmel:
The ```link_name``` for access vlans does not get populated, resulting in the creation of loopback vlans in certain scenarios (ref https://github.com/ipspace/netlab/blob/a937fa2bb10b437d75da7869cc29ecad5f0c2396/netsim/modules/vlan.py#L652 - e.g. test case rt-vlan-mode-link-route:
```
- sw:
  br:
  vlan.access: red
  vlan.mode: route        # This one should be routed on both ends
```

However, setting the link_name uncovers other issues - e.g. the vrf-leak link in bgp-vrf-local-as looses its neighbors, as the link is not flagged as a "routed_link":
```
 interfaces: # VRF route leaking between red and blue, using BGP peering
  - node: r2
    vrf: red
    bgp.local_as: 65001
    vlan.access: vrf-leak
  - node: r2
    vrf: blue
    bgp.local_as: 65002
    vlan.access: vrf-leak
```

Not ready to merge, but wanted to share the issue

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1321,8 +1321,8 @@ class VLAN(_Module):
       for intf in link.interfaces:
         if 'vlan' in intf:
           intf.vlan.mode = 'route'
-          if len(link.interfaces)==2:
-            intf.vlan.routed_link = True                                          # Mark it as a routed link? has issues
+          # if len(link.interfaces)==2:
+          #  intf.vlan.routed_link = True                                          # Mark it as a routed link? has issues
     else:
       set_link_vlan_prefix(link,v_attr,topology)
 
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1302,6 +1302,7 @@ class VLAN(_Module):
     link_vlan = get_link_access_vlan(v_attr)
     routed_vlan = False
     if not link_vlan is None:
+      link.vlan_name = link_vlan                                                  # Mark the use of this vlan to avoid creating loopback vlan
       set_access_vlan_interface_mode(link,link_vlan,topology)
       routed_vlan = routed_access_vlan(link,topology,link_vlan)
       vlan_data = topology.get(f'vlans.{link_vlan}',None)                         # Get global VLAN data
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -1321,6 +1321,8 @@ class VLAN(_Module):
       for intf in link.interfaces:
         if 'vlan' in intf:
           intf.vlan.mode = 'route'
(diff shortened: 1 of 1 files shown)


ipspace:
> - e.g. test case rt-vlan-mode-link-route:
> ```
> - sw:
>   br:
>   vlan.access: red
>   vlan.mode: route        # This one should be routed on both ends
> ```

Please do remember that an access interface on a routed VLAN is indistinguishable from a regular routed interface. The packets are sent and received without the VLAN tag (because access link) and then forwarded using the routing table not the MAC table. I hope we agree that describes a regular interface on a router. I think I wrote about this a long while ago:

https://blog.ipspace.net/2022/09/interfaces-ports/ (it might be worth reading all three blog posts in the series)

netlab allows you to set VLAN forwarding mode to **route** on an access link, but turns that into a regular interface _because it is_ and because that's how we have to configure most devices. All the VLAN attributes are removed from such interfaces _because they are irrelevant_ -- the same interface could be modeled as a regular link and nothing would change.

The routed VLANs are useful to create router-on-a-stick scenarios (where the VLAN mode would be 'bridge' or 'irb' but a device would have the VLAN mode set to **route** on a trunk link) or a set of independent subnets over a shared VLAN trunk in a VRF-lite scenario.

Yes, you can create things like the one above, and _netlab_ won't complain, but why do you keep doing it?

> However, setting the link_name uncovers other issues - e.g. the vrf-leak link in bgp-vrf-local-as looses its neighbors, as the link is not flagged as a "routed_link":

Setting the **link_name** probably triggers all sorts of crazy things because the **link_name** should not be set?

jbemmel:
This is not "me doing it", this is a test case from CI/CD: https://github.com/ipspace/netlab/blob/3bcbbf325c0da549a9e9ec131327b123d4db0c7b/tests/topology/input/rt-vlan-mode-link-route.yml#L64

I was working on interface level attribute changes, and discovered that the above case first sets the ipv4/ipv6 flags to ```False``` and then next converts the link to a routed link (which it is). The reason this happens, is because the logic creates a vlan loopback interface for the 'red' vlan as it deems the VLAN "not used on any physical interface"

I agree this is equivalent to a regular router interface, so therefore it shouldn't trigger the loopback vlan processing code. I can create a test case that reproduces the issue more succinctly for your review

ipspace:
> I agree this is equivalent to a regular router interface, so therefore it shouldn't trigger the loopback vlan processing code. I can create a test case that reproduces the issue more succinctly for your review

I agree with that. What's the impact of that?

jbemmel:
After the recent changes I can no longer reproduce what I saw before - closing

Issue #1457, opened 2024-10-28
[BUG] duplicate IP assigned on "rt" in rt-vlan-mode-link-route.yml


jbemmel:
On vlan ```red``` node ```rt``` is assigned the ip ```172.16.0.2/24``` on 2 different interfaces

https://github.com/ipspace/netlab/blob/352e8ba94664b5a3e92e23c9b40bf3efc6cbde19/tests/topology/expected/rt-vlan-mode-link-route.yml#L660

https://github.com/ipspace/netlab/blob/352e8ba94664b5a3e92e23c9b40bf3efc6cbde19/tests/topology/expected/rt-vlan-mode-link-route.yml#L695

FRR accepts this, but SR Linux does not

It arises from these 2 test cases:
```
defaults.device: frr
module: [ vlan ]

vlans:
  red:
    mode: route           # red VLAN is primarily routed

nodes:
  rt:
  br:
    vlan.mode: bridge     # And set the default VLAN mode to bridge

links:
- rt:
  br:
  vlan.trunk:
    red:                  # This one should be routed on S1 and bridged on S3
  name: Trunk with bridge
#  disable: True
- rt:
  br:                               # replace with a Linux host -> works, 2 different IPs
  vlan.access: red        # This one should be routed on S1 and bridged on S3
  name: Access link to bridge
```

Granted, this is not a common or "sane" topology. Replace ```br``` with a Linux host in the second link and it works; you get 2 different IPs

ipspace:
Here's what I'm guessing is going on: as the VLANs are bridged on br, netlab assigns the same IP prefix to both segments (we could trace where that's done), which one could argue is or is not the right thing to do, but it does make sense.

Then, as the VLANs work in ID-based allocation, the same prefix gets assigned to both interfaces of RT.

If you replace "br" with a Linux host, the two segments are no longer bridged and get different IPs. I would argue netlab does its job better than I would expect.

Finally, can we please stop discussing weird topologies and spend our time doing something more productive?

ipspace:
The one thing we can check though is the attempt to downgrade forwarding modes -- if the VLAN has **mode: route** then no device connected to that VLAN should have VLAN **mode** set to **bridge** or **irb**.

Would this make sense?

jbemmel:
> The one thing we can check though is the attempt to downgrade forwarding modes -- if the VLAN has **mode: route** then no device connected to that VLAN should have VLAN **mode** set to **bridge** or **irb**.
> 
[...]

PR #1459, opened 2024-10-29
Detect duplicate IP address on links and bridge/irb VLANs

### LINKED ISSUE #1351: [BUG] Node interface IP/gateway IP overlap on VLAN links

ipspace:
Closes #1351, makes #1457 impossible

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -1029,6 +1029,34 @@ def set_linkindex(topology: Box) -> None:
     linkindex = linkindex + 1
 
 '''
+check_duplicate_address: Check whether any two nodes on the link got duplicate IP
+'''
+def check_duplicate_address(
+      link: Box,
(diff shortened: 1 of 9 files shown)


jbemmel (approved):
Much improved, fwiw this is productive to me ;)

PR #1460, opened 2024-10-29
Add support for 'clab.dns' (DNS servers and search domain settings)

jbemmel:
See https://containerlab.dev/manual/nodes/#dns

### DIFF
--- a/netsim/utils/templates.py
+++ b/netsim/utils/templates.py
@@ -155,5 +155,8 @@ def load_ansible_filters() -> None:
   add_filters(get_ansible_filter_map('ansible_collections.ansible.netcommon.plugins.filter.ipaddr'))
   add_filters(get_ansible_filter_map('ansible_collections.ansible.netlog.plugins.filter.ipaddr'))
 
+  # Add Ansible core filters like 'to_yaml'
+  add_filters(get_ansible_filter_map('ansible.plugins.filter.core'))
+
   if debug_active('template'):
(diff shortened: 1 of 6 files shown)


ipspace (changes requested):
I think it would be much better to treat **dns** like any other **node_config_attributes** parameter in `clab.j2`

https://github.com/ipspace/netlab/blob/dev/netsim/providers/clab.yml#L6

If the value of an attribute is a dictionary (or a list), the expression generated by Jinja2 is usually a valid JSON expression (needs to be tested, maybe we need `|to_json` filter).

Anyway, it doesn't make sense to keep extending the `clab.j2` template every time we're adding an attribute if we already have a generic mechanism.

Oh, and while we're at it -- **license** should get in there as well.

netsim/utils/templates.py
```diff
@@ -155,5 +155,8 @@ def load_ansible_filters() -> None:
   add_filters(get_ansible_filter_map('ansible_collections.ansible.netcommon.plugins.filter.ipaddr'))
   add_filters(get_ansible_filter_map('ansible_collections.ansible.netlog.plugins.filter.ipaddr'))
 
+  # Add Ansible core filters like 'to_yaml'
+  add_filters(get_ansible_filter_map('ansible.plugins.filter.core'))
```
jbemmel, line 159: Not strictly needed anymore, up to you if you prefer to remove it

ipspace (approved):

PR #1463, opened 2024-10-29
Make container init more robust against failures

jbemmel:
When the init script fails, temporary network namespace links weren't getting cleaned up, and subsequent runs would fail because the link already exists

This PR uses Ansible error handling to always cleanup temporary namespace links


ipspace:
OMG, that ancient hack is still in there :( I should have removed it ages ago, but simply forgot about it :((

A much cleaner solution is in #1464

jbemmel:
Closed due to better solution

PR #1464, opened 2024-10-30
Fix: remove the 'temporary net namespace' hack

ipspace:
Docker containers do not have their network namespaces mapped into the /var/run/netns directory, so we cannot use 'ip netns exec' command with regular containers, and we used a dirty hack to provision Linux container networking.

Containerlab does map the container network namespace into that directory, so we can replace the hack with a much cleaner and more robust solution, using clab container name as the network namespace.

Replaces #1463

### DIFF
--- a/netsim/ansible/tasks/linux/initial-clab.yml
+++ b/netsim/ansible/tasks/linux/initial-clab.yml
@@ -3,17 +3,17 @@
 # the /etc/hosts file is created as a clab bind.
 #
 # The commands to set up Linux networking are executed on the container host in the context
-# of container namespace so we don't need the 'ip' command or Python in the container.
+# of container networking namespace so we don't need the 'ip' command or Python in the container.
 #
 # This task list:
-# 
+#
 # * Creates a random file name that will be used for the initialization script
 # * Creates the networking setup script in the temporary file
-# * Executes a hack that maps Docker container namespace into a well-known namespace
-#   and then executes the networking initialization script within that namespace
-# * Cleans up the mess it made
+# * Executes the setup script within the containerlab container namespace
 #
+---
 - set_fact:
+    # yamllint disable-line rule:line-length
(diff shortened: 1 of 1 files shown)


netsim/ansible/tasks/linux/initial-clab.yml
```diff
@@ -25,13 +25,9 @@
 - name: "Initial container configuration via {{ exec_script }}"
   shell: |
     set -e
-    pid=$(docker inspect -f {% raw %}'{{.State.Pid}}'{% endraw %} clab-{{ netlab_name }}-{{inventory_hostname }})
-    mkdir -p /var/run/netns
-    ln -s /proc/$pid/ns/net /var/run/netns/$pid
-    ip netns exec $pid bash {{ exec_script }}
-    rm /var/run/netns/$pid
+    ip netns exec clab-{{ netlab_name }}-{{ inventory_hostname }} bash {{ exec_script }}
```
jbemmel, line 28: I looked into changing the ```prefix``` in Containerlab to get 'clean' DNS names for the nodes such that name resolution would work the way it does in the physical network, but this "clab-xxx" naming convention is hard-coded in too many places. Hence the PR for DNS support in clab such that I could point at that instead

Still, I think we should set ```netlab_container_name``` or something similar at some right place in the script, and use that everywhere instead (such that we might change it easier in the future)

jbemmel (approved):
Suggestion to cleanup the container name globally, but it will work without that

netsim/ansible/tasks/linux/initial-clab.yml
ipspace: I'm assuming you wanted to get rid of the `clab-` prefix, not changing `netlab-name` (because that should be easy to do with the **name** attribute).

That's way too much work, and we would never be absolutely sure whether we caught all the instances. I don't think it's worth the effort.

Issue #1466, opened 2024-10-30
Cleanup container name references


jbemmel:
I looked into changing the ```prefix``` in Containerlab to get 'clean' DNS names for the nodes such that name resolution would work the way it does in the physical network, but this "clab-xxx" naming convention is hard-coded in too many places. Hence the PR for DNS support in clab such that I could point at that instead

Still, I think we should set ```netlab_container_name``` or something similar at some right place in the script, and use that everywhere instead (such that we might change it easier in the future)

![image](https://github.com/user-attachments/assets/2fea7a56-a4c5-4ad6-a495-2f2ecc6b4325)

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/pull/1464#discussion_r1822254472_

PR #1467, opened 2024-10-30
Cleanup container names

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1466

Usage: ```netlab up 02-vlan-bridge-multiple.yml -p clab -d frr -s defaults.providers.clab.lab_prefix=""```

### DIFF
--- a/netsim/providers/clab.yml
+++ b/netsim/providers/clab.yml
@@ -3,8 +3,7 @@
 #
 description: containerlab with Docker
 config: clab.yml
-lab_prefix: "clab"
-no_propagate: [ lab_prefix ]
+lab_prefix: "clab" # Containerlab default, set to "" to remove prefix
 node_config_attributes: [ type, cmd, env, ports, startup-delay, restart-policy ]
 template: clab.j2
 # Preserve env to allow user to configure PATH
@@ -28,8 +27,6 @@ kmods:
   vxlan: [ vxlan, udp_tunnel, ip6_udp_tunnel ]
   vrf: [ vrf ]
 attributes:
-  global:
-    lab_prefix:     # Prefix string to be added to all node names, defaults to "clab" if not set
   node:
     type: dict
     _keys:          # Make keys explicit to get around the 'type' attribute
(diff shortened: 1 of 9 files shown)


netsim/providers/clab.yml
```diff
@@ -26,6 +28,8 @@ kmods:
   vxlan: [ vxlan, udp_tunnel, ip6_udp_tunnel ]
   vrf: [ vrf ]
 attributes:
+  global:
```
ipspace, line 31: Why are you defining global attribute if you read the value from the system defaults?

ipspace (changes requested):
The global attribute is probably not required, and the documentation is missing.

netsim/providers/clab.yml
jbemmel: I was thinking it would be nice to have type validation including string length limitation, but I didn't finish my thought. Will rework

ipspace: We don't do type validation for default settings

jbemmel: Hence my thought to make it a global attribute instead

ipspace: Then you should change all references to it

jbemmel: It turned out to be more complicated to do a global attribute, and none of the providers currently have any. Decided to stick with the defaults section

Added a paragraph to documentation

ipspace (approved):

jbemmel:
Multilab plugin could check that the lab_prefix is not empty

ipspace:
> Multilab plugin could check that the lab_prefix is not empty

Let me unpack this a bit: you introduced a feature nobody asked for (because most devices get static host maps as part of initial device configuration) and that feature could break something that has been working for years and is used in production. As a cherry on a cake, I told you it's not worth doing.

Considering all that, who do you think should check what?

As I don't think many people will use that nerd knob, I'm perfectly fine with a documented caveat.

jbemmel:
I was only reading your edit to the docs, the multilab plugin is not something I have used so it didn’t come to my mind. It occurred to me that a caveat is great, but not everyone reads the (updated) docs and so that same warning added to the plugin could be helpful in communicating said caveat

The default behavior isn’t changed so this PR doesn’t affect anyone unless they ask for it.

PR #1473, opened 2024-10-31
Promote  iosvl2 to layer 3 switch. Passes [01,61]-vlan* integration tests.

DanPartelly:


ipspace:
Thank you, missed that one.

ipspace (approved):

PR #1474, opened 2024-11-01
Fix: check the VLAN/interface fwd mode compatibility

### LINKED ISSUE #1461: Check VLAN/node/interface forwarding mode compatibility
We should check the attempt to downgrade forwarding modes -- if the VLAN has **mode: route** then no device connected to that VLAN should have VLAN **mode** set to **bridge** or **irb**.

_Originally posted by @ipspace in https://github.com/ipspace/netlab/issues/1457#issuecomment-2443463909_

ipspace:
Connecting a bridge/IRB interface to two segments of a routed VLAN results in a shared IP prefix used on both segments.

This change checks for the forwarding mode mismatch and triggers an error that can be disabled with a defaults setting (because someone might want to use the shared IP prefix)

It turns out several routed VLAN transformation tests relied on this behavior, so they got the VLAN default tweak.

Fixes #1461

### DIFF
--- a/netsim/defaults/hints.yml
+++ b/netsim/defaults/hints.yml
@@ -15,14 +15,21 @@ evpn:
 bgp:
   igp: >
     Add a supported IGP (ospf, isis, eigrp) to the list of modules.
+
 report:
   source: >
     A report can be specified in a file with .j2 suffix within 'reports' subdirectory in
     package-, system-, user- or current directory. You can also specify a report in a
     defaults.outputs.report setting.
+
 quirks:
   junos_lb: >
     Junos devices cannot have more than one loopback interface per routing instance
+
+vlan:
+  fwd_mode_check: >
+    You can disable this error with "defaults.vlan.warnings.mixed_fwd: False", but then
+    you might get the same IP prefix assigned to multiple segments of the routed VLAN.
 
 vrf:
   inactive: >
--- a/netsim/defaults/hints.yml
+++ b/netsim/defaults/hints.yml
@@ -27,8 +27,8 @@ quirks:
     Junos devices cannot have more than one loopback interface per routing instance
 
 vlan:
-  fwd_mode_check: >
-    You can disable this error with "defaults.vlan.warnings.mixed_fwd: False", but then
+  mixed_fwd_check: >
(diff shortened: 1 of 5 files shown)


netsim/defaults/hints.yml
```diff
@@ -15,15 +15,22 @@ evpn:
...
     package-, system-, user- or current directory. You can also specify a report in a
     defaults.outputs.report setting.
+
 quirks:
   junos_lb: >
     Junos devices cannot have more than one loopback interface per routing instance
 
+vlan:
+  fwd_mode_check: >
+    You can disable this error with "defaults.vlan.warnings.mixed_fwd: False", but then
```
jbemmel, line 31: It may be confusing to have a "fwd_mode_check" that can be disabled through "mixed_fwd"

Suggestion: Either use ```mixed_fwd_check``` or ```fwd_mode_check``` for both?

tests/topology/input/rt-vlan-mode-link-route.yml
```diff
@@ -2,6 +2,7 @@
 # of link-, node- and vlan forwarding modes, in particular with mode = route
 #
 
+defaults.vlan.warnings.mixed_fwd: False
```
jbemmel, line 5: Could scope this flag to the vlan for which we want to allow the mixed forwarding, instead of globally?

or introduce a new ```mode``` value that represents this behavior, e.g. ```mode: mixed```

jbemmel (approved):
Consider introducing a new ```mode``` value instead of a global warning, this would allow more accurate modelling of user intent

tests/topology/input/rt-vlan-mode-link-route.yml
ipspace: No. This an extremely niche edge case, and the only reason I didn't just shut it down is because every other blue moon someone might have a good reason to do that. I will not make a hierarchy of nerd knobs out of this ;)

ipspace:
> Consider introducing a new `mode` value instead of a global warning, this would allow more accurate modelling of user intent

How about we condense the user intent into "I want to use _netlab_ the way it was designed" or "I want to do something you guys don't want to hear about and I will not complain about the results" 😇

PR #1479, opened 2024-11-02
Edgeshark support as external tool.

DanPartelly:
- tool template
- initial documentation with caveats


ipspace (approved):
Great job. Thank you!

PR #1484, opened 2024-11-03
Support OS10 configs without virtual networks

jbemmel:
Virtual networks are the more modern option, but they come with limitations. In our network we have Dell OS10 devices deployed with more "traditional" configurations that use (e.g.) OSPF on SVI interfaces, but the current Netlab devices don't support that

Rather than fork the current ```dellos10``` devices, this PR introduces a flag ```use_virtual_networks``` (default: True) that allows users to disable them where needed

One could see this as a bit of a hack - open to less hacky alternatives if any


ipspace:
I understand your personal dilemma, but maybe it's time you discover the infinite extensibility of netlab ;) You can define your own templates anywhere in the netlab search path and do whatever you wish with them.

https://netlab.tools/dev/config/deploy/#configuration-deployment-search-paths

jbemmel:
> I understand your personal dilemma, but maybe it's time you discover the infinite extensibility of netlab ;) You can define your own templates anywhere in the netlab search path and do whatever you wish with them.
> 
> https://netlab.tools/dev/config/deploy/#configuration-deployment-search-paths

But that would only work for me, not for other users also having OS10 devices that need features made impossible by Netlab's current template limitations.

I can easily make it work for me, that's not the issue. I would like it to work out of the box for everyone

ipspace:
> But that would only work for me, not for other users also having OS10 devices that need features made impossible by Netlab's current template limitations.

So point the search paths into a Git repo and publish it on GitHub.

> I can easily make it work for me, that's not the issue. I would like a solution that can work out of the box for everyone

And how many issues/discussions have you seen saying "I need old-style Dell OS10 VLANs"? Stop making your challenges "it should work for everyone" issues.

> Could we include support for alternative templates - checked into GitHub - that can be enabled using some flag?

Of course, it's already there. The "some flag" is called "template search path"

jbemmel:
> > I can easily make it work for me, that's not the issue. I would like a solution that can work out of the box for everyone
> 
> And how many issues/discussions have you seen saying "I need old-style Dell OS10 VLANs"? Stop making your challenges "it should work for everyone" issues.
> 

"Everyone" including (at least) my colleagues working on the same networks and devices

Could it be that we don't see those issues being raised because people try it, it doesn't work for what they need to do, and they move on to some other project?

ipspace:
> "Everyone" including (at least) my colleagues working on the same networks and devices

I'm pretty sure you'll find a way to share the extra templates with them.

> Could it be that we don't see those issues being raised because people try it, it doesn't work for what they need to do, and they move on to some other project?

And why do we see issues for other devices and unusual Linux distros? Please, stop.

jbemmel:
Closing in anticipation of #1501

PR #1485, opened 2024-11-03
Update iosxr-libvirt.yml

jbemmel:
Another case of no change


ipspace (approved):
Did you test the change, because I sure won't waste my time on it ;) Some things are not worth fixing.

PR #1488, opened 2024-11-03
Another example of device support being checked later than expected

jbemmel:
Being pedantic here, feel free to ignore or de-prioritize - but here's another example of the EVPN module producing an error while the "real" error is lack of device support

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -186,7 +186,7 @@ def register_static_transit_vni(topology: Box) -> None:
       continue
 
     for vrf_name,vrf_data in n.vrfs.items():
-      if vrf_data.get('evpn.transit_vni',None):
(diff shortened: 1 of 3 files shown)


ipspace:
So, using your test case, I got

```
[ERRORS]  Errors found in test.yml
[VALUE]   evpn: VRF test in node n1 cannot have evpn.bundle attribute
[HINT]    evpn.bundle attribute can be used only in global VRF definition
```

What's your point?

jbemmel:
The ```vmx``` device doesn't support the EVPN module, that should be the error

ipspace:
> The `vmx` device doesn't support the EVPN module, that should be the error

Theoretically, you're correct. In practice, you get a valid error message, and if you find an EVPN attribute that can be applied to a node VRF, you'll probably get the expected error message as well.

PR #1489, opened 2024-11-03
Fix order of vrf normalization

jbemmel:
If a topology contains only node local vrfs, normalization is skipped

### DIFF
--- a/netsim/modules/vrf.py
+++ b/netsim/modules/vrf.py
@@ -416,12 +416,12 @@ class VRF(_Module):
 
     init_vrf_static_ids(topology)
 
+    normalize_vrf_ids(topology)                         # Normalize global and node vrfs, if any
     if not 'vrfs' in topology:                          # No global VRFs, nothing to do
       return
 
     create_vrf_links(topology)                          # Create VRF links (and remove 'links' attribute)
     log.exit_on_error()
-    normalize_vrf_ids(topology)
     populate_vrf_static_ids(topology)
     set_vrf_ids(topology,topology)
     set_import_export_rt(topology,topology)
--- a/tests/topology/expected/null-vrfs.yml
+++ b/tests/topology/expected/null-vrfs.yml
@@ -1,3 +1,21 @@
+bgp:
+  advertise_loopback: true
+  as: 65000
+  community:
+    ebgp:
+    - standard
+    ibgp:
+    - standard
+    - extended
+  next_hop_self: true
+evpn:
+  session:
+  - ibgp
+groups:
+  as65000:
+    members:
+    - n1
+    - n2
 input:
 - topology/input/null-vrfs.yml
 - package:topology-defaults.yml
@@ -18,13 +36,42 @@ links:
     ipv4: 10.1.0.0/30
   type: p2p
 module:
+- bgp
 - vrf
+- evpn
 name: input
 nodes:
   n1:
     af:
       ipv4: true
       vpnv4: true
+    bgp:
+      advertise_loopback: true
+      as: 65000
+      community:
+        ebgp:
+        - standard
+        - large
+        ibgp:
+        - standard
+        - large
+        - extended
+        ibgp_localas:
+        - standard
+        - large
+        - extended
+      ipv4: true
+      neighbors:
+      - _source_ifname: lo
+        activate:
+          ipv4: true
+        as: 65000
+        evpn: true
+        ipv4: 10.0.0.2
+        name: n2
+        type: ibgp
+      next_hop_self: true
+      router_id: 10.0.0.1
     box: quay.io/frrouting/frr:10.0.1
     clab:
       binds:
@@ -35,6 +82,9 @@ nodes:
       - hosts:/etc/hosts
       kind: linux
     device: frr
+    evpn:
+      session:
+      - ibgp
     hostname: clab-input-n1
     id: 1
     interfaces:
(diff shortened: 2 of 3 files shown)


ipspace:
Yeah, this probably makes sense. Looking at the code, I can't see any drawbacks. What do the tests say?

jbemmel:
Tests passed locally but failed on CI/CD, maybe another exception being thrown at the wrong moment ;)

I’ll investigate and push corrections

ipspace:
Before merging this: please help me understand what the problem is. I took your sample topology (because there was no associated bug report) and it generated the same results with or without your patch.

OTOH, the FRR initial configuration crashed in both cases, so you did uncover an interesting behavior, but I don't see how your fix helps.

ipspace:
Also, your test case succeeds without your fix, which makes it worse than useless.

jbemmel:
> Also, your test case succeeds without your fix, which makes it worse than useless.

```
loading netsim.modules.vrf...
... found <class 'netsim.modules.vrf.VRF'> 
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jeroen/Projects/netlab/netsim/cli/create.py", line 81, in run
    augment.main.transform(topology)
  File "/home/jeroen/Projects/netlab/netsim/augment/main.py", line 109, in transform
    transform_data(topology)
  File "/home/jeroen/Projects/netlab/netsim/augment/main.py", line 66, in transform_data
    modules.pre_transform(topology)
  File "/home/jeroen/Projects/netlab/netsim/modules/__init__.py", line 77, in pre_transform
    module_transform("pre_transform",topology)
  File "/home/jeroen/Projects/netlab/netsim/modules/__init__.py", line 462, in module_transform
    mod_load[m].call("module_"+method,topology)
  File "/home/jeroen/Projects/netlab/netsim/utils/callback.py", line 42, in call
    return method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jeroen/Projects/netlab/netsim/modules/evpn.py", line 401, in module_pre_transform
    register_static_transit_vni(topology)
  File "/home/jeroen/Projects/netlab/netsim/modules/evpn.py", line 189, in register_static_transit_vni
    if vrf_data.get('evpn.transit_vni',None):
       ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```
I simplified the test case down to the minimum, eliminating the problem I saw. It's in the ```evpn``` module

ipspace:
> I simplified the test case down to the minimum, eliminating the problem I saw. It's in the `evpn` module

Oh, and what's the value of a test case that is so simplified that it does not generate the error you saw? What are you testing with it?

Looks like we have to go back to the basics:

* Identify something that looks like a bug
* Document the bug with a topology that reproduces it (not the one that is "simplified down to the minimum" so it does not reproduce stuff) as a GitHub issue
* If you feel like it, fix the bug and submit a PR. If needed, make the minimum topology that **reproduces the bug** a new test case (there's a reason a certain methodology is called "regression testing"), or integrate the elements that trigger the bug into another test case (if you're absolutely sure your won't remove some other relevant code path).
* As we both know, a proposed fix might not be the best way to solve the problem.

In this particular case, EVPN module crashes on dirty data, but I would never have found that out if I accepted your PR at face value. That's why we'll start from "open an issue with a topology that triggers the bug" from now on.

It might also be worth asking "as it looks like we're not expecting dirty data in the EVPN module, what else might go wrong" and (also) fix the EVPN module, or move the processing into a later stage.

Finally, your test case uses FRR, but if you try to generate a configuration out of it, the Ansible playbook crashes. Effectively, it looks from where I'm sitting that you proposed to fix the bare minimum to get your stuff working, potentially masking other problems.

I will remove the useless test case and merge the PR, and when I'll have time I'll fix the other related bugs.

jbemmel:
> > I simplified the test case down to the minimum, eliminating the problem I saw. It's in the `evpn` module
> 
> Oh, and what's the value of a test case that is so simplified that it does not generate the error you saw? What are you testing with it?
> 
> Looks like we have to go back to the basics:
> 
> * Identify something that looks like a bug
> * Document the bug with a topology that reproduces it (not the one that is "simplified down to the minimum" so it does not reproduce stuff) as a GitHub issue
> * If you feel like it, fix the bug and submit a PR. If needed, make the minimum topology that **reproduces the bug** a new test case (there's a reason a certain methodology is called "regression testing"), or integrate the elements that trigger the bug into another test case (if you're absolutely sure your won't remove some other relevant code path).
> * As we both know, a proposed fix might not be the best way to solve the problem.
> 
> In this particular case, EVPN module crashes on dirty data, but I would never have found that out if I accepted your PR at face value. That's why we'll start from "open an issue with a topology that triggers the bug" from now on.
> 
> It might also be worth asking "as it looks like we're not expecting dirty data in the EVPN module, what else might go wrong" and (also) fix the EVPN module, or move the processing into a later stage.
> 
> Finally, your test case uses FRR, but if you try to generate a configuration out of it, the Ansible playbook crashes. Effectively, it looks from where I'm sitting that you proposed to fix the bare minimum to get your stuff working, potentially masking other problems.
> 
> I will merge the PR, and when I'll have time I'll fix the other related bugs.

It wasn't like this at all - I hit the issue in https://github.com/ipspace/netlab/pull/1488/files, solved it "quick and dirty" by changing the evpn module, but I knew you'd get on my case for not thinking about the root cause, so I investigated deeper and found the issue with the normalization.

I simplified the test case too much, that was an error in my attempt to make things easier for you to review/reproduce. Of course the test case has to reproduce the bug.

FWIW I'm not working on anything related to EVPN these days, it was just something I found while playing around.

PR #1490, opened 2024-11-04
[FYI] Script to run template generation against topology test cases

jbemmel:
Not to be merged as-is, but wanted to share some interesting issues:

* "Fatal error in netlab: libvirt is not installed" - not as fatal as it is made out to be, exit code is 0 (and libvirt isn't needed to generate output files only)
* 14 total cases of "failed=1"
* 8 cases of "Executing Ansible playbook failed"

```
Creating configs based on topology/input/unmanaged-device.yml...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jeroen/Projects/netlab/netsim/cli/create.py", line 93, in run
    output_module.write(topology)
  File "/home/jeroen/Projects/netlab/netsim/outputs/provider.py", line 35, in write
    providers.mark_providers(topology)
  File "/home/jeroen/Projects/netlab/netsim/providers/__init__.py", line 333, in mark_providers
    l.provider[node.provider] = True
    ~~~~~~~~~~^^^^^^^^^^^^^^^
  File "box/box.py", line 667, in box.box.Box.__setitem__
  File "box/box.py", line 386, in box.box.Box.__hash__
box.exceptions.BoxTypeError: unhashable type: "Box"
```


ipspace:
Nice idea.

However, those topologies were never meant to be used to generate configurations, and if we use "device: none" just so we don't change test results when a particular device changes image name, you cannot expect the configurations to work correctly, can you?

As for "Fatal error in netlab: libvirt is not installed", did you investigate which command displays that? Also, how do you know the exit code is zero?

Anyway, the crash is worth investigating.

jbemmel:
Someone else found and fixed that one crash: https://github.com/ipspace/netlab/pull/1552

PR #1497, opened 2024-11-06
OS10: Add a check for anycast gateways only being supported on SVI interfaces, using new report_quirk

jbemmel:


ipspace:
This check could use the new "report_quirk" API.

jbemmel:
> This check could use the new "report_quirk" API.

Implemented

ipspace (approved):
Nicely done

PR #1498, opened 2024-11-06
VRRP version plugin for gateway module

jbemmel:
Provides control over the VRRP protocol version to run, specifically by downgrading to (obsolete) v2


ipspace (changes requested):
Looks good to me. The only missing bit seems to be a link from "docs/plugin.md" to the plugin documentation or it won't be displayed.

ipspace (approved):

PR #1506, opened 2024-11-07
Fix: Check for unnumbered/LLA loopback addresses

ipspace:
This fix implements additional checks on main loopback IP addresses:
it cannot be true/false (unnumbered/LLA loopbacks make no sense) and
it cannot be int (node ID is used in that case).

Prevents https://github.com/ipspace/netlab/pull/1495 for good

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -335,6 +335,44 @@ def augment_node_device_data(n: Box, defaults: Box) -> None:
     n.pop('_daemon_config',None)
 
 '''
+Set addresses of the main loopback interface
+'''
+def loopback_interface(n: Box, pools: Box, topology: Box) -> None:
+  n.loopback.type = 'loopback'
+  n.loopback.neighbors = []
+  n.loopback.virtual_interface = True
+  lbname = devices.get_loopback_name(n,topology)
+  if lbname:
+    n.loopback.ifname = lbname
+    n.loopback.ifindex = 0
+
+  pool = n.get('loopback.pool','loopback')
+  prefix_list = addressing.get(pools,[ pool ],n.id)
+
+  for af in prefix_list:
+    if prefix_list[af] is True:
+      log.error(
+        f"Address pool {pool} cannot contain unnumbered/LLA addresses",
+        category=log.IncorrectType,
+        module='nodes')
+    elif not n.loopback[af] and not (prefix_list[af] is False):
+      if af == 'ipv6':
+        if prefix_list[af].prefixlen == 128:
+          n.loopback[af] = str(prefix_list[af])
+        else:
+          n.loopback[af] = addressing.get_addr_mask(prefix_list[af],1)
+      else:
+        n.loopback[af] = str(prefix_list[af])
+      n.af[af] = True
+
+  for af in log.AF_LIST:
+    if af in n.loopback and not isinstance(n.loopback[af],str):
+      log.error(
+        f'{af} address on the main loopback interface of node {n.name} must be a CIDR prefix',
+        category=log.IncorrectType,
+        module='nodes')  
+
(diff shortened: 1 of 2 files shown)


netsim/augment/nodes.py
```diff
@@ -334,6 +334,44 @@ def augment_node_device_data(n: Box, defaults: Box) -> None:
...
+Set addresses of the main loopback interface
+'''
+def loopback_interface(n: Box, pools: Box, topology: Box) -> None:
+  n.loopback.type = 'loopback'
+  n.loopback.neighbors = []
+  n.loopback.virtual_interface = True
+  lbname = devices.get_loopback_name(n,topology)
+  if lbname:
+    n.loopback.ifname = lbname
+    n.loopback.ifindex = 0
```
jbemmel, line 347: Shouldn't the loopback have ifindex=0 regardless whether it has a name?
Code elsewhere depends on each interface having an ```ifindex```

netsim/outputs/ansible.py
```diff
@@ -81,7 +81,7 @@ def get_host_addresses(topology: Box) -> Box:
     if 'loopback' in node:                                  # Create a list of all usable interfaces
       intf_list = [ node.loopback ] + node.interfaces       # ... starting with loopback
```
jbemmel, line 82: Would be a bit of work to check the templates, but we could consider to include node.loopback in node.interfaces (just like secondary loopbacks)

netsim/augment/nodes.py
```diff
@@ -334,6 +334,44 @@ def augment_node_device_data(n: Box, defaults: Box) -> None:
...
+      log.error(
+        f"Address pool {pool} cannot contain unnumbered/LLA addresses",
+        category=log.IncorrectType,
+        module='nodes')
+    elif not n.loopback[af] and not (prefix_list[af] is False):
+      if af == 'ipv6':
+        if prefix_list[af].prefixlen == 128:
+          n.loopback[af] = str(prefix_list[af])
+        else:
+          n.loopback[af] = addressing.get_addr_mask(prefix_list[af],1)
```
jbemmel, line 363: There was a little hack I had to make in order to get around a loopback issue: https://github.com/ipspace/netlab/blob/bccc52f0d250c0e4ecb8f16ca4b77732ac5296b5/netsim/ansible/templates/initial/sros.j2#L41

SR OS requires a /128 on its primary loopback. For consistency, it would be better if there were a device property expressing any loopback prefix lengths restrictions

jbemmel (approved):
Up to you if you want to include addressing loopback prefix restrictions on some (one?) platform

netsim/augment/nodes.py
ipspace: That sounds like a device quirk to me ;)

netsim/augment/nodes.py
ipspace: The loopback ifindex is never checked in the transformation code (and a physical interface could have ifindex 0 on some platforms), the **ifname** and **ifindex** are there just to make templates that use **netlab_interfaces** (loopback + physical intf) work.

netsim/augment/nodes.py
ipspace: What's wrong with setting **prefix6** in **addressing.loopback** (https://netlab.tools/addressing/#specifying-address-pools) or its corresponding system default?

jbemmel: That didn't exist at the time I wrote that template - it was only introduced on October 16th 2023 https://github.com/ipspace/netlab/commit/4a6fd95c8e0c43bab9fc0c31d8be27d38798f114

ipspace: 🤷‍♂️

PR #1507, opened 2024-11-07
Fix: Node VRFs should not have any EVPN attributes

ipspace:
This commit simplifies the EVPN node VRF checks: the node VRFs should not have any EVPN attributes (at the moment).

Inspired by #1489

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -148,20 +148,21 @@ def create_vlan_service(vname: str, topology: Box) -> None:
     vlan_aware_bundle_service(vlan,vname,topology)
 
 """
-Validate that there's no evpn.bundle setting on node VRFs
+Validate that there are no evpn parameters on node VRFs
 """
 
-def validate_no_node_vrf_bundle(node: Box, topology: Box) -> None:
-  if not 'vrfs' in node:                                            # No VRFs defined in the node, move on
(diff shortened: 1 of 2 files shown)


netsim/modules/evpn.py
```diff
@@ -407,7 +397,7 @@ def module_pre_transform(self, topology: Box) -> None:
       module='evpn')
 
   def node_pre_transform(self, node: Box, topology: Box) -> None:
-    validate_no_node_vrf_bundle(node,topology)
+    validate_no_node_vrf_attributes(node,topology)
```
jbemmel, line 400: Did group data processing happen already before pre_transform? Any reason why we cannot move this to node_post_transform? We're ignoring any transit VNIs specified there anyway, the warning is merely a courtesy to the user no?

jbemmel (approved):
May be better to postpone to post_transform if possible, but it should work like this too

netsim/modules/evpn.py
ipspace: The per-node EVPN data makes no sense -- bundles mode and transit VNI should be in sync across all nodes in the same VRF.

This check prevents per-node evpn attributes (and resulting troubleshooting) and has to be done before the global VRF values are merged with the node VRF values.

PR #1508, opened 2024-11-07
Make quirk errors configurable

### LINKED ISSUE #1501: Make quirk errors configurable
Based on discussions in #1494, it became evident we have to have a mechanism to disable individual quirks, and to tell the user how to do that.

Will do a sample implementation for the OS10 SVI interfaces, and fix other quirks once we have a stable API.

_Source: https://github.com/ipspace/netlab/issues/1494#issuecomment-2459767631_

ipspace:
The generic report_quirk function provides an abstraction on top of log.error that checks whether the specified quirk should be reported, giving the user means of turning off device-specific warnings.

Implemented for OSPF on DellOS10 SVI as a proof-of-concept

Fixes #1501

### DIFF
--- a/netsim/devices/dellos10.py
+++ b/netsim/devices/dellos10.py
@@ -3,11 +3,13 @@
 #
 from box import Box
 
-from . import _Quirks,need_ansible_collection
+from . import _Quirks,need_ansible_collection,report_quirk
 from ..utils import log
 from ..augment import devices
 
-def check_vlan_ospf(name: str, iflist: list, vname: str) -> None:
+def check_vlan_ospf(node: Box, vname: str) -> None:
+  name = node.name
+  iflist = node.get('interfaces',[])
   err_data = []
   for intf in iflist:
     if 'ospf' not in intf or intf.type != 'svi':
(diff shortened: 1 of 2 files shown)


netsim/devices/dellos10.py
```diff
@@ -3,31 +3,33 @@
...
 
-from . import _Quirks,need_ansible_collection
+from . import _Quirks,need_ansible_collection,report_quirk
 from ..utils import log
 from ..augment import devices
 
-def check_vlan_ospf(name: str, iflist: list, vname: str) -> None:
+def check_vlan_ospf(node: Box, vname: str) -> None:
+  name = node.name
+  iflist = node.get('interfaces',[])
```
jbemmel, line 12: This is wrong - it should take into account the vrf, if not 'default'

jbemmel (changes requested):
Need to fix VRF interface list

netsim/devices/dellos10.py
ipspace: Indeed. Thank you!

netsim/devices/dellos10.py
```diff
@@ -3,31 +3,33 @@
...
       more_data=err_data,
-      module='ospf')
+      node=node)
 
 class OS10(_Quirks):
 
   @classmethod
   def device_quirks(self, node: Box, topology: Box) -> None:
-    check_vlan_ospf(node.name,node.get('interfaces',[]),'default')
+    check_vlan_ospf(node,'default')
```
jbemmel, line 30: Perhaps we should make the name of the default VRF a device property, and issue a warning (error?) if it's used in the topology?

As-is, a topology with a VRF called "default" would behave differently on OS10 versus (say) SR OS

jbemmel: image

Looks like I can create a vrf called "default", not sure if this is the base context

Issue #1505, opened 2024-11-07
STP port priority: Refactor to accept values 0..15, multiply by 16 if needed


jbemmel:
The STP module currently multiplies the configured value by 4 for some platforms, to get the same 4-bit value on the wire.
However, this doesn't allow for the full range of values to get utilized, and causes the value on the wire to be different from what the user put.

Instead, platforms that currently do "* 4" should use the value provided by the user, and the Linux platforms should divide that by 4 (rather than e.g. Cumulus multiplying by 4)

ipspace:
You're saying that the wire value of STP port priority is 4 bits, but then claim that you will get 4 bits of you divide 240 by 4. There's something wrong with this math.

As we try to stay away from vendor-specific hacks, the STP port priority should be 0..15 (4 bits) and the device templates should do whatever needs to be done (multiply by 16, not 4).

BTW, Arista EOS got the math right:

```
s1(config-if-Et1)#spanning-tree port-priority ?
  <0-240>  Port priority in increments of 16
```

jbemmel:
Yes, it's 16: https://github.com/ipspace/netlab/blob/bccc52f0d250c0e4ecb8f16ca4b77732ac5296b5/netsim/devices/eos.yml#L84

PR #1511, opened 2024-11-08
Refactor stp port priority

jbemmel:
Tested use case 03 using cEOS and FRR

Fixes https://github.com/ipspace/netlab/issues/1505


ipspace (approved):
It's so nice to see how simple things become with the right data model/abstraction ;)

PR #1514, opened 2024-11-08
Dell OS10: Remove escaping of commas in interface descriptions

jbemmel:
Tested - there is no need, the names look odd with the escaped commas

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -41,7 +41,7 @@ interface {{ l.ifname }}
  mtu {{ l.mtu + 32 }}
 {% endif %}
 {% if l.name is defined %}
- description "{{ l.name | replace(",", "\\\\,") }}{{ " ["+l.role+"]" if l.role is defined else "" }}"
+ description "{{ l.name }}{{ " ["+l.role+"]" if l.role is defined else "" }}"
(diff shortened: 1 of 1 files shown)


ipspace:
@ssasso: You created the original template, so you're the only one who knows why those commas had to be escaped.

ssasso:
On Dell OS10 **10.5.6.4** I still see the OS10 bug when description has commas:

```
r1# configure terminal
r1(config)# interface ethernet 1/1/1
r1(conf-if-eth1/1/1)# description "r1 -> [r2,r3]"
CA_STATUS_ERROR
```

(and the description is not applied)

ssasso:
you can try it yourself with this topology:

```
provider: clab
defaults.devices.dellos10:
  clab:
    image: vrnetlab/dell_ftos10v:10.5.6.4
  group_vars:
    ansible_user: admin
    ansible_ssh_pass: admin

addressing:
  loopback:
    ipv4: 10.255.0.0/24
    prefix: 32
  router_id:
    ipv4: 10.255.0.0/24
    prefix: 32

module: [ ospf ]

nodes:
  r1:
    device: dellos10
  r2:
    device: frr
  r3:
    device: frr

links:
- r1:
  r2:
  r3:
  ospf.area: 0
```

jbemmel:
I see - I use ```10.5.6.5``` and the issue is fixed there. Closing

PR #1519, opened 2024-11-09
Cumulus NVUE vlan support

jbemmel:
Adds support for VLANs, passing most of the integration test cases except for the routed native VLAN ones which are unsupported

* Need to load the ```ebtables``` kernel module
* Includes some groundwork for lag support


netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
```diff
@@ -13,8 +13,8 @@
     service_facts:
     register: svc
     until: svc.ansible_facts.services['nvued.service'].state == 'running'
-    retries: 10
-    delay: 5
+    retries: 5
```
ipspace, line 16: I prefer having smaller delay and more retries. On average, that reduces the waiting time.

netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -21,17 +21,17 @@
...
+{% for l in interfaces|default([]) if l.type in ['lan','p2p','stub','svi','lag'] %}
       {{ l.ifname }}:
         link:
 {%  if l.mtu is defined %}
           mtu: {{ l.mtu }}
 {%  endif %}
           state:
             up: {}
-        type: swp
+        type: {{ 'svi' if l.type=='svi' else 'bond' if l.type=='lag' else 'swp' }}
```
ipspace, line 32: You're mixing LAG changes with PR titled "VLAN support", but whatever.

netsim/devices/cumulus.py
```diff
@@ -15,10 +15,24 @@ def check_ospf_vrf_default(node: Box) -> None:
         category=log.IncorrectType,
         module='quirks')
 
+"""
+Checks whether any node VLAN is in the (changeable) reserved range
+"""
+def check_reserved_vlan_range(node: Box) -> None:
```
ipspace, line 21: Why is this change in a Cumulus NVUE VLAN PR?

netsim/devices/cumulus.py
```diff
@@ -15,10 +15,24 @@ def check_ospf_vrf_default(node: Box) -> None:
...
+  for vname,vdata in node.get('vlans',{}).items():
+    vid = vdata.get('id',0)
+    if vid>=3725 and vid <=3999:
+      log.error(
```
ipspace, line 25: And why are you not using the new report_quirk API?

netsim/devices/cumulus.py
```diff
@@ -15,10 +15,24 @@ def check_ospf_vrf_default(node: Box) -> None:
         category=log.IncorrectType,
         module='quirks')
 
+"""
+Checks whether any node VLAN is in the (changeable) reserved range
+"""
+def check_reserved_vlan_range(node: Box) -> None:
+  for vname,vdata in node.get('vlans',{}).items():
+    vid = vdata.get('id',0)
+    if vid>=3725 and vid <=3999:
```
ipspace, line 24: Where would Cumulus Linux 4.x enforce that limitation? We're configuring VLANs "by hand"

netsim/devices/cumulus_nvue.yml
```diff
@@ -23,14 +23,18 @@ features:
...
     unnumbered: True
+  vlan:
+    model: switch
+    svi_interface_name: "vlan{vlan}"
+    subif_name: "{ifname}.{vlan.access_id}"
   vrf: True
 clab:
   mtu: 1500
-  kmods:
+  kmods: [ ebtables ]
```
ipspace, line 33: I don't think this does what you think it will do ;)

ipspace (changes requested):
The Cumulus quirk is probably a mistake, and you should check (and fix if necessary) the kmods parameter. I did not go through the NVUE VLAN template.

netsim/devices/cumulus_nvue.yml
jbemmel: It didn't, but it will once we fix it: https://github.com/ipspace/netlab/pull/1520

netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
jbemmel: you got it ;)

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: just letting you know what to expect next...removed

netsim/devices/cumulus.py
jbemmel: Removed for now, may revisit (more generally?) in the future

netsim/devices/cumulus.py
jbemmel: n/a

ipspace (approved):

PR #1522, opened 2024-11-11
Refactor Linux bridge function to configure LACP forwarding for Libvirt too

jbemmel:
Using Libvirt forwarding for LACP frames is not enabled for the bridges interconnecting the nodes. This PR removes some duplicate code, and consolidates proper bridge forwarding settings in a utils package

### DIFF
--- a/netsim/providers/clab.py
+++ b/netsim/providers/clab.py
@@ -8,7 +8,7 @@ import pathlib
 import argparse
 
 from . import _Provider,get_forwarded_ports
-from ..utils import log, strings
+from ..utils import log, strings, linuxbridge
 from ..data import filemaps, get_empty_box, append_to_list
 from ..data.types import must_be_dict
 from ..cli import is_dry_run,external_commands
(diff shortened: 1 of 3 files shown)


jbemmel:
In the clan case Netlab creates the bridges and they start with STP off (I checked - ip link add name testbr type bridge). In libvirt case they get created for us

I did not fully test LACP on libvirt yet, but I believe we need this setting (and in any case it’s better to be consistent as you say)

ipspace:
> In the clab case Netlab creates the bridges and they start with STP off

👍 Thank you!

> I did not fully test LACP on libvirt yet, but I believe we need this setting 

Oh, this sounds so comforting. You do realize we need Linux bridges only for multi-access links (where LACP wouldn't work anyway) and VM-to-container links?

> and in any case it’s better to be consistent as you say

That may be the case. Please rebase this PR and I'll merge it (believing it works)

ipspace (approved):

PR #1529, opened 2024-11-13
VLAN and VRF groups

ipspace:
This PR generalizes "node groups" into "object groups" and adds VLAN and VRF groups.
The new group types are identified with the **type** attribute, whereas a group with no
**type** attribute is assumed to be a node group.

### DIFF
--- a/netsim/augment/groups.py
+++ b/netsim/augment/groups.py
@@ -21,7 +21,7 @@ from . import nodes
 '''
 Return members of the specified group. Recurse through child groups if needed
 '''
-def group_members(topology: Box, group: str, count: int = 0) -> list:
+def group_members(topology: Box, group: str, grp_type: str = 'node', count: int = 0) -> list:
   members: typing.List[str] = []
   if not group in topology.groups:  # pragma: no cover (just-in case catch, impossible to get here)
     log.error(
@@ -36,11 +36,16 @@ def group_members(topology: Box, group: str, count: int = 0) -> list:
       module='groups',
       header=True)
 
-  for m in topology.groups[group].members:
-    if m in topology.nodes:
+  gdata = topology.groups[group]
+  if gdata.get('type','node') != grp_type:
+    return []
+
+  for m in gdata.members:
+    m_ns = grp_type + 's'
+    if m in topology[m_ns]:
       members = members + [ m ]
     if m in topology.groups:
-      members = members + group_members(topology,m,count + 1)
+      members = members + group_members(topology,m,grp_type,count + 1)
 
   return members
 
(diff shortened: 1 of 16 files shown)


netsim/augment/groups.py
```diff
@@ -128,6 +132,19 @@ def check_group_data_structure(
       continue
 
     gpath=f'{gpath}.{grp}'
+    g_type = gdata.get('type','node')
+    gt_values = ['node','vlan','vrf']
```
jbemmel, line 136: Why can this not be defined in YAML using ```valid_values```?

netsim/augment/groups.py
```diff
@@ -264,11 +281,16 @@ def add_node_level_groups(topology: Box) -> None:
     must_be_list(n,'group',f'nodes.{name}')
 
     for grpname in n.group:
-      if not grpname in topology.groups:
-        topology.groups[grpname] = { 'members': [] }      # Create an empty new group if needed
+      # Sanity check for node groups
+      if topology.groups[grpname].get('type','node') != 'node':
```
jbemmel, line 285: No need to check if ```grpname``` exists?

jbemmel (commented):
Need to clarify and test the semantics of instance attribute override: If a vlan has ospf.cost=X and gets put in a group with ospf.cost=Y, what happens?

netsim/augment/groups.py
ipspace: Because we select the data type to validate against based on g_type in https://github.com/ipspace/netlab/pull/1529/files/060574491a548f534c1c18f76bff102e4098a366#diff-3ef4088ee26c6ab1b27ac441ecf2cb14b48299ee8414cf65f84efccb391e2beeR144

netsim/defaults/attributes.yml
```diff
@@ -136,6 +138,22 @@ group:
...
+  _namespace: [ vlan, link ]
+  type: str
+  members:
+    type: list
+    _subtype: id
+
+vrf_group:
+  _description: VRF groups
+  _namespace: [ vrf ]
+  type: str
```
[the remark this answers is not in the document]
ipspace: Doesn't matter, the data type (vrf_group) is selected based on type value (vrf), so we know what the value should be. The **type** key is there just to pass the validation.

tests/topology/input/groups-vlan-vrf.yml
```diff
@@ -0,0 +1,37 @@
+---
+module: [ vlan, vrf, ospf ]
+
+groups:
+  _auto_create: True
+  ce_vlan:
+    members: [ red_vlan ]
```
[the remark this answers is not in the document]
ipspace: Good point, will do.

ipspace:
> Need to clarify and test the semantics of instance attribute override: If a vlan has ospf.cost=X and gets put in a group with ospf.cost=Y, what happens?

This is no different than any other group, explained in https://github.com/ipspace/netlab/pull/1529/files#diff-ce3fd91438d1613c389915da67500215790d50ca4629d051ed66bc2bb87d3fbfR81

I do agree it would be nice to test multiple members and attribute inheritance though. Will add.

netsim/augment/groups.py
ipspace: No, that's a way to create a group (for example, to have one in Ansible inventory) straight from node data.

DanPartelly:
Could this mechanism be extended with plugins to classify links as underlay/overlay so for example you could generate for a dmvpn all required tunnels and nhrp maps ?

jbemmel:
> Could this mechanism be extended with plugins to classify links as underlay/overlay so for example you could generate for a dmvpn all required tunnels and nhrp maps ?

We have ```role``` as implied link groups (e.g. used for consistent addressing), but I agree explicit ```link``` type groups could make sense

ipspace:
> We have `role` as implied link groups (e.g. used for consistent addressing)

That's exactly what I wanted to suggest to identify (for example) underlay links.

> but I agree explicit `link` type groups could make sense

We had them for quite a while: https://netlab.tools/links/#link-groups

Will add a pointer to them into the groups.md documentation

tests/topology/expected/groups-vlan-vrf.yml
```diff
@@ -501,30 +686,32 @@ ospf:
...
+    mode: route
+    ospf:
+      cost: 1
+    prefix:
+      allocation: id_based
+      ipv4: 172.16.1.0/24
+    vrf: red_vrf
   red_vlan:
-    host_count: 2
+    host_count: 0
```
jbemmel, line 699: blue_vlan is missing ```host_count: 2``` (!)

jbemmel (changes requested):
Need to investigate why the blue_vlan created implicitly through the group is missing the ```host_count``` attribute that the red_vlan had before it became an inter-switch trunk

PR #1531, opened 2024-11-14
Fix typo in bridge name variable

jbemmel:
Sorry, my bad


ipspace:
Thank you. What worries me, though, is that it seems there's been a somewhat suboptimal testing of the original change :(

PR #1535, opened 2024-11-15
Dell OS10: Add lag support

jbemmel:
Adds passive mode test

TODO: Could rename ```lag_interface_name``` to ```features.lag.interface_name``` for consistency with vlan module

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -31,7 +31,7 @@ interface {{ mgmt.ifname|default('mgmt1/1/1') }}
 {% for l in interfaces|default([]) %}
 interface {{ l.ifname }}
  no shutdown
-{%   if l.virtual_interface is not defined %}
+{%   if l.virtual_interface is not defined or l.type=='lag' %}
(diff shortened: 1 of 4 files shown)


ipspace:
> Adds passive mode test

Which seems to be useless -- it succeeds regardless of whether the device under test uses passive mode. To test whether the passive mode works you should have two links, one where the probe (FRR) is active (so we know LAG works) and another one where the probe (FRR) is also passive. If that one works, we have a problem.

> TODO: Could rename `lag_interface_name` to `features.lag.interface_name` for consistency with vlan module

Sounds like a good idea. We should have thought about that sooner 🤷‍♂️

jbemmel:
> > Adds passive mode test
> 
> Which seems to be useless -- it succeeds regardless of whether the device under test uses passive mode. To test whether the passive mode works you should have two links, one where the probe (FRR) is active (so we know LAG works) and another one where the probe (FRR) is also passive. If that one works, we have a problem.
> 
Except that FRR currently doesn’t support ‘’’passive’’’. There is no validation yet, but I manually confirmed that the passive configuration gets applied

PR #1541, opened 2024-11-17
Lag fixes & cleanup

jbemmel:
In preparation for a PR adding MLAG support, factoring out these fixes that are not specific to MLAG:

* Cleanup ```lag``` attributes for member interfaces, only include ```lag._parentindex``` and defer to the parent
* Disallow setting ```lag``` attributes on member interfaces; although some platforms like OS10 could support setting a different LACP speed on different members, other platforms don't support this
* Stop VLAN module from removing the ```virtual_interface``` and ```lag``` attribute from lag links; include regression test case
* Check both L3 and L2-only lag cases

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -31,11 +31,9 @@ interface {{ mgmt.ifname|default('mgmt1/1/1') }}
 {% for l in interfaces|default([]) %}
 interface {{ l.ifname }}
  no shutdown
-{%  if l.virtual_interface is not defined or l.type=='lag' %}
-{%   if l.ipv4 is defined or l.ipv6 is defined %}
+{%   if l.virtual_interface is not defined or l.type=='lag' %}
  no switchport
 {%   endif %}
-{%  endif %}
 {% if l.vrf is defined %}
  ip vrf forwarding {{ l.vrf }}
 {% endif %}


netsim/devices/dellos10.yml
```diff
@@ -23,6 +23,7 @@ features:
   gateway:
     protocol: [ anycast, vrrp ]
   lag:
+    requires: [ vlan ]  # Need to configure 'switchport' before certain (m)lag settings
```
ipspace, line 26: This is wrong. If LAG needs extra configuration on Dell OS10, then the LAG template should provide that configuration, unless you're telling me that you cannot have pure routed LAG on OS10.

netsim/modules/lag.py
```diff
@@ -26,21 +26,27 @@ def populate_lag_id_set(topology: Box) -> None:
 """
 def create_lag_member_links(l: Box, topology: Box) -> None:
   lag_members = l.lag.members
-  l.lag.pop("members",None)                      # Remove explicit list of members
-  l2_ifdata = { 'type': "p2p", 'prefix': False } # Construct an L2 member link
+  l.lag.pop("members",None)                                    # Remove explicit list of members
+  l2_ifdata = data.get_box({ 'type': "p2p", 'prefix': False }) # Construct an L2 member link
```
ipspace, line 30: I can't figure out why you need this change.

netsim/modules/lag.yml
```diff
@@ -26,6 +26,6 @@ attributes:
 
   # Copy only these L2 attributes into LAG physical link members
   lag_l2_ifattr:
-    lag.ifindex:
     mtu:
     bandwidth:
+    stp:
```
ipspace, line 31: This is wrong. STP is not running on individual member links (otherwise one of them would be blocking). STP parameters SHOULD NOT be copied into LAG member links.

netsim/modules/vlan.yml
```diff
@@ -55,13 +55,17 @@ attributes:
...
+    lag:                # Keep lag parameters such as lacp settings
+  
   #
   # Keep these subinterface attributes
   keep_subif:
     vlan:
     ifindex:
     ifname:
     type:
+    virtual_interface:
```
ipspace, line 68: This seems wrong. There is no need to keep the "virtual_interface" setting on a subinterface (which is a virtual interface anyway).

netsim/modules/vlan.yml
```diff
@@ -55,13 +55,17 @@ attributes:
     bandwidth:
     _selfloop_ifindex:
     stp:
+    virtual_interface:  # Use case: VLAN on lag interface
+    lag:                # Keep lag parameters such as lacp settings
```
ipspace, line 59: Are you telling me you never tested LAG with VLAN interfaces before saying "the LAG module is ready"? Otherwise, I agree this change is needed as LAG is below VLANs in the IEEE protocol stack.

ipspace (changes requested):

ssasso:
> @ssasso: asking for your opinion as this PR involves what seems to be dubious OS10 stuff to me -- does OS10 support routed LAG links (physical interfaces in a LAG, with IP address on a port channel interface), or does it really need VLANs to make LAG work?

```
r2# configure terminal
r2(config)# interface port-channel 3
r2(conf-if-po-3)# interface range ethernet 1/1/3-1/1/4
r2(conf-range-eth1/1/3-1/1/4)# channel-group 3 mode active
r2(conf-range-eth1/1/3-1/1/4)# interface port-channel 3
r2(conf-if-po-3)# no switchport
r2(conf-if-po-3)# ip address 10.20.30.2/24
r2(conf-if-po-3)# no shutdown
r2(conf-if-po-3)# end
r2# ping 10.20.30.1
PING 10.20.30.1 (10.20.30.1) 56(84) bytes of data.
64 bytes from 10.20.30.1: icmp_seq=1 ttl=64 time=2027 ms
64 bytes from 10.20.30.1: icmp_seq=2 ttl=64 time=1000 ms
64 bytes from 10.20.30.1: icmp_seq=3 ttl=64 time=1.54 ms
64 bytes from 10.20.30.1: icmp_seq=4 ttl=64 time=1.08 ms
64 bytes from 10.20.30.1: icmp_seq=5 ttl=64 time=1.41 ms
64 bytes from 10.20.30.1: icmp_seq=6 ttl=64 time=1.05 ms

--- 10.20.30.1 ping statistics ---
6 packets transmitted, 6 received, 0% packet loss, time 37ms
rtt min/avg/max/mdev = 1.046/505.445/2027.150/772.161 ms, pipe 2
```

ipspace:
@ssasso: Thank you ;)

@jbemmel: As you can see, there is no need to require **vlan** module for OS10 LAG to work ;))

jbemmel:
> @ssasso: Thank you ;)
> 
> @jbemmel: As you can see, there is no need to require **vlan** module for OS10 LAG to work ;))

How I got there: FRR needs the lag module configured before VLANs, as VLANs are created on top of bond interfaces. However, OS10 needs the proper switchport settings before it allows you to configure certain lag settings, and those are configured here https://github.com/ipspace/netlab/blob/e5f6268db24f70a5d0abaa5564ee5fa30a21068e/netsim/ansible/templates/vlan/dellos10.j2#L17

I now realize a better solution is to have the lag module configure the switchport settings it needs, and remove the ```requires```

netsim/modules/vlan.yml
jbemmel: I did test lags with vlans: ```netlab create integration/lag/02-lag-vlan-trunk.yml -p clab -d frr```

Validation passes on ```dev``` - but I agree that apparently the test wasn't rigorous enough.

When I add this patch, the only thing that changes in that 02 integration test, is that ```vlan_member``` interfaces are no longer marked as ```virtual_interface```. Adding that under ```keep_subif``` fixes that (that is why I added that)

I did try adding ```virtual_interface``` as an interface attribute in ```defaults/attributes.yml``` and setting it in ```create_vlan_member_interfaces``` but that didn't work. I figured it's essentially part of the ```type```, so copying it along with the ```type``` attribute makes some sense

netsim/modules/lag.py
jbemmel: I removed ```lag.index``` from ```lag_l2_ifattr``` because it wasn't behaving the way I expected it to - it copied all the ```lag``` attributes. After removing that, the ```member.lag._parentindex = l.linkindex``` statement throws an error because it's trying to use '.lag' as an attribute

netsim/devices/dellos10.yml
jbemmel: Removed

netsim/modules/lag.yml
jbemmel: removed

netsim/modules/vlan.yml
jbemmel: removed

ipspace (changes requested):
Right now, I see a number of changed test results. I don't know how you got there, but the "virtual_interface" setting must be kept on the VLAN interface and VLAN subinterfaces, or some platforms try to configure "no switchport" on those interfaces and that fails.

Unfortunately I cannot merge this PR until that is fixed. Typically, a LAG PR should not cause a change in any non-LAG test results.

PR #1545, opened 2024-11-19
Cumulus_nvue: basic lag support

jbemmel:
* Enables bonded interfaces with dynamic (LACP) or static modes
* Refactor IP address config as macro; used only in 'initial' script
* Explicitly disable ipv6 when not configured
* Define MTU in lag OSPF test such that it works on libvirt too

Note: Fixes ipv6-only unnumbered config, would have been skipped prior to this PR

### DIFF
--- a/docs/module/lag.md
+++ b/docs/module/lag.md
@@ -9,7 +9,7 @@ LAG is currently supported on these platforms:
 | Operating system      | LACP | Static | Passive<br>LACP |
 | --------------------- |:--:|:--:|:--:|
 | Arista EOS [❗](caveats-eos) | ✅ | ✅ | ✅ |
-| Cumulus Linux         | ✅ | ✅ | ❌  |
+| Cumulus Linux 4.x/5.x | ✅ | ✅ | ❌  |
 | FRR                   | ✅ | ✅ | ❌  |
 
 ## Parameters
--- a/docs/module/lag.md
+++ b/docs/module/lag.md
@@ -9,7 +9,8 @@ LAG is currently supported on these platforms:
 | Operating system      | LACP | Static | Passive<br>LACP |
 | --------------------- |:--:|:--:|:--:|
 | Arista EOS [❗](caveats-eos) | ✅ | ✅ | ✅ |
-| Cumulus Linux 4.x/5.x | ✅ | ✅ | ❌  |
+| Cumulus Linux 4.x     | ✅ | ✅ | ❌  |
+| Cumulus 5.x (NVUE)    | ✅ | ✅ | ❌  |
 | FRR                   | ✅ | ✅ | ❌  |
 
 ## Parameters
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -14,25 +14,24 @@
 {%  if i.ipv4 is defined or i.ipv6 is defined %}
         ip:
           address:
-{%   if i.ipv4 is defined %}
-{%    if i.ipv4 == True %}
+{%    if i.ipv4 is defined %}
+{%      if i.ipv4 == True %}
             {{ loopback.ipv4 }}: {}
-{%    else %}
+{%      else %}
             {{ i.ipv4 }}: {}
+{%      endif %}
 {%    endif %}
-{%   endif %}
-{%     if i.vrf is defined %}
+{%    if i.vrf is defined %}
           vrf: {{ i.vrf }}
-{%  endif %}
-{%  if i.ipv6 is defined %}
-{%   if i.ipv6 == True %}
-{%   else %}
+{%    endif %}
+{%    if i.ipv6 is defined %}
+{%      if i.ipv6 is string %}
             {{ i.ipv6 }}: {}
-{%   endif %}
-{%   else %}
+{%      endif %}
+{%    else %}
           ipv6:
             enable: off
-{%   endif %}
+{%    endif %}
 {%  endif %}
 {% endmacro %}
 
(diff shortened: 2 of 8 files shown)


docs/module/lag.md
```diff
@@ -9,7 +9,7 @@ LAG is currently supported on these platforms:
 | Operating system      | LACP | Static | Passive<br>LACP |
 | --------------------- |:--:|:--:|:--:|
 | Arista EOS [❗](caveats-eos) | ✅ | ✅ | ✅ |
-| Cumulus Linux         | ✅ | ✅ | ❌  |
+| Cumulus Linux 4.x/5.x | ✅ | ✅ | ❌  |
```
ipspace, line 12: As we have two distinct devices and completely different configuration methods (and quite often different functionality), it would be better to keep Cumulus Linux 4.x and Cumulus Linux NVUE apart (and yes, we should be consistent, any suggestions?)

netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -1,3 +1,40 @@
...
+        ip:
+          address:
+{%   if i.ipv4 is defined %}
+{%    if i.ipv4 == True %}
+            {{ loopback.ipv4 }}: {}
+{%    else %}
+            {{ i.ipv4 }}: {}
+{%    endif %}
+{%   endif %}
+{%     if i.vrf is defined %}
```
ipspace, line 24: I prefer having a 2-space indents (makes things easier to read, at least for me), but whatever you do, please be consistent.

netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -1,3 +1,40 @@
...
+          vrf: {{ i.vrf }}
+{%  endif %}
+{%  if i.ipv6 is defined %}
+{%   if i.ipv6 == True %}
```
ipspace, line 28: I would rewrite this into `i.ipv6 is string` or some such

netsim/ansible/templates/lag/cumulus_nvue.j2
```diff
@@ -0,0 +1,26 @@
+{% from "initial/cumulus_nvue.j2" import decl_interface with context %}
+
+{% for i in interfaces if i.type == 'lag' %}
```
ipspace, line 3: Is there a reason the LAG interfaces are defined here and not in the initial template? This might require most modules to be executed after LAG in the config sequence, and right now we don't have that dependency in place (unless the VLAN-LAG dependency forces the config order even if the VLAN module is not used)

netsim/devices/cumulus_nvue.yml
```diff
@@ -1,5 +1,6 @@
 description: Cumulus VX 5.x configured with NVUE
 interface_name: swp{ifindex}
+lag_interface_name: "bond{lag.ifindex}"
```
ipspace, line 3: Didn't we discuss going to **lag.interface_name**?

ipspace (changes requested):
Minor quirks. Did not test the configuration templates, I'm hoping they work...

docs/module/lag.md
jbemmel: Copied the entry from vlan module

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: Would it help to use something like https://www.djlint.com/docs/configuration/?

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: done

netsim/ansible/templates/lag/cumulus_nvue.j2
jbemmel: I was thinking to separate concerns, and not 'pollute' the initial template with logic understanding to configure 'lag' interfaces with type 'bond', etc.

We can move it to initial if you prefer

jbemmel: refactored

docs/module/lag.md
ipspace: I'm not saying that one is correct 🤷‍♂️

netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -19,40 +60,11 @@
...
-          vrf: {{ l.vrf }}
-{%     endif %}
-{%     if l.ipv6 is defined %}
-{%       if l.ipv6 == True %}
-{%       else %}
-            {{ l.ipv6 }}: {}
-{%       endif %}
-{%     endif %}
-{%   endif %}
+{% for l in interfaces|default([]) if l.type in ['lan','p2p','stub','svi','lag'] %}
```
ipspace, line 68: Why do you need to check interface type here? What interfaces do you not want to configure and why?

ipspace (changes requested):
One more minor issue: why do you check the interface type when configuring interfaces? What interfaces do you want to skip?

I'm guessing the PR is good-to-go as it is (assuming the config templates work), but the test of interface types (unless needed for some reason) might bite us in the future.

netsim/ansible/templates/initial/cumulus_nvue.j2
ipspace: Probably. Thank you, will check it out.

netsim/devices/cumulus_nvue.yml
ipspace: Thank you!

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: It’s a historic leftover, “loopback” type comes to mind (though that would be caught elsewhere)

I can submit another PR to add loopback support and remove (test) this check on type

ipspace: Please do. Thank you!

Issue #1546, opened 2024-11-19
[BUG] Libvirt pre_transform hook is too aggressive in setting link.type='lan'


jbemmel:
Encountered in the mlag feature branch
```
provider: clab

groups:
  _auto_create: True
  core:
    members: [ c1,c2 ]
    module: [ lag, vlan ]
    device: dellos10
    provider: clab

  tor:
    members: [ s1 ]
    device: linux
    provider: libvirt  # <- mixed provider causes issues with mlag_peer link

links:
- lag:
   mlag.peergroup: 255
   members: [c1-c2]
```

Without ```provider: libvirt``` this creates a link of type ```mlag_peer``` as intended. However, adding libvirt causes the link type to be set to ```lan```

The issue is here: https://github.com/ipspace/netlab/blob/e5f6268db24f70a5d0abaa5564ee5fa30a21068e/netsim/providers/libvirt.py#L281

This link is not a multi-provider link, both nodes are clab

ipspace:
I think you will find that if you stop using **link.type** for all sorts of things it was never intended to be used for, your life will become much easier ;)

The **mlag_peer** link type idea is plain wrong.

jbemmel:
> 
> I think you will find that if you stop using **link.type** for all sorts of things it was never intended to be used for, your life will become much easier ;)
> 
> The **mlag_peer** link type idea is plain wrong.
>

I forgot about this one: https://github.com/ipspace/netlab/commit/91f6416778a176747b7d8f50a874823e64e0c500
In retrospect, it would be better to avoid creating the bridges earlier in the process, and in the template simply execute what's defined (rather than second-guessing whether a bridge needs to be created or not)

I agree that because of the way link.type is embedded in core logic all across the platform, re-purposing it wasn't a great idea

PR #1542, opened 2024-11-18
MLAG support feature, initially for OS10 and EOS

jbemmel:
Depends on https://github.com/ipspace/netlab/pull/1541 (rebased)

* Add ```lag.mlag.peergroup``` on LAG links to model peerlinks
* Showcases 2 platforms with different approaches to peerlinks, modeling each case (fully configurable)
* Checks that the 'M' side of MLAG supports mlag and is the same device type
* Numbers lag interfaces consistently (same ifindex on all M-sides, starting from 1 on all 1-sides)
* Validated anycast gateway support

This works, on both OS10 and EOS:
```
vlans:
 red:
  gateway.protocol: anycast
links:
- lag: # MLAG peer links between S1 and S2
   members: [s1-s2,s1-s2]
   mlag.peergroup: 1
- lag:
   members: [ h1-s1,h1-s1,h1-s2,h1-s2 ]
   mlag: True
  vlan.access: red
- lag:
   members: [ h2-s1,h2-s1,h2-s2,h2-s2 ]
   mlag: True
  vlan.access: red
```

PR #1548, opened 2024-11-20
Fix multi-provider link logic

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1546

As stated in the ```libvirt-clab-complex``` test case comments:
```
- h1-r1-r2          # multi-access link that should not turn into a P2P tunnel
- r1-r2             # Link that has to be libvirt P2P tunnel
- r2-r3             # P2P link that should be a LAN link
- r3-h2             # clab-only link that does not need a Linux bridge
- h1-r3-h2          # clab-only link that requires a Linux bridge
```

This PR ensures that the h1-r3-h2 case and the r3-h2 case are properly identified as 'clab-only'

In ```links.py``` ```set_link_bridge_name``` there is a further optimization that removes the Linux bridge in the r3-h2 case, but as it has broad impact on many test cases I left it commented out.

Corrects some minor typos and adds some debugging enhancements

Code overlaps with https://github.com/ipspace/netlab/blob/e5f6268db24f70a5d0abaa5564ee5fa30a21068e/netsim/providers/__init__.py#L323 we may want to refactor

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -868,6 +868,8 @@ def set_link_type_role(link: Box, pools: Box, nodes: Box, defaults: Box) -> None
 def set_link_bridge_name(link: Box, defaults: Box) -> None:
   if link.type in ['p2p','loopback','vlan_member']:                   # No need for bridge names on P2P links, loopbacks and virtual links
     return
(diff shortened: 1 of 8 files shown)


jbemmel:
Changed https://github.com/ipspace/netlab/pull/1542 to not overload ```link.type```

Issue #1551, opened 2024-11-21
[BUG] unmanaged device causing exception: box.exceptions.BoxTypeError: unhashable type: "Box"


sdargoeuves:
### Describe the bug

Starting to think how I am going to connect the Cloud to netlab, I started to play with unmanaged devices. But using the example in the [documentation](https://netlab.tools/example/external/#unmanaged-devices), it fails [here](https://github.com/ipspace/netlab/blob/dev/netsim/providers/__init__.py#L333)

### To Reproduce

You can use the example in the doc:

```yaml
defaults.device: eos
module: [ bgp ]
nodes:
  vm:
    bgp.as: 65000
  sw:
    unmanaged: True
    bgp.as: 65001
links:
- vm:
    ipv4: 10.42.0.2/24
  sw:
    ipv4: 10.42.0.1/24
  libvirt.public: True
```

### Output

```shell
  ~/code/netsim-main-lab on   revival_q3_2024 !15 ❯ netlab create                                                                                             netsim-main-lab with sa@sa-netlab01a
Traceback (most recent call last):
  File "/home/sa/code/netsim-main-lab/venv/bin/netlab", line 14, in <module>
    netsim.cli.lab_commands(__file__)
  File "/home/sa/code/netsim-main-lab/venv/lib/python3.11/site-packages/netsim/cli/__init__.py", line 351, in lab_commands
    mod.run(sys.argv[arg_start:])   # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/sa/code/netsim-main-lab/venv/lib/python3.11/site-packages/netsim/cli/create.py", line 93, in run
    output_module.write(topology)
  File "/home/sa/code/netsim-main-lab/venv/lib/python3.11/site-packages/netsim/outputs/provider.py", line 35, in write
    providers.mark_providers(topology)
  File "/home/sa/code/netsim-main-lab/venv/lib/python3.11/site-packages/netsim/providers/__init__.py", line 333, in mark_providers
    l.provider[node.provider] = True
    ~~~~~~~~~~^^^^^^^^^^^^^^^
  File "box/box.py", line 667, in box.box.Box.__setitem__
  File "box/box.py", line 386, in box.box.Box.__hash__
box.exceptions.BoxTypeError: unhashable type: "Box"

```

### Version

* netlab version: `1.9.2` and the current `main`

### Additional context

I've done this as a workaround, it allows the `netlab create` to complete:

```python
[...]

PR #1552, opened 2024-11-21
Error when using unmanaged node

sdargoeuves:
linked to #1551

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -329,9 +329,9 @@ def mark_providers(topology: Box) -> None:
 
   for l in topology.links:                          # Set 'providers' attribute on all links
     for intf in l.interfaces:
-      node = topology.nodes[intf.node]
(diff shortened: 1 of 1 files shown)


netsim/providers/__init__.py
```diff
@@ -330,6 +330,8 @@ def mark_providers(topology: Box) -> None:
   for l in topology.links:                          # Set 'providers' attribute on all links
     for intf in l.interfaces:
       node = topology.nodes[intf.node]
+      if node == {}:                                # take into account unmanaged nodes
```
jbemmel, line 333: This code adds non-existing nodes to the topology.

To avoid this:
```
if intf.node not in topology.nodes:
  continue
```

jbemmel (changes requested):
Need to revise the code to avoid adding non-existent nodes

ipspace (approved):

PR #1556, opened 2024-11-23
Cumulus nvue multiple loopback support

jbemmel:
* Puts additional ipv4/6 loopback addresses on 'lo' interface (only supported device name)
* Fix IPv6 ND/RA
* Remove check on most interface types (except 'loopback' which is still special)
* Moves interface_name under ```features.<module>.interface_name```
* Enables gradual migration to new interface_name layout, adding ```deprecated``` debug flag to call out remaining legacy settings

Tested:
integration/initial/02-loopbacks (Cumulus_nvue and FRR)
integration/lag/01-l3-lag (Cumulus_nvue and FRR)

(PS - if you prefer to move all platforms to the new scheme in 1 PR, we can do that too. I simply cannot test all of them personally)

Note: May need additional fixes for loopbacks in vrfs (didn't test those yet), let me know if you'd like to include those

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -2,9 +2,9 @@
       {{ i.ifname }}:
         type: {{ 'svi' if i.type=='svi' else 'bond' if i.type=='lag' else 'swp' }}
         link:
-{%  if i.mtu is defined %}
+{%    if i.mtu is defined %}
           mtu: {{ i.mtu }}
-{%  endif %}
+{%    endif %}
           state:
             up: {}
 {%  if i.name is defined %}
@@ -14,6 +14,9 @@
 {%  endif %}
 {%  if i.ipv4 is defined or i.ipv6 is defined %}
         ip:
+{%    if i.vrf is defined %}
+          vrf: {{ i.vrf }}
+{%    endif %}
           address:
 {%    if i.ipv4 is defined %}
 {%      if i.ipv4 == True %}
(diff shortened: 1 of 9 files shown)


jbemmel:
NB ```bandwidth``` is not supported - or at least I couldn't find the config setting. Could add a quirk for that

netsim/devices/cumulus_nvue.yml
```diff
@@ -1,6 +1,5 @@
 description: Cumulus VX 5.x configured with NVUE
 interface_name: swp{ifindex}
```
jbemmel, line 2: Candidate for moving under ```features.initial.interface_name```

netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -2,9 +2,9 @@
       {{ i.ifname }}:
         type: {{ 'svi' if i.type=='svi' else 'bond' if i.type=='lag' else 'swp' }}
         link:
-{%  if i.mtu is defined %}
+{%    if i.mtu is defined %}
```
ipspace, line 5: Seems like an indentation error to me

ipspace (changes requested):
Could you please explain why you felt the need to change the way loopback interface names are generated? Instead of a simple change to device settings and configuration template you submitted changes in three core modules and two configuration modules.

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: See https://github.com/ipspace/netlab/pull/1557

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/1557

netsim/ansible/templates/initial/cumulus_nvue.j2
ipspace: I have to admit my understanding of a "ready to merge PR" differs significantly from your "input for discussion" perspective.

This PR claims it implements multiple loopbacks on Cumulus NVUE, so let's focus on that.

PR #1557, opened 2024-11-23
Cumulus_nvue: Add loopback and VRF support, fix IPv6

jbemmel:
* Removes interface ```type``` test
* Puts interfaces in the correct VRF
* Fixes IPv6 to send out router-advertisements ( "off" means enable them ... )
* Updates docs (note I could not find ```bandwidth``` setting in NVUE)

Note that Cumulus supports multiple loopbacks by putting multiple IPs on a single loopback device; ```set``` commands are cumulative and don't replace existing addresses

This also means applying Netlab configs to a system that already has some existing configs won't replace those; this would require an 'unset' command

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -99,9 +99,6 @@
 {%     endif %}
 {%     if 'ipv6' in lb %}
               {{ lb.ipv6 }}: {}
-{%     else %}
-            ipv6:
-              enable: off
 {%     endif %}
 {%   endif %}
 {% endfor %}
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -14,6 +14,9 @@
 {%  endif %}
 {%  if i.ipv4 is defined or i.ipv6 is defined %}
         ip:
+{%    if i.vrf is defined %}
+          vrf: {{ i.vrf }}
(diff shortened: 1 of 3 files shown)


netsim/ansible/templates/initial/cumulus_nvue.j2
```diff
@@ -50,21 +59,49 @@
...
           ipv6:
             enable: off
-{% endif %}
-        type: loopback
-{% for l in interfaces|default([]) if l.type in ['lan','p2p','stub','svi','lag'] %}
-{{      decl_interface(l) }}
+{%     endif %}
+{%   else %}
+    vrf:
+      {{ lb.vrf }}:
```
ipspace, line 91: It looks to me like this logic might incorrectly handle the scenarios where there are multiple loopbacks in a VRF. I know those should be rare, but we still have to do it right (not my fault that CL NVUE has a suboptimal data model ;)

ipspace (changes requested):
Thanks for focusing on the NVUE loopbacks. However, I think we might still have a problem if someone configures more than one loopback per VRF (for example, to generate more prefixes)

netsim/ansible/templates/initial/cumulus_nvue.j2
jbemmel: image

image

Multiple ```set``` commands with different IPs are cumulative. Did find an issue with disabling ipv6 on vrf loopbacks - that's not needed

ipspace (approved):

PR #1561, opened 2024-11-25
Update libvirt-clab-complex.yml

jbemmel:
Fix invalid port forward


ipspace (approved):
Thank you! It would be nice to add a note to the documentation saying "this is an advanced feature for which we're not validating the data model"

Issue #1562, opened 2024-11-25
Undefined behavior for BGP in case of ipv4:True with numbered ipv6


jbemmel:
https://github.com/ipspace/netlab/blob/a1e1076bb4459eb85c75fe23bc13e2a7889a8fb9/tests/topology/expected/bgp-unnumbered-dual-stack.yml#L205

```
# interface 6 - IPv6 AF over IPv6 numbered, no IPv4 session/AF
- test:
  x4:
  prefix:
    ipv4: True
    ipv6: 2001:db8:1::/64
```

As the test case states, no IPv4 session should be created here. However, it does get created

ipspace:
Fixed in #1575

PR #1563, opened 2024-11-25
BGP: Define behavior in case of ipv6 with unnumbered ipv4 as IPv4+6 AF over ipv6 session (only)

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1562

This PR modifies 2 test cases from https://github.com/ipspace/netlab/blob/dev/tests/topology/input/bgp-unnumbered-dual-stack.yml:
```
# interface 2 - activate only IPv4 AF over IPv4 unnumbered session, no IPv6
- test:
  x1:
  prefix:
    ipv4: True
```
In this case, there should be no IPv6 nor RFC8950 behavior

```
# interface 6 - IPv4+IPv6 AF over IPv6 numbered, no IPv4 session
- test:
  x4:
  prefix:
    ipv4: True
    ipv6: 2001:db8:1::/64
```
Previously undefined, now defined as:
1. Enable RFC8950 feature to advertise ipv4 prefixes over an ipv6 session
2. Activate ipv4 and ipv6 address families over ipv6 session
3. Remove ipv4 bgp session
4. Remove ipv4 flag from the interface (as there is no ```_parent_ipv4``` defined)

Documentation updated; note that RFC8950 seems oblivious to numbered or unnumbered IPv6, either should work

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -308,7 +308,7 @@ See the [Simple BGP](bgp_example/simple.md) and [EBGP Data Center Fabric](bgp_ex
 
 Unnumbered EBGP sessions are supported on a few platforms. *netlab* creates an IPv6 LLA EBGP session when the **unnumbered** link- or interface attribute is set, or when **ipv6** interface address or link prefix is set to *True* (IPv6 LLA).
 
-*netlab* can use an IPv6 LLA EBGP session to transport IPv4 address family with IPv6 next hops (RFC 8950) -- the functionality commonly used to implement *unnumbered EBGP sessions*. *netlab* will enable IPv4 AF over IPv6 LLA EBGP session when the **unnumbered** link- or interface attribute is set, or when **ipv4** interface address or link prefix is set to *True*.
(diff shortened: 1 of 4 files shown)


ipspace:
If we go for "no IPv4 session" in #1562, then this PR is correct.

jbemmel:
> If we go for "no IPv4 session" in #1562, then this PR is correct.

Thinking from a user perspective, my vote would be to try and do the ipv4 session in this case. If the user didn't want ipv4, they wouldn't put ```ipv4: True```

ipspace:
I'm sorry I pulled you down this rabbit trail, but things are WAY more complex than they look:

* Enabling IPv4 AF over IPv6 EBGP session usually involves additional configuration, and none of the configuration templates (that I'm aware of) contain that logic as we always checked the "is this an interface EBGP session" flag
* Some devices might support interface EBGP sessions, but not regular RFC8950 sessions (it all has to do with faked IPv4 next hops)
* We might need an extra feature flag to say "device supports RFC8950 over numbered IPv6" as opposed to "device supports interface EBGP sessions", or modify the meaning of existing feature flags. We could go for "ipv6_lla" means "dual-stack interface EBGP" and "rfc8950" means "RFC 8950 over regular IPv6 BGP session"
* We need an extra integration test.

I will shelve this PR for the moment until we figure out what we want to do.

jbemmel:
Replaced by #1569

ipspace:
> Replaced by #1569

Not too bad that I did not merge them in order ;))

PR #1564, opened 2024-11-25
Cumulus_nvue: Use password variable instead of hardcoded string, declare container image unsupported

jbemmel:
* Use variable for cumulus password provisioned into the box
* Declare Cumulus container images unsupported (warning to be added)

### DIFF
--- a/netsim/templates/provider/libvirt/cumulus_nvue-domain.j2
+++ b/netsim/templates/provider/libvirt/cumulus_nvue-domain.j2
@@ -1,9 +1,10 @@
     $cumulus_script = <<-SCRIPT
 
 function set_cumulus_password(){
-    echo "### Setting cumulus user password to 'cumulus' ###"
+{% set _password = defaults.devices.cumulus_nvue.group_vars.ansible_ssh_pass %}
+    echo "### Setting cumulus user password to '{{ _password }}' ###"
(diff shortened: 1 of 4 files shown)


jbemmel:
Ok, but let’s include the password fix which applies to both

jbemmel:
PR modified

netsim/templates/provider/libvirt/cumulus_nvue-domain.j2
```diff
@@ -1,9 +1,10 @@
     $cumulus_script = <<-SCRIPT
 
 function set_cumulus_password(){
-    echo "### Setting cumulus user password to 'cumulus' ###"
+{% set _password = defaults.devices.cumulus_nvue.group_vars.ansible_ssh_pass %}
+    echo "### Setting cumulus user password to '{{ _password }}' ###"
     echo "cumulus ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/10_cumulus
-    echo "cumulus:GetLost1!" | chpasswd
+    echo "cumulus:{{ _password }}" | chpasswd
```
ipspace, line 7: FWIW, this works as long as nobody changes the ansible_ssh_pass variable for a node or a group. Better than hardcoded value though ;)

jbemmel: Agreed, not perfect but better

Issue #1565, opened 2024-11-26
RFC 8950-style IPv4 next hops on regular IPv6 EBGP sessions


ipspace:
#1563 that tries to fix the edge case of "unnumbered IPv4, numbered IPv6" (#1562) forced me to realize how complex this particular can of worms is:

* Enabling IPv4 AF over IPv6 EBGP session usually involves additional configuration, and none of the configuration templates (that I'm aware of) contain that logic as we always checked the "is this an interface EBGP session" flag
* Some devices might support interface EBGP sessions, but not regular RFC8950 sessions (it all has to do with faked IPv4 next hops)
* We might need an extra feature flag to say "device supports RFC8950 over numbered IPv6" as opposed to "device supports interface EBGP sessions", or modify the meaning of existing feature flags. We could go for "ipv6_lla" means "dual-stack interface EBGP" and "rfc8950" means "RFC 8950 over regular IPv6 BGP session"
* We would need an extra integration test.

We have to decide whether to support RFC 8950 over regular IPv6 BGP sessions (and do all the above changes) or whether to say "that does not work, you must use IPv6 LLA sessions" and add a corresponding error to the BGP code.

_Originally posted by @ipspace in https://github.com/ipspace/netlab/issues/1563#issuecomment-2499971070_

jbemmel:
I'd say let's try and see what breaks: https://github.com/ipspace/netlab/pull/1569

I tested the platforms I have access to, seems to work fine (with some minor tweaks to the templates in case of EOS)

ipspace:
Implemented for FRR/CL 4.4 in 78f5a516e55e8b4443bc4b1a9ff0be5a36f02919

PR #1569, opened 2024-11-26
Rfc8950 over numbered ipv6

jbemmel:
Context: https://github.com/ipspace/netlab/issues/1565

Adds a numbered ipv6 RFC8950 test case. Tested the following devices:
* FRR
* Cumulus
* Cumulus NVUE
* EOS
* SR Linux

EOS and Cumulus NVUE required some template adjustments, not related to RFC8950 in the Cumulus case. 
But overall, RFC8950 over numbered ipv6 seems to work fine for the basic case


jbemmel:
> What you did is not "IPv4 AF over numbered IPv6 BGP session". It's IPv4 over LLA EBGP session plus IPv6 over numbered IPv6 EBGP session. This is the FRR config I get when using this PR, and EOS config is very similar.
> 
> ```
> router bgp 65000
>  bgp router-id 10.0.0.1
>  no bgp default ipv4-unicast
>  bgp bestpath as-path multipath-relax
>  neighbor eth1 interface remote-as 65100
>  neighbor eth1 description x1
>  neighbor eth2 interface remote-as 65101
>  neighbor eth2 description x2
>  neighbor 2001:db8:1::2 remote-as 65100
>  neighbor 2001:db8:1::2 description x1
>  neighbor 2001:db8:2::2 remote-as 65101
>  neighbor 2001:db8:2::2 description x2
> ```
> 
> It makes no sense to merge this, it's better to say "you're not supposed to do this" than the spaghetti mess this would create.

ok I see what you mean, the ```validate``` check should allow only a single numbered IPv6 session. Back to the drawing board - will try to adjust and see if we can make this work with reasonable effort. If not, we can always decide to not merge it

jbemmel:
> 
> It makes no sense to merge this, it's better to say "you're not supposed to do this" than the spaghetti mess this would create.
>

After fixing the test, I've come to share your conclusion that this makes no sense. FRR only supports RFC8950 in combination with ipv6 LLA, it is implied by the 'interface' neighbor construct and there seems to be no way to say "I want to do RFC8950 over numbered IPv6"

It may be possible to force other vendor platforms to do this, but the test cases all use FRR nodes for reference.

I'll rework the PR to say "don't try this at home"

jbemmel:
Replaced by #1575

Issue #1135, opened 2024-04-30
[BUG] Dell OS10: Route leaking BGP<->OSPF is not working


ssasso:
```
core1# show running-configuration
!
[cut]
!
ip vrf blue
 ip route-export 65000:2
 ip route-import 65000:2
 ip route-import 65000:1
!
ip vrf common
 ip route-export 65000:1
 ip route-import 65000:3
 ip route-import 65000:2
 ip route-import 65000:1
!
ip vrf red
 ip route-export 65000:3
 ip route-import 65000:3
 ip route-import 65000:1
!
[cut]
!
router ospf 102 vrf red
 redistribute bgp 65000
 redistribute connected
 router-id 10.0.0.1
!
[cut]
!
router bgp 65000
 router-id 10.0.0.1
 !
 address-family ipv4 unicast
  network 10.0.0.1/32
 !
 template unnumbered_ebgp
 !
 vrf blue
  router-id 10.0.0.1
  !
  address-family ipv4 unicast
   redistribute connected
  !
  neighbor 10.1.0.5
   description aggr2
   remote-as 65111
   send-community standard
   no shutdown
   !
   address-family ipv4 unicast
    soft-reconfiguration inbound
   !
   address-family ipv6 unicast
 !
 vrf common
  router-id 10.0.0.1
  !
  address-family ipv4 unicast
   redistribute connected
 !
 vrf red
  router-id 10.0.0.1
  !
  address-family ipv4 unicast
   redistribute connected
   redistribute ospf 102
```

BUT

```
core1# show ip route vrf common
Codes: C - connected
       S - static
       B - BGP, IN - internal BGP, EX - external BGP, EV - EVPN BGP
       O - OSPF, IA - OSPF inter area, N1 - OSPF NSSA external type 1,
       N2 - OSPF NSSA external type 2, E1 - OSPF external type 1,
       E2 - OSPF external type 2, * - candidate default,
       + - summary route, > - non-active route
Gateway of last resort is not set
  Destination                 Gateway                                        Dist/Metric       Last Change
----------------------------------------------------------------------------------------------------------
  B  EX 10.0.0.3/32         via 10.1.0.5,blue                                20/0              00:10:04
  C     10.1.0.0/30         via 10.1.0.2,red         ethernet1/1/1           0/0               00:10:38
  C     10.1.0.4/30         via 10.1.0.6,blue        ethernet1/1/2           0/0               00:10:42
[...]

PR #1571, opened 2024-11-26
Dell OS10: Fix test cases 33-vrf-common-ospf and 34- by redistributing imported OSPF routes too

jbemmel:
Might be related to https://github.com/ipspace/netlab/issues/1135

Fixes both 33-vrf-common-ospf and 34-vrf-common-mixed integration test cases

Technically this flag is only required if there are multiple VRFs running OSPF and they import from each other - but I think it's safe to enable whenever OSPF is used in VRFs

Note: Also fixes reference_bandwidth attribute


ipspace:
@ssasso You know way more about OS10 than I do. Are these changes OK?

jbemmel:
> @ssasso You know way more about OS10 than I do. Are these changes OK?

Not to put words in his mouth, but https://github.com/ipspace/netlab/issues/1135#issuecomment-2088152212 explicitly includes ```redistribute imported-ospf-routes```

ssasso (approved):

PR #1574, opened 2024-11-27
Fix: Print a warning when using CL 5.x containers

ipspace:
As promised in #1564

### DIFF
--- a/netsim/devices/cumulus_nvue.py
+++ b/netsim/devices/cumulus_nvue.py
@@ -64,3 +64,10 @@ class Cumulus_Nvue(_Quirks):
     
     if 'vrf' in mods:
       nvue_check_vrf_route_leaking(node)
+
+    if devices.get_provider(node,topology) == 'clab':
+      log.error(
+        f"Cumulus VX 5.x container used for node {node.name} is not supported and might not work correctly",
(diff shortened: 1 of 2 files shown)


netsim/devices/cumulus_nvue.py
```diff
@@ -64,3 +64,10 @@ def device_quirks(self, node: Box, topology: Box) -> None:
     
     if 'vrf' in mods:
       nvue_check_vrf_route_leaking(node)
+
+    if devices.get_provider(node,topology) == 'clab':
```
jbemmel, line 68: Same applies to ```cumulus``` device, I imported ```check_ospf_vrf_default``` from there and something similar could be done here (or create 1 method for common quirks and import that)

jbemmel (changes requested):

netsim/devices/cumulus_nvue.py
ipspace: Yeah, I wanted to do that as well, but then realized that the container and the VM use the same (or very similar) version of CL 4.4, so who cares. I did add a caveat, but I don't think it's worth generating a warning because it would say only "we did not test the container" as opposed to "we know it doesn't work"

Issue #1583, opened 2024-11-30
[BUG] CLNVUE fails to configure extra loopback interfaces


ipspace:
Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/initial/02-loopback.yml

Configuration attempt: https://tests.netlab.tools/cumulus_nvue/libvirt/initial/02-loopback.yml-initial.log

Seems to be related to #1557

Execution environment:

```
$ pip3 list | ack 'nsible|inja'
ansible                  10.6.0
ansible-core             2.17.6
ansible-pylibssh         1.1.0
Jinja2                   3.1.4
$ python3 --version
Python 3.10.12
```

PR #1588, opened 2024-11-30
Cumulus_nvue: Don't try to configure MTU for loopback interfaces

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1583

Also modifies Netlab link augmentation code to remove MTU settings from loopbacks, since they generally don't make sense and might trip over other platform templates

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -69,10 +69,6 @@
     interface:
       lo:
         type: loopback
-{%     if lb.mtu is defined %}
-        link:
-          mtu: {{ lb.mtu }}
-{%     endif %}
         ip:
           address:
 {%     if 'ipv4' in lb %}
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -326,7 +326,7 @@ def add_node_interface(node: Box, ifdata: Box, defaults: Box) -> Box:
       if not sys_mtu:                           # .. does the device support system MTU?
         ifdata.mtu = node.mtu                   # .... no, copy node MTU to interface MTU
 
-  if ifdata.get('type',None)=='loopback':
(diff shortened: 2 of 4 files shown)


ipspace:
> Also modifies Netlab link augmentation code to remove MTU settings from loopbacks

I wouldn't do that, although I admin I have no idea at the moment where it might come handy. Also, why should we change the existing behavior just because you made an error in a template?

> since they generally don't make sense and might trip over other platform templates

That's why the configuration templates are supposed to test interface types and virtual interface flags to avoid configuration that's incompatible with the interface (like "no switchport" on VLAN interface) -- see https://netlab.tools/dev/config/initial/#virtual-interfaces

Also, why did you assume you have to deal with MTU on the loopback (see https://netlab.tools/dev/config/initial/#loopback-configuration)

ipspace (changes requested):
Please remove the changes to links.py and changes to the test case.

jbemmel:
> > Also modifies Netlab link augmentation code to remove MTU settings from loopbacks
> 
> I wouldn't do that, although I admin I have no idea at the moment where it might come handy. Also, why should we change the existing behavior just because you made an error in a template?
> 

Because I figured that if one template had this issue, others might too. I found https://github.com/ipspace/netlab/blob/7209bd6083dd2bb61ab4b8378720aafe6b37f096/netsim/ansible/templates/initial/ios.j2#L65 for example and I even tried to bring up a topology to check whether it fails, but it took too much effort to get an image

I was trying to be proactive - just like ipv6 lla doesn't make sense on loopbacks, MTU doesn't make sense either and many platforms don't even allow you to configure it as illustrated by this PR

ipspace:
> Because I figured that if one template had this issue, others might too. I found

Fair point. We should have added MTU to the integration test ;)

OK, let's remove the MTU from loopback interfaces.

Issue #1594, opened 2024-12-01
[BUG] IBGP not working on CL NVUE


ipspace:
IBGP sessions never worked on CL NVUE because we could not set the source IP address of an IBGP session with early versions of NVUE data model (https://blog.ipspace.net/2022/10/cumulus-linux-nvue/). Now they want the AS number to be set to "internal", so even the config fails.

* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp/03-ibgp-rr.yml
* Validation results: https://tests.netlab.tools/cumulus_nvue/libvirt/bgp/03-ibgp-rr.yml-initial.log

@jbemmel: I can try to fix this if you don't have time.

jbemmel:
I'll fix it

PR #1598, opened 2024-12-01
Cumulus NVUE: Set neighbor BGP AS to "internal" when it is the same as bgp.as

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1594


jbemmel:
Just noticed one test still failed, reverting to draft

ipspace (approved):
Looks good to me. Will rerun the integration tests. Thank you!

PR #1601, opened 2024-12-01
Cumulus: Update OSPFv3 support check mark

jbemmel:
Validated that Cumulus passes the first OSPFv3 integration test


ipspace (approved):
Thank you!

Issue #1602, opened 2024-12-02
LAG: Support M:N MLAG configurations


jbemmel:
The lag module currently assumes MLAG is between 1 device and N other devices. It should be possible to configure M:N mlags:

```
links:
- lag:
  mlag.peergroup: 1
  members: [ a1-a2 ]
- lag:
  mlag.peergroup: 1
  members: [ b1-b2 ]
- lag:
    mlag: True
    members: [ a1-b1, a2-b2 ]
```

The above example shows 2 pairs of switches, each configured as an MLAG pair and interconnected

PR #1604, opened 2024-12-03
lag: Support M:M MLAG topologies

jbemmel:
* Enables 3 lag scenarios: 1:1 regular lag, 1:2 mlag and 2:2 dual mlag
* Include ```lag.ifindex``` attributes on interfaces
* Check node devices being the same for peerlinks
* Add check for peerlink when MLAG is used

Fixes https://github.com/ipspace/netlab/issues/1602

To be discussed:
* Most intuitive name for peer group id - currently "lag.mlag.peergroup" but could also be "peerlink" or plain "lag.mlag.id"
* Include explicit ```mlag``` flag or not? Currently left out
* Should we calculate the set of vlans to temporarily trunk across peerlinks, to enable correct address allocation?


ipspace:
The 'determine_mlag_sides' function makes my head explode. There must be an easier way to get that done.

Its sole purpose is to figure out which nodes are on the left- and the right- side of the bundle, right?

jbemmel:
> The 'determine_mlag_sides' function makes my head explode. There must be an easier way to get that done.
> 
> Its sole purpose is to figure out which nodes are on the left- and the right- side of the bundle, right?

Probably because it's done incrementally, we could do it in one shot upfront:

* For all members, count how often each node appears
* If there is 1 node that occurs as many times as there are members, that's the unique "1" side node of an mlag
* If there are 2 such nodes, it's a regular lag
* Else, it's a M:M mlag (absent self-links which should report an error)

ipspace (commented):
I can't review this. The diffs are a spaghetti mess. I will make a few suggestions and after you look at them merge it as-is (it shouldn't break anything else anyway) and then maybe take a look at individual functions.

docs/module/lag.md
```diff
@@ -69,10 +69,15 @@ links:
 ## Multi-chassis Link Aggregation (MLAG)
 
 For platforms that support it, link level redundancy can be provided by creating multiple links to different switches. The switches have an internal *peer link* to synchronize state related to the link aggregation, allowing them to present a single consistent network interface to the connected system.
+![image](https://github.com/user-attachments/assets/9c683f6e-0203-42e5-b5d6-91a2dfe711f8)
```
ipspace, line 72: There is no need for links to third-party websites. Add the image file to the 'module' directory and then link to it (see **netlab up** for an example).

docs/module/lag.md
```diff
@@ -83,17 +88,15 @@ groups:
   device: dellos10
  hosts:
   members: [h1,h2]
-  device: frr        # 'linux' does not support the lag module yet
+  device: frr          # 'linux' does not support the lag module yet
```
ipspace, line 91: If the hosts don't support the LAG module, create an access switch to two core switches example.

netsim/modules/lag.py
```diff
@@ -50,186 +63,207 @@ def check_lag_config(node: str, linkname: str, topology: Box) -> bool:
...
+"""
+normalized_members - builds a normalized list of lag member links, checking various conditions
+"""
+def normalized_members(l: Box, topology: Box, peerlink: bool = False) -> list:
+  members = []                                      # Build normalized list of members
+  _name = "peerlink" if peerlink else "lag"
+  for idx,member in enumerate(l.lag.members):
+    member = links.adjust_link_object(member,f'{l._linkname}.{_name}[{idx+1}]',topology.nodes)
+    if 'lag' in member:                             # Catch potential sources for inconsistency
+      if peerlink or ('ifindex' not in member.lag): # ...but allow for custom bond numbering
```
ipspace, line 87: Maybe I got it all wrong, but this seems to accepts any LAG attribute on a member link as long as there's 'ifindex' on it. Also, why would you allow a user to specify bond interface index on the LAG member link?

netsim/modules/lag.py
```diff
@@ -50,186 +63,207 @@ def check_lag_config(node: str, linkname: str, topology: Box) -> bool:
...
+                  category=log.IncorrectAttr,
+                  module='lag')
+        return []
+    if len(member.interfaces)!=2:                   # Check that there are exactly 2 nodes involved
+      log.error(f'Link {member._linkname} in LAG {l.lag.ifindex} must have exactly 2 nodes',
+                category=log.IncorrectValue,
+                module='lag')
+      return []
+    if member.interfaces[0].node==member.interfaces[1].node:
+      log.error(f'Link {member._linkname} in LAG {l.lag.ifindex} must have exactly 2 different nodes',
```
ipspace, line 98: I would change that to "a LAG link cannot connect a node to itself"

netsim/modules/lag.py
```diff
@@ -50,186 +63,207 @@ def check_lag_config(node: str, linkname: str, topology: Box) -> bool:
...
+    for i in member.interfaces:                     # Check that they all support LAG
+      if not check_lag_config(i.node,member._linkname,topology):
+        return []
+    members.append(member)
+  return members
+
 """
 create_lag_member_links -- expand lag.members for link l and create physical p2p links
 """
 def create_lag_member_links(l: Box, topology: Box) -> None:
```
ipspace, line 111: Is this a function with four (or so) embedded functions? Why did you use embedded functions? It would be much easier to read if you would take those functions out of this one.

netsim/modules/lag.py
```diff
@@ -327,16 +372,27 @@ def module_pre_transform(self, topology: Box) -> None:
...
+        has_peerlink = True
       elif i.type=='lag':
         i.lag = node.get('lag',{}) + i.lag  # Merge node level settings with interface overrides
         lacp_mode = i.get('lag.lacp_mode')  # Inheritance copying is done elsewhere
         if lacp_mode=='passive' and not features.lag.get('passive',False):
           log.error(f'Node {node.name} does not support passive LACP configured on interface {i.ifname}',
             category=log.IncorrectAttr,
             module='lag',
-            hint='lag')
+            hint='passive LACP')
```
ipspace, line 388: The 'hint' attribute should refer to a key in "defaults.hints.module", otherwise it will not do anything.

ipspace (changes requested):
At least the documentation should be changed to include the diagram.

netsim/modules/lag.py
jbemmel: The ```ifindex``` may be needed in case of 2:2 mlag where there is a different bond index on each pair (we have some examples like that in our network).

Rewritten to only allow 'ifindex', no other lag attributes

netsim/modules/lag.py
jbemmel: The embedded functions are only used from this context, and use many of the same arguments. I have moved them out (except ```analyze_lag```)

netsim/modules/lag.py
jbemmel: removed

docs/module/lag.md
jbemmel: ok, added

docs/module/lag.md
jbemmel: I'd prefer to include https://github.com/ipspace/netlab/pull/1624 and change the example to use Linux hosts

ipspace:
Merged as-is.

Issue #1581, opened 2024-11-29
[BUG] lag module fails in case of Multiprovider labs


jbemmel:
Just a note that Multiprovider labs don't combine with the lag module at the moment, due to libvirt changing the link type from "lag" to "lan"

```
groups:
 _auto_create: True
 core:
  members: [ c1, c2 ]
  device: dellos10
  provider: clab
  module: [ lag ]
 edge:
  members: [ e1, e2 ]
  device: cumulus_nvue
  provider: libvirt
  module: [ lag ]
 hosts:
  members: [ h1, h2, h3, h4 ]
  device: linux

links:
- lag:
   mlag.peergroup: 1
   members: [c1-c2]
- lag:
   mlag: True
   members: [e1-c1,e1-c2]
- lag:
   mlag: True
   members: [e2-c1,e2-c2]
```

ipspace:
Even if we get that fixed (and based on the recent SNAFU I caused in bf8ac79e594499fd4623c6242baddb2a337fe614 and fixed in e58b8eb8829233451159575e34680cb2fc8e7bbc I'm a bit reluctant to try and get around this limitation) we don't know if LACP works over Linux bridges, so it might be worth trying that out first.

If my hunch is right and LACP does not work across bridges, then we just have to document the caveat in the LAG module documentation and move on.

jbemmel:
> 
> If my hunch is right and LACP does not work across bridges, then we just have to document the caveat in the LAG module documentation and move on.
>
Your 2020 article still applies - see https://github.com/ipspace/netlab/pull/1582

LACP gets discarded, but FRR happily creates the bond and doesn't alert the user that LACP isn't working. Cumulus declares the bond status as 'DOWN' on the libvirt side of the link, not sure if this "MII status"

I can add a quirk to the lag module warning about this limitation of LACP with Linux bridges. However, static bonding could still work in multi-provider cases

ipspace:
> LACP gets discarded, but FRR happily creates the bond and doesn't alert the user that LACP isn't working. Cumulus declares the bond status as 'DOWN' on the libvirt side of the link, not sure if this "MII status" is due to LACP failing

Let's just say that LAGs without LACP are as reliable as parallel static routes pointing to multiple uplinks, and adding a bridge in the middle only makes it worse.

Alas, unpleasant facts never stopped "creative" people.

> I can add a quirk to the lag module warning about this limitation of LACP with Linux bridges.

Your choice.

[...]

PR #1605, opened 2024-12-03
Draft: Improve multiprovider p2p links by removing Linux bridge

jbemmel:
As observed, Linux bridges don't work well with STP or LACP, posing challenges especially in multi-provider topologies

This draft PR proposes a workaround: Implement Libvirt-Clab P2P links (i.e. links having 2 nodes) without a bridge by connecting the clab Macvlan device directly to the Libvirt virtual interface (deleting the Linux bridge created by Vagrant)

A PR like this obviously needs thorough testing, it is not ready to be merged. But before I spend more time doing that, I'd like to get some feedback

Rather than marking multi-provider links as 'lan', this PR annotates the links to orchestrate a different implementation of the underlying connectivity. It never made sense that P2P links suddenly get IPs from a different pool when a different provider is selected, and as a side benefit, this also allows us to keep the 'lag' type links.

It is nicer if both STP and lag can work in multi-provider scenarios, as some platforms only come as containers and others only as Vagrant box

Fixes https://github.com/ipspace/netlab/issues/1581

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -50,7 +50,7 @@
 
 - set:
     system:
-      hostname: {{ inventory_hostname }}
+      hostname: {{ inventory_hostname | replace("_","-") }}
     interface:
       eth0:
         ip:
--- a/netsim/providers/clab.py
+++ b/netsim/providers/clab.py
@@ -21,35 +21,6 @@ def list_bridges( topology: Box ) -> typing.Set[str]:
 def use_ovs_bridge( topology: Box ) -> bool:
     return topology.defaults.providers.clab.bridge_type == "ovs-bridge"
 
-def create_linux_bridge( brname: str ) -> bool:
-  if external_commands.run_command(
-       ['brctl','show',brname],check_result=True,ignore_errors=True) and not is_dry_run():
-    log.print_verbose(f'Linux bridge {brname} already exists, skipping')
-    return True
-
-  status = external_commands.run_command(
-      ['sudo','ip','link','add','name',brname,'type','bridge'],check_result=True,return_stdout=True)
-  if status is False:
-    return False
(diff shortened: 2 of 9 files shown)


ipspace:
> As observed, Linux bridges don't work well with STP or LACP, posing challenges especially in multi-provider topologies

Didn't you get STP to work across a Linux bridge?

> This draft PR proposes a workaround: Implement Libvirt-Clab P2P links (i.e. links having 2 nodes) without a bridge by connecting the clab Macvlan device directly to the Libvirt virtual interface (deleting the Linux bridge created by Vagrant)

No. Absolutely not. Every time we added complexity we were burned further down the line (for example, it was "interesting" to get the code that recovers the labs after a server reboot to work). Adding unnecessary complexity is totally out of question.

> Rather than marking multi-provider links as 'lan', this PR annotates the links to orchestrate a different implementation of the underlying connectivity.

I don't like the idea to change working unrelated stuff just because your stuff doesn't work.

> It never made sense that P2P links suddenly get IPs from a different pool when a different provider is selected

I agree with that, and maybe it's time we stop using link.type to select the default pool. That would also be a breaking change, but I would have no problem with that, as it makes sense.

> and as a side benefit, this also allows us to keep the 'lag' type links.

Well, I already told you I would have blocked the 'lag' type idea if it could ever work over multi-provider links.

> It is nicer if both STP and lag can work in multi-provider scenarios, as some platforms only come as containers and others only as Vagrant box

Not worth the effort. No platform (apart from CL and VyOS) comes as a ready-to-use libvirt Vagrant box, and the only platform that comes only as a container is SR Linux. Also, most of the platforms supported by netlab can also be converted into a vrnetlab container. We will not channel our inner MacGyver to deal with a few exceptions.

jbemmel:
As we discovered, Cumulus doesn't really work that well as a container and there is no vrnetlab container for it yet.
Closing this for now

PR #1607, opened 2024-12-03
Cumulus NVUE: Fix BGP communities being advertised

### LINKED ISSUE #1606: [BUG] CL NVUE: BGP community propagation is not configured correctly
* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp/05-community.yml
* Validation results: https://tests.netlab.tools/cumulus_nvue/libvirt/bgp/05-community.yml-validate.log

I don't think this ever worked, and I'm OK to declare this a caveat and move on. What do you think @jbemmel?

jbemmel:
* Explicitly set to 'on' or 'off'
* Add neighbor description
* Treat ibgp_localas as ibgp
* Check bgp.next_hop_self

Fixes #1606

### DIFF
--- a/netsim/ansible/templates/bgp/cumulus_nvue.j2
+++ b/netsim/ansible/templates/bgp/cumulus_nvue.j2
@@ -80,9 +80,9 @@
 {%       if bgp.rr|default('') and not n.rr|default('') %}
                     route-reflector-client: on
 {%       endif %}
-{{       advertise_communities( bgp.community.ibgp|default([]) ) }}
-{%     else %}
-{{       advertise_communities( bgp.community.ebgp|default([]) ) }}
+{%     endif %}
+{%     if n.type in bgp.community %}
+{{       advertise_communities( bgp.community[n.type] ) }}
 {%     endif %}
 {%   endfor %}
 {% endfor %}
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
While this would pass the integration tests, it's different from most other configuration templates that use something along these lines:

```
{% if n.type in bgp.community|default({}) %}
{%   for c_type in bgp.community[n.type] %}
  neighbor {{ ip }} send-community {{ c_type }}
{%   endfor %}
{% endif %}
```

Important differences:

* Missing bgp.community.x means "we don't care" not "no communities"
* ibgp_localas is different from ibgp, allowing finer granularity if needed

jbemmel:
Updated PR 

Could consider to extend the integration test to include ```ibgp_localas``` with a different set of communities

ipspace (approved):
Thank you!

ipspace:
> Could consider to extend the integration test to include `ibgp_localas` with a different set of communities

Not all devices support IBGP local-as, so we'd have to create a new test just for that, and the integration tests already take around 48 hours to run, so I tend to skip edge cases (I could create way more integration tests were it not for time constraints)

PR #1609, opened 2024-12-04
Dell OS10: Disable ipv6 on ipv4-only interfaces

jbemmel:
Dell OS10 enables ipv6 unless told not to. On ipv4-only interfaces it should be disabled

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -74,6 +74,8 @@ interface {{ l.ifname }}
 {%   else %}
 ! Invalid IPv6 address {{ l.ipv6 }}
 {%   endif %}
+{% elif 'ipv4' in l %}
+ no ipv6 enable
 {% endif %}
 !
 {% endfor %}


ipspace (changes requested):
I can't possibly figure out what you'd want to disable IPv6 only when IPv4 is enabled on an interface.

I also can't figure out why you'd care, but I guess that's a metaphysical question.

jbemmel:
> I can't possibly figure out what you'd want to disable IPv6 only when IPv4 is enabled on an interface.
> 

Because on L2-only interfaces or MLAG peerlinks OS10 won't allow you to configure ```no ipv6 enable```.
It's only needed on IP-enabled interfaces

> I also can't figure out why you'd care, but I guess that's a metaphysical question.

In labs with several VLANs, my packet traces get flooded with useless IPv6 ICMP6 packets that I'd like to filter out

ipspace:
> Because on L2-only interfaces or MLAG peerlinks OS10 won't allow you to configure `no ipv6 enable`. It's only needed on IP-enabled interfaces

I can't figure out in which universe that would make sense. How does configuring IPv4 address enable IPv6 LLA by default? Never minds, moving on...

> In labs with several VLANs, my packet traces get flooded with useless IPv6 ICMP6 packets that I'd like to filter out

Are you saying that Dell switches send bogus ICMP6 packets from link-local addresses 🤦‍♂️ and that those packets get delivered over multiple hops? 🤦‍♂️🤦‍♂️

Merging... and I'm sorry you have to work with such a "creative" platform.

ipspace (approved):

PR #1612, opened 2024-12-05
OSPF: Enable global default passive

jbemmel:
* Allow ```ospf.passive``` to be set globally or per node
* Extend OSPFv2 validation plugin for FRR with a ```count``` parameter that checks the number of adjacencies
* Add integration test case to test OSPF passive

### DIFF
--- a/docs/module/ospf.md
+++ b/docs/module/ospf.md
@@ -144,6 +144,7 @@ OSPF routing daemons support these interface-level features:
 
 * **ospf.reference_bandwidth** sets the OSPF auto-cost reference bandwidth (in Mbps) for all devices in the network.
 * **ospf.bfd.strict** enables RFC9355 BFD Strict-Mode (default: False)
+* **ospf.passive** sets the global default for passive (default: False)
 
 ## Node Parameters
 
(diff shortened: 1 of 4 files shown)


ipspace (changes requested):
The documentation and module parameter changes seem OK. However:

* `04-passive.yml` integration test already checks the "passive interface" implementation. There is no need for another node-level integration test as the parameter eventually appears on the interfaces and thus does not impact the config templates.
* Making VLAN module a prerequisite to pass an OSPF integration test is stupid.
* If we get to a point where another OSPF integration test would be warranted, I expect to see OSPFv2 and OSPFv3 versions of the test.
* The only reason to check the number of adjacencies would be to deal with parallel links. A simple workaround is not to use parallel links ;)
* It's unacceptable that the validation plugins for different devices support different arguments. It destroys the whole idea of "you can validate X using devices Y,W, or Z"

Please remove the unnecessary changes and resubmit the PR.

jbemmel:
ok, it may not be suitable as an ```integration``` test, I thought it would be nice to include a use case where one might use this - having multiple VLANs is another way of getting multiple adjacencies.

Submitted another stripped down version of this PR

ipspace:
> ok, it may not be suitable as an `integration` test, I thought it would be nice to include a use case where one might use this - having multiple VLANs is another way of getting multiple adjacencies.

You can always submit it to 'netlab-examples' repository.

PR #1616, opened 2024-12-05
node.clone plugin - simplify repetitive topologies by cloning nodes

jbemmel:
A plugin to clone nodes and any links/interfaces they are attached to, to simplify provisioning of large clusters of identical devices (like servers in a rack)

Context: https://github.com/ipspace/netlab/discussions/1570

Now includes support for *lag* links
```
---
plugin: [ node.clone ]

module: [vlan]

vlans:
  red:
    links: [r-h]        # VLAN links get updated

nodes:
 r:
  device: frr
 h:
  device: linux
  clone.count: 32       # Create 32 hosts
```


jbemmel:
Ready for review

jbemmel:
Apologies for the additional notifications

Astrotokii:
Could this be applied to link groups for simplifying etherchannels and redundancy? 

I.e. 

-group: etherchannel
 members: [r-h]
 repeat: 1

ipspace:
Just a suggestion on the name change (as "repeater" can mean anything -- see https://en.wikipedia.org/wiki/Repeater_(disambiguation) ):

* Rename the plugin to "node.clone"
* Use "clone" attributes like "clone.count" (for a total count) and "clone.start" (potentially also "clone.step")

If I read the code right, the original node is retained, which I would find confusing -- I would get "h, h-01, h-02, ... h-31" for a total of 32 nodes. I think the first node should be "h-01" (or "h-00" -- this is where "clone.start" would come handy) and the count (however it is called) should specify the total number of nodes.

netsim/extra/repeater/defaults.yml
```diff
@@ -0,0 +1,9 @@
+# repeater plugin attributes and defaults
+#
+---
+attributes:
+  node:
+    repeat: int                     # Create N copies of this node anywhere it's used in groups or links
+
+repeater:
+  node_name_pattern: "{name[:13]}-{id:02x}"
```
ipspace, line 9: The generated node name is not an identifier, which obviously works for you because the code is executed in just the right place in the transformation process, but it might bite us in some unexpected places. It might be a good idea to replace "-" with "_"

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+    clone = data.get_box(node)
+    clone.name = strings.eval_format(topology.defaults.repeater.node_name_pattern, node + { 'id': c } )
+    clone.interfaces = []                                                   # Start clean, remove reference to original node
+    if 'id' in node:
+      clone.id = node.id + c - 1                                            # Update any explicit node ID sequentially
+    topology.nodes += { clone.name: clone }
+    clones.append( clone.name )
+
+  for link in list(topology.get('links',[])):
+    link_data = links.adjust_link_object(link,f'repeater{c}',topology.nodes)
```
ipspace, line 59: This would need to change if you go for "clone". Also:

* I would add some separator between "repeater" (or whatever) and the number, for example "clone[3]"
* If you have multiple cloned objects, you'd get duplicate link names. Not a problem from the transformation perspective, but potentially confusing to the user. "clone.{node.name}[{c}]" might be a better option.

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+  def update_ifindex(intf:Box, c:int) -> Box:
+    ifindex = intf.get('ifindex',None)
+    if ifindex is not None:
+      intf = data.get_box(intf)                                             # Need to make a fresh copy
+      intf.ifindex = ifindex + c + 1
+    return intf
+  
+  ifs = []
+  if nodename in [ i.node for i in link_data.interfaces ]:
+     link_data.pop('_linkname',None)
```
ipspace, line 19: I would leave the link name, that would help the user figure out where the link came from if the link generates errors.

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+     atts = [ i for i in link_data.interfaces if i.node==nodename ][0]      # Copy any interface atts
```
ipspace, line 20: This wouldn't work for nodes connected to themselves (someone loved to use them IIRC). It might be better to iterate over interfaces looking for the target nodename)

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+     for c,clone in enumerate(clones):
+       l = data.get_box(link_data)
+       l.interfaces = [ atts+{ 'node': clone } ] + [ update_ifindex(i,c) for i in link_data.interfaces if i.node!=nodename ]
```
ipspace, line 23: And this would become much simpler if you do in-place replacement of the node name instead of rearranging the list.

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+       ifs.append(l)
+  return ifs
+
+def update_vlan_access_links(topology: Box, nodename: str, clones: list) -> None:
```
ipspace, line 27: This illustrates the major problem of this approach (not that I think it could be done in some other way): you're playing whack-a-mole with other data structures. You handled VLAN links (because you needed them), but there are also VRF links, LAG links, link groups, and topology components...

At the very minimum, the limitations have to be well-documented.

FWIW, not using the original node name (replacing "h" with "h_00", for example) would also uncover all cases where the cloning does not work as the node name used on a link would not be valid.

netsim/extra/repeater/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+    if 'lag' in link_data:
+      log.error("LAG links not yet supported by repeater plugin, ignoring...",
+                category=Warning, module='repeater')
+      continue  
+    ifs = clone_interfaces(link_data,node.name,clones)
+    topology.links += ifs
+
+  if 'groups' in topology:
+    for group in topology.groups.values():
+      if isinstance(group,Box) and node.name in group.get('members',[]):   # Groups can include 'bool flags too
```
ipspace, line 71: I think you have to skip group names starting with "_"

ipspace (changes requested):
Made a few suggestions (if nothing else, I would strongly recommend using "_" in node names). Also, the plugins needs documentation.

netsim/extra/repeater/defaults.yml
jbemmel: I specifically used '-' and not '_' because of some issues, see https://github.com/ipspace/netlab/pull/1619

'_' is not a valid hostname character

Note that '-' *is* valid in node ids, the logic used to verify them is at https://github.com/ipspace/netlab/blob/5dc32bf8ef219c650184be9ec652f5e13523fa23/netsim/data/types.py#L462 (and not https://github.com/ipspace/netlab/blob/5dc32bf8ef219c650184be9ec652f5e13523fa23/netsim/utils/strings.py#L196)

jbemmel: How about we call ```make_id``` on whatever the (user provided) string template generates

jbemmel:
> Could this be applied to link groups for simplifying etherchannels and redundancy?
> 
> I.e.
> 
> -group: etherchannel members: [r-h] repeat: 1

It could, but the ```lag``` module already greatly simplifies the definition of multi-link etherchannels; the attributes can be specified once for the group, and then the list of links is very short for typical deployment cases

```
links:
- lag.members: [r-h, r-h, r-h, r-h]
```
is simpler (and less characters) than:
```
links:
- lag:
   members: [r-h]
   clone.count: 4
```

netsim/extra/repeater/plugin.py
jbemmel: done

netsim/extra/repeater/plugin.py
jbemmel: Done, we now get errors when an unsupported construct is used in the topology (as it doesn't get updated)

netsim/extra/repeater/plugin.py
jbemmel: ok

netsim/extra/repeater/defaults.yml
ipspace: I'm also OK with '-' in the default template. As you said, it makes no sense to call **make_id** (also: defaults are not sanity-checked).

netsim/extra/repeater/plugin.py
jbemmel: We wouldn't want to disappoint our more sophisticated users - refactored

netsim/extra/repeater/plugin.py
jbemmel: ok

netsim/extra/repeater/plugin.py
jbemmel: The 'c' reference was actually a leftover from when the code was under the for loop - changed

jbemmel:
Ready for review

* Perhaps link groups could be supported transparently if they were transformed before the plugin(s) get invoked?
* It would simplify things if link normalization would happen before invoking topology-changing plugins

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+  if nodename in [ i.node for i in link_data.get('interfaces',[]) ]:
+    for c,clone in enumerate(clones):
+      l = data.get_box(link_data)
+      l.interfaces = []
+      for intf in link_data.interfaces:
+        intf_clone = data.get_box(intf)
+        if intf.node == nodename:
+          intf_clone.node = clone
+        elif 'ifindex' in intf:
+          intf_clone.ifindex = intf.ifindex + c
```
ipspace, line 17: This could easily generate overlapping ifindex values. Just saying.

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+  if 'include' in node:                                                      # Check for components
+    log.error("Cannot clone component {node.name}, only elementary nodes",
+              category=AttributeError, module='node.clone')
+    return
+
+  name_format = topology.defaults.clone.node_name_pattern
+
+  orig_name = node.name
+  node.name = strings.eval_format(name_format, node + { 'id': _p.start } )  # Rename first node as clone_01
+  topology.nodes += { node.name: node }
```
ipspace, line 62: I would find `topology.nodes[node.name] = node` easier to understand. Also you never check whether the node already exists.

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+
+  clones = [ node.name ]
+  for c in range(_p.start+_p.step,_p.start+_p.count*_p.step,_p.step):       # Existing node is '1'
```
ipspace, line 65: I don't see why you treat the first node differently than the others? At the very minimum, it would be nice to have a comment to explain it.

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+    if 'id' in node:
+      clone.id = node.id + c - 1                                            # Update any explicit node ID sequentially
+    topology.nodes += { clone.name: clone }
+    clones.append( clone.name )
+
+  for link in list(topology.get('links',[])):                               # Make a copy of links list
+    link_data = links.adjust_link_object(link,f'clone[{orig_name}]',topology.nodes)
+    if link_data is None:
+      continue
+    if 'lag' in link_data:
```
ipspace, line 78: Shouldn't you check this only if the link includes the cloned node? Also, wouldn't it be better to have a common 'link adjustment' function that would work for **links**, **vlans.X.links** and **vrfs.X.links**?

ipspace (changes requested):
Mostly suggestions, the only truly annoying part is the "vlans" link name in VRF links.

netsim/extra/node.clone/plugin.py
jbemmel: Leftover from earlier embodiment, it should be removed

netsim/extra/node.clone/plugin.py
jbemmel: If ```'lag' in link_data``` the cloned node could be mentioned in ```lag.members``` - but checking if it's in there amounts to doing the work to support lag links

I was holding off to add ```lag``` support as it's a bit of a moving target, but I can add it. Will build an adjustment function that can also handle ```lag.members```

ipspace: Changing the error message to something like "You cannot use node.clone plugin in a topology with LAG links" might make it easier to understand for non-native speakers like myself. I understood the error message the wrong way.

jbemmel: Added lag support and removed the error

netsim/extra/node.clone/plugin.py
jbemmel: added

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+    if not 'links' in vdata:                                                # No VLAN links?
+      continue                                                              # ... no problem, move on
+
+    for cnt,l in enumerate(list(vdata.links)):                              # So far so good, now iterate over the links
+      link_data = links.adjust_link_object(                                 # Create link data from link definition
+                    l=l,
+                    linkname=f'vlans.{vname}.links[{cnt+1}]',
+                    nodes=topology.nodes)
+      if link_data is not None:
+        ifs = clone_interfaces(link_data,nodename,clones)
```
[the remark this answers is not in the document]
jbemmel: ```clone_link```

netsim/extra/node.clone/plugin.py
```diff
@@ -0,0 +1,109 @@
...
+  for vname,vdata in topo_items.items():                                    # Iterate over global VLANs or VRFs
+    if not isinstance(vdata,Box):                                           # VLAN not yet a dictionary?
+      continue                                                              # ... no problem, skip it
+    if not 'links' in vdata:                                                # No VLAN links?
+      continue                                                              # ... no problem, move on
+
+    for cnt,l in enumerate(list(vdata.links)):                              # So far so good, now iterate over the links
+      link_data = links.adjust_link_object(                                 # Create link data from link definition
+                    l=l,
+                    linkname=f'vlans.{vname}.links[{cnt+1}]',
```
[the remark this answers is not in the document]
jbemmel: fixed

netsim/extra/node.clone/plugin.py
jbemmel: Added a caveat note

jbemmel:
Hit an issue with ```lag.ifindex``` value overlap (see https://github.com/ipspace/netlab/pull/1634) but ready for review

ipspace (approved):

PR #1617, opened 2024-12-06
STP: Add stp.port_type configuration for Cumulus 4.x/NVUE, EOS and Dell OS10

jbemmel:
* Option to auto-configure stp.port_type as 'edge' for ports with only hosts connected, default disabled
* Feature flag stp.port_type (not supported by FRR)

### DIFF
--- a/docs/module/stp.md
+++ b/docs/module/stp.md
@@ -31,7 +31,7 @@ MSTP/RSTP ports fall back to regular STP upon receiving a plain STP BPDU.
 ## Global Parameters
 
 * **stp.protocol** (one of stp, mstp, rstp or pvrst) -- Global STP flavor to run on supporting nodes, default **stp**
-* **stp.stub_port_type** (one of 'normal','edge','network','auto' or 'none') -- Port type to configure on ports with only hosts connected, default **'edge'**
+* **stp.stub_port_type** (one of 'normal','edge','network','auto' or 'none') -- Port type to configure on ports with only hosts connected, default **'none'**
 
 ## Global, Node, Link, Interface, and VLAN Parameters
 
--- a/docs/module/stp.md
+++ b/docs/module/stp.md
@@ -44,7 +44,7 @@ You can set the **‌stp.enable** parameter in the **‌vlans** dictionary to en
 ## Node Parameters (global or per VLAN)
 
 * **stp.priority** (int 0..61440 in increments of 1024) -- STP priority for root election, by default, all nodes have equal priority 32656.  In case of equal priority, the bridge with the lowest MAC address becomes root; note that MAC addresses are assigned randomly in Netlab
-* **stp.port_type** (one of 'normal','edge','network' or 'auto') -- STP port type for all interfaces connected to this link
+* **stp.port_type** (one of 'normal','edge','network' or 'auto') -- STP port type for all interfaces connected to this node
 
 ## Interface Parameters
 
--- a/docs/module/stp.md
+++ b/docs/module/stp.md
@@ -13,9 +13,10 @@ The following table describes per-platform support of individual STP features:
 | ------------------ |:---:|:---:|:---:|:---:|:---:|
 | Arista EOS[^EOS]   | ✅  | ✅  | ✅  | ✅ |  ✅ |
 | Aruba AOS-CX[^AOSCX] | ❗  | ✅  | ❌  | ✅ |  ✅ |
-| Cumulus Linux[^CL] | ✅  |  ❌  | ✅  | ❌   |  ✅ |
-| Dell OS10[^OS10]   | ✅  | ✅  | ✅  | ✅ |  ✅ |
-| FRR[^FRR]          | ✅  |  ❌  |  ❌  |  ✅ | ❌   |
+| Cumulus Linux 4.x[^CL] | ✅  |  ❌  | ✅  | ❌   |  ✅ |  ✅ |    
+| Cumulus 5.x (NVUE)[^CL] | ✅  |  ❌  | ✅  | ❌   |  ✅ |  ✅ |  
+| Dell OS10[^OS10]   | ✅  | ✅  | ✅  | ✅ |  ✅ |  ✅ |
+| FRR[^FRR]          | ✅  |  ❌  |  ❌  |  ✅ | ❌   | ❌   |
 
 [^EOS]: MSTP is enabled by default
 [^AOSCX]: MSTP is enabled by default; STP is stated as not supported, but it is configured as MSTP (see tip below).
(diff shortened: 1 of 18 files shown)


docs/module/stp.md
```diff
@@ -9,12 +9,12 @@ Many platforms already support and enable STP by default; this module provides e
...
-| Operating system   | STP | MSTP | RSTP | Per-VLAN (R)STP | Enable per port
-| ------------------ |:---:|:---:|:---:|:---:|:---:|
-| Arista EOS[^EOS]   | ✅  | ✅  | ✅  | ✅ |  ✅ |
-| Cumulus Linux[^CL] | ✅  |  ❌  | ✅  | ❌   |  ✅ |
-| Dell OS10[^OS10]   | ✅  | ✅  | ✅  | ✅ |  ✅ |
-| FRR[^FRR]          | ✅  |  ❌  |  ❌  |  ✅ | ❌   |
+| Operating system   | STP | MSTP | RSTP | Per-VLAN (R)STP | Enable per port | Port type
+| ------------------ |:---:|:---:|:---:|:---:|:---:|:---:|
+| Arista EOS[^EOS]   | ✅  | ✅  | ✅  | ✅ |  ✅ |  ✅ |
+| Cumulus Linux[^CL] | ✅  |  ❌  | ✅  | ❌   |  ✅ |  ✅ |
```
ipspace, line 15: It might be worth expanding this to "CL" and "CL NVUE" to match other documentation.

docs/module/stp.md
```diff
@@ -28,6 +28,7 @@ MSTP/RSTP ports fall back to regular STP upon receiving a plain STP BPDU.
 ## Global Parameters
 
 * **stp.protocol** (one of stp, mstp, rstp or pvrst) -- Global STP flavor to run on supporting nodes, default **stp**
+* **stp.host_edge_port** (bool) -- Configure ports with only hosts connected as 'edge' port_type on platforms that support it, default **True**
```
ipspace, line 31: A cosmetic suggestion: I don't like the name of this parameter. How about 'stub_port_type' with the same parameters as the 'port_type', and then you apply it to stub links that have no 'port_type'?

netsim/ansible/templates/stp/eos.j2
```diff
@@ -36,4 +36,5 @@ interface {{ ifdata.ifname }}
 #
 {%    endif %} 
 {%   endif %}
+ spanning-tree portfast {{ ifdata.stp.port_type|default('auto') }}
```
ipspace, line 39: You should either configure this conditionally or set the default value to 'normal' (the default EOS value)

ipspace (changes requested):

netsim/ansible/templates/stp/eos.j2
jbemmel: Added 'auto' as a valid value, other platforms enable 'auto' by default. Changed the EOS template to default to 'normal'

docs/module/stp.md
jbemmel: done

docs/module/stp.md
jbemmel: Changed, was looking to use ```link.role == 'stub'``` but it doesn't seem to get set in case of VLANs? Code still checking all connected nodes for ```role == 'host'```

ipspace: IIRC the "single router on a segment" logic sets some extra flags, but maybe they are removed before you get to them. Anyway, don't expect this logic to work on VLANs unless you thoroughly tested all possible scenarios (just saying).

docs/module/stp.md
```diff
@@ -40,8 +42,9 @@ You can set the **‌stp.enable** parameter in the **‌vlans** dictionary to en
 ## Node Parameters (global or per VLAN)
 
 * **stp.priority** (int 0..61440 in increments of 1024) -- STP priority for root election, by default, all nodes have equal priority 32656.  In case of equal priority, the bridge with the lowest MAC address becomes root; note that MAC addresses are assigned randomly in Netlab
+* **stp.port_type** (one of 'normal','edge','network' or 'auto') -- STP port type for all interfaces connected to this node
```
ipspace, line 45: I don't see **stp.port_type** attribute defined for nodes or VLANs. Have you tested it?

netsim/ansible/templates/stp/cumulus.j2
```diff
@@ -36,6 +36,9 @@ iface {{ ifdata.ifname }}
 # Newer versions 5.x support mstpctl-port-vlan-priority <vlan id>={{ ifdata.stp.port_priority }}
 #
 {%  endif %}
+{%  set port_type = ifdata.stp.port_type|default("normal") %}
```
ipspace, line 39: Wouldn't it be better to start with `if ifdata.stp.port_type is defined`? Otherwise, if you insist on using stp.port_type without checking whether it's defined, it might be better to set the default value in the STP module instead of having `default(normal)` filter in every single template.

Also, this effectively overrides device default value even in case where the user does not care (or does not want to change the value).

Applies to all templates.

docs/module/stp.md
```diff
@@ -28,6 +29,7 @@ MSTP/RSTP ports fall back to regular STP upon receiving a plain STP BPDU.
 ## Global Parameters
 
 * **stp.protocol** (one of stp, mstp, rstp or pvrst) -- Global STP flavor to run on supporting nodes, default **stp**
+* **stp.stub_port_type** (one of 'normal','edge','network','auto' or 'none') -- Port type to configure on ports with only hosts connected, default **'edge'**
```
ipspace, line 32: Do you think it's a good idea to assume everyone wants "portfast" on the edge ports as the "don't care about this" value?

netsim/modules/stp.py
```diff
@@ -31,19 +47,26 @@ def node_post_transform(self, node: Box, topology: Box) -> None:
...
+          log.error(
+            f'node {node.name} (device {node.device}) does not support enabling/disabling STP only on a specific port ({intf.ifname})',
+            log.IncorrectValue,
+            'stp')
+        if 'port_type' in intf.stp and not features.get('stp.port_type',False):
+          log.error(
+            f'node {node.name} (device {node.device}) does not support configuration of STP port_type on ({intf.ifname})',
+            log.IncorrectValue,
+            'stp')
+      if stub_port_type != 'none' and not intf.get('stp.port_type',None):
```
ipspace, line 68: BTW, I can't see how the node **port_type** value is propagated to interfaces.

docs/module/stp.md
ipspace: ... oops, you're using _vlan_saved_neighbors, so this would work as that attribute is created in 'post_link_transform' and you're using it in the final 'post_transform' hook.

ipspace (changes requested):
It looks to me like the node.port_type and vlan.port_type attributes would not work. Also, I'm wondering whether it's OK to set the global default to "edge".

As for the templates, do whatever you feel is best.

docs/module/stp.md
jbemmel: added a test case for node and vlan level ```stp.port_type```

docs/module/stp.md
jbemmel: Changed it to ```none```

netsim/modules/stp.py
jbemmel: It is now, conditionally based on the node supporting the feature and the interface supporting STP port_type config

netsim/ansible/templates/stp/cumulus.j2
jbemmel: added ```if defined```

ipspace:
Looks like we got merge requests on this PR. Please resolve them first.

PR #1618, opened 2024-12-06
Replace use of 'netlab_provider' with 'node_provider' in templates

jbemmel:
Needed to fix multi-provider deployments; FRR clab nodes in a libvirt topology currently try to run ```modprobe``` for example


ipspace (approved):
Good catch. Thanks a million!

PR #1619, opened 2024-12-07
Consistently replace '_' with '-' in hostnames

jbemmel:
It is not a valid FQDN character, and some platforms (like Cumulus NVUE) fail with an error

We could consider restricting node names to be valid FQDNs (of at most 16 characters), and/or apply the 'replace' normalization in Python rather than all templates (either define a ```node.hostname``` property and use that everywhere instead of ```inventory_hostname```, or change ```inventory_hostname``` (and issue an error when troublesome node names are used)

### DIFF
--- a/netsim/ansible/templates/initial/arcos.j2
+++ b/netsim/ansible/templates/initial/arcos.j2
@@ -1,4 +1,4 @@
-system hostname {{ inventory_hostname }}
+system hostname {{ inventory_hostname.replace('_','-') }}
 !
 !
 interface loopback0
--- a/netsim/ansible/templates/initial/asa.j2
+++ b/netsim/ansible/templates/initial/asa.j2
@@ -1,4 +1,4 @@
-hostname {{ inventory_hostname }}
+hostname {{ inventory_hostname.replace('_','-') }}
 domain-name lab.local
 !
 {% for k,v in hostvars.items() if k != inventory_hostname and v.af.ipv4|default(False) %}
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -50,7 +50,7 @@
 
 - set:
     system:
-      hostname: {{ inventory_hostname }}
+      hostname: {{ inventory_hostname.replace("_","-") }}
     interface:
       eth0:
         ip:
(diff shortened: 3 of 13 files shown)


ipspace (changes requested):
There is no need for such a change. Many platforms accept both.

Whichever platform cannot deal with a valid node identifier should solve its own issues in a configuration template.

Also:

* Did you test all these changes? Are you sure they work?
* Why did you not change the hostname mappings if you want to enforce consistency?

Just because you have a problem with NVUE does not mean everything else has to change. Submit a PR for those platforms that cannot handle '_' and I'll merge it.

jbemmel:
> There is no need for such a change. Many platforms accept both.
> 
> Whichever platform cannot deal with a valid node identifier should solve its own issues in a configuration template.
> 
> Also:
> 
> * Did you test all these changes? Are you sure they work?
> * Why did you not change the hostname mappings if you want to enforce consistency?
> 
> Just because you have a problem with NVUE does not mean everything else has to change. Submit a PR for those platforms that cannot handle '_' and I'll merge it.

Many platforms accept both, Ubuntu for example simply ignores the '_' characters and a ```r_1``` hostname becomes ```r1```.

Most of the platforms I have access to - eos, cumulus, dell os10, linux - use the character substitution. The immediate issue was with Cumulus NVUE, but - for example - Cisco AVA is [reported](https://github.com/GNS3/gns3-gui/issues/3524) to have similar issues.

I think users would be better off if they stayed away from using '_' in hostnames, lacking that the next best thing is - imho - to replace '_' with '-', rather than let it be ignored and hope it doesn't cause issues. We could make that configurable, while I did not test the other platforms I think it is reasonable to expect the hostname with a '-' to have a better chance to work.

I can submit a PR for only Cumulus NVUE, I do not have access to most other platforms so I don't really care for myself. But I thought it'd be nice for the community

ipspace:
> I can submit a PR for only Cumulus NVUE

You should have done that in the first place.

> I do not have access to most other platforms so I don't really care for myself.

Copiously illustrated in #1623.

Also: stop "fixing" stuff you don't care about and have no intention of testing.

> But I thought it'd be nice for the community

Maybe it's just me, but in my personal opinion the community might appreciate a tested and working software more.

ipspace:
Still waiting for the changes to be implemented. Converting to draft.

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/1682

PR #1621, opened 2024-12-08
Change the FRR Vagrant box to Debian bookworm

### LINKED ISSUE #1595: Find a better VM to run FRRouting
FRRouting currently uses Ubuntu 20.04 as later versions of Ubuntu tend to be multi-interface-challenged. Ubuntu 20.04 comes with Python 3.8 which barely works with Ansible 11.0 (see #1590 for details).

We need to find a long-term solution for FRR virtual machines (Debian?)

ipspace:
Fixes #1595

Also: includes a YAML validation fix for an unrelated file

### DIFF
--- a/netsim/ansible/tasks/readiness-check/vsrx.yml
+++ b/netsim/ansible/tasks/readiness-check/vsrx.yml
@@ -7,6 +7,7 @@
 - name: Execute local ssh command to check vSRX readiness
   local_action:
     module: command
+    # yamllint disable-line rule:line-length
     cmd: "sshpass -p '{{ ansible_ssh_pass }}' ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {{ ansible_user }}@{{ ansible_host }} 'show system information'"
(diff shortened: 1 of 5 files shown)


jbemmel (approved):
Should we change it for the generic Linux device as well?
```netsim/devices/linux.yml``` is the last one still using old Ubuntu

jbemmel (approved):
Should we also change the default Linux device?

jbemmel:
> Should we also change the default Linux device?

See https://github.com/ipspace/netlab/pull/1622

ipspace:
> Should we also change the default Linux device?

Absolutely. Same challenge.

jbemmel:
In hindsight, this PR is missing some required changes to SSH private key handling - see https://github.com/ipspace/netlab/pull/1623

ipspace:
> In hindsight, this PR is missing some required changes to SSH private key handling - see #1623

In hindsight, this PR works and does not break everything else as opposed to #1623.

PR #1623, opened 2024-12-08
Linux: Update libvirt image to debian/bookworm64 and clab image to python:3.11-alpine

jbemmel:
* Update Python to version 3.11, match across providers
* Fix SSH private key handling for Libvirt provider (set ```ansible_ssh_private_key_file``` in inventory)
* Update ```netlab_linux_distro``` to debian, add custom init script

I commented out the DNS fixes for systemd resolvd, not sure if those are still needed. We may want to refactor ```netplan``` init into its own script - I copied the Ubuntu bits

### DIFF
--- a/netsim/ansible/templates/initial/linux/debian.j2
+++ b/netsim/ansible/templates/initial/linux/debian.j2
@@ -1,6 +1,25 @@
 echo -n 'Starting initial config ' && date
 
+# Set persistent hostname
+hostnamectl set-hostname {{ inventory_hostname }}
+
+# Update APT and install netplan
+apt-get update -qq
+apt-get install -qq nplan
+
 # (Overwrite resolved config to remove DNS stuff)
+cat <<SCRIPT >/etc/netplan/01-netcfg.yaml
+network:
+  version: 2
+  renderer: networkd
+  ethernets:
+    eth0:
+      dhcp4: true
+      dhcp6: false
+      optional: true
+SCRIPT
+netplan apply
+
 # cat <<SCRIPT > /etc/systemd/resolved.conf 
 # [Resolve]
 # DNS=
@@ -11,13 +30,6 @@ echo -n 'Starting initial config ' && date
 # DNSStubListener=yes
 # SCRIPT
 # systemctl restart systemd-resolved
-
-# Set persistent hostname
-hostnamectl set-hostname {{ inventory_hostname }}
-
-# Update APT and install netplan
-apt-get update -qq
-apt-get install -qq nplan
 
 {% if netlab_net_tools|default(False) %}
 #
@@ -72,7 +84,7 @@ net.ipv6.conf.{{ l.ifname }}.disable_ipv6=0
 SCRIPT
 sysctl -p /etc/sysctl.d/10-netsim.conf
 
-# Loopback addressing, JvB commented out
+# Loopback addressing
 {% if loopback.ipv4 is defined or loopback.ipv6 is defined %}
 cat <<SCRIPT > /etc/netplan/02-loopback.yaml
 network:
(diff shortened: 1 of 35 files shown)


ipspace:
Did this work for you? I can't even connect to the Linux VM due to a mismatch in SSH authentication expectations. This is from the Ansible initial configuration playbook:

```
TASK [template] ********************************************************************************************************************
fatal: [h1]: UNREACHABLE! => changed=false
  msg: 'Failed to authenticate: Bad authentication type; allowed types: [''publickey'']'
  unreachable: true
fatal: [h2]: UNREACHABLE! => changed=false
  msg: 'Failed to authenticate: Bad authentication type; allowed types: [''publickey'']'
  unreachable: true
```

**netlab connect** reports a very similar error.

netsim/templates/provider/libvirt/Vagrantfile.j2
```diff
@@ -1,6 +1,7 @@
 VAGRANT_COMMAND = ARGV[0]
 
 Vagrant.configure("2") do |config|
+  config.ssh.insert_key = true
```
ipspace, line 4: We have 21 Vagrant boxes and most of them are not Linux and thus cannot have their SSH key replaced, and you go ahead and change the system default? Congratulations, great thinking.

FWIW, according to Vagrant documentation, the "true" value is equal to the default behavior, so this was completely unnecessary, but great job anyway.

netsim/providers/libvirt.yml
```diff
@@ -41,3 +41,6 @@ attributes:
     uplink: str
   global:
     providers:
+
+inventory:
+  ansible_ssh_private_key_file: .vagrant/machines/{ name }/libvirt/private_key
```
ipspace, line 46: This is set for **all** Vagrant devices and probably breaks **every single device that uses Vagrant default key** (as in: most of them, I did a spot check on Cisco IOSv). Awesome. Just awesome.

Would you once in a while stop for a microsecond, think about the wider implications of your changes, and make the minimum amount of changes necessary?

netsim/ansible/templates/initial/linux/debian.j2
```diff
@@ -0,0 +1,152 @@
+echo -n 'Starting initial config ' && date
```
ipspace, line 1: So you just copied the "ubuntu.j2" file, added "get netplan" and removed a few things. So when we figure out something needs to be changed and we fix it in one of the files everyone knows it has to be fixed in the other file as well, right?

There's a reason Jinja2 has "include" functionality.

netsim/cli/connect.py
```diff
@@ -103,7 +103,9 @@ def ssh_connect(
   if data.netlab_ssh_args:
     c_args.extend(data.netlab_ssh_args.split(' '))
 
-  if data.ansible_ssh_pass:
+  if data.ansible_ssh_private_key_file:
+    c_args.extend(['-i', strings.eval_format(data.ansible_ssh_private_key_file,{'name': data.host})])
```
ipspace, line 107: What's wrong with passing the whole node data, and if you need the 'name' parameter, you can adjust the 'adjust_inventory_host' call to retain it. This is yet another example of the "let's fix my immediate problem" mentality.

ipspace (changes requested):
This is probably one of the worst rush jobs (at least considering its impact) I've seen so far.

ipspace:
After we decide how to deal with Vagrant private keys (#1628) we'll implement a "minimum impact" move to Debian/Bookworm.

netsim/templates/provider/libvirt/Vagrantfile.j2
jbemmel: I figured it's the system default, so this only makes explicit what's already happening. However, Vagrant has an elaborate merging process for its settings - see https://developer.hashicorp.com/vagrant/docs/vagrantfile - and so we need to make sure it's set to ```true``` for the private key file to exist in the location we expect it to, in case users may override system defaults. I experimented with setting this to ```false```

jbemmel:
> This is probably one of the worst rush jobs (at least considering its impact) I've seen so far.

FWIW, I thought I was rushing to fix something that was also affecting your FRR PR. You could have just told me that I missed the brute-force password authentication override patch

PR #1624, opened 2024-12-08
Linux: add lag support and bonding plugin for non-lag cases

jbemmel:
* Refactor and reuse FRR logic for bond creation on Linux, both for the ```lag``` module and for the ```bonding``` plugin
* Support all the Linux bonding modes, split into:
  + lag module: [ 802.3ad, balance-rr, balance-xor, broadcast ] - these require switch support
  + bonding plugin: [ active-backup, balance-tlb, balance-alb ]
  
In all cases, bonding devices are created in ```initial``` such that any IP addresses can be assigned. Bond/lag members are then added in the respective ```lag``` module and ```bonding``` plugin

Executes in netns for containers -> no need to install iproute2

Tested:
* netlab up integration/lag/05-bonding-active-standby.yml -p clab
* netlab up integration/lag/05-bonding-active-standby.yml -p libvirt
* netlab up integration/lag/06-host-mlag-anycast-gateway.yml -p clab -d dellos10


ipspace (changes requested):
A few minor details plus a major problem: we don't get the Linux static routes for the lab pools (they are supposed to point to the first data-plane interface), so the Linux host cannot reach anything but the adjacent router.

Fix the nits and I'll add the static routes part.

netsim/devices/linux.yml
```diff
@@ -9,6 +10,8 @@ features:
       ipv6: true
     server: true
     relay: true
+  lag:
```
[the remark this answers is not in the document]
jbemmel: Current FRR: https://github.com/ipspace/netlab/blob/7d80e7cabd32bcd16ddc668d6ff662cdab63ffe2/netsim/devices/frr.yml#L23

I vaguely remember you suggesting to use the generic mechanism, looks like that change may have gotten lost somehow.

Current template manually loads the kernel module with custom parameters in case of a VM, could add ```kmods``` for lag (to both Linux and FRR)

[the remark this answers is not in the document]
ipspace: > Current FRR

... is exactly what you need. The mapping between **lag** module and the **bonding** driver is here:

https://github.com/ipspace/netlab/blob/dev/netsim/providers/clab.yml#L24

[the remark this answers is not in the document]
jbemmel: Didn't you add ```kmods``` 4 months ago? https://github.com/ipspace/netlab/commit/4cf8c7b1d1dc13a32953111f610f7a85c785d42e

netsim/ansible/templates/lag/linux.j2
```diff
@@ -0,0 +1,51 @@
...
+{%     if node_provider != 'clab' %}
+#
+# Make sure 'bonding' module is loaded
+#
+if [[ ! -e /sys/module/bonding ]]; then
+modprobe bonding miimon=100 mode=802.3ad lacp_rate=fast
+fi
+{%     endif %}
+if [[ ! -e /sys/class/net/{{l.ifname}} ]]; then
+ip link add dev {{l.ifname}} type bond mode {{_m}}
```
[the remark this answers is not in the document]
jbemmel: added caveat

netsim/devices/linux.yml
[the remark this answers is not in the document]
jbemmel: I remember suggesting to remove it and put it back with lag support, but you refused arguing that we might forget to put it back…

ipspace:
> Fix the nits and I'll add the static routes part.

Static route issue is tracked in #1641. I will probably go for a simple fix in the configuration template (that code is ancient)

jbemmel:
Ready to add the static route bits

docs/caveats.md
```diff
@@ -280,6 +280,10 @@ We're not testing Fortinet implementation as part of the regular integration tes
 * Junos cannot have more than one loopback interface per routing instance. Using **loopback** links on Junos devices will result in configuration errors.
 * Junos configuration template configures BFD timers within routing protocol configuration, not on individual interfaces
 
+(caveats-linux)=
```
ipspace, line 283: All other Linux caveats are in 'labs/linux.md'. Please move this one there and insert a pointer in this file (all other documentation should also point to that file, and at least some of the platform tables do).

docs/plugins/bonding.md
```diff
@@ -0,0 +1,66 @@
+(plugin-bonding)=
+# Host-side link bonding
```
ipspace, line 2: Use consistent heading capitalization

docs/plugins/bonding.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Using the Plugin
+
+* Add `plugin: [ bonding ]` to the lab topology.
+* Include the **bonding.ifindex** attribute in any links that need to be bonded
+
+### Supported attributes
+
+The plugin adds the following attributes:
+* **bonding.mode** (string, one of active-backup, balance-tlb, or balance-alb) -- the bonding mode to use, default `active-backup`
```
ipspace, line 22: global, node, or interface?

docs/plugins/bonding.md
```diff
@@ -0,0 +1,66 @@
...
+
+Additional interface level attributes:
+* **bonding.ifindex** (int,mandatory) -- the interface index for the bonding device; links with matching ifindex are bonded together
+* **bonding.primary** (bool) -- optional flag to mark this interface as primary, default *False*
```
ipspace, line 26: Document what happens if this is not set on any interface

netsim/augment/links.py
```diff
@@ -1168,7 +1168,8 @@ def transform(link_list: typing.Optional[Box], defaults: Box, nodes: Box, pools:
       continue
 
     set_link_bridge_name(link,defaults)
-    link_default_pools = ['p2p','lan'] if link.type in ['p2p','lag'] else ['lan']
+    link_default_pools = ['p2p','lan'] if link.type=='p2p' or (
+                                          link.type=='lag' and not 'vlan' in link) else ['lan']
```
ipspace, line 1172: This has been obsoleted by #1673

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+from netsim.augment import devices
+
+_config_name = 'bonding'
+
+'''
+add_bond_interfaces - append interface data to node.interfaces for bonding template to read and implement
+'''
+def add_bond_interfaces(node: Box, bonds: dict[int,Box]) -> None:
+  for c,(ifindex,bond) in enumerate(bonds.items()):
+    ifname = f'bond{ifindex}'  # XXX hardcoded, could make this a device template attribute
```
ipspace, line 14: You should ;)

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+    #
+    # TODO: Could make sure the layout is exactly like the 'lag' module needs it, to reuse provisioning logic
```
ipspace, line 17: You call...

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if intf.get('bonding.primary',False):
+          bonds[node.name][bond_ifindex]['primary'] = intf.ifname
+
+        intf.neighbors = [ { 'ifname': i.ifname, 'node': i.node } for i in link.interfaces if i.node!=node.name ]
+        intf.type = 'p2p'
+        intf.prefix = False                            # L2 p2p interface
+        intf.pop('name',None)
+
+  for node in topology.nodes.values():                 # For each node
```
ipspace, line 80: I will not even try to understand what's going on here. I hope it works.

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+Apply plugin config to nodes with interfaces marked with 'bonding.ifindex', for devices that support this plugin
+'''
+def post_transform(topology: Box) -> None:
+# def post_link_transform(topology: Box) -> None: # Use 'pre_node_transform' such that only 1 IP address gets assigned, nothing to move?
+  global _config_name
+  bond_mode = topology.get('bonding.mode','active-backup')
+  bonds : Box = data.get_empty_box()                   # Map of bonds per node, indexed by bonding.ifindex
+  for node in topology.nodes.values():
+    features = devices.get_device_features(node,topology.defaults)
```
ipspace, line 47: This seems to work along the lines of "ignore bonding attributes on nodes that do not support bonding". Not exactly the best approach to the Principle of Least Astonishment.

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+          continue
+
+        clone = data.get_box(intf)
+        if node.name in bonds and bond_ifindex in bonds[node.name]:
+          bonds[node.name][bond_ifindex]['members'].append( clone.ifname )
+          for att in ['ipv4','ipv6']:
+            intf.pop(att,None)
+        else:
+          mode = intf.get('bonding.mode',bond_mode)
+          bonds[node.name][bond_ifindex] = { 'interfaces': intf.neighbors, 'members': [ clone.ifname ], 'mode': mode }
```
ipspace, line 67: You're setting "interfaces" key to neighbor list. That's at least confusing.

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+          for att in ['ipv4','ipv6']:                  # Move any ips (from first member link)
+            if att in intf:
+              bonds[node.name][bond_ifindex][att] = intf.pop(att,None)
+
+        if intf.get('bonding.primary',False):
+          bonds[node.name][bond_ifindex]['primary'] = intf.ifname
+
+        intf.neighbors = [ { 'ifname': i.ifname, 'node': i.node } for i in link.interfaces if i.node!=node.name ]
```
ipspace, line 75: Is this "clean up the neighbor attributes" (if so, why?) or "avoid links to myself" (in which case you should report an error)

netsim/extra/bonding/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+        intf.type = 'p2p'
```
ipspace, line 76: Why exactly do you need this?

netsim/modules/lag.yml
```diff
@@ -12,19 +12,21 @@ attributes:
   global:
     lacp: { type: str, valid_values: [ "off", "slow", "fast" ] }
     lacp_mode: { type: str, valid_values: [ "passive", "active" ] }
-    mode: { type: str, valid_values: [ "802.3ad", "balance-xor", "active-backup" ] }
+
+    # All Linux bonding modes that require lag configuration on the switch side
```
ipspace, line 16: The only bonding mode that requires switch configuration is "802.3ad" or am I missing something? Why exactly do you think we need Linux bonding modes in the generic LAG module?

ipspace (changes requested):
A bunch of confused comments and a few minor changes. Also, it looks like it's time for a rebase.

netsim/modules/lag.yml
jbemmel: Based on https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/networking_guide/overview-of-bonding-modes-and-the-required-settings-on-the-switch#overview-of-bonding-modes-and-the-required-settings-on-the-switch

```
Table 7.1. Switch Configuration Settings Depending on the Bonding Modes
Bonding Mode	Configuration on the Switch
0 - balance-rr	Requires static Etherchannel enabled (not LACP-negotiated)
1 - active-backup	Requires autonomous ports
2 - balance-xor	Requires static Etherchannel enabled (not LACP-negotiated)
3 - broadcast	Requires static Etherchannel enabled (not LACP-negotiated)
4 - 802.3ad	Requires LACP-negotiated Etherchannel enabled
5 - balance-tlb	Requires autonomous ports
6 - balance-alb	Requires autonomous ports
```

So 0, 2, 3 and 4 require lag configuration on the switch side, with only 4 including LACP.

1, 5 and 6 are handled by the bonding plugin (server-side config only)

jbemmel:
One issue: We need to execute the ```ip``` command on the host, from within the container namespace context. This happens for ```initial``` during bond device creation, but the bond members are added during ```lag``` module execution which (currently) happens inside the container (hence depends on ```ip``` etc.)

I'm thinking we need a mechanism to control the execution context of provisioning scripts in case of Linux containers: Either on the host (for ```initial``` and ```lag``` modules), or "normal" as is done for ```dhcp```. Perhaps a convention in the filename: e.g. ```linux-host.j2``` for network namespace execution, and regular ```linux.j2``` for current practice

Thoughts?

netsim/extra/bonding/plugin.py
jbemmel: I forgot that plugins don't automatically get these checks, unlike modules

netsim/extra/bonding/plugin.py
jbemmel: Guess we don't, removed

netsim/extra/bonding/plugin.py
jbemmel: It's updating the neighbor interface names, after all bonds have been processed

netsim/extra/bonding/plugin.py
jbemmel: Yes, the cleanup is needed because the bond now gets all the indirect neighbors - like vlan peers at other hosts - while the bond member links only get directly attached peers on the link

netsim/extra/bonding/plugin.py
jbemmel: It should be 'neighbors'

docs/caveats.md
jbemmel: done

docs/plugins/bonding.md
jbemmel: ok

docs/plugins/bonding.md
jbemmel: ok

netsim/augment/links.py
jbemmel: merged

ipspace:
Looks like you're still working on this one. When you're done, please mark it as "ready for review"

jbemmel:
> Looks like you're still working on this one. When you're done, please mark it as "ready for review"

It's ready, all I did was rebase it. I'm waiting on this one before merging #1646

PR #1626, opened 2024-12-09
STP: Don't include module attributes with each neighbor

jbemmel:
Interpret ```link_to_neighbor: False``` to mean "don't include interface attributes in neighbors"

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -784,6 +784,11 @@ def create_node_interfaces(link: Box, addr_pools: Box, ndict: Box, defaults: Box
     print(f'... link data: {link}')
     print(f'... interface data: {interfaces}\n')
 
+  # link_to_neighbor returns True iff a given module's attributes should be included with each neighbor
+  def link_to_neighbor(m:str) -> bool:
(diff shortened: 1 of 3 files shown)


ipspace:
And why exactly is that behavior bothering you?

jbemmel:
> And why exactly is that behavior bothering you?

Someone told me to "stop for a microsecond, think about the wider implications of your changes, and **make the minimum amount of changes necessary**?"

We could call it data model hygiene - I prefer to not include attributes that don't get used

ipspace:
> Someone told me to "[...] **make the minimum amount of changes necessary**?"

And yet you feel the urge to change the core functionality for cosmetic reasons. There's no reason to merge this.

FWIW, a plugin COULD use the neighbor **stp** attributes to detect mismatches ;))

jbemmel:
> FWIW, a plugin COULD use the neighbor **stp** attributes to detect mismatches ;))

FWIW, such a hypothetical plugin could also modify the ```link_to_neighbor``` setting of the ```stp``` module to get its data included

PR #1636, opened 2024-12-10
node.clone plugin: Combine tests into 1 file

jbemmel:
As the tests were getting hard to read due to the nature of the cloning feature, in particular when combined with lag, I split them up so they were more readable. 

This PR combines them into 1 test case


jbemmel:
It got messed up, I hit that issue with ```lag.ifindex``` as I was testing ```node.clone``` and things went down hill from there.

I prefer to have the 2 files, cleaned up. Will submit another PR

PR #1638, opened 2024-12-10
Fix node.clone 'count' attribute

jbemmel:
Not sure how this got changed, and strangely it doesn't affect the topology tests so attributes aren't getting validated as one would expect

### DIFF
--- a/netsim/extra/node.clone/defaults.yml
+++ b/netsim/extra/node.clone/defaults.yml
@@ -6,6 +6,6 @@ clone:
 
   attributes:
     node:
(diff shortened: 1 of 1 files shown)


ipspace:
A) The way you specified attributes is all wrong (and I completely missed that during code review, my bad). The plugin YAML file is merged with the defaults, see `extra/fabric/defaults.yml` for an example.

B) Your code runs before the node attributes are validated and you remove the "clone" attribute anyway, so nobody ever sees it.

You could either remove the whole thing (because it doesn't matter) or implement your own validation. For example, you could define a new data type (like the "routing" module does) and validate just that.

PR #1645, opened 2024-12-13
Use --allow-downgrades to install specific Vagrant version

jbemmel:
There is a later release, if we want people to use this specific one the user may need to downgrade


ipspace:
Do we want people to use that specific release?

jbemmel:
> Do we want people to use that specific release?

I don't know, I did not pin the version but I figured it is there for a reason? Vagrant just came out with a new release and the install script stopped working

ipspace:
OK, let's put this in perspective:

* We're pinning the Vagrant version to 2.4.0-1
* 2.4.1 came out in January (11 months ago)
* 2.4.2 came out on November 2nd quickly followed by 2.4.3 (November 12th).
* Vagrant is usually not included in a Linux distro, so it has to be installed by some installation script.
* Nobody so far found this to be a problem (while we got several reports of Cumulus/Ansible SNAFU)

The big question thus remains: what exactly were you doing to hit this bug and is it relevant to a regular user?

ipspace:
It might be more productive to start with "An upgrade broke my Vagrant setup and **netlab install libvirt** could fix it... but we need the `--allow-downgrades` option to make it work"... or not 🤷‍♂️

PR #1646, opened 2024-12-13
lag: Remove 'lag' link type, cleanup virtual links

jbemmel:
Conceptually a 'lag' link type does not make a whole lot of sense; the ```lag``` module temporarily uses a link object as a grouping construct until the node lag interfaces are created, but there is no reason to keep this around once that is done

This PR cleans things up a little:
* Replace the "lag" link.type with a "_virtual_lag" flag, and remove those links during post processing
* Link lag members to the parent device through ```lag.ifindex```, not ```linkindex``` (which is now removed from the lag interface)
* Temporarily uses '_type: lag' interface attribute to pass validation, then renames it to 'type' for backwards compatibility

Includes various bug fixes:
* Testing and handling of lag attribute inheritance (interface level attributes weren't being considered)
* Remove MTU from lag interfaces
* Refactoring of lag module structure (separate 'lag' interface creation)


netsim/defaults/attributes.yml
```diff
@@ -80,6 +80,7 @@ interface:                  # Interface (node-to-link attachment) attributes
   ipv6: { type: ipv6, use: interface }
   ifindex: int
   ifname: str
+  type: { type: str, valid_values: [ p2p, lan, stub, loopback, vlan_member, svi, lag ] }
```
ipspace, line 83: The "type" attribute is set when the interface is created during the link transformation, so it does not need to be defined (the validation is done before that).

It also makes no sense to allow the user to set the interface type in the lab topology, and not having an attribute defined helps to enforce that ;) This needs to be removed.

netsim/augment/links.py
```diff
@@ -1167,7 +1167,7 @@ def transform(link_list: typing.Optional[Box], defaults: Box, nodes: Box, pools:
       continue
 
     set_link_bridge_name(link,defaults)
-    link_default_pools = ['p2p','lan'] if link.type in ['p2p','lag'] else ['lan']
+    link_default_pools = ['p2p','lan'] if link.type in ['p2p','virtual_lag'] else ['lan']
```
ipspace, line 1170: It might be better to implement #1647 than to tweak this list every time we have a new idea.

ipspace (changes requested):
The "type" interface attribute should not be defined, otherwise it looks reasonable.

ipspace:
Also, what happened to "I will submit a PR when I'm done programming"? Got three notifications about this PR.

netsim/defaults/attributes.yml
jbemmel: When I leave it out, I get this:
```
Reading topology/input/lag-l2.yml
[ERRORS]  Errors found in lag-l2.yml
[VALUE]   links: attribute links[1].r1.type has invalid value(s): lag
[HINT]    valid values are: lan,p2p,stub,loopback,tunnel,vlan_member,virtual_lag
          use 'netlab show attributes interface' to display valid attributes
[VALUE]   links: attribute links[1].r2.type has invalid value(s): lag
[HINT]    valid values are: lan,p2p,stub,loopback,tunnel,vlan_member,virtual_lag
```

So either it does need to be defined, or that validation logic needs to be modified

jbemmel: The ```lag``` module is currently adding interfaces with type ```lag``` during ```module_pre_transform``` - need to postpone that to ```module_pre_link_transform```

jbemmel: ...which does not work either, because then the ```vlan``` module must be changed from ```pre_link_transform``` to ```link_pre_link_transform``` (as it assumes link.interfaces is populated) but then the attribute validation stage misses user provided attributes

jbemmel: Finally arrived at the following solution:

1. Create lag interfaces during ```module_pre_transform``` such that they are present when ```vlan.link_pre_transform``` executes
2. Use ```_type: lag``` to create them such that validation ignores the attribute, then rename in ```link_pre_link_transform``` (after validation)

It is not possible to create the interfaces after validation, because validation needs to consider user provided interface attributes, some of which depend on vlan transformation

ipspace:
Looks like you're still working on this one. Change it to "ready for review" when you're done.

ipspace:
What's going on with this one? Is it ready to merge (I won't even try to review it)?

jbemmel:
> What's going on with this one? Is it ready to merge (I won't even try to review it)?

Waiting on the other lag PR to be merged, then I'll update this one

Issue #1647, opened 2024-12-13
Decouple default pool selection from link type


ipspace:
From the very early days, _netlab_ used **link.type** (p2p, stub, lan) to select the default pool to address the link. In the meantime, **link.type** evolved into an attribute that is either used to provision the physical aspects of the link or signals special link types like **vlan_member** or **loopback**.

It might be best to decouple the two and use the conditions initially used to set link.type to set default pools:

* Two nodes, both routers, no **gateway** attribute → `p2p`
* One router → `stub`
* Anything else → `lan`

This would be a breaking change for multi-provider P2P links (that got prefixes from the **lan** pool due to the multi-provider logic) and to old lab topologies that use **link.type** as pool selection mechanism.

Any thoughts?

jbemmel:
There is no ```stub``` pool?

PR #1648, opened 2024-12-14
links: Decouple default address pool from link type

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1647

* Modifies ```dual-stack``` test case to use ```pool: lan``` instead of ```type: lan```
* ```vrrp-interface-granularity``` test no longer references ```addressing.p2p.prefix``` setting


jbemmel:
Obsolete

PR #1653, opened 2024-12-14
Proposal: rename 'interface.type' to 'interface._type' during transformation

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/1650

In preparation for https://github.com/ipspace/netlab/pull/1646 it would greatly simplify things if this internal attribute were named starting with "_" such that validation is properly skipped


jbemmel:
Never mind - I think I have an even simpler solution

PR #1659, opened 2024-12-17
Enable deployment of topologies containing 'none' devices, remove ghosts from links

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/1657

```none``` devices are useful for validation, but they could also be used to deploy scaled deployments representing 100s of nodes. This PR treats 'none' devices as 'unmanaged' for the purpose of ```netlab up```

* Modify the node.clone plugin to merge settings from existing nodes, to allow selective override of attributes.
   For example: model a switch with 32 servers, but launch only 2 of those as 'linux' containers

* Fix ```netlab up``` for clab topologies containing ```unmanaged``` nodes (the ```links``` section would contain semi-connected entries), by removing ghosts from links as well.
Test:  ```netlab up topology/input/unmanaged-device.yml -p clab -v```

Handling unmanaged devices could possibly be done in different ways, without removing them from the links. ```libvirt``` deployment does succeed. For example, we could modify clab.j2 to use Containerlab ```dummy``` devices (https://containerlab.dev/manual/topo-def-file/#dummy)

The current output in https://github.com/ipspace/netlab/blob/dev/tests/topology/expected/unmanaged-device.yml is invalid, as there is a link that refers to a non-existent node 'external'

### CODE REVIEW

File: netsim/augment/nodes.py [hunk 1]
```diff
@@ -466,10 +466,24 @@ def cleanup(topology: Box) -> None:
...
+    return node.get("unmanaged",False) or node.device=='none'
+
+  # Remove unmanaged and/or 'none' type nodes from links
+  for link in topo_copy.get('links',[]):
+    count = len(link.interfaces)
+    link.interfaces = [ i for i in link.interfaces if not is_ghost(topo_copy.nodes[i.node]) ]
+    new_count = len(link.interfaces)
+    if new_count < count:
+      link.node_count -= (count-new_count)
+      if link.node_count==1:
```
jbemmel, line 483: Should remove the link if the count drops to 0

netsim/augment/nodes.py
```diff
@@ -466,10 +466,24 @@ def cleanup(topology: Box) -> None:
...
+        link.type = 'dummy'
+        link.bridge = f'ghost_{link.linkindex}'
```
ipspace, line 485: This won't work with the multilab plugin. Please rework.

netsim/extra/node.clone/plugin.py
```diff
@@ -105,14 +105,11 @@ def clone_node(node: Box, topology: Box) -> None:
...
 
-    if clone.name in topology.nodes:                                         # Check for overlapping names
-      log.error("Generated clone name '{clone.name}' conflicts with an existing node",
-                category=AttributeError, module='node.clone')
-      return
-
     clone.interfaces = []                                                    # Start clean, remove reference to original node
     if 'id' in node:
       clone.id = node.id + c - 1                                             # Update any explicit node ID sequentially
+    if clone.name in topology.nodes:                                         # Check for existing nodes that may override
```
ipspace, line 111: This seems to be a change unrelated to the topic of this PR.

ipspace (changes requested):
Your idea to use a hard-coded bridge name won't work with the multilab plugin. Also, the integration tests do not belong to the "initial device configuration" suite. Move them into a new directory (if needed) of the "platform-integration" tree.

netsim/augment/nodes.py
jbemmel: done

netsim/augment/nodes.py
jbemmel: Using ```links.set_link_bridge_name```

netsim/extra/node.clone/plugin.py
jbemmel: It was the driving use case for the change, but moved to a separate PR

jbemmel:
Closing as a similar effect can be achieved by using ```unmanaged``` nodes

PR #1660, opened 2024-12-17
EOS: Fix variable name typo in bfd template

jbemmel:
bfd echo wasn't being configured when requested on a link


ipspace:
Thank you!

PR #1668, opened 2024-12-19
OSPF v2/v3 router priority (CL, EOS, FRR, IOS, NXOS)

ipspace:
Includes integration tests and minor tweaks to the OSPF validation plugins

### DIFF
--- a/tests/integration/ospf/ospfv3/31-priority.yml
+++ b/tests/integration/ospf/ospfv3/31-priority.yml
@@ -1,0 +1,66 @@
+---
+message: |
+  This lab tests the OSPF interface timers. The device under test must establish
+  an OSPF adjacency with the probe having custom OSPF hello/dead timers.
+
+  The test assumes the device under test supports OSPFv2 in VRFs and configurable
+  OSPF router ID.
+
+vrfs:
+  customer:
+
+groups:
+  probes:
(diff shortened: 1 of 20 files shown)


tests/integration/ospf/ospfv3/31-priority.yml
```diff
@@ -0,0 +1,66 @@
+---
+message: |
+  This lab tests the OSPF interface timers. The device under test must establish
```
jbemmel, line 3: OSPF interface priority

jbemmel (commented):
VRF test should be revised

ipspace:
> VRF test should be revised

Yeah, I have no idea why it works ;)

jbemmel:
#1667 has proposed changes - I think it's better to check the neighbor priority directly

ipspace:
> #1667 has proposed changes - I think it's better to check the neighbor priority directly

Absolutely. I missed that bit of information.

I will cherry-pick the test into 'dev' branch while we're waiting for @ssasso to review #1667.

ipspace:
> #1667 has proposed changes - I think it's better to check the neighbor priority directly

FWIW, it would have been so much easier (for me) if you also fixed the OSPFv3 test 😜

Issue #1670, opened 2024-12-19
BUG: Inconsistent "ibgp_localas" key in bgp.community


jbemmel:
https://github.com/ipspace/netlab/blob/0bc588eb532e83d60c484bb0d9f0c4656ed0d81e/netsim/modules/bgp.py#L605

The ```bgp.community``` attribute is indexed by bgp.neighbor.type, which is one of https://github.com/ipspace/netlab/blob/0bc588eb532e83d60c484bb0d9f0c4656ed0d81e/netsim/modules/bgp.py#L58

Lines 605-606 should be:
```
if 'localas_ibgp' not in clist:
    clist.localas_ibgp = clist.ibgp
```

Issue #1663, opened 2024-12-18
[BUG] Unable to explicitly activate localas_ibgp sessions


jbemmel:
Consider the following subtopology:
```
module: [ bgp ]

bgp.sessions:
  ipv4: [ebgp]

nodes:
  a:
    bgp.as: 65001
  b:
    bgp.as: 65002
  c:
    bgp.as: 65003

links:
- a-b
- a-c
- b:
    bgp.local_as: 65003
  c:
```

If left as default, ```bgp.sessions.ipv4``` is ```['ibgp','ebgp','localas_ibgp']```. However, "localas_ibgp" is not a valid value in https://github.com/ipspace/netlab/blob/c88777ae908877a4074cb2b522b9e51fa9d21ec9/netsim/modules/bgp.yml#L31

PR #1671, opened 2024-12-20
bgp: Fix localas_ibgp inconsistencies and issues

jbemmel:
This PR fixes some corner case issues around ```localas_ibgp``` sessions:
* The ebgp.multihop plugin does not handle ```localas_ibgp``` sessions that might occur
* There is a misspelled key used to populate the ```bgp.community``` dictionary
* Users by default get settings that include ```localas_ibgp```, but cannot explicitly configure settings for this type of sessions

Fixes:
* #1663 
* #1670 

Note: The fact that the ```bgp.community``` key mismatch has gone undetected for 11 months implies a lack of coverage for CI/CD test cases, happy to add a test case if it is deemed important enough

The PR adds an entry to the ```integration/bgp.multihop/01-global``` test but no actual validation; it might be better to put this in a dedicated test case instead

### DIFF
--- a/netsim/extra/ebgp.multihop/plugin.py
+++ b/netsim/extra/ebgp.multihop/plugin.py
@@ -202,13 +202,6 @@ def fix_vrf_loopbacks(ndata: Box, topology: Box) -> None:
       ngb._source_ifname = ndata.loopback.ifname
 
 '''
-fix_localas_ibgp_neighbors - remove 'multihop' from any localas_ibgp neighbors
-'''
-def fix_localas_ibgp_neighbors(ndata: Box, topology: Box) -> None:
-  for ngb in _bgp.neighbors(ndata,vrf=True,select=['localas_ibgp']):    # Iterate over all localas_ibgp neighbors
-    ngb.pop('multihop',None)                                            # Remove 'multihop' config if any
-
-'''
 post_transform processing:
 
 * Remove fake interfaces and links
@@ -219,7 +212,6 @@ def post_transform(topology: Box) -> None:
   global _config_name
   for ndata in topology.nodes.values():
     fix_vrf_loopbacks(ndata,topology)
-    fix_localas_ibgp_neighbors(ndata,topology)
     cleanup_interfaces(ndata,topology)
 
   topology.links = [ link for link in topology.links if not link.get('_bgp_session',None) ]
--- a/netsim/extra/ebgp.multihop/plugin.py
+++ b/netsim/extra/ebgp.multihop/plugin.py
@@ -202,6 +202,13 @@ def fix_vrf_loopbacks(ndata: Box, topology: Box) -> None:
       ngb._source_ifname = ndata.loopback.ifname
 
 '''
+fix_localas_ibgp_neighbors - remove 'multihop' from any localas_ibgp neighbors
+'''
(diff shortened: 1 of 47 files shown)


netsim/extra/ebgp.multihop/plugin.py
```diff
@@ -201,6 +201,13 @@ def fix_vrf_loopbacks(ndata: Box, topology: Box) -> None:
     else:
       ngb._source_ifname = ndata.loopback.ifname
 
+'''
+fix_localas_ibgp_neighbors - remove 'multihop' from any localas_ibgp neighbors
+'''
+def fix_localas_ibgp_neighbors(ndata: Box, topology: Box) -> None:
```
ipspace, line 207: I could not find where these changes are justified. Please remove them from this PR.

tests/integration/bgp.multihop/01-global.yml
```diff
@@ -29,6 +29,10 @@ bgp.multihop.sessions:
 - dut:
   pe2:
     passive: True
+- dut:
```
ipspace, line 32: This makes bgp.multihop functionality dependent on the bgp.local_as functionality. Remove from the PR.

netsim/modules/bgp.yml
```diff
@@ -29,11 +29,11 @@ attributes:
...
     as_list: dict
     sessions:
-      ipv4: [ ibgp, ebgp ]
-      ipv6: [ ibgp, ebgp ]
+      ipv4: [ ibgp, ebgp, localas_ibgp ]
+      ipv6: [ ibgp, ebgp, localas_ibgp ]
     activate:
-      ipv4: [ ibgp, ebgp ]
-      ipv6: [ ibgp, ebgp ]
+      ipv4: [ ibgp, ebgp, localas_ibgp ]
```
ipspace, line 35: Most of the templates treat ibgp and localas_ibgp sessions the same way. What is the justification for them being different?

ipspace (changes requested):
* Several files have to be removed from the PR.
* Most templates do not make the difference between ibgp and localas_ibgp for the AF activation purposes. We will not change the templates, but the behavior has to be documented. Please adjust the `bgp.md` if needed.

netsim/modules/bgp.yml
jbemmel: We could decide to remove the ```localas_ibgp``` enum value and only support ```[ ibgp, ebgp ]```, but at the moment bgp neighbors of type ```localas_ibgp``` are being generated when ```bgp.local_as``` is set equal to the neighbor AS. The default ```activate``` and ```sessions``` values - if not set in a lab - include such 'localas_ibgp' sessions due to the Python string constant, but if I set the global sessions to some value there is no way to get such localas_ibgp sessions activated. They disappear silently

The justification is based on being able to configure all valid values, with consistency between the YAML files and what's defined by the Python source that accompanies them

netsim/extra/ebgp.multihop/plugin.py
jbemmel: replaced by separate PR

tests/integration/bgp.multihop/01-global.yml
jbemmel: moved to separate test case in different PR

PR #1673, opened 2024-12-20
Use # of nodes on a link (not link type) to select default pool

### LINKED ISSUE #1647: Decouple default pool selection from link type

ipspace:
netlab used the link.type attribute to select the default prefix allocation pool, resulting in LAN prefixes assigned to some P2P links.

This commit removes the dependency on link.type and consistently assigns prefixes from p2p pool to inter-router P2P links.

Closes #1647, replaces #1648

### DIFF
--- a/docs/addressing.md
+++ b/docs/addressing.md
@@ -11,13 +11,13 @@ You can assign a static prefix to a link with a **prefix** link attribute and a 
 
 * **mgmt** pool: Management IPv4 and MAC addresses. IPv6 addresses are assigned if specified, but not used at the moment.
 * **loopback** pool: Loopback IPv4 and IPv6 addresses.
-* **lan** pool: IPv4 and IPv6 addresses used all links apart from P2P links (links with more or less than two nodes attached to them), or links with **type** set to *lan*.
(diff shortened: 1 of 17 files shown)


jbemmel (approved):
Those changes to ```tests/topology/expected/bgp.yml``` look odd - why ipv6 instead of ipv4?

ipspace:
> Those changes to `tests/topology/expected/bgp.yml` look odd - why ipv6 instead of ipv4?

That test topology had the "stub" pool that was never used. Now it gets used, and it has only an IPv6 prefix.

Issue #1675, opened 2024-12-20
[BUG] Vagrant fails to start when VM name starts with UPPERcase


adamcirillo:
### Describe the bug

My lab I made can't start up using libvirt for nexus getting the following error. using the image nexus9300v64-lite.10.5.2.F.box
i used the instructions in the documentation to mutate the box and deleted the virtual box version as well.

plugins listed below and versions

```
vagrant-libvirt (0.12.2, global)
  - Version Constraint: 0.12.2
vagrant-mutate (1.2.0, global)
```

```
[CREATED] creating libvirt management network vagrant-libvirt
provider libvirt: executing vagrant up --provider libvirt
Vagrant failed to initialize at a very early stage:

There is a syntax error in the following Vagrantfile. The syntax error
message is reproduced below for convenience:

/home/netlab/adamlab/Vagrantfile:9: formal argument cannot be a constant
...onfig.vm.define vm_name do |NX1|
...                            ^~~
/home/netlab/adamlab/Vagrantfile:108: formal argument cannot be a constant
...onfig.vm.define vm_name do |NX2|
...                            ^~~
/home/netlab/adamlab/Vagrantfile:207: formal argument cannot be a constant
...onfig.vm.define vm_name do |NX3|
...                            ^~~
/home/netlab/adamlab/Vagrantfile:330: formal argument cannot be a constant
...onfig.vm.define vm_name do |NX4|
...                            ^~~
Error executing vagrant up --provider libvirt:
  Command '['vagrant', 'up', '--provider', 'libvirt']' returned non-zero exit status 1.
[FATAL]   netlab up: vagrant up --provider libvirt failed, aborting...
```

also if i do provider via defaults.device.nxos.provider: libvirt i get this error instead.

```
[ERRORS]  Errors found in topology.yml
[ATTR]    links: Incorrect link attribute 'provider' in links.CE107[2].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE107[3].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE107[4].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE107[5].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE107[6].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE108[2].clab
[ATTR]    links: Incorrect link attribute 'provider' in links.CE108[3].clab
[...]

PR #1676, opened 2024-12-20
Libvirt/Vagrant: Fix vm name constant to be lowercase

jbemmel:
Ruby requires at least the first character to be lower case, easier to just make the entire string ```lower()```

Fixes https://github.com/ipspace/netlab/issues/1675


ipspace:
Thank you!

PR #1681, opened 2024-12-23
Generic host static routing framework

### LINKED ISSUE #1641: [BUG] Host static routes don't work if the first host interface is not a L3 interface
The Linux configuration template assumes that the first data-plane interface is used for default in-lab routing. When that interface does not have an IP address, the static routes are not configured and the Linux host cannot reach anything but directly-connected subnets.

### Expected behavior

We need either a better Linux configuration template or additional logic in the "default gateway" code to set a flag when an interface can be used for default routing.

### Lab topology

```
---
provider: clab

nodes:
  rtr:
    device: frr
  h:
    device: linux

links:
- rtr:
  h:
  pool: l2only
- rtr:
  h:
```

ipspace:
This commit prepares the static routes that have to be configured on lab devices not running a routing protocol (hosts). It mirrors the functionality currently implemented in the Linux configuration template but makes it generic and ensures the static routes always have a next hop if at least one interface has a default gateway.

The data structures for all hosts in a lab topology are significantly modified:

* The 'gateway' data structure is always copied from the link to the hosts to ensure the next step has all the relevant information available
* The host data contains 'routing.static' list of static routes pointing to the first usable gateway.ipv4 address.

The change prepares the data structures needed to fix #1641 but provides a more generic framework that will be used to implement 'network devices as hosts'

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -1048,8 +1048,7 @@ def set_default_gateway(link: Box, nodes: Box) -> None:
     if nodes[ifdata.node].get('role','') == 'host':       # Set gateway only for hosts
       for interface in nodes[ifdata.node].interfaces:     # Find the corresponding host interface
         if link.linkindex == interface.linkindex:
-          if interface.ifindex == 1:                      # Set the default gateway only on the first host interface
-            interface.gateway = link.gateway
+          interface.gateway = link.gateway                # Copy link gateway to host interface
 
 """
 Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family configured
--- a/netsim/augment/main.py
+++ b/netsim/augment/main.py
@@ -87,6 +87,7 @@ def transform_data(topology: Box) -> None:
 
 def post_transform(topology: Box) -> None:
   augment.validate.process_validation(topology)
+  augment.nodes.post_transform(topology)
   modules.post_transform(topology)
   augment.plugin.execute('post_transform',topology)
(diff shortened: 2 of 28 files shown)


netsim/augment/links.py
```diff
@@ -1048,8 +1048,7 @@ def set_default_gateway(link: Box, nodes: Box) -> None:
     if nodes[ifdata.node].get('role','') == 'host':       # Set gateway only for hosts
       for interface in nodes[ifdata.node].interfaces:     # Find the corresponding host interface
         if link.linkindex == interface.linkindex:
-          if interface.ifindex == 1:                      # Set the default gateway only on the first host interface
-            interface.gateway = link.gateway
+          interface.gateway = link.gateway                # Copy link gateway to host interface
```
jbemmel, line 1051: Should a "generic" host static routing framework depend on having gateway functionality enabled in the topology?

It is perfectly valid for a static route to have any router interface as next hop, gateway or no gateway. When present, the gateway makes for a natural candidate - but it should be possible to do static routing in topologies that do not use the gateway module

jbemmel (approved):
Overall it's heading in the right direction, but I would personally like to see routing also being possible without requiring the use of the `gateway` module

ipspace:
The "gateway" link attribute predates the "gateway" module and was always used to generate static routes on Linux hosts. It's set here: https://github.com/ipspace/netlab/blob/dev/netsim/augment/links.py#L1020

PR #1683, opened 2024-12-23
Implement static routing on Linux and use it to configure host static routes

ipspace:
This PR:

* Replaces the hardcoded static routes in Linux initial configuration template with configuration generated from 'routing.static' (see #1681)
* Enables static routing part of generic routing module for Linux
* Implements "limit maximum next hops for a static route" functionality because **ip route** command cannot handle more than one route to a prefix.

It's a spin-off from #1681 and will be rebased once that one is merged.

### DIFF
--- a/docs/module/routing.md
+++ b/docs/module/routing.md
@@ -22,22 +22,22 @@ This configuration module implements generic routing features:
 The following table describes high-level per-platform support of generic routing features:
 
 | Operating system      | Routing<br>policies | Prefix<br>filters| AS-path<br>filters | BGP<br>communities | Static<br>routes|
-| ------------------ | :-: | :-: | :-: |:-: | :-: |
-| Arista EOS          |  ✅  |  ✅  |  ✅  |  ✅  |
-| Aruba AOS-CX        |  ✅  |  ✅  |  ✅  |  ✅  |
-| Cisco IOSv/IOSvL2   |  ✅  |  ✅  |  ✅  |  ✅  |
-| Cisco IOS-XE[^18v]  |  ✅  |  ✅  |  ✅  |  ✅  |
-| Cumulus Linux       |  ✅  |  ✅  |  ✅  |  ✅  |
-| FRR                 |  ✅  |  ✅  |  ✅  |  ✅  |
-| Nokia SR Linux      |  ✅  |  ✅ [❗](caveats-srlinux) |
-| Nokia SR OS         |  ✅  |
-| VyOS                |  ✅  |  ✅  |  ✅  |  ✅  |
+| ------------------ |:--:|:--:|:--:|:--:|:--:|
+| Arista EOS         | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Aruba AOS-CX       | ✅ | ✅ | ✅ | ✅ |
+| Cisco IOS/XE[^18v] | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Cumulus Linux      | ✅ | ✅ | ✅ | ✅ | ✅ |
+| FRR                | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Linux              | ❌  | ❌  | ❌  | ❌  | ✅ |
+| Nokia SR Linux     |  ✅ |  ✅ [❗](caveats-srlinux) |
+| Nokia SR OS        |  ✅ |
+| VyOS               |  ✅ |  ✅ | ✅ |  ✅  |
 
 ```{tip}
 See [Routing Integration Tests Results](https://release.netlab.tools/_html/coverage.routing) for more details.
 ```
 
-[^18v]: Includes Cisco CSR 1000v, Cisco Catalyst 8000v, Cisco IOS-on-Linux (IOL), and IOL Layer-2 image.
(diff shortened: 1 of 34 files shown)


docs/module/routing.md
```diff
@@ -22,22 +22,22 @@ This configuration module implements generic routing features:
...
-| Cumulus Linux       |  ✅  |  ✅  |  ✅  |  ✅  |
-| FRR                 |  ✅  |  ✅  |  ✅  |  ✅  |
-| Nokia SR Linux      |  ✅  |  ✅ [❗](caveats-srlinux) |
-| Nokia SR OS         |  ✅  |
-| VyOS                |  ✅  |  ✅  |  ✅  |  ✅  |
+| ------------------ |:--:|:--:|:--:|:--:|:--:|
+| Arista EOS         | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Aruba AOS-CX       | ✅ | ✅ | ✅ | ✅ |
+| Cisco IOS/XE[^18v] | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Cumulus Linux      | ✅ | ✅ | ✅ | ✅ | ✅ |
```
jbemmel, line 29: Both regular and NVUE?

jbemmel:
```ip route``` does support multiple next hops: https://www.webconn.tech/kb/how-to-add-equal-cost-multipath-route-ecmp-to-ip-routing

```
ip route add 172.16.0.0/24 
    nexthop via 10.0.1.1 dev ethX weight 1 
    nexthop via 10.0.2.1 dev ethY weight 1
```

or

```
ip route add 172.16.0.0/24 nexthop via 10.0.1.1 dev ethX weight 1
ip route append 172.16.0.0/24 nexthop via 10.0.2.1 dev ethY weight 1
```

**Comment by jbemmel** [comment]
Also supported in Netplan: https://github.com/canonical/netplan/blob/main/examples/static_singlenic_multiip_multigateway.yaml

File: netsim/modules/routing.py [hunk 2]
```diff
@@ -868,7 +867,7 @@ def check_static_routes(idx: int,o_name: str,node: Box,topology: Box) -> None:
     for af in log.AF_LIST:
       if af not in sr_data:
         continue
-      for nh_entry in sr_data.nexthop.nhlist:
+      for nh_entry in sr_data.nexthop.nhlist[:sr_features.get('max_nexthop',256)]:
         sr_entry = { af: sr_data[af], 'nexthop': nh_entry }
```
jbemmel, line 871: It may be easier to make ```nexthop``` a list - then the Jinja2 template can pick it up from there

jbemmel (changes requested):
Static routes with multiple next hops (ECMP) can be supported on Linux using ```ip route```, wouldn't want Netlab to be witholding that from its users ;)

ipspace:
> Static routes with multiple next hops (ECMP) can be supported on Linux using `ip route`, wouldn't want Netlab to be witholding that from its users ;)

Thank you, will fix.

netsim/ansible/templates/initial/linux/vanilla.j2
```diff
@@ -46,24 +46,50 @@ ip link set {{ l.ifname }} mtu {{ l.mtu }}
...
-{%     for af,pfx in pool.items() if af == 'ipv4' %}
-# {{ name }} prefix: {{ pfx }} local subnet: {{ ifdata.gateway.ipv4|ipaddr('subnet') }}
-{%       if name not in ['mgmt','router_id'] and pfx is string and
-            pfx != ifdata.gateway.ipv4|ipaddr('subnet') %}
+{% for pfx in sr_list|map(attribute='ipv6',default='')|unique if pfx %}
+{%   if loop.first %}
 set +e
+echo Removing existing IPv6 routes
+{%   endif %}
 ip route del {{ pfx }} 2>/dev/null
```
jbemmel, line 68: If there are routes with multiple next hops, this will only delete the first such route. We may be better off simply flushing the entire route table at the start

netsim/modules/routing.py
```diff
@@ -868,12 +867,15 @@ def check_static_routes(idx: int,o_name: str,node: Box,topology: Box) -> None:
     for af in log.AF_LIST:
       if af not in sr_data:
         continue
-      for nh_entry in sr_data.nexthop.nhlist:
-        sr_entry = { af: sr_data[af], 'nexthop': nh_entry }
+      for (nh_idx,nh_entry) in enumerate(sr_data.nexthop.nhlist[:sr_features.get('max_nexthop',256)]):
```
jbemmel, line 870: Is there any device left that has this ```max_nexthop``` feature defined?

jbemmel (changes requested):
Now that Linux supports multiple next hops, I believe the ```max_nexthops``` feature is no longer needed? (Hence should be removed to avoid creating unnecessary complexity)

netsim/ansible/templates/initial/linux/vanilla.j2
ipspace: > If there are routes with multiple next hops, this will only delete the first such route.

Thank you. Fixed. A while loop makes the whole thing even easier ;)

> We may be better off simply flushing the entire route table at the start

That would also flush the static default route configured by containerlab plus whatever other custom stuff the user might have configured.

PR #1688, opened 2024-12-24
AOS-CX: support for static routing

ssasso:
(merge after #1687)

**AOS-CX: support for static routing**

Tests:

* 20-static: OK
* 21-static-vrf: OK
* 22-static-inter-vrf: caveat for IPv6


ipspace (approved):

ipspace:
Thank you! Two minor details:

* I would make "Inter-VRF static routes are supported only for IPv4" more explicit, for example "Aruba AOS-CX does not support inter-VRF IPv6 routes"
* Check whether there's "nexthop-vrf" option at the end of the "ipv6 route" command. Cisco IOS has it and it seems to be severely underdocumented.

ssasso:
> * Check whether there's "nexthop-vrf" option at the end of the "ipv6 route" command. Cisco IOS has it and it seems to be severely underdocumented

Unfortunately there is not, tried multiple options.

Issue #1684, opened 2024-12-23
Linux: Generalize the execution of provisioning scripts from container netns context


jbemmel:
Initial provisioning for Linux containers is done in ```ansible/tasks/linux/initial-clab.yml```, while remaining modules use ```ansible/tasks/deploy-config/linux-clab.yml```

My proposal is to include the functionality from ```ansible/tasks/linux/initial-clab.yml``` in ```ansible/tasks/deploy-config/linux-clab.yml```, and conditionally select ```ip netns``` from host or "native" execution within the container.

The condition could be: If the install script ends with "-clab.j2", use ```ip netns```
So we would have under ansible/templates/
* initial/linux-clab.j2
* lag/linux-clab.j2
* dhcp/linux.j2 ( i.e. regular )

ipspace:
> The condition could be: If the install script ends with "-clab.j2", use `ip netns`

You might have forgotten all the daemons that are based on Linux, but after analyzing their implementation, it seems that your idea makes sense. Submit the proposed changes as a separate PR.

PR #1690, opened 2024-12-24
Linux containers: generalize netns execution for any script template ending with "-clab.j2"

jbemmel:
The introduction of lag support for Linux hosts results in a need for netns execution beyond the ```initial``` module. This PR generalizes the use of ```ip netns exec``` for any template name ending with ```-clab.j2```

* Use the Ansible tempfile mechanism instead of a custom password hash based unique string
* Render the template on the Linux host in both cases (avoiding any dependency on Python in Linux container)

Addresses https://github.com/ipspace/netlab/issues/1684


ipspace (approved):
Nice job. I love that you provided documentation and the total decoupling from Python in container ;)

Issue #1689, opened 2024-12-24
[BUG] LAG tests not working


ssasso:
While trying to implement LAG in AOS-CX, I encountered this error in running the tests:

```
root@hippo:~/GIT_H/netlab/tests/integration/lag# netlab create 01-l3-lag.yml -p clab -d arubacx
Traceback (most recent call last):
  File "/usr/local/bin/netlab", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/root/GIT_H/netlab/netlab", line 14, in <module>
    netsim.cli.lab_commands(__file__)
  File "/root/GIT_H/netlab/netsim/cli/__init__.py", line 353, in lab_commands
    mod.run(sys.argv[arg_start:])   # type: ignore
  File "/root/GIT_H/netlab/netsim/cli/create.py", line 81, in run
    augment.main.transform(topology)
  File "/root/GIT_H/netlab/netsim/augment/main.py", line 109, in transform
    transform_data(topology)
  File "/root/GIT_H/netlab/netsim/augment/main.py", line 66, in transform_data
    modules.pre_transform(topology)
  File "/root/GIT_H/netlab/netsim/modules/__init__.py", line 77, in pre_transform
    module_transform("pre_transform",topology)
  File "/root/GIT_H/netlab/netsim/modules/__init__.py", line 462, in module_transform
    mod_load[m].call("module_"+method,topology)
  File "/root/GIT_H/netlab/netsim/utils/callback.py", line 42, in call
    return method(*args, **kwargs)
  File "/root/GIT_H/netlab/netsim/modules/lag.py", line 372, in module_pre_transform
    process_lag_links(topology)             # Expand lag.members into additional p2p links
  File "/root/GIT_H/netlab/netsim/modules/lag.py", line 324, in process_lag_links
    create_lag_member_links(l,topology)
  File "/root/GIT_H/netlab/netsim/modules/lag.py", line 178, in create_lag_member_links
    def analyze_lag(members: list, node_count: dict) -> tuple[bool,bool,str]:
TypeError: 'type' object is not subscriptable
```

(note: same happens using device type 'eos', so this is not device dependant)

Same for test *02*:

```
root@hippo:~/GIT_H/netlab/tests/integration/lag# netlab create 02-lag-vlan-trunk.yml -p clab -d eos
Traceback (most recent call last):
  File "/usr/local/bin/netlab", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
[...]

PR #1691, opened 2024-12-24
Python: Update minimum required version to 3.9, check version at runtime

jbemmel:
Triggered by https://github.com/ipspace/netlab/issues/1689

### DIFF
--- a/netsim/__init__.py
+++ b/netsim/__init__.py
@@ -6,18 +6,18 @@ __version__ = "1.9.3-dev1"
 
 abort = False
 
-if sys.version_info < (3,9,0):
+if sys.version_info < (3,9):
   print(f"FATAL ERROR: Netlab requires Python version 3.9 or newer - installed {sys.version}")
   abort = True
 
 try:
   import box
 
-  if box.__version__ < '7.2.0':                 # netlab needs Python Box version 7.0 or higher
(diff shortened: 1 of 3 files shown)


ssasso:
Ubuntu 20.04 has python 3.8 by default.

Either we force the install script to install python3.9 (available also for 20.04), or we remove also Ubuntu 20.04 from the supported ubuntu versions.

ipspace:
> Either we force the install script to install python3.9 (available also for 20.04), or we remove also Ubuntu 20.04 from the supported ubuntu versions.

We'll go for the third option and keep writing code that still works on Python 3.9 ;)

netsim/__init__.py
```diff
@@ -6,14 +6,18 @@
...
 
+if sys.version_info < (3,9):
+  print(f"FATAL ERROR: Netlab requires Python version 3.9 or newer - installed {sys.version}")
+  abort = True
+
 try:
   import box
 
-  if box.__version__ < '7.2.0':                 # netlab needs Python Box version 7.0 or higher
+  if box.__version__ < '7.2.0':                 # netlab needs Python Box version 7.2 or higher
```
ipspace, line 16: And you chose to change the minimum Box version as well. I wonder what the justification was.

Actually, I don't want to know.

jbemmel: I did not change the version, I corrected the comment to match the code

PR #1696, opened 2024-12-25
Fix: Deal with hosts attached to unnumbered links

### LINKED ISSUE #1692: [BUG] unnumbered interfaces are not transformed into "ipv4: True" for host-node routing daemons
bird does not support unnumbered interfaces or BGP unnumbered. However, netlab does not report that leaves "unnumbered: True" in interface data.

### Lab topology

```
module: [ bgp ]

provider: clab
bgp.as: 65000

nodes:
  dut:
    device: bird
  x1:
    device: iosv

links:
- dut:
  x1:
  unnumbered: True
```

### Output

The "unnumbered interfaces are unsupported" error is generated for IOSv but not for bird

```
$ netlab create test.yml
[ERRORS]  Errors found in test.yml
[VALUE]   interfaces: Device iosv does not support unnumbered IPv4 interfaces
[HINT]    Used on node x1 interface GigabitEthernet0/1 (link x1 -> dut)
```

Also, the x1 interface has "ipv4: True" (when changing X1 device type to FRR) while the bird interface has "unnumbered: True"

```
$ netlab inspect --node x1 interfaces
- _parent_intf: lo
  _parent_ipv4: 10.0.0.2/32
  bridge: X_1
  ifindex: 1
  ifname: eth1
  ipv4: true
  linkindex: 1
  mtu: 1500
  name: x1 -> dut
  neighbors:
  - ifname: eth1
    node: dut
  type: lan
  unnumbered: true

$ netlab inspect --node dut interfaces
- bridge: X_1
  ifindex: 1
  ifname: eth1
  linkindex: 1
  mtu: 1500
  name: dut -> x1
  neighbors:
  - ifname: eth1
    ipv4: true
    node: x1
  type: lan
  unnumbered: true
```

ipspace:
Netlab link augmentation code uses address family from the loopback interface to figure out what to do on old-style unnumbered links.

That clearly does not work on hosts that have no loopback interface. This fix enables IPv6 LLA on hosts attached to unnumbered links, hoping someone might eventually figure out whether that makes sense.

It also removes 'unnumbered' attribute from the link, as no device configuration template checks it. We'll clean up other references to the 'unnumbered' flag in another commit.

Fixes #1692

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -568,8 +568,13 @@ If the interface address is set, validate that it's a valid address (can't be in
 """
 def IPAM_unnumbered(link: Box, af: str, pfx: typing.Optional[bool], ndict: Box) -> None:
   for intf in link.interfaces:
-    if not af in intf:            # No static address, set it to link bool value or use loopback AF presence for old-style unnumbereds
-      intf[af] = pfx if isinstance(pfx,bool) else bool(ndict[intf.node].get(f'loopback.{af}',False))
+    if not af in intf:                            # No static address
+      if isinstance(pfx,bool):                    # Set it to link bool value if it exists
+        intf[af] = pfx
+      elif 'loopback' in ndict[intf.node]:        # Use loopback AF presence for old-style unnumbereds
+        intf[af] = bool(ndict[intf.node].get(f'loopback.{af}',False))
+      elif af == 'ipv6':                          # Else enable IPv6 LLA for hosts
+        intf[af] = True
     elif data.is_true_int(intf[af]):
       log.error(
         f'Node {intf.node} is using host index {intf[af]} for {af} on an unnumbered link',
@@ -626,6 +631,7 @@ def assign_interface_addresses(link: Box, addr_pools: Box, ndict: Box, defaults:
       IPAM_unnumbered(link,af,None,ndict)
 
     link.pop('prefix')
+    link.pop('unnumbered')
     return
 
   for af in ('ipv4','ipv6'):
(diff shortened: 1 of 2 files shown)


netsim/augment/links.py
```diff
@@ -568,8 +568,13 @@ def set_interface_address(intf: Box, af: str, pfx: netaddr.IPNetwork, node_id: i
 """
 def IPAM_unnumbered(link: Box, af: str, pfx: typing.Optional[bool], ndict: Box) -> None:
   for intf in link.interfaces:
-    if not af in intf:            # No static address, set it to link bool value or use loopback AF presence for old-style unnumbereds
-      intf[af] = pfx if isinstance(pfx,bool) else bool(ndict[intf.node].get(f'loopback.{af}',False))
+    if not af in intf:                            # No static address
+      if isinstance(pfx,bool):                    # Set it to link bool value if it exists
+        intf[af] = pfx
```
jbemmel, line 573: Shouldn't this be set only if ```True```?

netsim/augment/links.py
```diff
@@ -568,8 +568,13 @@ def set_interface_address(intf: Box, af: str, pfx: netaddr.IPNetwork, node_id: i
...
+      elif 'loopback' in ndict[intf.node]:        # Use loopback AF presence for old-style unnumbereds
+        intf[af] = bool(ndict[intf.node].get(f'loopback.{af}',False))
```
jbemmel, line 575: Same comment - only if ```True```

netsim/augment/links.py
ipspace: The 'false' values are scrubbed later in the process.

This also allows someone overly-creative to define (for example) `unnumbered: True, ipv6: False`. I know this is contrived, but I've seen worse 🤷‍♂️

netsim/augment/links.py
ipspace: As above...

Issue #1698, opened 2024-12-27
[BUG] Error in test validation for tests/integration/lag - tests 07,09


ssasso:
For 07-mlag-trunk.yml:

```
root@hippo:~/GIT_H/netlab/tests/integration/lag# netlab validate
[ping]    Pinging H2 from H1 [ node(s): h1 ]
[VALUE]   validate: 'Box' object is not callable
[HINT]    device: frr, action: exec
          plugin expression: exec_ping(nodes.h2.interfaces[0].ipv4,af='ipv4')
[MISSING] validation: Test ping: have no idea what command to execute on node h1 / device frr
[WAITING] Waiting for STP to enable the ports (retrying for 45 seconds)
[VALUE]   validate: 'Box' object is not callable
[MISSING] validation: Test ping: have no idea what command to execute on node h1 / device frr
[VALUE]   validate: 'Box' object is not callable
```

It seems the test is pointing at the wrong interface(s).

Looking at the netlab snapshot with *yq*:
```
cat netlab.snapshot.yml | yq .nodes.h2.interfaces[0].ipv4
null
```

```
cat netlab.snapshot.yml | yq .nodes.h2.interfaces[0]
{
  "ifindex": 30000,
  "ifname": "bond1",
  "lag": {
    "ifindex": 1,
    "lacp": "fast",
    "lacp_mode": "active",
    "mode": "802.3ad"
  },
  "linkindex": 3,
  "mtu": 1500,
  "name": "[Access VLAN red] h2 -> [s1,s2]",
  "neighbors": [
    {
      "ifname": "lag 2",
      "ipv4": "172.16.0.1/24",
      "node": "s1"
    },
    {
      "ifname": "lag 2",
      "ipv4": "172.16.0.2/24",
      "node": "s2"
    }
  ],
  "type": "lag",
  "virtual_interface": true,
  "vlan": {
    "access": "red",
    "access_id": 1000
  }
}
```

ssasso:
Side note: second test of **09-mlag-m-to-m.yml** is broken as well:

```
root@hippo:~/GIT_H/netlab/tests/integration/lag# netlab validate
[ping]    Pinging H2 from H1 [ node(s): h1 ]
[PASS]    h1: Ping to h2 (172.16.0.6) succeeded
[PASS]    Test succeeded

[ping_gw] Pinging gateway from H1 [ node(s): h1 ]
[VALUE]   validate: list index out of range
[HINT]    device: linux, action: exec
[...]

PR #1699, opened 2024-12-27
AOS-CX: support for LAG and MCLAG

ssasso:
Tested with all 'lag' integration tests except for *11-host-lag-active-standby.yml* (which complains about unsupported lag module on linux).

Will re-run the tests after #1698 - but there should be no problems.


ipspace (approved):
Looks good to me, but I know nothing about Aruba CX anyway ;)

PR #1703, opened 2024-12-27
lag: Fix wrong interface indices in tests 07/09, general cleanup

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1698


ssasso (approved):
both tests are working fine now!

PR #1704, opened 2024-12-27
gateway: Remove config/module if not used on a node

jbemmel:
Similar to IGP routing protocols, the module should be removed when not used

(Note: Same could be done for the ```vlan``` module on ```r3``` in this test case)

### DIFF
--- a/netsim/modules/gateway.py
+++ b/netsim/modules/gateway.py
@@ -85,6 +85,9 @@ def cleanup_intf_protocol_parameters(node: Box,topology: Box) -> list:
       elif k in node.get('gateway'):                                  # Do we have node-level parameters for this protocol?
         intf.gateway[k] = node.gateway[k] + intf.gateway[k]           # ... copy them to all interfaces
 
+  if not active_proto:                                                # No gateways in use anywhere?
+    node.pop('gateway',None)                                          # Remove its configuration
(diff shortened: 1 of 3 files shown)


netsim/modules/gateway.py
```diff
@@ -255,7 +255,10 @@ def node_post_transform(self, node: Box, topology: Box) -> None:
 
     cleanup_unicast_ip(node)
     active_proto = cleanup_intf_protocol_parameters(node,topology)    # Cleanup interface parameters and get a list of active protocols
-    if 'gateway' in node:
-      for k in list(node.gateway):                                    # Iterate over node-level gateway parameters
+    if active_proto:                                                  # Any gateway active?
```
ipspace, line 258: Leave this as is...

netsim/modules/gateway.py
```diff
@@ -255,7 +255,10 @@ def node_post_transform(self, node: Box, topology: Box) -> None:
...
+      for k in list(node.get('gateway',{})):                          # Iterate over node-level gateway parameters
         if not k in active_proto:                                     # Not a parameter for a FHRP active on this node?
           node.gateway.pop(k,None)                                    # ... zap it!
+    else:
```
ipspace, line 262: And make this into "if not active_proto"

ipspace (changes requested):

ipspace (approved):

PR #1711, opened 2024-12-30
Dell OS10: interface description fixes & mlag switchport fix

jbemmel:
Updated after feedback:
* (still) limit interface descriptions to 255 characters when adding local string
* Enable switchport on mlag interfaces in case there is no VLAN (-trunk)
* Replace comma's in descriptions, if any

### DIFF
--- a/netsim/ansible/templates/lag/dellos10.j2
+++ b/netsim/ansible/templates/lag/dellos10.j2
@@ -14,6 +14,9 @@ vlt-domain {{ intf.lag.mlag.peergroup }}
 interface {{ intf.ifname }}
 {%  if '_mlag' in intf.lag %}
  description "{{ intf.name }} in vlt-port-channel {{ intf.lag.ifindex }}"
+{%    if 'vlan' not in intf %}
+ switchport mode access
+{%    endif %}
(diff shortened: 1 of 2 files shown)


ssasso (changes requested):
The description length limitation imho should be addressed on the netlab core, as Ivan proposed.

I am wondering if any platform allows for "routed" mode in a MCLAG interface (I would say no?) - so maybe instead of adding a device quirk only for OS10 this can be handled directly on the LAG module? (i.e., disabling it at all, or with a feature flag?)

(i.e., ArubaOS-CX has the same "limitation")

ipspace:
> I am wondering if any platform allows for "routed" mode in a MCLAG interface (I would say no?) - so maybe instead of adding a device quirk only for OS10 this can be handled directly on the LAG module? (i.e., disabling it at all, or with a feature flag?)

I would go for "disable it", eventually introducing a feature flag only after someone documents support on a major platform and a good use case.

ipspace:
FWIW, stop making unnecessary spurious changes to the code base just because you think that will make it look nicer. Unfortunately, I can't trust your testing methods and it's not worth wasting anyone's time to try and figure out whether your changes are idempotent.

PR #1712, opened 2024-12-31
Fix: restrict interface description length

### LINKED ISSUE #1709: [BUG] Long interface descriptions cause provisioning errors
Encountered in large topologies with ```node.clone``` and VLANs: Descriptions longer than 255 characters cause issues on (at least) Dell OS10 and Cumulus NVUE devices

ipspace:
The interface description is limited to 255 characters due to SNMP MIB-2 limitations. This fix introduces two system const parameters (maxlength and max neighbors) that limit the length of the interface description.

Fixes #1709

### DIFF
--- a/netsim/defaults/const.yml
+++ b/netsim/defaults/const.yml
@@ -4,3 +4,6 @@
 routing_protocols: [ bgp, connected, eigrp, isis, ospf, ripv2 ]
 vrf_igp_protocols: [ connected, ospf, isis, ripv2 ]
 multi_provider: [ libvirt, clab ]
+ifname:
+  neighbors: 5
+  maxlength: 255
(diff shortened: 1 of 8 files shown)


netsim/defaults/const.yml
```diff
@@ -4,3 +4,6 @@
 routing_protocols: [ bgp, connected, eigrp, isis, ospf, ripv2 ]
 vrf_igp_protocols: [ connected, ospf, isis, ripv2 ]
 multi_provider: [ libvirt, clab ]
+ifname:
+  neighbors: 5
+  maxlength: 255
```
jbemmel, line 9: Perhaps add a note on SNMP driving this value

netsim/defaults/const.yml
```diff
@@ -4,3 +4,6 @@
 routing_protocols: [ bgp, connected, eigrp, isis, ospf, ripv2 ]
 vrf_igp_protocols: [ connected, ospf, isis, ripv2 ]
 multi_provider: [ libvirt, clab ]
+ifname:
+  neighbors: 5
```
jbemmel, line 8: ```maxneighbors``` would be slightly clearer

jbemmel (approved):

PR #1714, opened 2024-12-31
(v)Junos VLAN Major Fixes and refactoring

ssasso:
**(v)Junos VLAN Major Fixes and refactoring**

(unfortunately for multiple stuff it was hard to split in different PR, so I preferred to keep all the things together)

Re-tested vJunos-switch, vJunos-router, vJunos-EVO (vPTX) and an old version of vSRX (unable to access newer ones)(only for specific supported features) with: 

* 01-vlan-bridge-single.yml
* 02-vlan-bridge-multiple.yml
* 21-vlan-irb-single.yml
* 22-vlan-irb-multiple.yml
* 23-vlan-mixed-multiple.yml
* 31-vlan-bridge-trunk.yml
* 32-vlan-bridge-trunk-router.yml
* 33-vlan-irb-trunk.yml
* 41-vlan-bridge-native.yml
* 42-vlan-irb-native.yml
* 51-vlan-routed-trunk.yml
* 52-vlan-vrf-lite.yml
* 61-vlan-routed-native.yml
* 62-vlan-mixed-trunk.yml (mixed trunk not supported on switch devices)
* 63-vlan-mixed-native.yml (mixed trunk not supported on switch devices)

### DIFF
--- a/netsim/ansible/templates/initial/junos.j2
+++ b/netsim/ansible/templates/initial/junos.j2
@@ -21,7 +21,12 @@ system {
 interfaces {
 {# Handle MTU outside the main interface loop, since it needs to be applied only to the master interface #}
 {% for l in interfaces|default([]) %}
-{%   if l.mtu is defined %}
+{%   if l.mtu is defined and l._vlan_master|default(false) and l._vlan_mtu_override is defined %}
+{#   - i.e., can be 22 bytes more (depending on mode) #}
+  {{ l.junos_interface }} {
+    mtu {{ l.mtu + l._vlan_mtu_override }};
+  }
+{%   elif l.mtu is defined %}
   {{ l.junos_interface }} {
     mtu {{ l.mtu + 14 }};
   }
--- a/netsim/ansible/templates/initial/junos.j2
+++ b/netsim/ansible/templates/initial/junos.j2
@@ -19,16 +19,14 @@ system {
 {% endif %}
 
 interfaces {
-{# Handle MTU outside the main interface loop, since it needs to be applied only to the master interface #}
+{#
+# Handle MTU outside the main interface loop, since it needs to be applied only to the master interface
+# JunOS MTU with headers is calculated on device quirks python code, depending on interface and model type.
+#}
 {% for l in interfaces|default([]) %}
-{%   if l.mtu is defined and l._vlan_master|default(false) and l._vlan_mtu_override is defined %}
-{#   - i.e., can be 22 bytes more (depending on mode) #}
+{%   if l._junos_mtu_with_headers is defined %}
   {{ l.junos_interface }} {
-    mtu {{ l.mtu + l._vlan_mtu_override }};
-  }
-{%   elif l.mtu is defined %}
-  {{ l.junos_interface }} {
-    mtu {{ l.mtu + 14 }};
+    mtu {{ l._junos_mtu_with_headers }};
   }
 {%   endif %}
 {% endfor %}
--- a/netsim/ansible/templates/initial/junos.vlan.j2
+++ b/netsim/ansible/templates/initial/junos.vlan.j2
@@ -43,7 +43,7 @@ interfaces {
 
 {%     endif %}
 
-{%   elif l.junos_interface != 'irb' and l.junos_unit == '0' and l._vlan_master is defined and not(l.vlan is defined) %}
+{%   elif l.junos_interface != 'irb' and l.junos_unit == '0' and l._vlan_master is defined and 'vlan' not in l %}
 {#
 # in this case unit 0 is still present, but not handling any real vlan traffic (no vlan definition apart from _vlan_master)
 # --> we need to force a vlan id in any case on the sub unit 0, so let's put 1
@@ -57,7 +57,7 @@ interfaces {
 {%   endif %}
 
 {# handling for devices with vlan model router #}
-{%   if netlab_device_type == 'vmx' or netlab_device_type == 'vsrx' or netlab_device_type == 'vjunos-router' %}
+{%   if netlab_device_type in [ 'vmx', 'vsrx', 'vjunos-router' ] %}
 
 {%     if l.junos_interface != 'irb' and l.junos_unit == '0' and l._vlan_master is defined and l.vlan.access_id is defined %}
 {# this is a irb port in access mode on unit 0 #}
--- a/netsim/ansible/templates/initial/junos.vlan.j2
+++ b/netsim/ansible/templates/initial/junos.vlan.j2
@@ -42,9 +42,22 @@ interfaces {
   }
 
 {%     endif %}
+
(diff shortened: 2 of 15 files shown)


netsim/ansible/templates/initial/junos.j2
```diff
@@ -21,7 +21,12 @@ system {
 interfaces {
 {# Handle MTU outside the main interface loop, since it needs to be applied only to the master interface #}
 {% for l in interfaces|default([]) %}
-{%   if l.mtu is defined %}
+{%   if l.mtu is defined and l._vlan_master|default(false) and l._vlan_mtu_override is defined %}
+{#   - i.e., can be 22 bytes more (depending on mode) #}
+  {{ l.junos_interface }} {
+    mtu {{ l.mtu + l._vlan_mtu_override }};
```
jbemmel, line 27:
```yaml
{%   if l.mtu is defined %}
  {{ l.junos_interface }} {
    mtu {{ l.mtu + (l._vlan_mtu_override|default(14) if l._vlan_master|default(false) else 14) }};
```

netsim/ansible/templates/initial/junos.vlan.j2
```diff
@@ -42,9 +42,22 @@ interfaces {
...
+
+  {{ l.junos_interface }}.0 {
+    vlan-id 1;
+  }
+
 {%   endif %}
 
-{%   if netlab_device_type != 'vptx' %}
+{# handling for devices with vlan model router #}
+{%   if netlab_device_type == 'vmx' or netlab_device_type == 'vsrx' or netlab_device_type == 'vjunos-router' %}
```
jbemmel, line 60:
```
if netlab_device_type in [ 'vmx', 'vsrx', 'vjunos-router' ]
```

netsim/devices/junos.py
```diff
@@ -47,6 +47,16 @@ def fix_unit_0(node: Box, topology: Box) -> None:
...
         intf._vlan_master = True
+        # define vlan mtu override for VLAN tagging
+        junos_vlan_kind = topology.defaults.devices[node.device].features.vlan.get('model', '')
+        # in case of "flexible vlan tagging", 22 bytes more, depending on link type
+        if junos_vlan_kind == 'router':
+          # for a router, always add it
+          intf._vlan_mtu_override = 22
+        elif junos_vlan_kind == 'l3-switch':
+          # for a l3-switch, add it on vlan mode route or if no vlan structure is defined (flex vlan tagging with native only)
+          if not intf.get('vlan') or intf.get('vlan', {}).get('mode', '') == 'route':
```
jbemmel, line 58:
```
if 'vlan' not in intf or intf.get('vlan.mode', "") == 'route':
```

netsim/ansible/templates/initial/junos.vlan.j2
```diff
@@ -42,9 +42,22 @@ interfaces {
   }
 
 {%     endif %}
+
+{%   elif l.junos_interface != 'irb' and l.junos_unit == '0' and l._vlan_master is defined and not(l.vlan is defined) %}
```
jbemmel, line 46:
```
... and 'vlan' not in l
```

netsim/ansible/templates/initial/junos.j2
jbemmel: Though I would suggest to put this logic of setting ```_vlan_mtu_override``` in the Python code. and in the template only do
```
mtu {{ l.mtu + l._vlan_mtu_override|default(14) }}
```

ssasso:
Will try to fix (moving to draft for now)

netsim/ansible/templates/initial/junos.j2
ssasso: I will try to put all the logic for MTU calc on the python quirk file, it would be easier to perform all the required checks.

I will call the "final" variable with another name (i.e., `_junos_mtu_with_headers`), to avoid confusion if someone will inspect the topology (so both the real requested IP MTU and the calculated MTU can be shown).
In any case I will add proper comments on the template.

netsim/ansible/templates/initial/junos.vlan.j2
ssasso: Potentially I could also add a group var (i.e., `junos_device_kind` set to router or switch) and check for that instead of the device model  (or find a way to use from the Jinja file the `feature.vlan.mode` - any hint? @ipspace).

In that way there's no need to edit the template every time a new "routing" device is added - if there will be other ones in the future.

jbemmel: ```feature.vlan.mode``` is accessible from the device Python code, you could set a flag based on that (similar to how ```l.junos_interface``` is added)

ipspace: > Potentially I could also add a group var (i.e., `junos_device_kind` set to router or switch) and check for that instead of the device model (or find a way to use from the Jinja file the `feature.vlan.mode` - any hint? @ipspace).

You could also use two different generic templates and include (or symlink) them to the device configuration template, similar to what I'm doing for Cisco IOS/IOL/IOSL2/IOLL2.

jbemmel:
Not covered by this PR, but I noticed a ```unit_0_trick``` being applied to OSPF interfaces and I was wondering: What about ISIS?

netsim/ansible/templates/initial/junos.vlan.j2
ssasso: For now I will keep the same (simplified) logic in the current initial template, and will try to open a parallel PR for having all the device features exposed as host_vars. I think this could be useful not only for this template, but for any other template that inherit properties from "parent" devices.

ipspace: > will try to open a parallel PR for having all the device features exposed as host_vars.

If you're looking for something accessible by Jinja2 templates within Ansible, modify the ansible output module to add **features** dictionary to the device group vars.

ssasso: > If you're looking for something accessible by Jinja2 templates within Ansible, modify the ansible output module to add **features** dictionary to the device group vars.

Yap, that's exactly what I was thinking of (exposing the whole `fatures` dict via group v).

ssasso:
> Not covered by this PR, but I noticed a `unit_0_trick` being applied to OSPF interfaces and I was wondering: What about ISIS?

please open a separate issue on this and I will try to create a parallel PR - tnx :)

ssasso:
(on wait.... the transformation testing :/ )

ipspace (approved):
Looks good to me, although I would prefer having the image changes in a separate PR from the release-notes perspective. Will run the integration tests later today or tomorrow.

ssasso:
> Will run the integration tests later today or tomorrow.

thank you!

(If you are going to use vJunos-EVO with containerlab/vrnetlab, be sure your image is built with this patch applied: https://github.com/hellt/vrnetlab/pull/301 )

PR #1715, opened 2024-12-31
Cumulus_NVUE: Add mlag support

jbemmel:
* Generalize lag module features, adding ```global``` option to use per-node peerlink settings instead of per interface
* Add ```backup_ip``` option
* Update documentation


ipspace:
A merge conflict must be resolved first.

jbemmel:
Rebased and validated using ```netlab up integration/lag/04-host-mlag.yml -p clab -d cumulus_nvue -v``` (after adding #1716)

PR #1718, opened 2025-01-01
Lag: check mode support by platform

jbemmel:
Recent lag enhancements extend ```lag.mode```:
```mode: { type: str, valid_values: [ "802.3ad", "balance-rr", "balance-xor", "broadcast" ] }```

However, only Linux and FRR platforms support the full range of options; most platforms can only do 802.3ad with/without LACP and use a platform specific hashing algorithm which may or may not be "balance-xor"

Cumulus maps static lags to ```balance-xor``` mode at the Linux kernel level

This PR adds a check for platform lag mode support, preventing unsupported combinations


ipspace:
Will merge after you fix the merge conflict.

PR #1720, opened 2025-01-02
Add device features dict to ansible group_vars

ssasso:


netsim/outputs/ansible.py
```diff
@@ -170,6 +170,8 @@ def create(topology: Box) -> Box:
       if group in defaults.devices:
         devdata = defaults.devices[group]
         group_vars = devdata.group_vars + devdata[defaults.provider].group_vars
+        # add device features to device group_vars
+        group_vars['features'] = devdata.get('features', {})
```
ipspace, line 174: As 'group_vars' is a Box, use group_vars.features. Also, as we're dealing with Boxes, just go for devdata.features (I wouldn't expect to see a device with no features these days)

ipspace (changes requested):

ipspace (approved):

PR #1724, opened 2025-01-02
Node.clone: Allow override of properties by including cloned node name in topology

jbemmel:
Use case: Declare a large number of hosts as 'unmanaged', instantiate one or two for validation testing

Rather than reject clones matching existing node names, this PR merges their properties


ipspace (changes requested):
The 'device: none' hack is not going to get merged. I would suggest you rewrite the example to use 'unmanaged: true'. Also, you might want to document that you could change parameters of individual nodes.

PR #1728, opened 2025-01-03
Links: update hint to remove obsolete 'type' method of controlling addressing

jbemmel:
#1673 deprecated the use of ```link.type``` to select addressing pools


ipspace (approved):
Good catch. Thank you!

PR #1737, opened 2025-01-03
Ubuntu: Add static routes per interface instead of a separate single file

jbemmel:
Update: Modified to add static routes under each interface separately, such that the interface type is automatically correct


ipspace (changes requested):
No shortcuts. A Linux host could (theoretically) have static routes point to bond interfaces and native Ethernet interfaces.

jbemmel:
> No shortcuts. A Linux host could (theoretically) have static routes point to bond interfaces and native Ethernet interfaces.

ok, in that case I would propose to add the static routes in ```/etc/netplan/03-eth-[ifname].yaml``` instead of that separate single file

PR #1739, opened 2025-01-03
Validated STP/RSTP/PVRST integration tests

ipspace:
The original tests were moved to netlab-examples repository.


tests/integration/stp/02-stp-trunk.yml
```diff
@@ -0,0 +1,62 @@
...
+    mode: bridge
+    id: 42
+  blue:
+    mode: bridge
+    id: 76
+
+links:
+- dut:
+  x1:
+  vlan.trunk: [ red, blue ]
```
jbemmel, line 33: Test relies on the native untagged VLAN 1 being allowed on the trunk, could make that explicit

tests/integration/stp/03-stp-port-priority.yml
```diff
@@ -0,0 +1,73 @@
...
+validate:
+  eth3_fwd:
+    wait: 40
+    wait_msg: Waiting for STP
+    nodes: [ x1 ]
+    devices: [ frr ]
+    pass: Third link is forwarding; DUT set the correct port priority
+    fail: Third link is not forwarding; DUT probably did not set port priority
+    stop_on_error: True
+    exec: bridge link show|grep {{ interfaces[2].ifname }}
```
jbemmel, line 47: Is there a command that shows the port priority received from the neighbor? If so, could test for that explicitly (also to test correct multipliers)

tests/integration/stp/20-stp-port-type.yml
```diff
@@ -0,0 +1,53 @@
...
+  vlan.trunk: [ red, blue ]
+  stp.port_type: network
+
+validate:
+  edge:                     # The only thing we can test is the edge port type
+    wait: 3                 # It should be enabled in a few seconds
+    nodes: [ h1 ]
+    pass: H1 and H2 are connected to edge ports
+    fail: H1 and H2 are probably connected to regular STP-controlled ports
+    plugin: ping('h2',count=2)
```
jbemmel, line 53: This could benefit from a validation plugin that looks for a particular string in the device config...

jbemmel (approved):
Nice and comprehensive, some optional improvements suggested

tests/integration/stp/02-stp-trunk.yml
ipspace: With the "legacy" STP, BPDUs are always sent untagged, regardless of what you configure on the trunk. See https://lostintransit.se/2024/07/16/encapsulation-of-pdus-on-trunk-ports/ for details

tests/integration/stp/03-stp-port-priority.yml
ipspace: Good one. Found an EOS command. Changed.

tests/integration/stp/20-stp-port-type.yml
ipspace: That would assume we can somehow figure out what how interfaces and STP parameters are configured on every device out there _without having to create device-specific code_. If SNMP vendor MIBs have taught us anything, it's that this is Mission Impossible.

PR #1743, opened 2025-01-04
Fix: Start SVI ifindex values at 40000 (Fixes #1738)

### LINKED ISSUE #1738: ifindex assigned to SVI interfaces typically too low
https://github.com/ipspace/netlab/blob/81289041556c70a4b7ebe601d1fe0d7592ab5268/netsim/modules/vlan.py#L852

Depending on what the last interface happens to be, this can end up below 1000 resulting in https://github.com/ipspace/netlab/blob/81289041556c70a4b7ebe601d1fe0d7592ab5268/netsim/providers/libvirt.py#L210 counting them as physical interfaces, and not padding the VM interfaces correctly

ipspace:
There is no reason to start SVI ifindex values immediately after the last already-created interface ifindex. This change makes SVI ifindex processing align with loopback and tunnel interfaces.

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -16,7 +16,7 @@ from ..data.global_vars import get_const
 from . import devices,addressing
 
 VIRTUAL_INTERFACE_TYPES: typing.Final[typing.List[str]] = [
-  'loopback', 'tunnel', 'lag' ]
+  'loopback', 'tunnel', 'lag', 'svi' ]
 
 def adjust_interface_list(iflist: list, link: Box, nodes: Box) -> list:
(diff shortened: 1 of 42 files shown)


jbemmel (approved):
There is no reason AND it breaks https://github.com/ipspace/netlab/blob/48ae98cba254fc4bb71f54c231dc30bbcf53170c/netsim/augment/links.py#L204 (which assumes a given range is exclusively used by a given interface type)

The bug could lead to duplicate ifindex values in certain rare scenarios, we could check for that (like we check unique interface names). Not a big deal though

Issue #1745, opened 2025-01-04
[BUG] FRR STP errors


ipspace:
* FRR cannot run regular STP on a trunk as each bridge has its own STP instance (it runs PVRST) (disable it with a quirk?)
* Port priority seems to be failing (disable it?)
* PVRST is probably not configured correctly (probably per-VLAN STP root priority)

https://tests.netlab.tools/_html/frr-clab-stp

PR #1749, opened 2025-01-04
FRR: STP fixes

jbemmel:
* Add quirk to block STP in VLAN trunking scenario's; peers don't process tagged STP BPDUs
* Stop pretending that FRR supports PVRST; it only does tagged STP per VLAN
* Fix STP port priority multiplier; use ```bridge``` command instead of deprecated ```brctl```
* Apply per VLAN STP priority if configured

Fixes https://github.com/ipspace/netlab/issues/1745


ipspace:
As I've been telling you several times, Linux per-bridge STP effectively implements per-VLAN STP, but I have no problems merging this as-is.

ipspace (changes requested):
The documentation needs to be adjusted as well.

PR #1753, opened 2025-01-04
Validated LAG/MLAG integration tests

ipspace:
The original LAG scenarios have been moved to netlab-examples repo.


tests/integration/lag/11-mlag-anycast.yml
```diff
@@ -0,0 +1,54 @@
...
+    pass: The LAG is active
+    fail: X1 cannot establish a LAG with both switches in MLAG pair
+    show: port-channel | json
+    valid: >-
+      len(portChannels['Port-Channel1'].activePorts) == 2
+
+  ping:
+    description: End-to-end connectivity test
+    nodes: [ h1 ]
+    wait_msg: Waiting for STP to enable the ports
```
jbemmel, line 52: Dell OS10 has a (default) 90 second delay-restore timer for MLAG, we may be waiting for that here too

jbemmel (approved):

tests/integration/lag/11-mlag-anycast.yml
ipspace: Can you reduce that delay (who has time to wait forever for vendor ****?) Otherwise I can extend the wait period for OS10.

Issue #1752, opened 2025-01-04
[BUG] Multiple Dell OS10 LAG issues


ipspace:
* Probably a MTU issue on the LAG interface
* MLAG scenarios negotiate a single link in the LAG

Test results @ https://tests.netlab.tools/_html/dellos10-libvirt-lag

jbemmel:
Tests pass for me using vrnetlab/dell_ftos10v release 10.5.6.5

Will see if I can reproduce using Libvirt

jbemmel:
Using ```netlab up integration/lag/10-mlag.yml -p libvirt -d dellos10```
```
jeroen@j:~/Projects/netlab/tests$ netlab validate -v
Read YAML data from netlab.snapshot.yml
[WARNING] Initial wait time extended by 30 seconds required by dellos10
[lag]     Check LAG ports on X1 [ node(s): x1 ]
run_command executing: ['sshpass', '-p', 'vagrant', 'ssh', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no', '-o', 'LogLevel=ERROR', 'vagrant@192.168.121.103', 'show', 'port-channel', '|', 'json']
[PASS]    Validation succeeded on x1
[PASS]    The LAG is active

[ping]    End-to-end connectivity test [ node(s): h2,h3 ]
run_command executing: ['docker', 'exec', '-it', 'clab-lag-h2', 'sh', '-il', '-c', 'ping -c 5 -W 1 -A h1 || true']
[PASS]    h2: Ping to h1 (172.16.0.4) succeeded
run_command executing: ['docker', 'exec', '-it', 'clab-lag-h3', 'sh', '-il', '-c', 'ping -c 5 -W 1 -A h1 || true']
[PASS]    h3: Ping to h1 (172.16.0.4) succeeded
[PASS]    Test succeeded

[SUCCESS] Tests passed: 3
```

It took a while before the test passed - Dell has a 90 second delay-restore-timer for example. 

My Vagrant versions:
arista/veos                 (libvirt, 4.33.1F)
dell/os10                   (libvirt, 10.5.6.5)

jbemmel:
Same for 11-mlag-anycast - test does work, but takes a few moments to setup

ipspace:
The MTU bug is gone. I increased the wait time for OS10 MLAG tests to some ridiculous value, and the tests pass.

It's up to you whether you want to set a lower default value in the config template.

PR #1758, opened 2025-01-05
Dell OS10: Reduce lag delay-restore timer from 90 to 30 seconds

jbemmel:
Should help with https://github.com/ipspace/netlab/issues/1752

Updated to make it a configurable parameter for the lag module

### DIFF
--- a/docs/module/lag.md
+++ b/docs/module/lag.md
@@ -19,7 +19,7 @@ LAG is currently supported on these platforms:
 ## Parameters
 
 The following parameters can be set globally or per node:
-* **lag.mlag.delay_restore**: Time in seconds to wait before bringing up MLAG ports after disruptions or startup, default 30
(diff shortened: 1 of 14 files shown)


ipspace:
1. I don't think this is worth another configuration parameter. Set it to a reasonable value in the configuration template, and if someone wants to change it, let them deal with that in a custom config template. Worst case, if you have an irresistible itch to make this configurable for Dell OS10, use a parameter name with an underscore (so it does not need to be defined as an attribute) and a default value in the config template.
2. There's no need to set this for EOS. EOS works just fine as-is, and restarting a lab device is problematic anyway.
3. There's absolutely no need to set this as a global default value that's inherited by every MLAG implementation.

Please stop solving problems of a single crappy platform with a global solution. It's like I would have introduced a half-dozen OSPF and IS-IS parameters just because some vendors have ridiculously large default timers.

We're not working for a vendor and don't have to support a gazillion nerd knobs, plus the users always have an option of extending the platform if they want to have vendor nerd knobs configurable through lab topology.

PR #1761, opened 2025-01-05
Dell OS10: Reduce mlag delay-restore timer from 90 to 30

jbemmel:
Going back to my original quick fix


ipspace (approved):
Attaboy 👍 Short and efficient 😊

PR #1771, opened 2025-01-06
AOS-CX: optimize configuration task

ssasso:
Instead of invoking two times `aoscx_config`, we can add the `auto-confirm` as *before* param of the "real" module call.


ipspace (commented):
Looks reasonable. I'll run a bunch of integration tests on AOS-CX anyway, so I'm pretty sure any potential problem will quickly pop up ;)

PR #1777, opened 2025-01-07
Fix: Add check for static loopback interface address validity

### LINKED ISSUE #1767: Problematic loopback ipv4 address used in BGP 08-ibgp-localas test highlights issues
https://github.com/ipspace/netlab/blob/d1295e87ff7a6f0ac9fed1a68d8c836f720dc7c2/tests/integration/bgp/08-ibgp-localas.yml#L24

This causes issues with Dell OS10: https://github.com/ipspace/netlab/blob/d1295e87ff7a6f0ac9fed1a68d8c836f720dc7c2/netsim/ansible/templates/initial/dellos10.j2#L10

The template assumes it can use ```v.loopback.ipv4|ipaddr('address')``` to retrieve the address

There is more code like that: https://github.com/ipspace/netlab/blob/d1295e87ff7a6f0ac9fed1a68d8c836f720dc7c2/netsim/ansible/templates/bgp/junos.j2#L71

ipspace:
Fixes #1767

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -515,6 +515,40 @@ def get_nth_ip_from_prefix(pfx: netaddr.IPNetwork, n_id: int) -> str:
   return str(node_addr)
 
 """
+check_interface_host_bits: Check whether the IP addresses on an interface have host bits. The check is disabled
+for special prefixes (IPv4: /31, /32, IPv6: /127, /128)
+"""
+def check_interface_host_bits(intf: Box, node: Box) -> bool:
+  OK = True
+
+  for af in log.AF_LIST:
+    if af not in intf:                          # Skip unusued AFs
+      continue
+    if not isinstance(intf[af],str):            # Skip unnumbered interfaces
+      continue
+    pfx = netaddr.IPNetwork(intf[af])
+    if pfx.last <= pfx.first + 1:               # Are we dealing with special prefix (loopback or /31)
+      continue                                  # ... then it's OK not to have host bits
+
+    if str(pfx) == str(pfx.cidr):               # Does the IP address have host bits -- is it different from its CIDR subnet?
+      log.error(
+        f'Address {intf[af]} on interface {intf.ifname}/node {node.name} does not contain host bits',
+        log.IncorrectValue,
+        'links')
+      OK = False
+      continue
+
+    if str(pfx[-1]) == str(pfx.ip):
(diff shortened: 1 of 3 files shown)


netsim/augment/links.py
```diff
@@ -514,6 +514,40 @@ def get_nth_ip_from_prefix(pfx: netaddr.IPNetwork, n_id: int) -> str:
...
+
+    if str(pfx) == str(pfx.cidr):               # Does the IP address have host bits -- is it different from its CIDR subnet?
+      log.error(
+        f'Address {intf[af]} on interface {intf.ifname}/node {node.name} does not contain host bits',
+        log.IncorrectValue,
+        'links')
+      OK = False
+      continue
+
+    if str(pfx[-1]) == str(pfx.ip):
```
jbemmel, line 541: RFC4291 section 2: "In IPv6, all zeros and all ones are legal values for any field, unless specifically excluded.  Specifically, prefixes may contain, or end with, zero-valued fields."

It then goes on (in 2.6.1) to reserve the all-zeros value for the "Subnet-Router anycast address". But the all-1s value is available for usage

jbemmel:
Shouldn't we also disallow multicast ranges while we're at it? ```loopback.ipv4=224.0.0.1/32``` is silently ignored by FRR

In my mind, all these checks could/should be part of the ```interface``` type, to be enforced whenever we're doing a type check. As constraints on the allowed values, that's where they belong. As it is, someone has to remember to invoke the correct method at the appropriate moment

ipspace:
> Shouldn't we also disallow multicast ranges while we're at it? `loopback.ipv4=224.0.0.1/32` is silently ignored by FRR

That check should go into the must_be_X functions. Will add.

> In my mind, all these checks could/should be part of the `interface` type, to be enforced whenever we're doing a type check.

While you are correct (in principle), there's a gap between theory and practice. We could catch the all-zero case in the "interface" check, but not the all-ones (subnet broadcast) one because that depends on the link prefix.

Anyway, will see what I can do to get as close as feasible.

netsim/augment/links.py
ipspace: > It then goes on (in 2.6.1) to reserve the all-zeros value for the "Subnet-Router anycast address". But the all-1s value is available for usage

In theory, you're correct for the IPv6 case, so I will add the "AF == IPv4" check.

PR #1782, opened 2025-01-09
Cumulus_NVUE: Add gateway support

jbemmel:
* Anycast ipv4 (ipv6 also configured if provided, module officially only supports ipv4)
* VRRPv3 ipv4 and ipv6, with preempt and priority
* Update documentation

Validation:
1. netlab create integration/gateway/01-anycast.yml -p clab -d cumulus_nvue
2. netlab up integration/gateway/02-vrrp.yml -p libvirt -d cumulus_nvue (clab doesn't work due to the ipv6 RA bug in NVUE) - 2 warnings related to unsupported header checksum calculation changes needed for interop (FRR too old)


tests/integration/gateway/03-vrrp-ipv6.yml
```diff
@@ -0,0 +1,114 @@
...
+      template: ifdown
+      variable.ifstate: 'down'
+    pass: R2 has been disconnected from the VRRP LAN
+    stop_on_error: True
+  ra:
+    description: Check RA-generated default route
+    wait: 30
+    wait_msg: Waiting for RA message to generate the default route
+    nodes: [ h1, h2 ]
+    plugin: default6()
```
jbemmel, line 57: Issue here is multiple default routes

ipspace: Yeah, that's always the issue with multiple IPv6 routers attached to the same subnet. I wrote about that 11 years ago: https://blog.ipspace.net/2014/02/first-hop-load-balancing-in-ipv6/

ipspace (approved):
Passed the integration tests ;)

jbemmel:
> Passed the integration tests ;)

If the mountain 🏔️ won't go to J...

ipspace:
> > Passed the integration tests ;)
> 
> If the mountain 🏔️ won't go to J...

OK, with the expected warning ;)) We're not yet in the Full Volkswagen mode 🤪

PR #1794, opened 2025-01-09
Cumulus: Insert newline before 'bond-lacp-rate slow'

jbemmel:


netsim/ansible/templates/lag/cumulus.j2
```diff
@@ -32,4 +32,4 @@ echo "LAG: executing ifreload"
 #
 until ifreload -a; do
```
jbemmel, line 33: This loop is not a good idea, as it will hang indefinitely in case of bad config. It should reload once, and fail in case of error

Issue #1669, opened 2024-12-19
[BUG] IPv6 is not working over multi-access clab links


ipspace:
IPv6 is not working between three nodes connected to the same link. The same setup works with libvirt, meaning we probably have a problem with the setup of the netlab-created Linux bridge.

### To Reproduce

Run the following topology with **libvirt** provider. X1 can ping X2. Restart the same topology with **clab** provider. X1 can reach X2 over IPv4 but not over IPv6.

### Lab topology

```
---
addressing:
  loopback:
    ipv6: 2001:db8:1::/48
  lan:
    ipv6: 2001:db8:2::/48

defaults.device: linux

nodes:
  x1:
  x2:
  x3:

links:
- x1-x2-x3
```

### Version

```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
```

ipspace:
So far I've identified these **sysctl net.ipv6.conf._brname_** differences (working values)

```
accept_ra = 0
autoconf = 0
disable_ipv6 = 1
mtu = 1500
```

There must also be something in the **/sys/devices/virtual/net/_brname_** settings.

ipspace:
Snooping on **x1_eth1** and **x2_eth2** (the two clab interfaces connected to the netlab-created Linux bridge) confirms that the Linux bridge drops IPv6 multicast packets (neighbor solicitation is seen on x1 but not received on x2).

Before someone starts quoting decade-old hits from Google: yes, I did `echo 0 >/sys/class/net/X_1/bridge/multicast_snooping` and no, it did not help.

Any working solution would be highly appreciated, I'm out of ideas.

ipspace:
The mystery deepens: IPv6 MLD snooping works correctly. The **bridge mdb** table is populated:

```
dev X_1 port x1_eth1 grp ff02::1:ff00:1 temp
dev X_1 port x3_eth1 grp ff02::1:ff00:3 temp
dev X_1 port x2_eth1 grp ff02::1:ff00:2 temp
dev X_1 port X_1 grp ff02::fb temp
dev X_1 port X_1 grp ff02::1:3 temp
dev X_1 port x3_eth1 grp ff02::1:ff1c:7d48 temp
dev X_1 port x2_eth1 grp ff02::1:fffa:7833 temp
dev X_1 port x1_eth1 grp ff02::1:ff1c:997 temp
[...]

PR #1800, opened 2025-01-12
Fix clab ipv6 forwarding on bridges

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1669

Temporary fix until Containerlab catches up

* Using OVS bridge doesn't have the same issue
* Cleaning up rules is a little cumbersome, as it requires the handle

Note: Verbose debug prints don't work upon 'down' - see other issue

### DIFF
--- a/netsim/providers/clab.py
+++ b/netsim/providers/clab.py
@@ -40,6 +40,11 @@ def create_linux_bridge( brname: str ) -> bool:
   log.print_verbose( f"Enable Linux bridge '{brname}': {status}" )
 
   status = linuxbridge.configure_bridge_forwarding(brname)
+  if status:
+    status = external_commands.run_command(
+      ['sudo','nft',f'insert rule ip6 filter DOCKER-USER oifname "{brname}" counter accept'],
+      check_result=True,return_stdout=True)
+    log.print_verbose(f"Insert ipv6 nftables rule for Linux bridge '{brname}': {status}")
   return status
 
 def destroy_linux_bridge( brname: str ) -> bool:
--- a/netsim/providers/clab.py
+++ b/netsim/providers/clab.py
@@ -40,11 +40,15 @@ def create_linux_bridge( brname: str ) -> bool:
   log.print_verbose( f"Enable Linux bridge '{brname}': {status}" )
 
   status = linuxbridge.configure_bridge_forwarding(brname)
-  if status:
-    status = external_commands.run_command(
-      ['sudo','nft',f'insert rule ip6 filter DOCKER-USER oifname "{brname}" counter accept'],
-      check_result=True,return_stdout=True)
-    log.print_verbose(f"Insert ipv6 nftables rule for Linux bridge '{brname}': {status}")
+  if status is False:
+    return False
+  return insert_ipv6_forwarding_rule( brname )
+
+def insert_ipv6_forwarding_rule( brname: str ) -> bool:
+  status = external_commands.run_command(
+    ['sudo','nft',f'insert rule ip6 filter DOCKER-USER oifname "{brname}" counter accept'],
+    check_result=True,return_stdout=True)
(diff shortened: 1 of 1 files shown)


jbemmel:
image

Libvirt creates rules with both ```iifname``` and ```oifname```, perhaps that would be better?

ipspace:
We're not going to start playing the security whack-a-mole just because you don't find it convenient to wait for the containerlab fix, particularly considering that:

A) the problem depends on the security setup of the server (considering the same setup works on my server and on a fresh Ubuntu 24.04 Vagrant box)
B) you're the first one reporting it
C) nobody tested this across various Linux distros and I don't think that would be a good investment of anyone's time anyway.

We should document the potential issue in the clab documentation though.

jbemmel:
That’s fine, I agree and I figured you might feel that way - changing security settings is risky. But I wanted to give you that choice

ipspace:
I will definitely polish the clab caveats you wrote and merge them. I the meantime, please check:

* Whether you have br_netfilter module loaded (`lsmod|grep net`)
* What are the `sysctl net.bridge` settings?

ipspace:
FWIW, on my server **br_netfilter** definitely caused #1669. After loading it, IPv6 connectivity immediately stopped working.

Setting `sudo sysctl net.bridge.bridge-nf-call-ip6tables=0` "fixed" the problem, but I don't feel comfortable doing more than mentioning it in the caveats. Can you check whether this works for you?

jbemmel:
> FWIW, on my server **br_netfilter** definitely caused #1669. After loading it, IPv6 connectivity immediately stopped working.
> 

It's loaded on mine by the Docker snap:
```
jeroen@j:~/Projects/netlab/tests$ cat /etc/modules-load.d/snap.docker.conf
# This file is automatically generated.
arp_tables
br_netfilter
ip6table_filter
iptable_filter
overlay
```

> Setting `sudo sysctl net.bridge.bridge-nf-call-ip6tables=0` "fixed" the problem, but I don't feel comfortable doing more than mentioning it in the caveats. Can you check whether this works for you?

```
jeroen@j:~/Projects/netlab/tests$ cat /proc/sys/net/bridge/bridge-nf-call-ip6tables
1
```
Setting it to '0' indeed allows IPv6 test validation to succeed

However, there's also individual bridges:
```
jeroen@j:~/Projects/netlab/tests$ cat /sys/devices/virtual/net/virbr0/bridge/nf_call_ip6tables
0
jeroen@j:~/Projects/netlab/tests$ cat /sys/devices/virtual/net/docker0/bridge/nf_call_ip6tables
0
```
but these are already ```0``` also for Netlab created bridges.

jbemmel:
> > FWIW, on my server **br_netfilter** definitely caused #1669. After loading it, IPv6 connectivity immediately stopped working.
> 
> It's loaded on mine by the Docker snap:
> 
> ```
> jeroen@j:~/Projects/netlab/tests$ cat /etc/modules-load.d/snap.docker.conf
> # This file is automatically generated.
> arp_tables
> br_netfilter
> ip6table_filter
> iptable_filter
> overlay
> ```

FYI - ```snap remove docker --purge``` followed by ```netlab install containerlab``` also solves the issue; the ```br_netfilter``` module is no longer loaded at startup.

In summary: We could detect the presence of the 'br_netfilter' module and issue a warning about possible ipv6 interference, and we could recommend to try reinstalling Docker using ```netlab install``` in case of issues

ipspace:
> Setting it to '0' indeed allows IPv6 test validation to succeed

👍 Now we know what the issue is ;)

> However, there's also individual bridges:
> but these are already `0` also for Netlab created bridges.

I tried setting those variables, but it seems like they have no effect whatsoever.

jbemmel:
Containerlab released a fix: https://containerlab.dev/rn/0.62/#ipv6-allow-rules

PR #1804, opened 2025-01-14
Update to latest Containerlab version

jbemmel:
No need to downgrade


ipspace:
Will upgrade my server and then run it for a few days just to be on the safe side ;)

jbemmel:
Released today: https://containerlab.dev/rn/0.62/#release-062

ipspace (approved):

PR #1811, opened 2025-01-14
Check for ovs-vsctl presence when using OVS with clab (fixes #1799)

### LINKED ISSUE #1799: [BUG] Setting bridge_type to 'ovs-bridge' fails to check for binary ovs-vsctl
Setting ```bridge_type``` to "ovs-bridge" for Containerlab fails when ```ovs-vsctl``` cannot be found

ipspace:


netsim/cli/external_commands.py
```diff
@@ -42,6 +42,12 @@ def add_netlab_path() -> None:
 
   return
 
+"""
+has_command: figures out whether a command is available
+"""
+def has_command(cmd: str) -> typing.Union[bool,int,str]:
```
jbemmel, line 48:
```
-> bool:

return bool( ... )
```

jbemmel (approved):
Would suggest to simplify the return type

PR #1826, opened 2025-01-21
Implement 'normalize' configuration step (fixes #1825)

### LINKED ISSUE #1825: [BUG] Arista EOS and Cisco IOSvL2 boot as unconfigured bridges
Arista EOS (at least cEOS) and Cisco IOSv layer-2 image boot as bridges with data-plane interfaces enabled and bridging untagged traffic. This could (depending on timing) cause leaking of control-plane traffic, for example IPv6 RA, resulting in stuff like #1821.

We should add an extra pre-initial (`normalize` ?) step to the initial configuration deployment (**netlab initial**). It should be executed before the `initial` module and ignore devices that do not have the corresponding templates.

ipspace:
The 'normalize' configuration step is executed before the 'initial' step. It's triggered with the 'initial' tag (so it's always executed together with 'initial' step) and uses an extra variable to signal the deploy-module task list that it's OK if the template is missing (as it would be for most devices).

The commit also includes the normalization templates for EOS and IOS layer-2 images


netsim/ansible/templates/normalize/eos.j2
```diff
@@ -0,0 +1,6 @@
+{% for intf in interfaces if intf.virtual_interface is not defined %}
+!
+interface {{ intf.ifname }}
+ no switchport
+ mac-address {{ '52dc.cafe.%02d%02d' % ( id,intf.ifindex % 100 ) }}
```
jbemmel, line 5: This will break in topologies with more than 255 nodes, or when users use a custom node ID > 255

jbemmel (approved):

netsim/ansible/templates/normalize/eos.j2
ipspace: The attribute limit is 150 (based on the default management network settings), the hard-coded limit in nodes.py is 250 (because we use ID in things like MAC addresses). Anyway, the formatting string needs to be changed to hex, or anything over 99 will break things.

PR #1832, opened 2025-01-24
Cumulus NVUE: Add VXLAN support

jbemmel:
Passes all integration tests except 07 (IPv6 VTEP not supported, blocked on lack of support in ifupdown2)

* Test ```04-vxlan-irb-ospf.yml```: Added a quirk to remove ospf.passive configuration from VRF interfaces, with warning
* Test ```08-vxlan-alt-vtep.yml```: Added a quirk to merge loopback OSPF settings if they are identical, else throw error

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -205,6 +205,7 @@ ansible-galaxy collection install git+https://github.com/jmussmann/ansible_colle
 
 * The Cumulus VX 4.4.0 Vagrant box for VirtualBox is broken. *netlab* is using Cumulus VX 4.3.0 with *virtualbox* virtualization provider.
 * The Cumulus VX 4.x uses Python version 3.7, which recent versions of Ansible refuse to work with. The permanent fix is coming in release 1.9.3. Until then, use the **frrouting** device or [Cumulus VX 5.x image](caveats-cumulus-5x).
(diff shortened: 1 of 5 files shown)


jbemmel:
Test case 4 failure:
```
Feb 01 16:00:18 s1 frrinit.sh[14443]: This command is deprecated, because it is not VRF-aware.
Feb 01 16:00:18 s1 frrinit.sh[14443]: Please, use "ip ospf passive" on an interface instead.
Feb 01 16:00:18 s1 frrinit.sh[14443]: interface vlan1001 not found.
Feb 01 16:00:18 s1 frrinit.sh[14443]: line 47: Failure to communicate[13] to ospfd, line:  passive-interface vlan1001
```

vlan1001 is inside a VRF, so NVUE cannot handle it

ipspace:
The "check for OSPF" quirk would prevent hosts being attached to VRF interfaces (interfaces with a single router are always made passive). We could rely on the user clearing that flag in system defaults or clear the passive flag and create a warning (category=Warning in report_quirk).

ipspace:
IRB tests are failing on my end, although admittedly I had to run them with release 5.9.2.

ipspace:
Works with 5.12.0, adding a caveat

PR #1834, opened 2025-01-25
Cumulus NVUE: Fix case of custom loopbacks with OSPF

jbemmel:
Extra loopbacks are a problem with NVUE, due to its over-simplified model of having a single loopback interface "lo"

This fix is not great, could add a quirk to detect unsupported combinations, e.g. multiple loopbacks with different OSPF areas, or declare custom loopbacks unsupported in combination with OSPF. Not to be merged as-is, only a starting point for discussion

Not sure if VRF loopbacks could get advertised in OSPF at all

The OSPF module adds the area to the loopback interface, so I removed the custom area inclusion from the template.

Fixes https://github.com/ipspace/netlab/issues/1836

### DIFF
--- a/netsim/ansible/templates/ospf/cumulus_nvue.j2
+++ b/netsim/ansible/templates/ospf/cumulus_nvue.j2
@@ -58,7 +58,7 @@
 {% endmacro %}
 
 {% if ospf is defined %}
-{% set _lo = (loopback | combine( { 'ospf': { 'area': ospf.area|default('0.0.0.0') } } )) if 'ipv4' in loopback else {} %}
-{% set _ospf_intfs = [_lo] + interfaces|default([])|selectattr('ospf','defined')|list %}
+{% set _lo = [loopback] if 'ipv4' in loopback else [] %}
+{% set _ospf_intfs = _lo + interfaces|default([])|selectattr('ospf','defined')|selectattr('type','ne','loopback')|list %}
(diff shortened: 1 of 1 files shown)


ipspace:
> Extra loopbacks are a problem with NVUE, due to its over-simplified model of having a single loopback interface "lo"

Junos is very similar, so whatever we do for one, we can also do for the other.

> This fix is not great, could add a quirk to detect unsupported combinations, e.g. multiple loopbacks with different OSPF areas, or declare custom loopbacks unsupported in combination with OSPF.

I would allow solve it with a quirk (like with Junos). Right now, Junos quirk throws an "unsupported" error, but we could make that configurable (working on "disable error-generating quirks" right now) and in case the error is not thrown collect all loopback IPs, add them to _ipv4 and _ipv6 (or something similar) on node.loopback, and potentially generate a warning about what we did.

> Not sure if VRF loopbacks could get advertised in OSPF at all

Yes they can.

jbemmel:
> 
> > Not sure if VRF loopbacks could get advertised in OSPF at all
> 
> Yes they can.

I mean not sure if it would work using NVUE, it offers functionality to configure an IP as 'loopback' for a VRF but I haven't seen a way to configure its OSPF area for example

jbemmel:
Fixed in https://github.com/ipspace/netlab/pull/1832

PR #1837, opened 2025-01-25
vxlan: Avoid creating empty 'vni' attribute on VLANs

jbemmel:
Many templates use tests like ```if 'vni' in vlan ...```

### DIFF
--- a/netsim/modules/vxlan.py
+++ b/netsim/modules/vxlan.py
@@ -146,7 +146,7 @@ def build_vtep_list(vlan: Box, node: str, nodes: typing.List[str], topology: Box
     ndata = topology.nodes[n]
     if not 'vlans' in ndata:                                        # No VLANs on remote node, skip it
       continue
-    vni_match = filter(lambda x: x.vni == vlan.vni,ndata.vlans.values())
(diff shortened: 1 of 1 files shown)


ipspace:
What happened with 'I will start with a sample topology that demonstrates the problem'?

netsim/modules/vxlan.py
```diff
@@ -146,7 +146,7 @@ def build_vtep_list(vlan: Box, node: str, nodes: typing.List[str], topology: Box
     ndata = topology.nodes[n]
     if not 'vlans' in ndata:                                        # No VLANs on remote node, skip it
       continue
-    vni_match = filter(lambda x: x.vni == vlan.vni,ndata.vlans.values())
+    vni_match = filter(lambda x: x.get('vni',0) == vlan.vni,ndata.vlans.values())
```
ipspace, line 149: Not sure if we allow VNI value zero, but None would definitely be a better option.

jbemmel: This PR was carefully crafted considering all constraints: https://github.com/ipspace/netlab/blob/5b580c076b7f1077688f2475d0e1da278debeaad/netsim/modules/vlan.yml#L80

jbemmel: ...but I'll go with your suggestion

PR #1838, opened 2025-01-25
Dell OS10: Fix default VLAN case of 1

jbemmel:
When creating a vlan trunk on Dell OS10 containing the native vlan 1, OS10 complains when VLAN 1 is included in the trunk since it is also (implicitly) defined as an access VLAN

If the access_id is not set, it defaults to 1 - not 0

Current template raises an error about VLAN 1 being both an access vlan and a trunk vlan

Minimal topology:
```YAML
---
groups:
  _auto_create: True
  dell:
    members: [s1, s2]
    device: dellos10
    module: [vlan]

vlans:
  native:
    id: 1
  normal:
    id: 2

links:
- s1:
  s2:
  vlan.trunk: [ native, normal ]
```

Result:
```
fatal: [s2]: FAILED! => changed=false 
  command: switchport trunk allowed vlan 1,2
  msg: |-
    switchport trunk allowed vlan 1,2
    % Error: Interface cannot be an access and trunk member of the same VLAN. One of the vlan in the range provided is already an access vlan.
    s2(conf-if-eth1/1/1)#
  rc: -32603
  command: switchport trunk allowed vlan 1,2
  msg: |-
    switchport trunk allowed vlan 1,2
    % Error: Interface cannot be an access and trunk member of the same VLAN. One of the vlan in the range provided is already an access vlan.
    s1(conf-if-eth1/1/1)#
  rc: -32603
```

### DIFF
--- a/netsim/ansible/templates/vlan/dellos10.j2
+++ b/netsim/ansible/templates/vlan/dellos10.j2
@@ -21,7 +21,7 @@ interface {{ ifdata.ifname }}
  switchport mode trunk
 
 {%     set vid_list = [] %}
-{%     for vid in ifdata.vlan.trunk_id if vid != ifdata.vlan.access_id|default(0) %}
+{%     for vid in ifdata.vlan.trunk_id if vid != ifdata.vlan.access_id|default(1) %}
 {{       vid_list.append(vid) }}
 {%     endfor %}
 


ipspace:
What happens if someone wants to have VLAN 1 in the trunk?

jbemmel:
> What happens if someone wants to have VLAN 1 in the trunk?

Pick a different platform - it works fine on EOS...

It needs more analysis, the VLAN model on Dell appears to be subtly different than typical

ipspace:
> It needs more analysis, the VLAN model on Dell appears to be subtly different than typical. Same construct works fine on EOS, we may need to unset something in the template

Or add an OS10 quirk saying "you cannot have VLAN 1 in the trunk".

ssasso:
If I understand correctly the situation, and if I recall what I did in the past from my memory, doing `no switchport access vlan` on the trunk port should be enough to allow then to have any vlan id on the trunk itself.

jbemmel:
> If I understand correctly the situation, and if I recall what I did in the past from my memory, doing `no switchport access vlan` on the trunk port should be enough to allow then to have any vlan id on the trunk itself.

Apparently, when creating a port it starts with ```switchport access vlan 1``` enabled.
After switching the port to trunk mode, it is possible to remove the access vlan

However:
```
fatal: [s3]: FAILED! => changed=false 
  command: switchport trunk allowed vlan 1,1000,1001
  msg: |-
    switchport trunk allowed vlan 1,1000,1001
    % Error: Default vlan cannot be configured as tagged vlan
    s3(conf-if-po-1)#
  rc: -32603
```

So on Dell OS10, whatever happens to be the native untagged vlan cannot be part of the trunk, but can be configured as an access vlan

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/1869

PR #1840, opened 2025-01-25
Cumulus and Cumulus NVUE: Implement system default mtu

jbemmel:
The included OSPFv2 VLAN test 06 fails on current dev branch:
```
netlab up integration/ospf/ospfv2/06-vlan.yml -p libvirt -d cumulus_nvue -v
```

This happens because FRR properly implements the 1750 MTU on its VLAN interface, while Cumulus defaults to 9216.
IMHO implementing a system wide default MTU setting helps provide consistency; while it would be possible to explicitly configure VLAN MTU per interface, none of the current platforms does this today


ipspace:
Well, if you decide to fix things nobody asked for, do it right.

I modified the VLAN OSPFv2 test slightly and of course both Cumulus and Cumulus NVUE fail when you specify the MTU for a single VLAN (FWIW, Arista and Junos pass the test). Opening two bug reports and assigning them to you...

ipspace:
> While it would be possible to explicitly configure VLAN MTU per interface, none of the current platforms does this today

That is completely incorrect. ArubaCX, EOS, Junos and VyOS worked out of the box once I limited the MTU values to 1280..1500.

I had to fix FRR template because it tried to set interface MTU before the interface was created and Cisco IOS templates because I decided to use **ip mtu** instead of **mtu** on VLAN/BVI/BDI interface to avoid potential platform-specific quirks. IOSv and a few other platforms would work out-of-the-box, IOL probably would not.

There was also the problem of node MTU not being copied to VLAN interfaces (fixed as well).

I opened relevant bug reports for FRR, Cumulus Linux, and SR Linux, and assigned them to you.

PR #1841, opened 2025-01-26
Dell OS10: Already configures system wide mtu, enable unreachables

jbemmel:
Passes ```initial/04-mtu``` test

The Dell templates already configure a system wide MTU - this PR simplifies the configuration by removing per-interface MTU settings and declaring system MTU as supported

Some tweaks to the initial template to enable unreachables and MTU discovery on ipv6

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -54,7 +54,7 @@ interface {{ l.ifname }}
 {% if 'ipv4' in l %}
 {%   if l.ipv4|ipv4 %}
  ip address {{ l.ipv4 }}
-{%     if l.type|default('') != 'loopback' %}
+{%     if l.virtual_interface is not defined %}
  ip unreachables
 {%     endif %}
 {%   else %}
@@ -76,12 +76,13 @@ interface {{ l.ifname }}
 {%   endif %}
 {%   if l.ipv6 == True %}
  ipv6 enable
- ipv6 unreachables
 {%   elif l.ipv6|ipv6 %}
  ipv6 address {{ l.ipv6 }}
- ipv6 unreachables
 {%   else %}
 ! Invalid IPv6 address {{ l.ipv6 }}
+{%   endif %}
+{%   if l.virtual_interface is not defined %}
+  ipv6 unreachables
 {%   endif %}
 {% elif 'ipv4' in l %}
  no ipv6 enable
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -74,6 +74,7 @@ interface {{ l.ifname }}
 {%   endif %}
 {%   if l.ipv6 == True %}
  ipv6 enable
+ ipv6 unreachables
 {%   elif l.ipv6|ipv6 %}
  ipv6 address {{ l.ipv6 }}
  ipv6 unreachables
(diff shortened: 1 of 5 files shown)


ipspace:
Is this ready to be reviewed?

ipspace (changes requested):
Fails the `initial/02-loopback.yml` integration test:

```
TASK [dellos10_config: deploying initial from /home/pipi/net101/tools/netsim/ansible/templates/initial/dellos10.j2] ****************
fatal: [r]: FAILED! => changed=false
  command: ipv6 unreachables
  msg: |-
    ipv6 unreachables
    % Error: Unrecognized command.
    r(conf-if-lo-1)#
  rc: -32603
```

It's extremely annoying that I have to run tests on every PR you submit because I can never be sure that you didn't break something obvious that you simply didn't feel the need to test.

ipspace:
**ip unreachable** also fails on the **virtual-network** interface, which seems to indicate that you never tested this change with VLANs.

```
TASK [dellos10_config: deploying initial from /home/pipi/net101/tools/netsim/ansible/templates/initial/dellos10.j2] ****************
  command: ip unreachables
  msg: |-
    ip unreachables
    % Error: Unrecognized command.
    s1(conf-if-vn-700)#
  rc: -32603
```

jbemmel:
> Fails the `initial/02-loopback.yml` integration test:
> 
> ```
> TASK [dellos10_config: deploying initial from /home/pipi/net101/tools/netsim/ansible/templates/initial/dellos10.j2] ****************
> fatal: [r]: FAILED! => changed=false
>   command: ipv6 unreachables
>   msg: |-
>     ipv6 unreachables
>     % Error: Unrecognized command.
>     r(conf-if-lo-1)#
>   rc: -32603
> ```
> 
> It's extremely annoying that I have to run tests on every PR you submit because I can never be sure that you didn't break something obvious that you simply didn't feel the need to test.

I can see that, on my end what’s annoying is that Dell OS10 templates use virtual networks while for my usage I need plain VLANs - so sometimes I forget to switch back and run the tests with VNs too. It doubles the testing load

ipspace:
> I can see that, on my end what’s annoying is that Dell OS10 templates use virtual networks while for my usage I need plain VLANs - so sometimes I forget to switch back and run the tests with VNs too. It doubles the testing load

And how do you plan to explain that you fixed the IPv4 unreachables on the loopbacks in d87fe40 but not IPv6 ones?

jbemmel:
> > I can see that, on my end what’s annoying is that Dell OS10 templates use virtual networks while for my usage I need plain VLANs - so sometimes I forget to switch back and run the tests with VNs too. It doubles the testing load
> 
> And how do you plan to explain that you fixed the IPv4 unreachables on the loopbacks in [d87fe40](https://github.com/ipspace/netlab/commit/d87fe40f5d2bae344d22faadd62af2a7e5f6aac2) but not IPv6 ones?

As I stated in the original PR - ```passes initial/04-mtu test```. I don't have an easy way to run all integration tests, ```run-integration-tests.sh``` uses vagrant but I use ```clab``` for my Dell OS10 setups, so I have to pick and run tests manually one by one. I put it in draft after I found the issue with loopback interfaces, and then I left it as draft thinking I'd come back to run more tests. When you asked me whether it was ready, I forgot that I hadn't completed running all the tests

jbemmel:
> 
> It's extremely annoying that I have to run tests on every PR you submit because I can never be sure that you didn't break something obvious that you simply didn't feel the need to test.

As you have noticed (https://github.com/ipspace/netlab/pull/1869/commits/ddec12cdaf11c54d9586bd3b7d0e237cdf82bbd1) I have started putting more emphasis on testing in my PRs. Sometimes this shows up as listing the tests that I ran, sometimes I'll include a custom test case that I crafted for the purposes of that specific PR

It is not that I "didn't feel the need to test", I do. But as I have said before, Netlab has grown to a point where it has become really challenging to test all possible combinations of features, "obvious" or not.

I would suggest we create a script that allows me to run all integration tests in a given directory, and then in the PRs I'll state which tests I ran like that. How about that?

Issue #1843, opened 2025-01-26
[BUG] Cumulus Linux does not set VLAN interface MTU


ipspace:
OSPFv2: 
* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv2/06-vlan-mtu.yml
* Validation results: https://tests.netlab.tools/cumulus/libvirt/ospf/ospfv2/06-vlan-mtu.yml-validate.log

OSPFv3: 
* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv3/07-vlan-mtu.yml
* Validation results: https://tests.netlab.tools/cumulus/libvirt/ospf/ospfv3/07-vlan-mtu.yml-validate.log

PR #1851, opened 2025-01-27
Cumulus: Fix MTU setting on VLAN interfaces

jbemmel:
* Move all IP and MTU assignment to initial, remove it from vlan script
* Remove filter on interface types, simply do all

Fixes https://github.com/ipspace/netlab/issues/1843 (both those test cases pass now, I did test a couple of others but not all integration tests)

Could use some more testing, but I'm sharing for comments


ipspace (approved):
It passes all integration tests. Thank you!

It's interesting that you can add a "bogus" device to the /etc/network/interfaces.d and nothing happens ;)

Issue #1844, opened 2025-01-26
[BUG] SR Linux does not set MTU on VLAN interfaces


ipspace:
OSPFv2:

* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv2/06-vlan-mtu.yml
* Validation results: https://tests.netlab.tools/srlinux/clab/ospf/ospfv2/06-vlan-mtu.yml-validate.log

OSPFv3:

* Test topology: https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv3/07-vlan-mtu.yml
* Validation results: https://tests.netlab.tools/srlinux/clab/ospf/ospfv3/07-vlan-mtu.yml-validate.log

PR #1857, opened 2025-01-28
Srlinux: Configure vlan mtu

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1844

SR Linux only supports MTU values starting from 1436, propose to increase the values used in the integration tests


ipspace (approved):
Thanks for the fixes. I don't see why we should be changing the integration tests just to cope with the weird choices developers of a single device made.

I will probably add a caveat to the integration tests explaining what's going on.

PR #1860, opened 2025-01-28
Dell OS10: Add quirk to check for unsupported use of VRRP on virtual networks

jbemmel:
The use of virtual networks for VLANs really limits the available functionality - VRRP is another example


ipspace:
> The use of virtual networks for VLANs really limits the available functionality - VRRP is another example

We've heard that before. Repeating it won't help, it will just make everyone else annoyed and irritated.

PR #1865, opened 2025-01-30
Cumulus NVUE: Disable ipv6 on l2-only interfaces

### LINKED ISSUE #1864: [BUG] Cumulus_NVUE enables IPv6 LLA on L2-only interfaces
Similar to previous issue, modify ```31-vlan-bridge-trunk.yml```:
```
links:
- s1:
  s2:
  vlan.trunk: [ red, blue ]

- s1:
    ipv4: False
  h5:
  prefix:
    ipv4: 172.31.1.0/24
- s2:
    ipv4: False
  h6:
  prefix:
    ipv4: 172.31.1.0/24
```

On Cumulus NVUE, ```swp2``` gets an IPv6 LLA. Same applies to all VLAN interfaces

Suggested fix: https://github.com/ipspace/netlab/blob/d07e481703cc6fb5851a3699fb3c25059c6a7aed/netsim/ansible/templates/initial/cumulus_nvue.j2#L44

```
{%  else %}
        ip:
          ipv6:
            enable: off
{%    if i.type=='lag' %}
        bridge:
          domain:
            br_default: {}
{%    endif %}
{%  endif %}
```

jbemmel:
Fixes #1864

> It might be better to go for a loop over 'ipv4' and 'ipv6' checking if intf[ax] is defined

I considered that, but the conditions are different for each so I think writing them out separately works better in this case

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -14,27 +14,27 @@
 {%  endif %}
         ip:
 {%  if i.ipv4 is defined or (i.ipv6 is string and i.ipv6|ipv6) %}
-        address:
+          address:
 {%    if i.ipv4 is defined %}
-          {{ (loopback if i.ipv4 == True else i).ipv4 }}: {}
+            {{ (loopback if i.ipv4 == True else i).ipv4 }}: {}
 {%    endif %}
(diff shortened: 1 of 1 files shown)


jbemmel:
Apologies for the list of commits, I'm working from a place where I can only create PRs through the web interface

ipspace (approved):
Looks good, passed initial and VLAN integration tests

Issue #1866, opened 2025-01-30
[BUG] Cumulus_NVUE and Dell OS10 pass native VLAN on trunks by default


jbemmel:
Following up on https://github.com/ipspace/netlab/issues/1862

Modified ```31-vlan-bridge-trunk``` test:
```
vlans:
  red:
    mode: bridge
    prefix:
      ipv4: 172.31.1.0/24
    links: [ s1-h1, s2-h2 ]
  blue:
    mode: bridge
    prefix:
      ipv4: 172.31.1.0/24
    links: [ s1-h3, s2-h4 ]

  untagged: # Untagged links to test that device isn't passing the default VLAN on the trunk
    id: 1
    mode: bridge
    prefix:
      ipv4: 172.31.1.0/24
    links: [ s1-h5, s2-h6 ]

validate:
  ...
  untagged_vlan:
    description: Ping-based reachability test on untagged VLAN
    nodes: [ h5 ]
    plugin: ping('h6',expect='fail')
```

Both Cumulus_NVUE and Dell OS10 currently fail this test

jbemmel:
Suggested fix for Cumulus NVUE in VLAN templates:
```
- set:
    bridge:
      domain:
        br_default:
          type: vlan-aware
          mac-address: "{{ '08:4f:c2:a9:00:%02x' | format(id) }}" # The default 'auto' setting reads the MAC from the EEPROM, which is the same for multiple instances
          untagged: none
{% if vlans is defined %}
{%  for vname,vdata in vlans.items() %}
{%   if loop.first %}
          vlan:
{%   endif %}
            '{{ vdata.id }}': {}
{%  endfor %}
{% endif %}

{# Note: interface.xyz.vlan is only used for SVI interfaces in case of multiple bridges #}

{% for i in interfaces if i.vlan is defined and (i.virtual_interface is not defined or i.type=="lag") %}
{%   if loop.first %}
    interface:
{%   endif %}
     {{ i.ifname }}:
       bridge:
         domain:
           br_default:
{%   if i.vlan.trunk_id is defined +%}
             vlan:
{%     for v in i.vlan.trunk_id|sort %}
               '{{ v }}': {}
{%     endfor %}
             untagged: {{ i.vlan.access_id if 'native' in i.vlan else 'none' }}
{%   elif i.vlan.access_id is defined %}
             access: {{ i.vlan.access_id }}
{%   endif %}
{% endfor %}
```

jbemmel:
[...]

PR #1867, opened 2025-01-31
Cumulus NVUE: Don't trunk native VLAN unless requested

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1866 (Cumulus NVUE part)

Still passes VLAN tests 31 (modified), 32, 33 and 42 and initial test 01


ipspace (approved):
Passes the VLAN integration tests

PR #1869, opened 2025-01-31
Dell OS10: Don't allow untagged vlan on trunk unless configured

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1866

* There is no global definition for the default VLAN in Netlab, in theory it could be configurable but this PR assumes it is ```1```. 
* Dell OS10 does not accept the untagged VLAN (default:1) as part of the allow list on a trunk, but requires it to be put as an access vlan configured alongside the trunk
* The test case confirms interop between EOS and Dell OS10 for a trunk that contains VLAN 1


ssasso (approved):

ipspace:
In principle, there's a difference between VLAN 1 being a native VLAN on a VLAN trunk or VLAN 1 being part of a VLAN trunk, but if that's what it takes to make OS10 work... Whatever 🤦‍♂️

Issue #1868, opened 2025-01-31
[BUG] Dell OS10 enables IPv6 LLA on L2-only interfaces


jbemmel:
Similar to #1864, issue is here: https://github.com/ipspace/netlab/blob/d07e481703cc6fb5851a3699fb3c25059c6a7aed/netsim/ansible/templates/initial/dellos10.j2#L78

This statement was questioned during review, turns out it was indeed sub-optimal

Tentative fix:
```
{% elif 'lag' not in l %}
{# Don't configure IP related settings for discovery interfaces, L2-only port-channels or lag member links #}
 no ipv6 enable
{% endif %}
```

PR #1870, opened 2025-01-31
Dell OS10: Don't enable IPv6 for L2-only interfaces or VLANs

jbemmel:
Expand the types of interfaces for which to disable ipv6: Everything except LAG related interfaces such as peerlinks, port-channels and lag member interfaces.

NB L2-only interfaces include bridged VLANs

Fixes https://github.com/ipspace/netlab/issues/1868

Clear counters on virtual network interfaces (they remain 0)

Tested using OS Version 10.5.6.5 on clab:
* ```lag/11-mlag-anycast.yml```
* ```vlan/21-vlan-irb-single.yml```
* ```vlan/22-vlan-irb-multiple.yml``` (modified to add ipv6 tests in separate PR)
* ```vlan/41-vlan-bridge-native.yml```

Note: Only one of the VLAN integration tests includes ipv6 testing (21), and not between hosts - I figured it would be good to add one more

### DIFF
--- a/netsim/ansible/templates/initial/dellos10.j2
+++ b/netsim/ansible/templates/initial/dellos10.j2
@@ -82,9 +82,10 @@ interface {{ l.ifname }}
 ! Invalid IPv6 address {{ l.ipv6 }}
 {%   endif %}
 {%   if l.virtual_interface is not defined %}
-  ipv6 unreachables
+ ipv6 unreachables
 {%   endif %}
-{% elif 'ipv4' in l %}
+{% elif 'lag' not in l %}
+{# Don't configure IP related settings for discovery interfaces, L2-only port-channels or lag member links #}
  no ipv6 enable
 {% endif %}
 !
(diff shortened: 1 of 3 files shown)


ipspace (changes requested):
The title of the commit says "IPv6 something something on OS10", and then you modify an integration test. Is it so hard to keep the two separate?

Also, I would add the "wait for RA" step before IPv6 ping in the integration test, which, if you want to go down that path, goes into a separate commit.

jbemmel:
> The title of the commit says "IPv6 something something on OS10", and then you modify an integration test. Is it so hard to keep the two separate?
> 
> Also, I would add the "wait for RA" step before IPv6 ping in the integration test, which, if you want to go down that path, goes into a separate commit.

image

The modification to the integration test is a separate commit, do you mean you'd like a separate PR?

I put them together because the modification of the test serves to test the changes included in the PR

ipspace:
> The modification to the integration test is a separate commit, do you mean you'd like a separate PR?

Yes. When we merge the PR, the commits get squashed. Nobody wants to see the "I fixed a typo" commit messages.

> I put them together because the modification of the test serves to test the changes included in the PR

While I understand why you feel that way today, somebody trying to figure out why the integration test changed in a OS10 config template fix a year from now might disagree.

ipspace (approved):
I hope you got the indentation right ;)

PR #1873, opened 2025-01-31
EOS/IOS: Use 'shutdown' instead of 'no switchport' to avoid allocating internal VLAN resources

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1871

Summary: A user reported an issue with the 'initial' configuration step failing. The Ansible module reporting made it look like an issue with description length or invalid characters, but in reality the topology was using VLANs in the range 1006+<number of interfaces> (starting at 1011, causing issues on devices with 5 interfaces or more)

The ```normalize``` step introduced in ```1.9.4``` configures all ports as ```no switchport```, causing an internal VLAN to be allocated. This PR shuts down the port instead, avoiding the (useless) allocation of VLAN resources

image

### DIFF
--- a/netsim/ansible/templates/normalize/eos.j2
+++ b/netsim/ansible/templates/normalize/eos.j2
@@ -1,6 +1,7 @@
 {% for intf in interfaces if intf.virtual_interface is not defined %}
 !
 interface {{ intf.ifname }}
- no switchport
+{# 'no switchport' allocates an internal VLAN in range 1006-, causing issues when overlapping with topology vlans #}
+ shutdown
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Thank you!

FWIW, we wouldn't encounter this challenge on IOS platforms because the initial device configuration provisions VLANs before configuring interfaces.

jbemmel:
> Thank you!
> 
> FWIW, we wouldn't encounter this challenge on IOS platforms because the initial device configuration provisions VLANs before configuring interfaces.

Maybe we should do it that way on EOS too then?

Issue #1874, opened 2025-02-02
[BUG] Cumulus does not support passive OSPF interfaces inside VRFs


jbemmel:
Test: ```netlab up integration/vxlan/04-vxlan-irb-ospf.yml -p libvirt -d cumulus -v```

```vlan1001``` should be passive and the FRR template attempts to configure it, but ```ip ospf passive``` is silently ignored by the FRR version in Cumulus 4.4 (```FRRouting 7.5+cl4.4.0u4```)

ipspace:
Thank you. Just add it to Cumulus caveats.

PR #1881, opened 2025-02-04
Cumulus NVUE: Support remaining vlan use cases (as router)

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/1878

Add support for remaining test cases (https://tests.netlab.tools/_html/cumulus_nvue-libvirt-vlan):
* 51-vlan-routed-trunk
* 52-vlan-vrf-lite  ( uses ospf.passive which requires a quirk )
* 61-vlan-routed-native
* 62-vlan-mixed-trunk
* 63-vlan-mixed-native ( requires the creation of a sub interface for the native vlan )

I also tested STP 11 and 12, VLAN 21, 31, 33 at various points, though it is possible that adjustments during testing invalidated previous runs

Note that corner cases like a native VLAN interface with OSPF or BGP configuration "should" work, but the test cases don't cover those exceptions

Closes https://github.com/ipspace/netlab/issues/1874

### DIFF
--- a/docs/module/vlan.md
+++ b/docs/module/vlan.md
@@ -36,7 +36,7 @@ VLANs are supported on these platforms:
 | Cisco CSR 1000v       | ✅  | ✅  | ✅  | ✅ | ✅ |
 | Cisco Nexus OS        | ✅  | ✅  | ✅  | ✅  | ✅  |
 | Cumulus Linux 4.x     | ✅  | ✅  | ❌[❗](caveats-cumulus)  | ✅ | ✅ |
-| Cumulus 5.x (NVUE)    | ✅  | ✅  | ❌[❗](caveats-cumulus)  | ✅ | ✅ |
+| Cumulus 5.x (NVUE)    | ✅  | ✅  | ✅  | ✅ | ✅ |
 | Dell OS10             | ✅  | ✅  | ❌   | ✅ | ✅ |
 | FRR [❗](caveats-frr) | ✅  | ✅  | ✅  | ✅ | ✅ |
 | Juniper vMX           | ✅  | ✅  | ✅  | ✅ | ✅ |
--- a/docs/module/vxlan.md
+++ b/docs/module/vxlan.md
@@ -29,8 +29,7 @@ The following table describes per-platform support of individual VXLAN features:
 | Aruba AOS-CX       | ✅  | ✅  |  ❌  |
 | Cisco CSR 1000v    | ✅  | ✅  |  ❌  |
 | Cisco Nexus OS     | ✅  | ✅  |  ❌  |
-| Cumulus Linux 4.x  | ✅  | ✅  |  ❌  |
-| Cumulus 5.x (NVUE) | ✅  | ✅  |  ❌  |
+| Cumulus Linux      | ✅  | ✅  |  ❌  |
 | Dell OS10          | ✅  | ✅  |  ❌  |
 | FRR                | ✅  | ✅  | ✅  |
 | Nokia SR Linux     | ✅ [❗](caveats-srlinux)  |  ❌  |  ❌  |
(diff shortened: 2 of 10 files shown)


ipspace (changes requested):
* It would be simpler to just set `_cumulus_ifname` or `_nvue_ifname` on all interfaces and use that **everywhere** instead of convoluted tests sprinkled across a half-dozen templates.
* Yet again, the PR is full of unrelated stuff like "NVUE supports VXLAN" or "This is how we deal with OSPF passive interfaces"

jbemmel:
> * It would be simpler to just set `_cumulus_ifname` or `_nvue_ifname` on all interfaces and use that **everywhere** instead of convoluted tests sprinkled across a half-dozen templates.

ok, we can do it like that

> * Yet again, the PR is full of unrelated stuff like "NVUE supports VXLAN" or "This is how we deal with OSPF passive interfaces"

```
user@vm:~/netlab/tests$ netlab create integration/vlan/52-vlan-vrf-lite.yml -p libvirt -d cumulus_nvue
[WARNING] cumulus_nvue: Node r1 uses passive OSPF interfaces inside a VRF, which NVUE does not support; auto-converted to False
[DATA]    Interface swp2 inside VRF red using ospf.passive=True
          Interface swp3 inside VRF red using ospf.passive=True
          Interface swp4 inside VRF blue using ospf.passive=True
[HINT]    Set defaults.devices.cumulus_nvue.warnings.vrf_ospf_passive to False to disable this check
[WARNING] cumulus_nvue: Node r3 uses passive OSPF interfaces inside a VRF, which NVUE does not support; auto-converted to False
[DATA]    Interface swp2 inside VRF red using ospf.passive=True
          Interface swp3 inside VRF blue using ospf.passive=True
```

It's not unrelated. I tried to keep the 2 PRs separated, but test case 52 uses ospf.passive and needed the same quirk. I finally got tired of the long merge-test-checkout-change-merge-test cycle and decided to merge the two, cleaning up at the end - but I see I forgot 1 file.

jbemmel:
Refactored down to a minimum number of changes, replacing the "convoluted tests sprinkled across a half-dozen templates" with 30 lines of code to create sub-interfaces for native VLANs in case of mixed trunks.

Re-tested the 5 cases

ipspace (changes requested):
Here's a weird idea: if you change the VLAN configuration template, run all VLAN test cases, not just the ones you want to pass:

image

The current (dev branch) implementation of VLANs on CL NVUE passes all tests it supports.

The tests results from your changes are at https://tests.netlab.tools/_html/cumulus_nvue-libvirt-vlan

jbemmel:
> Here's a weird idea: if you change the VLAN configuration template, run all VLAN test cases, not just the ones you want to pass:
> 

Thanks for taking a look. In my defense, some of those tests were changed after I submitted this PR:
image

but I'll admit that I did not run all tests comprehensively like you did, and I should have. I'll take a look

ipspace:
> Thanks for taking a look. In my defense, some of those tests were changed after I submitted this PR: image

And who insisted we test that scenario? 😂

jbemmel:
With https://github.com/ipspace/netlab/pull/1955 applied, all VLAN and LAG tests now pass

ipspace:
> def nvue_rewrite_mgmt_mac(node: Box) -> None:
>   mac = netaddr.EUI(node.mgmt.mac,dialect=netaddr.mac_unix_expanded)
>   mac[3] = node.id                 # Make sure it differs in the 4th octet
>   node.mgmt.mac = str(mac)

After #1955, I have no problem having this as the default. We're going for release 2.0 anyway, so a minor breaking change like this one shouldn't be a big deal.

Also, making the change so late in the process breaks the duplicate MAC detection introduced by #1955. I will merge this as-is, but you might want to add a global change with another PR.

Issue #1877, opened 2025-02-02
[BUG] Combination of routed VLAN with lag results in index error


jbemmel:
Minimum topology:
```
---
message: |
  VLAN module bug when combined with lag

groups:
  _auto_create: True
  switches:
    members: [ s1, s2 ]
    module: [ vlan, lag ]

vlans:
  v1:
    mode: route

links:
- lag:
    members:
    - s1-s2
  vlan.trunk: [ v1 ]
```

Result:
```
File "/home/jeroen/Projects/netlab/netsim/modules/__init__.py", line 473, in module_transform
    mod_load[m].call("module_"+method,topology)
  File "/home/jeroen/Projects/netlab/netsim/utils/callback.py", line 42, in call
    return method(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jeroen/Projects/netlab/netsim/modules/vlan.py", line 1428, in module_post_link_transform
    rename_vlan_subinterfaces(n,topology)
  File "/home/jeroen/Projects/netlab/netsim/modules/vlan.py", line 1030, in rename_vlan_subinterfaces
    link_data = topology.links[intf.linkindex - 1]
                ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "box/box_list.py", line 77, in box.box_list.BoxList.__getitem__
IndexError: list index out of range
```

The ```intf```:
```
{'_global_merge': True, 'bridge_group': 1, '_selfloop_ifindex': 0, 'vlan': {'mode': 'route', 'access_id': 1000, 'name': 'v1'}, '_vlan_mode': 'route', 'ipv4': '10.1.0.1/30', 'parentindex': 1, 'type': 'vlan_member', 'linkindex': 3, 'vlan_name': 'v1', 'name': 's1 -> s2', 'ifindex': 2, 'ifname': 'port-channel1.1', 'clab': {'name': 'et2'}, 'neighbors': [{'ifname': 'Ethernet2', 'node': 's2', 'ipv4': '10.1.0.2/30'}], 'parent_ifindex': 30000, 'parent_ifname': 'port-channel1', 'virtual_interface': True}
```

ipspace:
Looks like someone changed the list of links, so the link index no longer works?

Also, I wonder how many platforms (apart from Linux where everything goes) allow you to create subinterfaces of LAG interfaces.

jbemmel:
[...]

PR #1882, opened 2025-02-04
lag: Postpone link cleanup until after vlan module_post_link_transform

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1877

There can still be vlan_members that reference the linkindex of the virtual lag links


ipspace (approved):
I hope this is good enough (i.e. there are no other modules using linkindex). Worst case, we can move it to "cleanup" action in the future.

Issue #1884, opened 2025-02-04
[BUG] Arista EOS switchport settings on port channels


ipspace:
For whatever reason we decided that Arista EOS cannot have routed LAG interfaces (lag/01-l3-lag.yml fails with a device quirk). Maybe it's just missing **no switchport**?

jbemmel:
We can do it like Dell OS10: https://github.com/ipspace/netlab/blob/d652860aa62baa828b5c78877a0bde5c17efa96f/netsim/ansible/templates/initial/dellos10.j2#L36

or modify lag/eos.j2 at https://github.com/ipspace/netlab/blob/d652860aa62baa828b5c78877a0bde5c17efa96f/netsim/ansible/templates/lag/eos.j2#L47 and add ```no switchport``` there - any preference?

ipspace:
> We can do it like Dell OS10:

That is totally overcomplicated. If the interface uses VLANs, then "no switchport" is configured in VLAN configuration template. There is no need to check whether IPv4 or IPv6 is defined on the interface.

jbemmel:
> > We can do it like Dell OS10:
> 
> That is totally overcomplicated. If the interface uses VLANs, then "no switchport" is configured in VLAN configuration template. There is no need to check whether IPv4 or IPv6 is defined on the interface.

lag/01-l3-lag.yml doesn’t use vlans

ipspace:
My bad :( This is what I had in mind:

- The **initial** template should configure everything that it can as L3 interfaces. Loopbacks and VLAN interfaces are already L3 interfaces, so "no switchport" fails there, but it should be applied to Ethernet and PortChannel interfaces.
- Whichever template needs VLANs on an interface should configure "switchport whatever" on the interface. That is usually done in the **vlan** template.

jbemmel:
> My bad :( This is what I had in mind:
> 
> * The **initial** template should configure everything that it can as L3 interfaces. Loopbacks and VLAN interfaces are already L3 interfaces, so "no switchport" fails there, but it should be applied to Ethernet and PortChannel interfaces.
> * Whichever template needs VLANs on an interface should configure "switchport whatever" on the interface. That is usually done in the **vlan** template.

ok, so that's the "Dell OS10 approach" ;)

PR #1886, opened 2025-02-04
EOS: Remove lag L3 quirk and configure 'no switchport' on l3 port-channels

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1884

Tested:
* lag/01-l3-lag
* lag/02-lag-vlan-trunk
* custom test with l2 over mlag

### DIFF
--- a/netsim/ansible/templates/initial/eos.j2
+++ b/netsim/ansible/templates/initial/eos.j2
@@ -33,7 +33,7 @@ interface {{ mgmt.ifname|default('Management1') }}
 !
 {% for l in netlab_interfaces %}
 interface {{ l.ifname }}
-{% if l.virtual_interface is not defined %}
+{% if l.virtual_interface is not defined or (l.type=='lag' and ('ipv4' in l or 'ipv6' in l)) %}
  no switchport
 {% endif %}
 {% if l.vrf is defined %}
--- a/netsim/devices/eos.py
+++ b/netsim/devices/eos.py
@@ -67,16 +67,6 @@ def check_dhcp_clients(node: Box, topology: Box) -> None:
       category=log.IncorrectType,
       module='quirks')
 
-def check_l3_lag(node: Box, topology: Box) -> None:
-  for intf in node.interfaces:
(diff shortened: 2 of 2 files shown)


netsim/ansible/templates/initial/eos.j2
```diff
@@ -33,7 +33,7 @@ interface {{ mgmt.ifname|default('Management1') }}
 !
 {% for l in netlab_interfaces %}
 interface {{ l.ifname }}
-{% if l.virtual_interface is not defined %}
+{% if l.virtual_interface is not defined or (l.type=='lag' and ('ipv4' in l or 'ipv6' in l)) %}
```
ipspace, line 36: I might be wrong, but I think you don't get it. Will make the change and run integration tests in a few days.

jbemmel: Sample topology:
```
---
message: |
 Tests a basic L2 link between 2 hosts

groups:
 _auto_create: True
 switches:
  members: [ dut ]
 hosts:
  members: [ h1, h2 ]
  device: linux
  provider: clab

links:
- dut:
    ipv4: False
  h1:
    ipv4: 10.0.0.1/24
- dut:
    ipv4: False
  h2:
    ipv4: 10.0.0.2/24

validate:
  ping:
    description: IPv4 ping H1 => H2
    wait_msg: Waiting for STP to enable the ports
    wait: 45
    nodes: [ h1 ]
    plugin: ping('h2')
```

This currently fails for both EOS and FRR, but works for EOS when I change the test into:
```
{% if (l.virtual_interface is not defined or l.type=='lag') and ('ipv4' in l or 'ipv6' in l) %}
```
(i.e. even more strict than this PR)

We can argue about whether this test should pass - in my view, this topology calls for L2 ports, hence 'no switchport' is incorrect. 

We could leave it as:
```
{% if (l.virtual_interface is not defined or l.type=='lag') %}
```
and then routed subinterfaces would "magically" start working - but the above l2 use case would break

Issue #1885, opened 2025-02-04
[BUG] Cumulus NVUE is failing L3 LAG test


ipspace:
01-l3-lag:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/lag/01-l3-lag.yml)
* [Test results](https://tests.netlab.tools/cumulus_nvue/libvirt/lag/01-l3-lag.yml-validate.log)

PR #1887, opened 2025-02-04
Cumulus NVUE: Fix l3 lag test case

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1885

* Don't attach l3 interfaces to the bridge
* Move lag bridging config to lag module script, and only connect to bridge for MLAG cases

### DIFF
--- a/netsim/ansible/templates/lag/cumulus_nvue.j2
+++ b/netsim/ansible/templates/lag/cumulus_nvue.j2
@@ -40,7 +40,8 @@
         link:
           mtu: {{ i.mtu }}
 {%  endif %}
-{%  if not (i.ipv4 is defined or i.ipv6 is defined) %}
+{# In case of MLAG, connect the bond interface to the bridge #}
+{%  if '_mlag' in i.lag %}
         bridge:
           domain:
             br_default: {}
--- a/netsim/ansible/templates/lag/cumulus_nvue.j2
+++ b/netsim/ansible/templates/lag/cumulus_nvue.j2
@@ -40,6 +40,11 @@
         link:
           mtu: {{ i.mtu }}
 {%  endif %}
+{%  if not (i.ipv4 is defined or i.ipv6 is defined) %}
+        bridge:
+          domain:
+            br_default: {}
+{%  endif %}
         bond:
 {%  if '_mlag' in i.lag %}
           mlag:
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -35,7 +35,7 @@
 {%  else %}
             enable: off
 {%  endif %}
-{%  if i.type=='lag' %}
+{%  if i.type=='lag' and not (i.ipv4 is defined or i.ipv6 is defined) %}
         bridge:
           domain:
             br_default: {}
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -35,11 +35,6 @@
 {%  else %}
             enable: off
 {%  endif %}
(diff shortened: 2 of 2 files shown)


netsim/ansible/templates/lag/cumulus_nvue.j2
```diff
@@ -39,6 +39,11 @@
 {%  if i.mtu is defined %}
         link:
           mtu: {{ i.mtu }}
+{%  endif %}
+{%  if not (i.ipv4 is defined or i.ipv6 is defined) %}
```
ipspace, line 43: This stinks, you're connecting LAG interface to a bridge for no good reason.

There should be no need to do anything here. If the LAG interface uses VLANs, then the VLAN module will do its job.

ipspace (changes requested):

netsim/ansible/templates/lag/cumulus_nvue.j2
jbemmel: I moved the attachment to the bridge to a different file and added a condition that there shouldn't be an IP address - but I did not introduce the connection

The connection is necessary for a L2 use case without VLANs - none of the integration tests covers that, but I have private tests that do. I'll concede that it is probably uncommon

```
message: |
  The device under is a pair of switches with a pair of L2 MLAG links connected
  to 2 Linux hosts, each connected through a LAG with L3 IP and no VLAN, with the switches in L2 mode.
  The hosts should be able to ping each other

groups:
  _auto_create: true
  switches:
    members: [ s1, s2 ]
    module: [ lag ]
  hosts:
    members: [ h1, h2 ]
    module: [ lag ]       # Host side must support lag to present single MAC on both interfaces
    device: linux

links:
- lag:
    members: [s1-s2, s1-s2]
    mlag.peergroup: 1
  # On OS10 in case of mlag, vlan.trunk is implied for all vlans - other platforms?
- lag.members:
  - h1:
      ipv4: 192.168.0.3
    s1:
      ipv4: False
  - h1:
    s2:
      ipv4: False
- lag.members:
  - h2:
      ipv4: 192.168.0.4
    s1:
      ipv4: False
  - h2:
    s2:
      ipv4: False

validate:
  ping:
    description: Pinging H2 from H1
    nodes: [h1]
    wait_msg: Waiting for STP to enable the ports
    wait: 45
    plugin: ping('h2',af='ipv4')
```

ipspace: > Moreover, I followed vendor instructions https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-55/Layer-2/Multi-Chassis-Link-Aggregation-MLAG/ step 4:

Written for people who do VLANs over (M)LAG.

> Granted, that only applies to MLAG. However, the connection is necessary for a L2 use case without VLANs

There is no L2 use case without VLANs or extra configs. The documentation of l2-only interfaces is too vague and I will fix that.

> none of the integration tests covers that

Because that's not a valid use case ;)

> but I have private tests that do. I'll concede that it is probably uncommon

I wonder how many platforms would pass those tests on regular interfaces (LAG interfaces don't count as you wrote the config templates to pass those tests).

jbemmel: > 
> I wonder how many platforms would pass those tests on regular interfaces (LAG interfaces don't count as you wrote the config templates to pass those tests).

EOS does (as we saw when we introduced the ```normalize``` scripts to avoid bridging ipv6 RAs)

Under the hood it associates the ports - without ```no switchport``` - with VLAN 1. I suppose that's an exception, and the way to model that properly in Netlab would be to use the vlan module

If it's not a valid use case, should Netlab issue a warning/error in this case?

jbemmel: Updated, passes tests 01-l3-lag and 10-mlag

PR #1888, opened 2025-02-05
lag: Add integration test for routed trunk over lag

jbemmel:
EOS currently fails this test (separate PR):

```
dut#show run int port-channel 1
interface Port-Channel1
   description dut -> xr
dut#show run int port-channel 1.1
interface Port-Channel1.1
   description dut -> xr
   encapsulation dot1q vlan 1000
   ip address 10.1.0.1/30
   ip ospf network point-to-point
   ip ospf area 0.0.0.0
dut#show int Port-Channel 1.1
Port-Channel1.1 is down, line protocol is dormant (notconnect)
  Hardware is Subinterface, address is 001c.73f0.28da
  Description: dut -> xr
  Internet address is 10.1.0.1/30
  Broadcast address is 255.255.255.255
  IP MTU 9194 bytes (default), BW 2000000 kbit
  Down 1 minute, 4 seconds
```

### DIFF
--- a/tests/integration/lag/04-lag-vlan-routed-trunk.yml
+++ b/tests/integration/lag/04-lag-vlan-routed-trunk.yml
@@ -1,0 +1,35 @@
+message: |
+  The device under test is a router connected with a routed VLAN trunk on a LAG to
+  another router. Hosts should be able to ping each other.
+
+  On routers, this case tests routed subinterfaces on a port-channel interface
+groups:
+  _auto_create: true
+  hosts:
+    members: [h1, h2]
+    device: linux
+    provider: clab
+  routers:
+    members: [dut, xr]
+    module: [lag, vlan, ospf]
+    vlan.mode: route
+  probe:
+    members: [xr]
+    device: frr
+
+vlans:
+  v1:
+    links: [dut-h1, xr-h2]
+
+links:
+- vlan.trunk: [v1]
+  lag:
+    members: [dut-xr, dut-xr]
+
+validate:
+  ping_v1:
+    description: Pinging H2 from H1 on VLAN v1
+    nodes: [h1]
+    wait_msg: Waiting for STP to enable the ports
+    wait: 45
+    plugin: ping('h2')
--- a/tests/integration/lag/04-lag-vlan-routed-trunk.yml
+++ b/tests/integration/lag/04-lag-vlan-routed-trunk.yml
@@ -6,30 +6,37 @@ message: |
 groups:
   _auto_create: true
   hosts:
-    members: [h1, h2]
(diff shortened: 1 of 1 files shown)


ipspace:
It would be much better to test OSPF adjacency. Never mind, I'll fix it in a few days.

jbemmel:
> It would be much better to test OSPF adjacency. Never mind, I'll fix it in a few days.

I agree, and I don’t mean to annoy anyone but the reason I shy away from OSPF in tests is that they don’t work on OS10 virtual networks…

ssasso:
> I agree, and I don’t mean to annoy anyone but the reason I shy away from OSPF in tests is that they don’t work on OS10 virtual networks…

Guys, seriously, if you find this is a huge limitation of OS10 template, maybe we should think about either:

* Have a new os10 "macro platform" with plain clans (like we have for cumulus)
* Completely change os10 vlan implementation - but then please don't complain about the fact that evpn/vxlan/mclag stopped working
* Have some quirk to change vlan & co behavior based on some parameter.

To be honest I don't think we have so much users using OS10, so - personal opinion - I'm ok if we try to solve real world problems.

jbemmel:
> 
> Guys, seriously, if you find this is a huge limitation of OS10 template, maybe we should think about either:
> 
> * Have a new os10 "macro platform" with plain vlans (like we have for cumulus nvue)
> * Completely change os10 vlan implementation - but then please don't complain about the fact that evpn/vxlan/mclag stopped working
> * Have some quirk to change vlan & co behavior based on some parameter.
> 
> To be honest I don't think we have so much users using OS10, so - personal opinion - I'm ok if we try to solve real world problems.

I'd be happy to sign up for revising the OS10 templates, and I believe I could do it in such a way that EVPN and VXLAN keep working. Our enterprise network is built on OS10, not my choice or preference but it is the real world for me

ssasso:
> I could do it in such a way that EVPN and VXLAN

I explicitly said even/vxlan/mlag for a reason ;)

/ == +

ipspace:
> I agree, and I don’t mean to annoy anyone but the reason I shy away from OSPF in tests is that they don’t work on OS10 virtual networks…

Says the guy who used OSPF to propagate routes between the two routers in the test topology 😂

ipspace:
> To be honest I don't think we have so much users using OS10, so - personal opinion - I'm ok if we try to solve real world problems.

Maybe sometime later this year I'll stop fixing broken stuff and implement something new like tracking what people use, and we'll move beyond having opinions.

In the meantime, I share your personal opinion, so let's solve some real-world problems (and I'm sorry @jbemmel that you have to deal with OS10 as part of your job).

PR #1889, opened 2025-02-05
EOS: Set 'no switchport' on parent interface of routed lag subinterface

jbemmel:
It is possible to have an L2 port-channel or an L3 parent with routed subinterfaces; in both cases, there is no IP address on the parent

Fixes test case https://github.com/ipspace/netlab/pull/1888


ipspace:
This is probably totally unnecessary and triggered by the overly restrictive test in #1886.

PR #1894, opened 2025-02-06
Cumulus NVUE: Move initialization of bridge MAC to 'initial' step

jbemmel:
…such that it also works in (mlag) topologies without VLANs

The MAC is read from the EEPROM, and if it's not set to a unique value each instance gets the same (duplicate) value which breaks mlag and STP to name a few

### DIFF
--- a/netsim/ansible/templates/initial/cumulus_nvue.j2
+++ b/netsim/ansible/templates/initial/cumulus_nvue.j2
@@ -54,6 +54,14 @@
             content: |
               { "address": { "defaults": { "mtu": "{{ mtu }}" } } }
 {% endif %}
+    bridge:
+      domain:
+        br_default:
+          type: vlan-aware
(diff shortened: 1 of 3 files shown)


jbemmel:
> Is there a reason you want to do this in the 'normalize' step and not as part of initial configuration? The 'normalize' step is used for things that need to be done before anything else (for example, shutting down interfaces to prevent leaked packets between IPv6 subnets).

It seemed consistent with the other 'normalize' scripts that configure MACs on interfaces. But it could be at the start of the 'initial' script as well if you prefer

ipspace:
> It seemed consistent with the other 'normalize' scripts that configure MACs on interfaces. But it could be at the start of the 'initial' script as well if you prefer

I put the "change MAC address" in the EOS normalization script to ensure nothing is ever sent from a different MAC address. Now that we do "shutdown" instead of "no switchport", that no longer matters, but there's no reason to move that back to initial device configuration solely for perceived purity.

PR #1897, opened 2025-02-06
Cumulus NVUE lag: Allow bond0, False == 0 but None != 0

jbemmel:


jbemmel:
Found more places where this needs to be fixed, will submit bulk PR

PR #1898, opened 2025-02-06
lag: Correct comparison with missing _parentindex using 'None', not 'False' (==0)

jbemmel:
Allows the use of lag.ifindex==0


ipspace (approved):
Do you remember that I told you using `|default(0)` is a bad idea? 🤷‍♂️

jbemmel:
> Do you remember that I told you using `|default(0)` is a bad idea? 🤷‍♂️

I do, but you did not mention ```False```

ipspace:
> I do, but you did not mention `False`

🥺🥺🥺 OMG, what a horrible omission 🤦‍♂️

But don't worry too much about being in the "confused by bool being int" club. We had to fix similar stuff in the past: https://github.com/ipspace/netlab/pull/414

Issue #1896, opened 2025-02-06
VLAN integration tests don't cover VLAN specific MTU settings


jbemmel:
```
jeroen@j:~/Projects/netlab/tests$ grep mtu integration/vlan/* -r
integration/vlan/configs/h1.initial.cfg:ip link set eth1 mtu 1500
integration/vlan/configs/h2.initial.cfg:ip link set eth1 mtu 1500
```

ipspace:
* https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv2/06-vlan-mtu.yml
* https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv3/07-vlan-mtu.yml

PR #1900, opened 2025-02-07
OSPFv2: Add VLAN trunk MTU test, highlights MTU issue in FRR

jbemmel:
Fails for Cumulus_NVUE with FRR:
https://github.com/ipspace/netlab/issues/1896 tests system MTU and access VLAN MTU, but not MTU on VLAN trunks

```netlab up integration/ospf/ospfv2/07-vlan-trunk-mtu.yml -p libvirt -d cumulus_nvue -v``` fails:

* MTU on FRR X2 vlan1001 interface is default 1500 instead of 1280; ```ip link set dev vlan1001 mtu 1280``` fixes it

### DIFF
--- a/tests/integration/ospf/ospfv2/07-vlan-trunk-mtu.yml
+++ b/tests/integration/ospf/ospfv2/07-vlan-trunk-mtu.yml
@@ -1,0 +1,51 @@
+---
+message: |
+  This lab tests OSPF adjacencies on VLAN interfaces with non-default MTU
+  settings. The adjacency establishment could fail due to MTU mismatch.
+
+  Failure to establish X2-DUT adjacency means that the VLAN interface does not
+  get the correct MTU. Failure to establish X1-DUT adjacency indicates a
+  potential problem with system MTU.
+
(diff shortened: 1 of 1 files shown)


jbemmel:
Issue is here: https://github.com/ipspace/netlab/blame/ed618f446ff80ab7795b436394936134122500b3/netsim/ansible/templates/vlan/frr.j2#L21

Introduced by https://github.com/ipspace/netlab/commit/41b09a499bf5ae7f2580756eaba72d25f06585bc

Perhaps there is something to say for standardizing the variable names used in templates, at least for a given platform?

ipspace:
Stop adding random integration tests for what was effectively a regression test for a single platform. A single integration test like the OSPFv2 ones takes over 30 minutes to run for all platforms we're testing, and we have 200+ tests.

I added the main idea from this test to the existing VLAN/MTU test, and I really can't see why you had to create a new one.

ipspace:
... oh, and if you write "fixes #PR" in another PR, merging that one closes the referenced PR 🤷‍♂️

PR #1901, opened 2025-02-07
FRR: Fix SVI MTU

jbemmel:
* Use 'i' for interfaces consistently

Fixes https://github.com/ipspace/netlab/pull/1900

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -92,17 +92,17 @@ fi
 
 # Disable IPv6 (for IPv4-only interfaces) or SLAAC (if the device is a router)
 #
-{% for l in interfaces if l.type in ['lan','p2p','stub','lag','bond'] %}
-{%   if l.ipv6 is not defined %}
-sysctl -qw net.ipv6.conf.{{ l.ifname }}.disable_ipv6=1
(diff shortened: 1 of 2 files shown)


ipspace:
Stop changing code for no good reason apart from eternal beauty and poetic consistency. As I've learned the very hard way, you're not always consistent and you might not thoroughly test all the changes you're making. I'm thus forced to eyeball every single change hoping to spot a potential unusual difference.

jbemmel:
> Stop changing code for no good reason apart from eternal beauty and poetic consistency. As I've learned the very hard way, you're not always consistent and you might not thoroughly test all the changes you're making. I'm thus forced to eyeball every single change hoping to spot a potential unusual difference.

I'm curious why the ```l``` did not result in an undefined variable error?

ipspace:
> I'm curious why the `l` did not result in an undefined variable error?

We did as if `l.mtu` was defined, right? Well, Ansible thinks it was not 🤷‍♂️

jbemmel:
> > I'm curious why the `l` did not result in an undefined variable error?
> 
> We did as if `l.mtu` was defined, right? Well, Ansible thinks it was not 🤷‍♂️

There is a reason to use ‘’’if ‘mtu’ in l’’’ rather than ‘’’if l.mtu is defined’’’

Issue #1853, opened 2025-01-27
[BUG] Gateway integration test might give false positives over IPv6


ipspace:
We don't configure IPv6 anycast gateway on Cumulus Linux, and yet the IPv6 ping works in the integration test. It looks like the adjacent FRR replies to the ND request, and then Linux kernel on CL takes whatever is sent to its MAC address and forwards it (yuck).

Somehow we have to modify the integration test to ensure this doesn't happen, for example by shutting down the VLAN interface in FRR during the initial tests?

PR #1905, opened 2025-02-08
Gateway test: Disable the FRR anycast interface before testing pings

jbemmel:
Reduces False positives - fixes https://github.com/ipspace/netlab/issues/1853

Also explicitly check the IPv6 neighbor entry for "REACHABLE" (not "PROBE")

### DIFF
--- a/tests/integration/gateway/01-anycast.yml
+++ b/tests/integration/gateway/01-anycast.yml
@@ -53,6 +53,11 @@ links:
   prefix: target
 
 validate:
+  x1_gw_down:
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
Nice try, but how are we going to check for duplicate packets if the anycast GW on X1 is down all the time? Also, we have to make sure the GW on X1 is operational before checking for duplicate packets or doing traceroutes over DUT.

Don't worry, will fix and merge tomorrow.

PR #1911, opened 2025-02-09
Dell OS10: Enable tagged VLAN 1 on trunks

jbemmel:
This PR modifies the VLAN templates to configure a new internal untagged VLAN different than 1, in case 1 is used as a tagged VLAN on any trunk

With this, Dell OS10 passes the new ```70-vlan-1-trunk``` test

### DIFF
--- a/netsim/ansible/templates/vlan/dellos10.j2
+++ b/netsim/ansible/templates/vlan/dellos10.j2
@@ -12,17 +12,31 @@ interface vlan {{ vlan.id }}
 {%   endfor %}
 {% endif %}
 !
+
+{# Check if the platform default vlan 1 is used as tagged on a trunk - if so, change the default #}
+{% for ifdata in interfaces if 1 in ifdata.vlan.trunk_id|default([]) %}
+{%   if loop.first %}
+{%     set used_vlans = vlans.values()|map(attribute='id')|list|sort %}
+{%     set new_default = 4093 if 4093 not in used_vlans else used_vlans|last + 1 %}
+interface vlan {{ new_default }}
+ description "New default internal untagged vlan, instead of platform default 1"
+!
+default vlan-id {{ new_default }}
+{%   endif %}
+{% endfor %}
+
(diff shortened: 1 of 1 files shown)


ipspace:
This seems to be a bit too much just to implement a weird scenario that should not be used anyway. I would much prefer a quirk saying "do not use tagged VLAN 1 on Dell OS10". What do you think @ssasso

ssasso:
> This seems to be a bit too much just to implement a weird scenario that should not be used anyway. I would much prefer a quirk saying "do not use tagged VLAN 1 on Dell OS10". What do you think @ssasso

Still on the "focus on the use case" - @jbemmel are you using VLAN 1 as trunk vlan?

otherwise I agree to have a quirk.

ipspace:
> Still on the "focus on the use case" - @jbemmel are you using VLAN 1 as trunk vlan?

I can't see a use case beyond "I have to recreate all the bad choices they made in my network as a digital twin" -- VLAN 1 should be remapped into something else before it becomes tagged (blog post coming tomorrow ;).

> otherwise I agree to have a quirk.

👍

jbemmel:
> > Still on the "focus on the use case" - @jbemmel are you using VLAN 1 as trunk vlan?
> 
> I can't see a use case beyond "I have to recreate all the bad choices they made in my network as a digital twin" -- VLAN 1 should be remapped into something else before it becomes tagged (blog post coming tomorrow ;).
> 

It's not a high priority or common use case, on Dell OS10 (as on many platforms) VLAN 1 is configured as the native untagged VLAN for trunk ports by default. Afaik it needs to be untagged for STP interop scenarios, I just wanted to see if it could be done if needed

> > otherwise I agree to have a quirk.
> 
> 👍

I'm ok with a quirk, I can put a reference to changing the default VLAN using custom config if the user is really desperate

ipspace:
> I'm ok with a quirk, I can put a reference to changing the default VLAN using custom config if the user is really desperate

👍

jbemmel:
To be replaced by quirk PR

PR #1912, opened 2025-02-10
Mlag VTEP plugin

jbemmel:
Context: https://github.com/ipspace/netlab/discussions/1835

This PR adds support for VTEP redundancy, by provisioning a shared IP across both MLAG peers.

Includes - tested - support for:
* Dell OS10
* Arista EOS
* Cumulus NVUE (depends on PR #1832)

### DIFF
--- a/netsim/extra/mlag.vtep/plugin.py
+++ b/netsim/extra/mlag.vtep/plugin.py
@@ -44,7 +44,7 @@ def post_transform(topology: Box) -> None:
           node = topology.nodes[i.node]
           features = devices.get_device_features(node,topology.defaults)
           if 'lag.mlag.vtep' not in features:
-            log.error(f'Node {node.name} does not support the mlag.vtep plugin',log.IncorrectValue,_config_name) 
+            log.error(f'Node {node.name} does not support the mlag.vtep plugin',log.IncorrectValue,_config_name)
           if 'vtep' in node.vxlan and node.get('lag.mlag.vtep',None) is not False:
             node.lag.mlag.vtep = change_ip[ node.vxlan.vtep ] = str(vtep_a.network)
 
@@ -71,7 +71,7 @@ def post_transform(topology: Box) -> None:
     for n, ndata in topology.nodes.items():
       # Update remote vtep list in case of static flooding
       if ndata.vxlan.get('flooding', '') == 'static':
-          
+
         def replace(ip: str) -> str:
            return change_ip[ip] if ip in change_ip else ip
         mlag_vtep = ndata.get('lag.mlag.vtep',None)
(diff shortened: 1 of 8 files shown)


ipspace:
Are you sure there's need for such a convoluted solution? Wouldn't it be better to create an extra loopback link per node, set static IP on it, and set **vxlan.vtep** to true. Also, use the existing address pool functionality instead of building your own.

Don't reinvent the wheel where it's not needed.

Closing this PR as I assume it will be cleaner to start from scratch.

netsim/extra/mlag.vtep/plugin.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+              # Add an extra loopback interface with the allocated VTEP IP
+              vtep_loopback = data.get_empty_box()
+              vtep_loopback.type = 'loopback'
+              vtep_loopback.name = f"MLAG VTEP VXLAN interface shared between {' - '.join([i.node for i in peers])}"
+              vtep_loopback.ipv4 = node.lag.mlag.vtep + "/32"
+              vtep_loopback.vxlan.vtep = True
+              links.create_virtual_interface(node, vtep_loopback, topology.defaults)
+              if 'ospf' in node.get('module',[]):    # Add it to OSPF when used, TODO ISIS
+                vtep_loopback.ospf = { 'area': "0.0.0.0", 'passive': True }
```
jbemmel, line 66: Should use ```node.get('ospf.area',"0.0.0.0")```

ipspace:
Look, if other people invest time into commenting your stuff, you should at least carefully read what they wrote. For example,

> Wouldn't it be better to create an extra loopback **link** per node

Here's the exact data structure you would have to use: https://netlab.tools/module/vxlan/#changing-the-vxlan-vtep-source-interface-address

Creating a loopback link _before link transformation_ would eliminate all the hassle and the dirty hacks you had to go through to make the plugin work _for one use case you had in mind_ -- OSPF running in area 0. Instead of reinventing a badly-designed wheel, everything would work out of the box.

There would be no need to add feature flags. Either multiple loopbacks, MLAG, and VXLAN work or not.

Similarly for the address pools. Why do you have to invent a new attribute and convert it into an address pool if the address pools work just fine out of the box? Just add an address pool in plugin defaults.

Finally, you should use **vrf_loopback** pool as the fallback. Addresses are taken sequentially from the **vrf_loopback** pool while the **loopback** address allocation uses node ID.

jbemmel:
> Look, if other people invest time into commenting your stuff, you should at least carefully read what they wrote. 
> 
I wasn't following what you were saying, but now I see what you mean. Will rework things

PR #1915, opened 2025-02-11
Mlag VTEP plugin simplified

jbemmel:
Based on suggestions from https://github.com/ipspace/netlab/pull/1912

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -15,6 +15,7 @@
    plugins/bgp.originate.md
    plugins/check.config.md
    plugins/fabric.md
+   plugins/mlag.vtep.md
    plugins/multilab.md
    plugins/node.clone.md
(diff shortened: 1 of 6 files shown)


jbemmel:
I tripped over a bug in the VXLAN module: https://github.com/ipspace/netlab/blob/fc79c057455a594abf294065f64aa58d0f58f83e/tests/topology/expected/vxlan-static.yml#L623

The local VTEP IP isn't supposed to be included in the flood list.

ipspace (approved):
Isn't it amazing how simple things become when you use existing mechanisms ;)

jbemmel:
> Isn't it amazing how simple things become when you use existing mechanisms ;)

Where would we be if Einstein had limited himself to existing theories?

PR #1919, opened 2025-02-12
vJunos: VLAN 1 trunk tweak/fix

### LINKED ISSUE #1910: [BUG] vjunos-router cannot configure tagged VLAN 1 in a trunk
70-vlan-1-trunk:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vlan/70-vlan-1-trunk.yml)
* [Test results](https://tests.netlab.tools/vjunos-router/clab/vlan/70-vlan-1-trunk.yml-config.log)

The root cause is the fake "let's put VLAN 1 on unit 0" part of the `initial/junos.vlan.j2` template

https://github.com/ipspace/netlab/blob/dev/netsim/ansible/templates/initial/junos.vlan.j2#L53

I'm perfectly OK with declaring this a quirk, or using a fake VLAN ID on unit 0, similar to what I'm doing in Cisco IOSvL2 case:

https://github.com/ipspace/netlab/blob/dev/netsim/devices/iosvl2.py#L24

ssasso:
Closes #1910

Tested with vjunos-router (and compat testing with vjunos-switch)

(side note, as reported: requires adding wait timer to test `70-vlan-1-trunk//ping_irb_1`)

### DIFF
--- a/netsim/ansible/templates/initial/junos.vlan.j2
+++ b/netsim/ansible/templates/initial/junos.vlan.j2
@@ -46,12 +46,14 @@ interfaces {
 {%   elif l.junos_interface != 'irb' and l.junos_unit == '0' and l._vlan_master is defined and 'vlan' not in l %}
 {#
 # in this case unit 0 is still present, but not handling any real vlan traffic (no vlan definition apart from _vlan_master)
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Passes all VLAN integration tests

PR #1920, opened 2025-02-12
Make IPv4/IPv6 address/prefix validation more granular

### LINKED ISSUE #1831: [BUG] We need 'pure_address' use type for IP address validation
Next hops in static routes are another example of IP addresses that cannot be prefixes and cannot take int or bool values. Looks like we have to differentiate between 'subnet_address' (which can be int or bool) and 'address' (which must be a string but without the prefix length).

I will try to change all current validation definitions to use 'subnet_address', and use 'address' for real IP addresses. This seems cleaner than inventing another 'really_an_address' keyword.

ipspace:
Add more granular IPv4/IPv6 attribute uses:

* Address and prefix
* Host and subnet prefix
* Interface address and ID

Closes #1831

### DIFF
--- a/netsim/defaults/attributes.yml
+++ b/netsim/defaults/attributes.yml
@@ -53,8 +53,8 @@ link:                       # Global link attributes
   prefix:
     type: dict
     _keys:
-      ipv4: { type: ipv4, use: prefix }
-      ipv6: { type: ipv6, use: prefix }
+      ipv4: { type: ipv4, use: subnet_prefix }
+      ipv6: { type: ipv6, use: subnet_prefix }
       allocation: { type: str, valid_values: [ p2p, sequential, id_based ] }
       _name: str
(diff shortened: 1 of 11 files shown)


netsim/defaults/attributes.yml
```diff
@@ -94,23 +94,23 @@ node:
   group: list
   role: id
   mgmt:
-    ipv4: { type: ipv4, use: id }
-    ipv6: { type: ipv6, use: id }
+    ipv4: { type: ipv4, use: address }
```
jbemmel, line 97: This is 'address' because the prefix is implicitly assumed to be a /24?

ipspace: No, because specifying subnet mask on an individual management IP address makes no sense. Either it's within the management network (we know how well that works), or it's an external IP address where the mask doesn't matter.

PR #1924, opened 2025-02-14
Migrate to FRR release 10.2

ipspace:
The only major change found between FRR release 10.0 and 10.2 is the change in the **show isis route** JSON field names.

However, 10.2.1 has a showstopper: BGP daemon crashes in EVPN deployments. The **frr:master** container works, which means they already found the problem and fixed it. We just have to wait for the new image.

### DIFF
--- a/netsim/validate/isis/frr.py
+++ b/netsim/validate/isis/frr.py
@@ -64,8 +64,8 @@ def show_isis_prefix(pfx: str, level: str = '2', **kwargs: typing.Any) -> str:
   return f'isis route level-{level} json'
 
 def check_prefix_cost(pfx: str, cost: int, p_info: Box) -> None:
-  if p_info.Metric != cost:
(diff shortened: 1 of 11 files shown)


netsim/validate/isis/frr.py
```diff
@@ -64,8 +64,8 @@ def show_isis_prefix(pfx: str, level: str = '2', **kwargs: typing.Any) -> str:
   return f'isis route level-{level} json'
 
 def check_prefix_cost(pfx: str, cost: int, p_info: Box) -> None:
-  if p_info.Metric != cost:
```
jbemmel, line 67: We could make this work with both versions:
```
metric = p_info.metric if 'metric' in p_info else p_info.Metric
```

ipspace:
10.2.2 results: EVPN bridging and asymmetric IRB work. Symmetric IRB is still crashing the BGP daemon. Not migrating yet...

PR #1925, opened 2025-02-14
NXOS: Move netlab_check_* variables to group_vars (where they get read)

jbemmel:
Correct typo in name ```netlaby_check_delay```

### DIFF
--- a/netsim/devices/nxos.yml
+++ b/netsim/devices/nxos.yml
@@ -14,13 +14,13 @@ clab:
     kind: cisco_n9kv
   interface.name: eth{ifindex}
   build: https://containerlab.dev/manual/kinds/vr-n9kv/
-  netlab_check_retries: 50
-  netlaby_check_delay: 10
 group_vars:
   ansible_user: vagrant
   ansible_ssh_pass: vagrant
   ansible_network_os: nxos
   ansible_connection: network_cli
+  netlab_check_retries: 50
+  netlab_check_delay: 10
 bfd:           # NXOS requires lower default timer values
   min_rx: 500
 evpn._start_transit_vlan: 3800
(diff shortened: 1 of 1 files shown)


ipspace (approved):
I'm assuming you tested that SSH works, right?

jbemmel:
> I'm assuming you tested that SSH works, right?

I had to add that bit because SSH wasn't working

ipspace:
> I had to add that bit because SSH wasn't working

Sure. The beauty of Cisco using ancient SSH algorithms :(

jbemmel:
> > I had to add that bit because SSH wasn't working
> 
> Sure. The beauty of Cisco using ancient SSH algorithms :(

```
dut# show ip ospf nei
 OSPF Process ID 1 VRF default
 Total number of neighbors: 2
 Neighbor ID     Pri State            Up Time  Address         Interface
 10.0.0.2          1 FULL/DR          00:00:39 10.1.0.2        Eth1/1 
 10.0.0.3          1 FULL/ -          00:00:57 10.1.0.6        Eth1/2 
dut# show run int e1/2

!Command: show running-config interface Ethernet1/2
!Time: Fri Feb 14 15:54:24 2025

version 7.0(3)I5(1)

interface Ethernet1/2
  description dut -> x2
  no switchport
  mac-address 52dc.cafe.0102
  ip address 10.1.0.5/30
  ip ospf network point-to-point
  ip router ospf 1 area 0.0.0.0
  no shutdown
```

I am troubleshooting OSPFv2 between Cisco NXOS 7.0(2)I3(1) - which is so old I couldn't even get a virtual image - and Dell OS10 5.something

It works fine virtually...

PR #1926, opened 2025-02-14
Migrate from netaddr to ipaddress

### LINKED ISSUE #1921: Cleanup: replace netaddr with ipaddress
ipaddress seems to be a better fit for our needs than netaddr (at least it has a clear distinction between IPv4 and IPv6 addresses). It might make sense to get rid of netaddr within netlab code (we will still have to install it for Ansible).

ipspace:
* Most netaddr calls have been replaced with ipaddress calls, the major exception is MAC address handling which is not available in ipaddress
* The 'get ip from interface' and 'get prefix from interface' functions have been moved to 'utils.routing' and most other modules use those functions
* Direct calls to ipaddress functions are use only for more advanced use cases.
* Error messages in augment.addressing and augment.links have been simplified and include parsing error information from ipaddress
* 'get_addr_mask' and 'get_nth_ip_from_prefix' functions did almost the same thing and have been merged into a single function
* Extra properties available in ipaddress make it easier to check if an address is a network address, multicast address, or a broadcast address

Note:
* Sometimes we still cheat and merge IP address string with a prefix length string to get a full prefix. There's no good way to get that done with ipaddress functions (you cannot change the prefix length in IPv4Network/IPv6Network)
* Some error checking code has been removed from old modules; that functionality is now available as generic data validation

Closes #1921

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -641,7 +641,7 @@ def IPAM_id_based(link: Box, af: str, pfx: ipaddress._BaseNetwork, ndict: Box) -
 
 def IPAM_loopback(link: Box, af: str, pfx: ipaddress._BaseNetwork, ndict: Box) -> bool:
   for intf in link.interfaces:
-    prefixlen = 128 if ':' in str(pfx) else 32
+    prefixlen = 128 if af == 'ipv6' else 32
     intf[af] = f'{str(pfx.network_address)}/{prefixlen}'
 
   return True
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -4,7 +4,7 @@
 #
 import typing
 
-import netaddr
+import ipaddress
 from box import Box
 
 # Related modules
@@ -373,7 +373,7 @@ def set_fhrp_gateway(link: Box, pfx_list: Box, nodes: Box, link_path: str) -> No
       continue
 
     try:                                                              # Now try to get N-th IP address on that link
-      link.gateway[af] = get_nth_ip_from_prefix(netaddr.IPNetwork(link.prefix[af]),link.gateway.id)
+      link.gateway[af] = addressing.get_nth_ip_from_prefix(link.prefix[af],link.gateway.id)
       fhrp_assigned = True
     except Exception as ex:
       log.error(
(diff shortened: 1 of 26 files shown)


netsim/augment/links.py
```diff
@@ -636,25 +624,25 @@ def IPAM_sequential(link: Box, af: str, pfx: netaddr.IPNetwork, ndict: Box) -> b
...
     OK = set_interface_address(intf,af,pfx,ndict[intf.node].id) and OK
 
   return OK
 
-def IPAM_loopback(link: Box, af: str, pfx: netaddr.IPNetwork, ndict: Box) -> bool:
+def IPAM_loopback(link: Box, af: str, pfx: ipaddress._BaseNetwork, ndict: Box) -> bool:
   for intf in link.interfaces:
-    pfx.prefixlen = 128 if ':' in str(pfx) else 32
-    intf[af] = str(pfx)
+    prefixlen = 128 if ':' in str(pfx) else 32
```
jbemmel, line 644: ```128 if af=='ipv6'```?

netsim/data/types.py
```diff
@@ -743,21 +743,14 @@ def transform_to_ipv6(value: typing.Any) -> dict:
     return { '_type': 'IPv4 or IPv6 prefix' }
 
   try:
-    parse = netaddr.IPNetwork(value)                                  # now let's check if we have a valid address
+    parse = ipaddress.ip_network(value)                              # now let's check if we have a valid prefix
   except Exception as ex:
-    return check_named_prefix(value) or { '_value': "IPv4, IPv6, or named prefix" }
+    return check_named_prefix(value) or { '_value': f"IPv4, IPv6, or named prefix" }
```
jbemmel, line 748: ? Format string without parameters?

jbemmel (approved):
Few minor tidbits, assuming the tests will be fixed...

netsim/data/types.py
ipspace: Doesn't hurt ;)

netsim/augment/links.py
ipspace: Of course. Thank you, changed.

netsim/modules/vxlan.py
```diff
@@ -120,18 +120,18 @@ def node_set_vtep(node: Box, topology: Box) -> bool:
         log.IncorrectValue,
         'vxlan')
       return False
+  else:
+    vtep_af = 'ipv4'
+    if not 'ipv4' in vtep_interface:
```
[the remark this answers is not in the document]
ipspace: I didn't want to do that because the error messages were different, but I guess making them the same is OK.

PR #1928, opened 2025-02-15
Gateway: Fix incorrect "merging" of non-protocol attributes

### LINKED ISSUE #1927: [BUG] Gateway protocol 'anycastanycast'
Lab:
```yaml
---

module: [ vlan, gateway ]

vlans:
  v1:
    gateway.id: 1
    links: [ s1-s2 ]

groups:
  _auto_create: True
  switches:
    members: [ s1,s2 ]
    gateway.protocol: anycast
```

Results in ```s1/topology.yml```
```yaml
- bridge_group: 1
  gateway:
    anycast:
      mac: 0200.cafe.00ff
      unicast: true
    id: 1
    ipv4: 172.16.0.1/24
    protocol: anycastanycast  <!-- ***
  ifindex: 40000
  ifname: vlan1000
  ipv4: 172.16.0.2/24
  mtu: 1500
  name: VLAN v1 (1000) -> [s2]
```

jbemmel:
Fixes #1927

### DIFF
--- a/netsim/modules/gateway.py
+++ b/netsim/modules/gateway.py
@@ -63,7 +63,7 @@ def cleanup_unicast_ip(node: Box) -> None:
 
 #
 # Default settings copied onto individual links have parameters for every known FHRP protocol.
-# We don't need those parameters on every interface -- this function cleans up unusud gateway protocol
+# We don't need those parameters on every interface -- this function cleans up unused gateway protocol
 # parameters from interfaces and returns a list of active protocols so we know what to clean on the
 # node level.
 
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Looked somewhat counter-intuitive at first (I would have gone with "isinstance(x,Box)"), but it passes the tests and I persuaded myself it's most probably legit, so off we go ;)

ipspace:
Also, thanks for the concise test case ;) Made my life significantly more pleasant ;))

jbemmel:
> Looked somewhat counter-intuitive at first (I would have gone with "isinstance(x,Box)"), but it passes the tests and I persuaded myself it's most probably legit, so off we go ;)

```
for k in proto_list:                                      # Now iterate over all keywords in interface gateway settings
      if k != gw_proto:                                               # ... found FHRP protocol that is NOT the active protocol
        intf.gateway.pop(k,None)                                      # ... useless, pop it
      elif k in node.get('gateway'):                                  # Do we have node-level parameters for this protocol?
        intf.gateway[k] = node.gateway[k] + intf.gateway.get(k,{}))           # ... copy them to all interfaces
```

Alternative option

PR #1929, opened 2025-02-15
EOS: Configure ipv6 anycast gateway IP when provided

jbemmel:
Allows EOS to pass ```gateway/01-anycast``` tests


ipspace (approved):
Thanks a million! Somehow I totally missed the **ipv6** version of the command 🤦‍♂️

Issue #1938, opened 2025-02-17
[BUG] Cumulus NVUE 5.x Vagrant box is FUBARed


ipspace:
It looks like the Cumulus Vagrant boxes have a 180-day password expiration policy that totally borks Vagrant's attempts to start the box (see #1781 for details).

Release 5.10.0 stopped working "today". It was published 166 days ago, and 180 = 166 + 14. Coincidence? I think not.

Fortunately, 5.9.2 was published a few days later, so we still have 12 days to go 🤪

Releases later than 5.10.0 are available for download from NVIDIA (registration required), but they do not contain the version info.

We could:

* Update the documentation to describe how the downloaded Vagrant boxes should be installed.
* Create a box-building procedure from QCOW image (also downloadable from NVIDIA web site)
* Change the clock of the Cumulus NVUE virtual machines (**[libvirt.clock_absolute](https://vagrant-libvirt.github.io/vagrant-libvirt/configuration.html#clock)** should do the trick, not tested)
* Give up and tell the world [Linus was right](https://www.youtube.com/watch?v=tQIdxbWhHSM)

Suggestions, @jbemmel @ddutt @petercrocker?

jbemmel:
It is the Cumulus **Community** boxes that have this issue with expiration (https://portal.cloud.hashicorp.com/vagrant/discover/CumulusCommunity/cumulus-vx/versions/5.10.0), and it looks like they stopped producing new versions after ```5.9.2```

Looks like we'll need to build our own box versions

ipspace:
> It is the Cumulus **Community** boxes that have this issue with expiration

FWIW, the official NVIDIA boxes have the same issue. I pushed the VM clock 6 months into the future and got the exact same error message. That means we'll enjoy downloading and replacing boxes every few months just because whatever.

jbemmel:
Found a solution:
1. Install guestfs-tools: ```sudo apt install guestfs-tools```
2. Find out where Vagrant put its .qcow2 disk image: ```virsh vol-list --pool default```
3. Disable password expiration for the image: ``` sudo virt-customize --run-command "chage -M -1 vagrant" -a /var/lib/libvirt/images/CumulusCommunity-VAGRANTSLASH-cumulus-vx_vagrant_box_image_5.10.0_box.img```

petercrocker:
I've got nothing I can add here @ipspace other than I know people I can ask if it's helpful. I've been removed from this for years!

PR #1939, opened 2025-02-17
Vagrant: Add instructions on how to fix box password expiration

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/1938


ipspace:
Wow. Great job. Will move it to Cumulus caveats though, as we haven't seen anyone else doing something as ~~stupid~~ creative as this.

jbemmel:
> Wow. Great job. Will move it to Cumulus caveats though, as we haven't seen anyone else doing something as ~stupid~ creative as this.

I suspect it may be default behavior by [Packer](https://developer.hashicorp.com/packer/integrations/hashicorp/vagrant/latest/components/builder/vagrant) so it may pop up in other places (because "security"...) - but you can put it wherever you feel is best

ipspace:
@jbemmel Are you OK with the changes?

jbemmel:
> @jbemmel Are you OK with the changes?

Fine with me

Issue #1934, opened 2025-02-17
[BUG] Fix configuration deployment on devices using Ansible task lists


ipspace:
#1933 would add several dummy templates and task lists to get around the safeguards we put in place that broke the devices that do not use the mainstream "config template + potentially module-specific deployment".

At least the following things have to be fixed:

* Do not check for template if the device metadata (probably group variable) tells you it's OK not to have a template.
* Figure out why the generic deploy-config gets involved.
* Create a reference implementation in tests/platform-integration that can be used to verify whether things work as they should

Optional:

* Add IGP information to the loopback interface to allow the IGP configuration task lists (or templates) to use **netlab_interfaces**
* Document how one could use a Jinja2 template to transform the data structure into something a device is willing to accept (to get rid of awkward data manipulation in Ansible task lists).

jbemmel:
I am a little surprised at this - in my experience, whenever something deviates from how things are done on most other platforms, it's "we are not going to change things just because of this one exotic corner case you're dealing with"

It should be possible to change the FortiOS code to use Jinja2 templates to generate Ansible task lists (as you pointed out, similar to how SR Linux templates generate YAML files)

PR #1940, opened 2025-02-17
Draft proposal to integrate Ansible tasklist based device configuration

jbemmel:
Context: https://github.com/ipspace/netlab/issues/1934

This is not ready to merge as-is, just input for discussion on how to restructure the code to facilitate Ansible tasklist based configurations. In particular, the clab image for fortios is a placeholder


ipspace:
The problem with this approach is that it assumes everything is fine if we find a deployment task list. However, that task list might rely on a configuration template that is missing.

An explicit feature flag (introduced in bf5de080e64588b8024fc1b52e65d550b59f4c3e) should be safer.

PR #1941, opened 2025-02-18
Cumulus NVUE: Place mlag backup IP in mgmt vrf when applicable

jbemmel:
For a backup connection to mgmt (eth0) to work, it needs to be placed in the ```mgmt``` vrf

### DIFF
--- a/netsim/ansible/templates/lag/cumulus_nvue.j2
+++ b/netsim/ansible/templates/lag/cumulus_nvue.j2
@@ -22,7 +22,12 @@
     mlag:
       init-delay: 10
       backup:
+{% if pools.mgmt.ipv4 | ansible.netcommon.network_in_usable(lag.mlag.peer_backup_ip) %}
+        {{ lag.mlag.peer_backup_ip }}:
+          vrf: mgmt
+{% else %}
         {{ lag.mlag.peer_backup_ip }}: {}
+{% endif %}
       enable: on
       mac-address: {{ lag.mlag.mac | hwaddr('linux') }}
       peer-ip: {{ lag.mlag.peer }}


ssasso:
> I don't think it's a good idea to run MLAG backup over management VRF. Have you ever seen someone recommending that?

This is only a keepalive/heartbeat connectivity.

Multiple vendors say this can be run on the oob network.

Cumulus doc for example:

```
The backup IP address is any layer 3 backup interface for the peer link, which the switch uses when the peer link goes down. You must add the backup IP address, which must be different than the peer link IP address. Make sure that any route that does not use the peer link can reach the backup IP address. Use the loopback or management IP address of the switch.
```

ssasso:
Cisco vPC:

```
When building a vPC peer-keepalive link, use the following in descending order of preference:

1. Dedicated link(s) (1-Gigabit Ethernet port is enough) configured as L3. Port-channel with 2 X 1G port is even better.
2. Mgmt0 interface (along with management traffic)
3. As a last resort, route the peer-keepalive link over the Layer 3 infrastructure
```

ssasso:
Arista Validated Design config file examples put the multi master detection on management vrf:

https://github.com/arista-netdevops-community/avd-evpn-webinar-june-11/blob/ade90f1ae3b1d381c91645c74f370c370527d327/intended/configs/LEAF2A.cfg#L259

jbemmel:
> I don't think it's a good idea to run MLAG backup over management VRF. Have you ever seen someone recommending that?

I am taking a Cumulus NVUE training this week, and the trainer said it's the most common way customers deploy this.

I was thinking like you, so the default: https://github.com/ipspace/netlab/blob/252fd1b53bc7e8ba741deccc5e7dcff097e5cab1/netsim/devices/cumulus_nvue.yml#L46
uses ```loopback.ipv4```. However, that requires an IGP

This PR merely says "if the backup is set to the management IP, add the mgmt vrf". It won't work without that

jbemmel:
> 
> This feels a lot like another one of those "let me fix my immediate problem without considering the wider impact" PRs.

:)

It does fix an immediate problem (i.e. broken promise in the code base) but it could use some additional documentation. Similar fixes may be required for other MLAG platforms

I don't believe supporting other VRFs for this item would make sense (if people need that, they should create a plugin...)

ipspace:
> It does fix an immediate problem (i.e. broken promise in the code base)

I don't know what you're talking about. **backup_ip** is not documented anywhere ;)

> but it could use some additional documentation. Similar fixes may be required for other MLAG platforms

I hope nobody is brave enough to try to guess which VRF to use automatically ;)

> I don't believe supporting other VRFs for this item would make sense (if people need that, they should create a plugin...)

Then that needs to be documented as well.

PR #1942, opened 2025-02-18
Fix IOS VRF route leaking

melvync:
Leaking routes to global from a VRF by setting nexthop.vrf to an empty value as described in the documentation is not working because the keyword "global" is not added to the static route.


ipspace:
Thanks a million for reporting/fixing this!

I have to dust off my IOS skills, but I'm assuming you encountered this challenge in a scenario where the next hop of a route is an indirect (remote) next hop, right?

Of course I'll merge this, I just have to understand what we're missing in the integration tests (one of the reasons why it's always nice to have a simple topology that reproduces the problem).

melvync:
> Thanks a million for reporting/fixing this!
> 
> I have to dust off my IOS skills, but I'm assuming you encountered this challenge in a scenario where the next hop of a route is an indirect (remote) next hop, right?
> 
> Of course I'll merge this, I just have to understand what we're missing in the integration tests (one of the reasons why it's always nice to have a simple topology that reproduces the problem).

I encountered this problem while creating GRE tunnels that originate from a VRF and connect to another VRF on a remote router… The endpoints of the tunnels must be leaked to the global routing table on both routers to successfully build the GRE tunnels over the existing infrastructure.

PR #1945, opened 2025-02-19
cEOS: bump version to 4.33

### LINKED ISSUE #1944: Upgrade EOS version?
Shall we update the (c)EOS version to `4.33.1F` (the one supporting ARM as well)?

ssasso:
Close #1944


ssasso:
> There is a 4.33.0F too, I've been running that for months

Yes, but why use 4.33.0 when we can go directly to 4.33.1?

ipspace (approved):
We need to fix those transformation tests to use NONE device. Will do ;)

PR #1946, opened 2025-02-19
FRR: Fix incorrect MTU setting on vxlan devices

### LINKED ISSUE #1943: [BUG] FRR fails EVPN test 20-vxlan-irb-ospf
```netlab up integration/evpn/20-vxlan-irb-ospf.yml -p clab -d frr -v``` fails (it works with EOS)

Root cause seems to be that there is no ```evpn``` attribute in the "tenant" vrf, even though the ```red``` and ```blue``` VLANs have EVPN enabled.

The EOS templates don't depend on ```vrfs.X.evpn``` being defined for enabling EVPN (which also seems questionable)

jbemmel:
Allows FRR to pass integration/evpn/20-vxlan-irb-ospf.yml

Fixes #1943

### DIFF
--- a/netsim/ansible/templates/vxlan/frr.j2
+++ b/netsim/ansible/templates/vxlan/frr.j2
@@ -30,13 +30,10 @@ ip link set {{ br_name }} master {{ vrf }}
 {% endif %}
 {% endmacro %}
 
-# Determine max MTU used on interfaces, to configure VXLAN interface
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
The MTU calculation assumes that the default **mtu** value is set for FRR nodes. That happens to be true today, but who knows what we might decide to do in the future.

Do not rely on an optional attribute (**mtu** is optional) always being present. You should add `|default(1500)` everywhere you mention the node MTU just to be on the safe side.

ipspace (approved):

PR #1951, opened 2025-02-20
Cumulus NVUE: Add EVPN support

jbemmel:
I ran ```NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test evpn```

8/14 tests validate ok. Some are failing due to missing bits that aren't part of EVPN (like missing 'redistribute connected' for VRFs); I could easily fix that, but this isn't the right place

Main things still missing:
* generalized redistribution of routes (separate PR)
* BGP in VRFs

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -205,7 +205,7 @@ ansible-galaxy collection install git+https://github.com/jmussmann/ansible_colle
 
 * The Cumulus VX 4.4.0 Vagrant box for VirtualBox is broken. *netlab* is using Cumulus VX 4.3.0 with *virtualbox* virtualization provider.
 * The Cumulus VX 4.x uses Python version 3.7, which recent versions of Ansible refuse to work with. The permanent fix is coming in release 1.9.3. Until then, use the **frrouting** device or [Cumulus VX 5.x image](caveats-cumulus-5x).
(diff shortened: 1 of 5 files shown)


ipspace:
> I ran `NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test evpn`

Thank you.

> 8/14 tests validate ok. Some are failing due to missing bits that aren't part of EVPN (like missing 'redistribute connected' for VRFs); I could easily fix that, but this isn't the right place

All devices supporting EVPN implemented the "missing bits" in the EVPN template (or in the VRF template) until we implemented the generic routing module for them.

At the very minimum, you should document your shortcomings in the caveats.

jbemmel:
> 
> All devices supporting EVPN implemented the "missing bits" in the EVPN template (or in the VRF template) until we implemented the generic routing module for them.
> 
> At the very minimum, you should document your shortcomings in the caveats.
>
OK, so what I propose to do: I'll put back those bits required to pass the tests, marking them as "to be moved" at a later point in time.

(while I was working on it, I could hear your comments about things not belonging there, so I anticipated :)

ipspace:
> OK, so what I propose to do: I'll put back those bits required to pass the tests, marking them as "to be moved" at a later point in time.

I'm running tests right now to get the baseline. Looking at the code, it seems we should just set **irb: false** for the moment.

ipspace:
The results are in: with **irb: false** in cumulus_nvue.yml all tests either pass or we get "not supported" message. Change that setting and the corresponding coverage in evpn.md and we can merge this (and add all the bells and whistles in another PR).

image

PR #1952, opened 2025-02-20
Cumulus NVUE: Generalized redistribute of OSPF/BGP/connected routes

jbemmel:
Small tweak to redistribute configs, still missing BGP in VRFs but will adjust when that gets added

Tested: ```integration/ospf/ospfv2/10-import.yml```


ipspace:
Based on the recent experience with #1881 I'm not touching this one until you tell me it passes all BGP, OSPF and routing integration tests.

ipspace:
BGP and OSPF tests passed, **features.routing** not set yet. Merging.

ipspace (approved):

jbemmel:
> BGP and OSPF tests passed, **features.routing** not set yet. Merging.

Thanks

Going forward, I think the bar should be: Doesn't break any tests that currently pass, and improves things in some way

Issue #1954, opened 2025-02-21
Increase auto-generated management MAC address diversity


jbemmel:
It turns out that the ```vlan/42-vlan-irb-native``` test fails for Cumulus NVUE, also in the ```dev``` branch. The reason is that it auto-generates duplicate MAC addresses on the VLAN interfaces.

Each VM gets a 'Base MAC Address' and a serial number; the latter is identical to the mgmt MAC address and the base is derived from that.
Here is an example based on ```vlan/42-vlan-irb-native``` in ```dev```:
```
cumulus@s1:mgmt:~$ decode-syseeprom 
TlvInfo Header:
   Id String:    TlvInfo
   Version:      1
   Total Length: 69
TLV Name             Code Len Value
-------------------- ---- --- -----
Vendor Name          0x2D  16 Cumulus Networks
Product Name         0x21   2 VX
Device Version       0x26   1 3
Part Number          0x22   5 5.9.2
MAC Addresses        0x2A   2 4
Base MAC Address     0x24   6 08:4F:C2:A9:00:00
Serial Number        0x23  17 08:4f:a9:00:00:05
CRC-32               0xFE   4 0x85379A81
(checksum valid)
```
On s2:
```
cumulus@s2:mgmt:~$ decode-syseeprom 
...
Base MAC Address     0x24   6 08:4F:C2:A9:00:00
Serial Number        0x23  17 08:4f:a9:00:00:06
```

As you can see, Base MAC Address = {{ mgmt.mac[0:2] }}:C2:{{ mgmt.mac[2:5] }}, and the 6th octet gets ignored.
If 2 VMs have a management MAC that only differs in the last octet, they get identical base MACs, and trouble results

One possible fix: https://github.com/ipspace/netlab/blob/a01920e8b985f4eb5f84688e6249e49c68e07c5f/netsim/augment/nodes.py#L147

Changing ```mac_eui[5]``` into ```mac_eui[3]``` would solve it

But knowing you a little bit, I'm guessing you'd prefer a device quirk that rewrites the mgmt.mac only for Cumulus NVUE nodes...

ipspace:
> It turns out that the `vlan/42-vlan-irb-native` test fails for Cumulus NVUE, also in the `dev` branch.

That was not true when I was doing the full tests for 1.9.4-post2, using release 5.10.0 at that time (sidebar: SW release info is already added to the new test reports, I "just" have to display it). There's a reason the test reports are under version control and are snapshotted when a release goes out.

[...]

PR #1955, opened 2025-02-21
Use assigned node.mgmt.mac/ipv4 instead of generating one

jbemmel:
In order to fix https://github.com/ipspace/netlab/issues/1954 with a quirk, ```node.mgmt.mac``` needs to be configurable. This PR replaces the hardwired MAC generation based on mgmt.mac - with the node.id in the last octet - with the already assigned node.mgmt.mac and the corresponding node.mgmt.ipv4

Testing: I ran 
* NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test vlan -v
* NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test lag -v

with this PR, and all tests passed


ipspace (approved):
This is really good idea. Half-baked with no guardrails to fix your immediate challenge, but that wasn't too hard to fix ;)

Opened two additional issues for the remaining bits and pieces.

tests/platform-integration/libvirt/02-mac-ip.yml
```diff
@@ -0,0 +1,31 @@
...
+module: [ ospf ]
+ospf.timers.hello: 1
+defaults.multilab.change.addressing.mgmt.ipv4: '192.42.42.0/24'
+addressing.mgmt.ipv4: '192.42.42.0/24'
+
+nodes:
+  r1:
+    mgmt:
+      mac: 0A:4F:A9:01:00:00
+      ipv4: 192.42.42.110
```
jbemmel, line 16: Would like to be able to use ```110``` here too - probably requires a small tweak

Issue #1949, opened 2025-02-19
Document MLAG peer_backup_ip parameter and its limitations


ipspace:
> It does fix an immediate problem (i.e. broken promise in the code base)

I don't know what you're talking about. **backup_ip** is not documented anywhere ;)

> but it could use some additional documentation. Similar fixes may be required for other MLAG platforms

I hope nobody is brave enough to try to guess which VRF to use automatically ;)

> I don't believe supporting other VRFs for this item would make sense (if people need that, they should create a plugin...)

Then that needs to be documented as well.

_Originally posted by @ipspace in https://github.com/ipspace/netlab/issues/1941#issuecomment-2669075046_

jbemmel:
Done https://github.com/ipspace/netlab/pull/1956

PR #1956, opened 2025-02-21
lag: Document lag.mlag.peer.backup_ip

jbemmel:
Addresses https://github.com/ipspace/netlab/issues/1949

### DIFF
--- a/docs/module/lag.md
+++ b/docs/module/lag.md
@@ -82,7 +82,14 @@ The above diagram illustrates the three supported topologies:
 * 2:2 dual MLAG between 2 pairs of nodes (4 nodes in total)
 
 MLAG related parameters:
-* **lag.mlag.mac**: Used at node or interface level to configure the MAC address for the peer link. *Netlab* can auto-generate this, so it usually is not necessary to set this
+
+Global:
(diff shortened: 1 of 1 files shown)


docs/module/lag.md
```diff
@@ -82,7 +82,14 @@ The above diagram illustrates the three supported topologies:
 * 2:2 dual MLAG between 2 pairs of nodes (4 nodes in total)
 
 MLAG related parameters:
-* **lag.mlag.mac**: Used at node or interface level to configure the MAC address for the peer link. *Netlab* can auto-generate this, so it usually is not necessary to set this
+
+Global:
```
ipspace, line 86: Looking at the code, it seems to me that this has to be defined in the device features, not as a global attribute. Am I missing something?

jbemmel: Globally as a device feature ;) Clarified

It implies the code does not support topologies in which some devices use ```loopback.ipv4``` and others use ```mgmt.ipv4``` - that could be generalized (by making it a node level attribute)

ipspace (approved):
Moved the "backup_ip" to a new "advanced parameters" section. It's not something a first-time user should change ;)

jbemmel:
> Moved the "backup_ip" to a new "advanced parameters" section. It's not something a first-time user should change ;)

If they’re reading the mlag documentation, they’re unlikely to be first time users…

PR #1961, opened 2025-02-21
Use Vagrant private key instead of username for FRR

ipspace:
This is a proof-of-concept for the 'use Vagrant private key' VM authentication. FRR is used as an example because it already uses Debian Bookworm; other devices that could use the same approach are Cumulus (4/5), VyOS, and Linux.

### DIFF
--- a/netsim/templates/provider/libvirt/frr-domain.j2
+++ b/netsim/templates/provider/libvirt/frr-domain.j2
@@ -1,15 +1,6 @@
 {% if _vagrant_scripts.frr is not defined %}
 {%   set _vagrant_scripts.frr = True %}
 
-    $frr_script = <<-SCRIPT
-echo "Setting password for user Vagrant"
-echo vagrant:{{ n.ansible_ssh_pass }} | chpasswd
-
-echo "Enabling SSH password authentication"
-sed -i -e "s#PasswordAuthentication no#PasswordAuthentication yes#" /etc/ssh/sshd_config
-service sshd restart
-    SCRIPT
-
     $frr_install_script = <<-SCRIPT
 set -e
 if which /usr/lib/frr/frrinit.sh; then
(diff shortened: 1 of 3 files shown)


ipspace:
After #1931, we know the netlab commands are always executed within the lab directory, so it's safe to use relative path to the Vagrant private key as an Ansible group variable.

netsim/templates/provider/libvirt/frr-domain.j2
```diff
@@ -1,15 +1,6 @@
 {% if _vagrant_scripts.frr is not defined %}
 {%   set _vagrant_scripts.frr = True %}
 
-    $frr_script = <<-SCRIPT
-echo "Setting password for user Vagrant"
-echo vagrant:{{ n.ansible_ssh_pass }} | chpasswd
```
jbemmel, line 6: Instead of deleting this, can't we do something like
```
{% if n.ansible_ssh_pass is defined %}
...
```

I think we should keep the option open to use plain passwords, if users prefer

jbemmel (commented):
I would suggest that we keep backwards compatibility, users may have a need to have SSH password authentication enabled for them

ipspace:
> I would suggest that we keep backwards compatibility, users may have a need to have SSH password authentication enabled for them

Every nerd knob is a noble idea until it has to be supported. Who will test that code when we change something else or go for another Linux distro? You?

Also, the users can easily change the Vagrant template or fix the Vagrant box the way you fixed Cumulus Linux password policy if they don't like what Debian is doing.

ipspace:
> users may have a need to have SSH password authentication enabled for them

Unfortunately, you're absolutely right, we can't reduce this complexity :( We need password-based authentication in all cases for network management tools. I don't want to deal with copying per-VM private keys into network management container volume.

ipspace:
Back to the drawing board...

PR #1963, opened 2025-02-22
Use Vagrant private key instead of username for FRR

ipspace:
This is a proof-of-concept for the 'use Vagrant private key' VM authentication. FRR is used as an example because it already uses Debian Bookworm; other devices that could use the same approach are Cumulus (4/5), VyOS, and Linux.

We're adding private-key-based-authentication functionality to netlab while retaining the password-based authentication that might be needed for network management tools.

The Vagrant provisioning script adds the commands to set the password for the 'vagrant' user (unless it's not defined) and change the Debian default SSH authentication policy.

The Debian-specific provisioning steps are stored in a separate template as we'll need them when migrating Linux node to Debian.

### DIFF
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -23,7 +23,7 @@ from . import external_commands, set_dry_run, error_and_exit, parser_lab_locatio
 
 from . import load_snapshot, parser_add_verbose
 from ..outputs import common as outputs_common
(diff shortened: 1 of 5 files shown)


ipspace:
This one retains password-based authentication while adding the support for VM-specific private key. I will update the documentation once I do the same thing for Linux.

PR #1964, opened 2025-02-22
Change the L3 LAG integration test to use static routing

ipspace:
The current L3 LAG integration test uses OSPF and thus fails on Linux. This change replaces OSPF with static routing on the probe, testing just pure L3 forwarding functionality over LAG.

### DIFF
--- a/tests/integration/lag/01-l3-lag.yml
+++ b/tests/integration/lag/01-l3-lag.yml
@@ -2,24 +2,54 @@ message: |
   The device under is a router with a L3 LAG link connected to a FRR device. The
   FRR device should be able to ping the loopback interface of DUT.
 
-nodes:
-  dut:
-    module: [ lag ]
-    role: router
-  xr:
-    device: frr
+groups:
+  probes:
+    device: eos
     module: [ lag, routing ]
+    members: [ x1, x2 ]
     routing.static:
     - ipv4: 0.0.0.0/0
       nexthop.node: dut
 
+nodes:
+  dut:
+    module: [ lag ]
+    role: router
+  x1:
+    role: host
+  x2:
+    role: host
+
 links:
-- lag.members: [dut-xr, dut-xr]
+- lag.members: [dut-x1, dut-x1]
   mtu: 1500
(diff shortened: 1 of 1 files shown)


tests/integration/lag/01-l3-lag.yml
```diff
@@ -1,25 +1,25 @@
...
-    description: Check OSPF adjacencies
-    wait_msg: Waiting for OSPF adjacency process to complete
-    wait: 50
-    nodes: [xr]
-    plugin: ospf_neighbor(nodes.dut.ospf.router_id)
+  ping:
+    wait_msg: Waiting for LAG to be established
+    wait: 15
+    nodes: [ xr ]
+    plugin: ping(nodes.dut.loopback.ipv4)
```
jbemmel, line 25: Could set MTU to 1600 and do a ping of size 1600, to verify that the device configures MTU correctly for the LAG interface

OSPF was testing this for us before, but not anymore

jbemmel (commented):
Suggestion to add an MTU test

Actually, for completeness we could do 2 pings - 1 too large (and verify it correctly fails, after setting the FRR side to 9000)

tests/integration/lag/01-l3-lag.yml
ipspace: Thank you. Will do (I'll probably creatively reuse the initial MTU test)

ipspace:
Changed the test to use two LAG links, one with lower MTU (1400). Unfortunately, the MTU detection logic relies on DUT sending ICMP unreachables, and some devices don't do that, so I made that part of the test optional (hoping someone would review the "warning" results).

jbemmel:
> Changed the test to use two LAG links, one with lower MTU (1400). Unfortunately, the MTU detection logic relies on DUT sending ICMP unreachables, and some devices don't do that, so I made that part of the test optional (hoping someone would review the "warning" results).

Are warnings reported in the integration test results overview? Assuming so, I think this is the best we can do

jbemmel (approved):
Assuming warnings are visible in the integration test report

PR #1965, opened 2025-02-22
Check the validity of management IP addresses

### LINKED ISSUE #1957: [BUG] Check assigned management addresses for libvirt and clab provider
After merging #1955, we have to check that:

* The IPv4 or IPv6 addresses assigned to containers are within the management subnet (otherwise containerlab fails)
* That the libvirt nodes have a management IPv4 address (Vagrant boxes are not configured to use DHCPv6) within the management subnet (otherwise the IPv4 address gets assigned but is not reachable)

ipspace:
clab and libvirt management IP addresses must be within the management subnet and the AF used on the node must be defined in the management pool.

Furthermore, libvirt nodes need IPv4 management address.

Closes #1957

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -10,6 +10,7 @@ import subprocess
 import os
 import typing
 import pathlib
+import ipaddress
 
 # Related modules
 from box import Box
@@ -378,7 +379,6 @@ def get_forwarded_ports(node: Box, topology: Box) -> list:
 """
 validate_images -- check the images used by individual nodes against provider image repo
 """
-
 def validate_images(topology: Box) -> None:
   p_cache: dict = {}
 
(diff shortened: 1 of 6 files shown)


netsim/providers/__init__.py
```diff
@@ -378,11 +379,48 @@ def get_forwarded_ports(node: Box, topology: Box) -> list:
...
     execute_node('validate_node_image',n_data,topology)
 
   log.exit_on_error()
+
+"""
+validate_mgmt_ip -- Validate management IP addresses
+"""
+def validate_mgmt_ip(
+      node: Box,
+      provider: str,
```
jbemmel, line 395: ```provider``` is only used as ```module``` parameter in logging, perhaps declare this to be part of the ```initial``` module (or ```addressing``` module, although that doesn't really exist) and leave out the ```provider``` parameter?

jbemmel: Looking at ```validate_pools``` in ```addressing.py``` which uses ```"addressing"```

jbemmel (approved):
Minor thoughts on logging module parameter, not material

netsim/providers/__init__.py
ipspace: We need the provider parameter for the first error message. Libvirt requires IPv4 management address whereas (in theory) clab could work just with IPv6 ones.

Issue #1975, opened 2025-02-23
[BUG] VRF IPv6 BGP configuration fails on Cumulus NVUE


ipspace:
22-multi-vrf-bgp-ipv6:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/22-multi-vrf-bgp-ipv6.yml)
* [Test results](https://tests.netlab.tools/cumulus_nvue/libvirt/vrf/22-multi-vrf-bgp-ipv6.yml-initial.log)

jbemmel:
I started running ```NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test vrf -v``` but I stopped short when I hit #1971 

Hit the same issue in another test case - will fix

Issue #1971, opened 2025-02-22
Validation plugin for device frr has no action for test 'bgp_prefix_community'


jbemmel:
https://github.com/ipspace/netlab/blob/583a55f58766da3964f2cd5e3ba73aba7420f28c/tests/integration/vrf/16-vrf-bgp-community.yml#L52

There seems to be no plugin called "bgp_prefix_community" - there is a ```bgp_prefix``` validation plugin

ipspace:
Thank you, will fix.

PR #1976, opened 2025-02-23
Cumulus NVUE: Support VRF route leaking

jbemmel:
* Include BGP module config from VRF module template when needed
* Fix BGP RD configuration
* Fix ```advertise_loopback``` (also fixes https://github.com/ipspace/netlab/issues/1975 - I hit the same issue)
* Use custom NVUE snippet to configure FRR route leaking (adding caveat)

Tested: ```NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test vrf -v```
- tests involving OSPFv3 fail (21,23 - not supported), but the rest all pass except for 16 (see #1971)


ipspace (approved):
Passed all BGP and VRF tests, merging

PR #1979, opened 2025-02-25
Linux: Fix LAG configuration on Ubuntu 24.04

### LINKED ISSUE #1978: [BUG] LACP might not work on Linux VM after changing Vagrant box to Ubuntu 24.04
The baseline LAG test fails with Linux VM but works with Linux container:

https://tests.netlab.tools/linux/libvirt/lag/01-l3-lag.yml-validate.log

The only Linux-related thing that changed recently was the change of Vagrant box from `generic/ubuntu2004` to `bento/ubuntu-24.04`, that might be the cause of the problem.

ipspace:
* Delete bond interface after setting the speed on member interfaces and recreate it with 'netplan'
* Set MTU on bond interfaces

Fixes #1978


jbemmel (commented):
The MTU fix may still be needed, but I think https://github.com/ipspace/netlab/pull/1980 is a cleaner solution to the second part

PR #1980, opened 2025-02-25
Ubuntu: Configure Systemd to disable auto-negotiation on Ethernet interfaces

jbemmel:
… and fix the speed to 1G full duplex

Fixes https://github.com/ipspace/netlab/issues/1978


ipspace (approved):
Nice one, thank you!

PR #1983, opened 2025-02-26
Use FRR/EOS devices as probes in VLAN integration tests

### LINKED ISSUE #1862: Use multiple device types in VLAN integration tests
From @jbemmel in #1861 

> All vlan tests are testing DUT against itself, which might fail to detect certain misconfigurations. For example, a device configuring a vlan as tagged while it should be untagged, or vice versa, would still succeed in talking to itself

We should use FRR or EOS as one of the switches and (when possible) get creative, for example checking whether VLAN 1 is bridged on a trunk port.

ipspace:
Closes #1862


tests/integration/vlan/41-vlan-bridge-native.yml
```diff
@@ -12,15 +12,22 @@ message: |
...
     device: linux
     provider: clab
   switches:
-    members: [ s1, s2 ]
+    members: [ s1 ]
+    module: [ vlan ]
+  probes:
+    members: [ s2 ]
     module: [ vlan ]
+    device: eos
```
jbemmel, line 29: Could ```frr``` be used too?

EOS has a hardened, production grade implementation, FRR is open source and more easily obtained.
Could add a comment explaining why ```eos``` is preferred here

jbemmel (approved):

tests/integration/vlan/41-vlan-bridge-native.yml
ipspace: Let's just say I trust EOS more than what we're hacking together for FRR ;)

PR #1984, opened 2025-02-26
ansible playbook for SONiC platform backup (for use with `netlab collect`)

sulrich:
## PR overview

this PR adds a SONiC-specific ansible playbook for use with the `netlab collect` process.

currently, when `netlab collect` command is run against a SONiC node, it only attempts to backup the FRR configuration. (and only the `frr.conf` file)  this extends the backup process to do the following:

- save both the SONiC and the FRR configs to their respective storage locations.
- capture the contents of the `/etc/sonic` directory to the host environment.
- capture the contents of the `/etc/sonic/frr` directory to the host environment.

an `frr` subdirectory is created in the `config_dir` to separate these configurations.  the current naming structure (`<config-file>-<inventory_hostname>`) is retained.

## test coverage

- all tests run clean.  though, in reviewing `tests/run-tests.sh` it doesn't look like ansible elements are touched on here.

### DIFF
--- a/netsim/ansible/tasks/fetch-config/sonic.yml
+++ b/netsim/ansible/tasks/fetch-config/sonic.yml
@@ -1,4 +1,3 @@
-
 # get the core SONiC config file
 - name: save current SONiC configuration to config_db.json
   command: sudo config save -y
@@ -26,7 +25,7 @@
   become: true
   register: file_content
   loop: "{{ sonic_files.files }}"
-  
+
 - name: Save files from /etc/sonic/
   copy:
     content: "{{ item.stdout }}"
@@ -52,7 +51,7 @@
   become: true
   register: frr_file_content
   loop: "{{ frr_files.files }}"
-  
+
 - name: Save files from /etc/sonic/frr/
   copy:
     content: "{{ item.stdout }}"
--- a/netsim/ansible/tasks/fetch-config/sonic.yml
+++ b/netsim/ansible/tasks/fetch-config/sonic.yml
@@ -3,8 +3,8 @@
   command: sudo config save -y
   become: true
 
-# Create inventory-specific frr directory if it doesn't exist
-- name: Create inventory-specific frr directory
+# Create node-specific frr directory if it doesn't exist
+- name: Create node-specific frr directory
   file:
     path: "{{ config_dir }}/{{ inventory_hostname }}-frr"
     state: directory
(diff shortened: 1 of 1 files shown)


ipspace:
Thank you! The only change I would make is the naming of the FRR directory.

It might be better to have FRR files under "{{ inventory-hostname }}-frr" directory so it's evident there are extra files FRR-related files collected from a particular node.

sulrich:
works for me.  will update later today.

ipspace:
> works for me. will update later today.

Thank you!

sulrich:
updated to align per discussion. (turned out my car _wasn't_ ready for pickup)

ipspace (approved):
It works ;) The amount of files collected is surprisingly high, but if that's what it takes 🤷‍♂️

ipspace:
> updated to align per discussion.

Thank you. Merged. Welcome to the club!

> (turned out my car _wasn't_ ready for pickup)

Yeah, I know that lonely feeling :( and how good it is to have a laptop with you ;)

jbemmel:
> Thank you! The only change I would make is the naming of the FRR directory.
> 
> It might be better to have FRR files under "{{ inventory-hostname }}-frr" directory so it's evident there are extra files FRR-related files collected from a particular node.

Wouldn't ```{{ inventory-hostname }}/frr``` (subdirectory) make more sense?

sulrich:
> > Thank you! The only change I would make is the naming of the FRR directory.
> > It might be better to have FRR files under "{{ inventory-hostname }}-frr" directory so it's evident there are extra files FRR-related files collected from a particular node.
> 
> Wouldn't `{{ inventory-hostname }}/frr` (subdirectory) make more sense?

i was holding off of placing things into discrete subdirectories broadly as it seemed like there wasn't this pattern in other collection actions. though to be fair, SONiC seems to be the patient `0` here wrt littering config across a range of files and directories.  happy to amend accordingly.

ipspace:
> Wouldn't `{{ inventory-hostname }}/frr` (subdirectory) make more sense?

No, as long as all other files are in the "config" directory. You would have a "rtr" directory with a single entry ("frr" subdirectory) and a bunch of files in the "frr" subdirectory.

ipspace:
> i was holding off of placing things into discrete subdirectories broadly as it seemed like there wasn't this pattern in other collection actions. though to be fair, SONiC seems to be the patient `0` here wrt littering config across a range of files and directories. happy to amend accordingly.

I was thinking about asking you to put SONiC files into "nodename-sonic" directory, but I don't care one way or the other as long as nobody is complaining 🤷‍♂️

PR #1988, opened 2025-02-28
Dell: Disable ipv6 on ipv4-only port-channels

jbemmel:


netsim/ansible/templates/initial/dellos10.j2
```diff
@@ -84,7 +84,7 @@ interface {{ l.ifname }}
 {%   if l.virtual_interface is not defined %}
  ipv6 unreachables
 {%   endif %}
-{% elif 'lag' not in l %}
+{% elif 'lag' not in l or 'ipv4' in l %}
```
ipspace, line 87: Is there a reason you cannot just say "otherwise disable IPv6"? Are there interfaces that would complain if you do **no ipv6 enable**?

ipspace: Anyway, I would love to see a more deterministic approach, not just a random list of exceptions put together as you find them.

jbemmel: image

This is the result of a deterministic approach:
```
|   Type                    | can set ```no ipv6 enable``` |
| ------------------------ | :-: |
| Regular links (no 'lag')   |  ✅ |
| L3 port-channels ('lag' and 'ipv4')   |  ✅ |
```

ipspace: This raises even more questions, but I'm not sure I want to know the answers. It looks like the less I know about OS10, the better I sleep.

Issue #1987, opened 2025-02-28
Make saving of config configurable


jbemmel:
I have a use case where I need the running configuration on a Dell OS10 node to stay different from the startup config, to determine the diff

Currently there is a hardcoded ```save: "yes"``` - I'd like to make that configurable (e.g. a ```--no-save``` option to 'up' and 'initial')

ipspace:
Can't you use "netlab up --no-config" followed by "netlab collect -o before", "netlab initial" and "netlab collect -o after"

jbemmel:
> Can't you use "netlab up --no-config" followed by "netlab collect -o before", "netlab initial" and "netlab collect -o after"

I'm creating a Netbrain plugin that retrieves the current configs of nodes and puts them as ```clab.startup-config```.
I do start with ```netlab up --no-config```, but then I'd like the ```initial``` config (staged changes to the startup-config) to remain as 'running' such that I can do ```show diff startup-config running-config``` (inside the box)

ipspace:
Thinking out loud:

* You could always copy the initial startup config into another file on the device and run diff against that
* On-device **show diff** is often just Linux diff in disguise.

Anyway:

* Adding an option to **netlab initial** would mean we have to implement the same functionality on all devices, and I think we have better things to do.
* However, using **netlab_save_config** node (or group/device) variable and checking that in the Ansible playbook with the default being the current behavior would not impact anything but the devices for which someone has the willpower to implement this ;)

PR #1989, opened 2025-02-28
Dell OS10: Add 'netlab_save_config' (default:True) to control whether initial config is saved to startup

jbemmel:
Test: ```netlab up integration/vlan/41-vlan-bridge-native.yml -p clab -d dellos10 -v -s defaults.devices.dellos10.group_vars.netlab_save_config=False```

```
TASK [dellos10_config: deploying initial from /home/jeroen/Projects/netlab/netsim/ansible/templates/initial/dellos10.j2, save to startup=no]

s1# show diff running-configuration startup-configuration 
!
interface ethernet1/1/1
 switchport mode trunk
 switchport access vlan 700
 switchport trunk allowed vlan 701
 switchport
!
interface ethernet1/1/2
 no ipv6 enable
 description "[Access VLAN red] s1 -> h1"
 switchport access vlan 700
 switchport
!

etc.
```

Fixes https://github.com/ipspace/netlab/issues/1987 (for Dell OS10)

### DIFF
--- a/netsim/ansible/tasks/deploy-config/dellos10.yml
+++ b/netsim/ansible/tasks/deploy-config/dellos10.yml
@@ -3,10 +3,12 @@
   when: |
     node_provider == 'clab' and ansible_connection == 'network_cli'
 
-- name: "dellos10_config: deploying {{ netsim_action }} from {{ config_template }}"
+- name: "dellos10_config: deploying {{ netsim_action }} from {{ config_template }}, save to startup={{ _save }}"
+  vars:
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Having the "|default" filter in Ansible expression while at the same time adding device variable is a bit of an overkill ;)

jbemmel:
> Having the "|default" filter in Ansible expression while at the same time adding device variable is a bit of an overkill ;)

I learned from the best - always minimize your assumptions (=dependencies)

PR #1992, opened 2025-03-04
JunOS: support for BGP local-as

ssasso:
Support for BGP local-as and specific AF activate in JunOS.

Passed testing
- bgp/07
- bgp/08
- bgp/09

### DIFF
--- a/netsim/ansible/templates/bgp/junos.j2
+++ b/netsim/ansible/templates/bgp/junos.j2
@@ -89,11 +89,15 @@ protocols {
     group ebgp-peers {
       export ebgp-export;
       advertise-inactive;
-{% for n in bgp.neighbors if n.type == 'ebgp' %}
+{# treat localas_ibgp as ebgp session --> no usage of local address #}
+{% for n in bgp.neighbors if (n.type == 'ebgp' or n.type == 'localas_ibgp') %}
 {%   for af in ['ipv4','ipv6'] if n[af] is defined %}
       neighbor {{ n[af] }} {
         peer-as {{ n.as }};
         description {{ n.name }};
+{%     if n.local_as is defined %}
+        local-as {{ n.local_as }}{% if n.replace_global_as|default(True) %} no-prepend-global-as{% endif +%};
+{%     endif %}
         family {{ 'inet' if af == 'ipv4' else 'inet6' }} {
           unicast;
         }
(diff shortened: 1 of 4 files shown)


ssasso:
moved to draft to see if I can easily implement "configurable activation of default BGP address"

ssasso:
Side note: test `bgp/20-dual-stack-activate.yml` expect that a session without the enabled AF is in Active or Idle state. Here JunOS is bringing up the session in any case, but not exchanging any update.

ipspace:
> Side note: test `bgp/20-dual-stack-activate.yml` expect that a session without the enabled AF is in Active or Idle state. Here JunOS is bringing up the session in any case, but not exchanging any update.

Opening a session but not negotiating any address family is "unusual" (I'm being diplomatic today)... but what else could one expect from a company that is willing to accept a BGP session from anyone just to tell them to go away.

Will fix the test plugins and the integration test.

ipspace (approved):
Looks good to me ;)

netsim/ansible/templates/bgp/junos.j2
```diff
@@ -79,24 +79,32 @@ protocols {
...
       }
 {%   endfor %}
     }
 {% endfor %}
     group ebgp-peers {
       export ebgp-export;
       advertise-inactive;
-{% for n in bgp.neighbors if n.type == 'ebgp' %}
+{# treat localas_ibgp as ebgp session --> no usage of local address #}
+{% for n in bgp.neighbors if (n.type == 'ebgp' or n.type == 'localas_ibgp') %}
```
jbemmel, line 95: Perhaps ```localas_ibgp ``` is similar to ```ebgp``` in that it does not use a local address, but similar to ```ibgp``` in that the ```ibgp-export``` policy should apply?

netsim/ansible/templates/bgp/junos.j2
```diff
@@ -79,24 +79,32 @@ protocols {
 {%   for n in bgp.neighbors if n[af] is defined and n.type == 'ibgp' %}
```
jbemmel, line 79: ``` 'ibgp' in n.type ``` (should also apply to localas_ibgp neighbors)

netsim/ansible/templates/bgp/junos.j2
ssasso: I didn't want to add overcomplexity with the need to declare a new group or too much per-neighbor options.

If we find use cases where this is not working we can either fix it or declare it as current limitation or caveat.

My 2c

PR #2002, opened 2025-03-05
JunOS: support for ebgp multihop and bgp allowas-in

ssasso:
Added an additional neigh attribute from the `ebgp.multihop` plugin to expose the entire source loopback data, in order to extract the IP addresses as well (and not only the source interface).

Updated transformation tests as well.

Tests passed:

* bgp.session/01-allowas-in.yml
* bgp.multihop/01-global.yml
* bgp.multihop/02-vrf.yml

### DIFF
--- a/netsim/extra/ebgp.multihop/junos.j2
+++ b/netsim/extra/ebgp.multihop/junos.j2
@@ -5,7 +5,7 @@ protocols {
 {% for n in bgp.neighbors if n.type == 'ebgp' and n.multihop is defined %}
 {%   for af in ['ipv4','ipv6'] if n[af] is defined %}
       neighbor {{ n[af] }} {
-        multihop;
+        multihop ttl {{ n.multihop }};
 {%     if n._source_lb_data[af] is defined %}
         local-address {{ n._source_lb_data[af]|ipaddr('address') }}
 {%     endif %}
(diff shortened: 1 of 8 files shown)


netsim/extra/ebgp.multihop/junos.j2
```diff
@@ -0,0 +1,43 @@
+{# global #}
+protocols {
+  bgp {
+    group ebgp-peers {
+{% for n in bgp.neighbors if n.type == 'ebgp' and n.multihop is defined %}
+{%   for af in ['ipv4','ipv6'] if n[af] is defined %}
+      neighbor {{ n[af] }} {
+        multihop;
+{%     if n._source_lb_data[af] is defined %}
+        local-address {{ n._source_lb_data[af]|ipaddr('address') }}
```
jbemmel, line 10: Would suggest to only add ```n._local_address```, not the entire loopback interface

ipspace: Potentially, we need IPv4 and IPv6 addresses, so we could either copy those from loopback data into a new variable, or just add a pointer to the loopback data (which is what you did). Your choice is simpler ;) and one of these days I might clean things up and change all the remaining templates to use just the loopback interface data.

ipspace (approved):

PR #2008, opened 2025-03-06
Cumulus NVUE: VRF local_as works given that the same BGP template is used

jbemmel:
Passes test ```bgp/09-vrf-localas```, couldn't find a documentation checkbox for it?


ipspace:
> Passes test `bgp/09-vrf-localas`, couldn't find a documentation checkbox for it?

There is none. The only reason we have that feature flag is to make sure someone checks both templates ;) I don't think I've seen a device where local-as would work in global routing table but not in a VRF

Issue #2007, opened 2025-03-06
Shut down a Junos BGP neighbor with no active address families


ipspace:
> The only idea I had was to use a quirk to remove BGP neighbors that have no AF active (and generate a configurable warning with report.quirk, so this can be turned off).
> 
> In that case, instead of completely removing it, I would suggest to simply configure it as `shutdown`. So the peer can be used for additional manual configuration, in case someone uses netlab only as a starting point for more complex configs.
> 
> I.e.,
> ```
> admin@r1> show configuration protocols bgp
> group ebgp {
>     type external;
>     neighbor 10.0.0.2 {
>         peer-as 2;
>         shutdown;
>     }
> }
> ``` 

 _Originally posted by @ssasso in [#1994](https://github.com/ipspace/netlab/issues/1994#issuecomment-2703853929)_

ipspace:
> This would be universally applicable, no need to make things work like that only for Junos

Every other device we encountered so far is less stubborn than Junos according to the integration tests. If they don't have an explicitly activated AF, they don't establish a BGP session.

https://tests.netlab.tools/_html/coverage.bgp

I never found the willpower to implement it on NXOS, and it should work on NVUE if you can configure it -- FRR works correctly.

jbemmel:
> > This would be universally applicable, no need to make things work like that only for Junos
> 
> Every other device we encountered so far is less stubborn than Junos according to the integration tests. If they don't have an explicitly activated AF, they don't establish a BGP session.
> 
> https://tests.netlab.tools/_html/coverage.bgp
> 
> I never found the willpower to implement it on NXOS, and it should work on NVUE if you can configure it -- FRR works correctly.

The other devices don't strictly need it, but it wouldn't hurt if ```shutdown``` would appear in their config. It makes explicit that there is no active session with that neighbor

Having neighbors with no active address families is a corner case anyway, so I don't really care either way

ssasso:
I can try to see if I can add a `shutdown` flag directly on the modules (bgp, and the modules using bgp).

Then, every template "owner" can decide to use it or not.

ipspace:
Hey @ssasso, just got a crazy idea. What if:

* You put a neighbor into **shutdown** state if the **activate** dictionary is empty
* Explicitly enable the neighbor in EVPN and MPLS/VPN configuration tasks

[...]

PR #2009, opened 2025-03-07
proposal: global shutdown flag for bgp neigh with no af

ssasso:
Proposal for #2007

Tested topology data result with:

**bgp/20-dual-stack-activate.yml**

```
bgp:
  neighbors:
  - name: x1
    as: 65000
    type: ibgp
    ipv4: 10.0.0.2
    ipv6: 2001:db8:1:2::1
    activate:
      ipv4: false
      ipv6: true
    shutdown:
      ipv4: true
  - name: x2
    as: 65100
    type: ebgp
    ipv4: 10.1.0.6
    ipv6: 2001:db8:3:1::2
    activate:
      ipv4: true
      ipv6: false
    shutdown:
      ipv6: true
  ipv4: true
  ipv6: true
```

**evpn/12-vxlan-ibgp-ebgp.yml**

```
bgp:
  neighbors:
  - rr: true
    name: spine
    as: 65000
    type: ibgp
    ipv4: 10.0.0.5
    activate:
      ipv4: false
    evpn: true
  - ifindex: 0
    local_as: 65201
    name: spine
    as: 65100
    type: ebgp
    ipv4: 10.1.0.2
    activate:
      ipv4: true
  ipv4: true
```

Tested also with:

* mpls/10-vpn-connected.yml
* mpls/20-vpnv6-connected.yml

to check backward compatibility, even if no actual result can be tested.

No tests for BGP-LU and 6PE, but I expect these to work as well.

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -361,6 +361,39 @@ def activate_bgp_default_af(node: Box, activate: Box, topology: Box) -> None:
         ngb.activate[af] = node.bgp.get(af) and af in activate and ngb.type in activate[af]
 
 """
+shutdown_neigh_no_af -- set shutdown flag on neighbors without active AF (per af)
+
+Based on transport session(s) with a BGP neighbor, local BGP AF, and BGP AF configuration
+parameters, set neighbor.shutdown.AF flags.
+To be used to force neigh shutdown on devices that enable the default AF.
+"""
+def shutdown_neigh_no_af(node: Box, topology: Box) -> None:
+  for ngb in node.bgp.neighbors:
+    ngb.shutdown = {}
+    for af in ('ipv4','ipv6'):
+      if af in ngb:
+        if not ngb.activate[af]:
+          ngb.shutdown[af] = True
+    if len(ngb.shutdown) == 0:
+      del ngb.shutdown
+
+"""
+prune_shutdown_flag -- prune the shutdown flag for a specific af in neigh
+
+Remove the shutdown flag for a specific Address Family in a BGP neighbor.
+If the resulting shutdown structure is empy, remove it as well.
+
+This function can be called by external modules which depend on BGP, i.e., EVPN or MPLS
+
+"""
+def prune_shutdown_flag(neigh: Box, af: str) -> None:
+  if af in neigh.get('shutdown', {}):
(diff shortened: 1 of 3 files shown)


jbemmel:
Or alternatively, in the ```bgp/20-dual-stack-activate.yml``` case for X1, we could simply remove the ```ipv4: 10.0.0.2``` from the neighbor

ipspace:
No problems with the "shutdown" flag. We could even make it into an attribute and have it configurable

However:
* The generation of the shutdown flag should happen after all modules (and plugins) had the chance to enable their AFs. I don't like the "make sure you don't shut me down" approach. I added a "cleanup" hook a while back, that might be the right place to do this.
* We need to know which AFs are valid, so either we hard-code that somewhere, or we have an attribute like `bgp.defaults._af_list` and every other module adds its stuff to that.

Finally: what happens for IPv4-AF-over-IPv6 sessions (LLA or otherwise) in case someone decides to disable IPv6 AF on them (people were known to try crazier stuff with netlab 😜)

jbemmel:
> * We need to know which AFs are valid, so either we hard-code that somewhere, or we have an attribute like `bgp.defaults._af_list` and every other module adds its stuff to that.

Isn't that simply all the keys under ```activate```?

And then generate ```_shutdown: True``` if ```activate``` has no ```True``` values for any AF

ipspace:
> Isn't that simply all the keys under `activate`?

How about doing **netlab create** on an EVPN or MPLS/VPN topology and inspecting the BGP neighbors with **netlab inspect --node X bgp** instead of guessing? 😉

jbemmel:
> How can the session be established without knowing the neighbor address?

That's why I cited ```X1``` and not the second example - it has an active AF, so the session should not be ```shutdown```

jbemmel:
> > Isn't that simply all the keys under `activate`?
> 
> How about doing **netlab create** on an EVPN or MPLS/VPN topology and inspecting the BGP neighbors with **netlab inspect --node X bgp** instead of guessing? 😉

What I meant was "Couldn't we make it such that..."

I know EVPN and other AFs don't set the ```activate``` flag today - but that would be relatively straightforward to fix

ipspace:
The original problem (#2007) has been solved in #2061

PR #2010, opened 2025-03-07
BGP: Generalize AF activation

jbemmel:
Context: https://github.com/ipspace/netlab/pull/2009

### DIFF
--- a/netsim/ansible/templates/bgp/arubacx.j2
+++ b/netsim/ansible/templates/bgp/arubacx.j2
@@ -42,7 +42,7 @@ router bgp {{ bgp.as }}
 {{     bgpcfg.bgp_network(af,pfx) }}
 {%   endfor %}
 !
-{%   for n in bgp.neighbors if n[af] is defined and n.activate[af] is defined and n.activate[af] %}
+{%   for n in bgp.neighbors if n[af] is defined and n._activate[af][af]|default(False) %}
 {{     bgpcfg.neighbor_af(n,n[af],bgp) }}
 {%   endfor %}
 {% endfor %}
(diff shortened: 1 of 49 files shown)


ipspace:
This is just throwing spaghetti at the wall. You created duplicate data, and half of that data is not used anywhere.

Now, tell me how you'll use that duplicate data to figure out whether to shut down an IPv4 or an IPv6 BGP session (which is what we started with in the first place)

jbemmel:
> 
> Now, tell me how you'll use that duplicate data to figure out whether to shut down an IPv4 or an IPv6 BGP session (which is what we started with in the first place)

Like this:
```Python
  #
  # Mark neigbors with no active address families as '_shutdown'
  #
  def node_cleanup(self, node: Box, topology: Box) -> None:
    for nb in node.bgp.get('neighbors',[]):
      if 'activate' in nb:
        active_af = [ k for k,v in nb.activate.items() if v ]
        if 'ipv4' in nb:
          if not active_af or ('ipv4' not in active_af and 'ipv6' in active_af):
            nb._shutdown.ipv4 = True
        if 'ipv6' in nb:
          if not active_af or ('ipv6' not in active_af and 'ipv4' in active_af and 'ipv4_rfc8950' not in nb):
            nb._shutdown.ipv6 = True
```

jbemmel:
> Now, tell me how you'll use that duplicate data to figure out whether to shut down an IPv4 or an IPv6 BGP session (which is what we started with in the first place)

Like this:
```Python
#
# Mark neigbors with no active address families as '_shutdown'
#
def node_cleanup(self, node: Box, topology: Box) -> None:
  for nb in node.bgp.get('neighbors',[]):
    for af in ['ipv4','ipv6']:
      if af not in nb and (af=='ipv4' or 'local_if' not in nb):
        nb._activate.pop(af,None)
        continue
      nb._activate[af] = { k:v for k,v in nb._activate[af].items() if v }  # Remove any 'False' flags
      if not nb._activate[af]:
        nb._activate.pop(af,None)
        if nb[af] is not True:
          nb._shutdown[af] = True
```

ipspace:
The original problem (#2007) has been solved in #2061

PR #2012, opened 2025-03-09
FRR: Only configure active LDP address families

jbemmel:
Still passes all MPLS tests - without this, FRR complains about incomplete ipv6 configuration


ipspace (approved):
Seems legit

PR #2017, opened 2025-03-13
Fix: max_length parameter of 'id' type can refer to a constant

### LINKED ISSUE #2015: [BUG] MAX_NODE_ID_LENGTH not applied to group members
In a topology representing an existing device with a 17-character name:
```
defaults.const.MAX_NODE_ID_LENGTH: 18

groups:
 _ auto_create: True
 nodes:
   members: [ device-017-length ]
```

This fails as there is still a 16-character limit being applied to ids in group member lists

ipspace:
Closes #2015

### DIFF
--- a/netsim/data/types.py
+++ b/netsim/data/types.py
@@ -17,6 +17,7 @@ from . import global_vars
 Common error checking routines:
 
 * get_element_path: given prefix and suffix (either one could be missing) return the full path
+* resolve_const_value: when given a string, try to use it as a constant, otherwise return initial value
 * init_wrong_type: initialize the error message caches
 * wrong_type_text: return the type of the data item, tranforming Box into dict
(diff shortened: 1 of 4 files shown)


jbemmel (approved):
Just a thought about
```
node:
  name: str                                         # Validity of node name is checked in the nodes module
```

Could this be ```id``` with length restriction similar to group members? Why the exception?

ipspace:
> Could this be `id` with length restriction similar to group members? Why the exception?

Because `name` is set from the **nodes** key (https://github.com/ipspace/netlab/blob/dev/netsim/augment/nodes.py#L40) and then the keys of the **nodes** dictionary are checked to be IDs with desired max-length in https://github.com/ipspace/netlab/blob/dev/netsim/augment/nodes.py#L81

PR #2023, opened 2025-03-14
Clarify addressing rules on routed VLANs

### LINKED ISSUE #2022: 'unnumbered' pool
https://github.com/ipspace/netlab/blob/bbc11db99743208f04cdcfc6681a9ad315009d27/docs/module/vlan.md?plain=1#L308

The above does not currently work, but could be made to work by defining a default ```unnumbered``` pool (for both ipv4 and ipv6)

Alternatively, the documentation could be corrected

ipspace:
Closes #2022


jbemmel (approved):
clear now

PR #2024, opened 2025-03-14
Add default gateway as a valid next-hop for static routes

### LINKED ISSUE #1678: Implement 'gateway' as a valid next hop of a static route
The 'nexthop.gateway' attribute could be 'true' (take gateway.ipv4 and gateway.ipv6 from the first interface that has it) or the name of a directly-connected link in case you want to be more specific.

ipspace:
Closes #1678

### DIFF
--- a/docs/module/routing.md
+++ b/docs/module/routing.md
@@ -433,11 +433,12 @@ A static route entry's **nexthop** attribute specifies the next hop used to reac
 
 * **ipv4** -- an IPv4 next hop (an address, not a prefix)
 * **ipv6** -- an IPv6 next hop (an address, not a prefix)
+* **gateway** (boolean) -- the next hop is the default gateway. This entry will be resolved into a list containing all directly connected default gateways in the specified next-hop VRF (a node might use more than one default gateway to reach a destination).
 * **node** -- The next hop is the specified node. The node name will be resolved into a list of directly connected next-hops or the control-plane endpoint of a distant node (allowing you to specify recursive static routes).
 
 ```{tip}
 * Static route entries with an **‌ipv4** prefix must have an **‌ipv4** next hop (and likewise for **‌ipv6**). _netlab_ does not support IPv6 next hops for IPv4 routes.
-* _netlab_ will configure several static routes with different next hops if your topology has multiple direct links between the source and the next-hop node.
+* _netlab_ will configure several static routes with different next hops if your topology has multiple direct links between the source and the next-hop node or if a node is attached to multiple links with default gateways.
 * Static routes cannot point to unnumbered IPv4 interfaces or LLA-only IPv6 interfaces.
 ```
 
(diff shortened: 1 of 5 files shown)


netsim/modules/routing.py
```diff
@@ -715,6 +715,37 @@ def process_static_route_includes(
...
+"""
+Create next-hop information from neighbor addresses and static route data
+"""
+def create_nexthop_data(sr_data: Box,ngb_addr: Box, intf: Box) -> Box:
+  nh_data = data.get_empty_box()
+
+  for af in log.AF_LIST:
+    if af not in ngb_addr:
+      continue
+    if isinstance(ngb_addr[af],str):
```
jbemmel, line 727: Shouldn't we check for neighbors being in the same VRF as the next hop entry?

ipspace: VRFs have local significance.

docs/module/routing.md
```diff
@@ -433,11 +433,12 @@ A static route entry's **nexthop** attribute specifies the next hop used to reac
 
 * **ipv4** -- an IPv4 next hop (an address, not a prefix)
 * **ipv6** -- an IPv6 next hop (an address, not a prefix)
+* **gateway** (boolean) -- the next hop is the default gateway. This entry will be resolved into a list containing all directly connected default gateways in the specified next-hop VRF (a node might use more than one default gateway to reach a destination).
 * **node** -- The next hop is the specified node. The node name will be resolved into a list of directly connected next-hops or the control-plane endpoint of a distant node (allowing you to specify recursive static routes).
```
jbemmel, line 437: What happens if the node has multiple addresses, for example through ```gateway```? Which address gets selected - the interface address I would presume?

ipspace: Ah, the precedence. **node** is before **gateway** which is before **ipv4** / **ipv6**. Also, have to add the **_valid_with** check on **gateway** (and wondering why it doesn't work on **node**)

jbemmel, line 437: For a node with ```ipv6: True``` on its interface, would it make sense to have ```node``` resolve to the interface towards that node?

ipspace: Not at the moment. That would require IPv6 proxy-arp on the other node. Also, static routes pointing to interfaces (with no next hops) are a REALLY BAD idea. Here are the impacts https://blog.ipspace.net/2009/10/follow-up-interface-default-route/ and this is how I found it out https://blog.ipspace.net/2009/10/my-stupid-moments-interface-default/

netsim/modules/routing.py
jbemmel, line 727:
```
 nh_data.intf = intf.ifname
 if isinstance(ngb_addr[af],str):
      nh_data[af] = ngb_addr[af]
```
To support interface-only static routes?

jbemmel (approved):

PR #2025, opened 2025-03-14
Fix: Set 'external' role on inter-AS VLANs

### LINKED ISSUE #1859: [BUG] Link role is not set to 'external' on inter-AS IRB VLAN links
When two routers from different AS are connected over an IRB VLAN (using SVI interfaces), the "role: external" attribute is not set on the SVI interface, potentially resulting in an IGP running over the same link.

A simple fix (propagation of 'role' attribute) would probably address the "direct connection" scenario but not the "intermediate L2 switch" scenario.

### Lab topology: direct connection

```
---
defaults.device: eos
provider: clab

vlans:
  core:

groups:
  pe:
    members: [ pe1, pe2 ]
    module: [ ospf, bgp, vlan ]

nodes:
  pe1:
    bgp.as: 65001
  pe2:
    bgp.as: 65002

links:
- pe1:
  pe2:
  vlan.trunk: [ core ]
```

### Output

```
$ netlab inspect --node pe1 interfaces[1]
bridge_group: 1
ifindex: 40000
ifname: Vlan1000
ipv4: 172.16.0.1/24
name: VLAN core (1000) -> [pe2]
neighbors:
- ifname: Vlan1000
  ipv4: 172.16.0.2/24
  node: pe2
ospf:
  area: 0.0.0.0
  network_type: point-to-point
  passive: false
type: svi
virtual_interface: true
vlan:
  mode: irb
  name: core
```

The **role** attribute is not set and there's OSPF running on that link.

ipspace:
Fixes #1859

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -604,6 +604,46 @@ def bgp_transform_community_list(node: Box, topology: Box) -> None:
   for s_type in list(clist.keys()):
     clist[s_type] = data.kw_list_transform(kw_xform,clist[s_type])
 
+"""
+link_ebgp_role_set -- set EBGP role for a regular link
+"""
+def ebgp_role_link(link: Box, topology: Box, EBGP_ROLE: str) -> None:
+  as_set = {}
+  for ifdata in link.get('interfaces',[]):                  # Collect BGP AS numbers from nodes
+    ndata = topology.nodes[ifdata.node]                     # ... connected to the link
+    node_as = ndata.get('bgp.as',None)
+    if node_as:
+      as_set[node_as] = True                                # ... and store them in a dictionary
+
+  if len(as_set) > 1 and not link.get("role"):              # If we have more than two AS numbers per link
+    link.role = EBGP_ROLE                                   # ... set link role unless it's already set
+
+"""
+vlan_ebgp_role_collect -- collect EBGP role information from a VLAN link
+"""
+def vlan_ebgp_role_collect(link: Box, topology: Box) -> None:
+  vlan_name = link.get('vlan_name',None)                    # Get the relevant VLAN name
+  if vlan_name not in topology.vlans:                       # Not a global VLAN, don't waste time
+    return
+
+  for ifdata in link.get('interfaces',[]):                  # Iterate over nodes attached to this VLAN segment
+    ndata = topology.nodes[ifdata.node]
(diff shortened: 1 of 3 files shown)


netsim/modules/bgp.py
```diff
@@ -604,6 +604,46 @@ def bgp_transform_community_list(node: Box, topology: Box) -> None:
...
+"""
+def ebgp_role_link(link: Box, topology: Box, EBGP_ROLE: str) -> None:
+  as_set = {}
+  for ifdata in link.get('interfaces',[]):                  # Collect BGP AS numbers from nodes
+    ndata = topology.nodes[ifdata.node]                     # ... connected to the link
+    node_as = ndata.get('bgp.as',None)
+    if node_as:
+      as_set[node_as] = True                                # ... and store them in a dictionary
+
+  if len(as_set) > 1 and not link.get("role"):              # If we have more than two AS numbers per link
```
jbemmel, line 618: Should check this first, and not bother with collecting AS numbers if it's already set

netsim/modules/bgp.py
```diff
@@ -628,26 +668,23 @@ def node_pre_transform(self, node: Box, topology: Box) -> None:
...
 
-    as_set = {}
-    for ifdata in link.get('interfaces',[]):
-      n = ifdata.node
-      if "bgp" in topology.nodes[n]:
-        node_as = topology.nodes[n].bgp.get("as")
-        if node_as:
-          as_set[node_as] = True
+    has_vlans = 'vlan' in topology.module and 'vlans' in topology
+    for link in topology.links:
```
jbemmel, line 680:
```
if 'role' in link:
  continue
```

jbemmel (changes requested):
Suggest to move the check for links that already have a 'role' set, and avoid the extra processing

netsim/modules/bgp.py
ipspace: Done

jbemmel:
I apologize for the haphazard review - my last suggestion was to move the 'role' check to line 684, instead of the previous one:
```
for link in topology.links:
  if 'role' in link:
    continue
  if not 'vlan_name' in link:
    ebgp_role_link(link,topology,EBGP_ROLE)
  elif has_vlans:
    vlan_ebgp_role_collect(link,topology)
```

That way, we skip both ```ebgp_role_link``` and ```vlan_ebgp_role_collect``` processing

PR #2026, opened 2025-03-14
Cumulus NVUE: Fix untagged vlan on br_default

jbemmel:
```_native_vlan``` is only set for routed interfaces, which aren't connected to the bridge. Need to look at ```vlan.native``` attribute instead

Since the untagged VLAN also gets added as a tagged VLAN, connectivity is not affected

### DIFF
--- a/netsim/ansible/templates/vlan/cumulus_nvue.j2
+++ b/netsim/ansible/templates/vlan/cumulus_nvue.j2
@@ -6,7 +6,7 @@
       domain:
         br_default:
           type: vlan-aware
-{% set native_vlans = interfaces|selectattr('_vlan_native','defined')|map(attribute='_vlan_native')|list %}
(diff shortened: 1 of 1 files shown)


ipspace:
Setting a generic untagged (default?) VLAN for the whole bridge seems fishy to me (usually, that would be VLAN 1), but if it works for you 🤷‍♂️

ipspace:
I can't see any changes in the test results: https://tests.netlab.tools/_html/cumulus_nvue-libvirt-vlan

Time for another attempt?

jbemmel:
> Time for another attempt?

You're right, it wasn't the correct solution. What was there was clearly wrong, but now I'm thinking that the better way to fix it, is to omit configuring ```untagged``` altogether and just leave it as default

PR #2031, opened 2025-03-15
No prefix on routed vlans

jbemmel:
Following the logic in https://github.com/ipspace/netlab/pull/2023 Netlab shouldn't allocate a prefix to routed VLANs, only to each of the links that refers to that VLAN

This PR avoids allocating a (unused) prefix to routed VLANs, and fixes a lingering typo ```topology.node[...]```

Tested: ```NETLAB_DEVICE=frr NETLAB_PROVIDER=clab ./device-module-test -v vlan```

### DIFF
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -100,10 +100,13 @@ def validate_vlan_attributes(obj: Box, topology: Box) -> None:
     if not 'id' in vdata:                                           # When VLAN ID is not defined
       vdata.id = _dataplane.get_next_id('vlan_id')                  # ... take the next free VLAN ID from the list
 
-    if vdata.get('mode',default_fwd_mode or 'irb')=='route':        # Don't assign a prefix to routed VLANs
+    if vdata.get('mode',default_fwd_mode or 'irb') == 'route':      # Don't assign a prefix to routed VLANs
       if 'prefix' in vdata:
-        log.error(f'Routed vlan {vname} in {obj_name} cannot be assigned a prefix ({vdata.prefix})',
-                  log.IncorrectAttr,'vlan',more_hints="use a pool instead")
+        log.error(
+          f'Routed vlan {vname} in {obj_name} cannot be assigned a prefix ({vdata.prefix})',
+          category=log.IncorrectAttr,
+          module='vlan',
+          more_hints="Use the 'pool' VLAN attribute to control VLAN prefix allocation")
       continue
 
     vlan_pool = [ vdata.pool ] if isinstance(vdata.get('pool',None),str) else []
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -105,21 +105,21 @@ def validate_vlan_attributes(obj: Box, topology: Box) -> None:
     We need to assign prefixes to:
     
     * VLANs in bridge/irb forwarding mode
-    * All VLANs if the mixed_fwd_check is disabled
-    * node-only VLANs
+    * All global VLANs if the mixed_fwd_check is disabled
     """
     #
     fwd_mode = vdata.get('mode',default_fwd_mode or 'irb')
     need_pfx = fwd_mode != 'route' or (mixed_fwd_mode and obj is topology)
     if log.debug_active('vlan'):
       print(f'validate_vlan: {obj_path}.{vname} fwd_mode: {fwd_mode} need_pfx: {need_pfx}')
-    if not need_pfx:                                               # Does this VLAN need a prefix?
-      if 'prefix' in vdata and not mixed_fwd_mode:                 # No, having one is an error
+    if not need_pfx:                                                # Does this VLAN need a prefix?
+      if 'prefix' in vdata and not mixed_fwd_mode:                  # No, having one is an error
         log.error(
           f'Routed vlan {vname} in {obj_name} cannot be assigned a prefix ({vdata.prefix})',
           category=log.IncorrectAttr,
           module='vlan',
-          more_hints="Use the 'pool' VLAN attribute to control VLAN prefix allocation")
+          more_hints="Use the 'pool' VLAN attribute to control VLAN prefix allocation, or change the global VLAN to " +
+                     "'irb' and set vlan.mode to 'route' at selected nodes")
       continue
 
     vlan_pool = [ vdata.pool ] if isinstance(vdata.get('pool',None),str) else []
--- a/netsim/modules/vlan.py
+++ b/netsim/modules/vlan.py
@@ -80,27 +80,30 @@ at this point, so most of the changes done on node VLANs would be applicable
 to single-node VLANs.
 """
 def validate_vlan_attributes(obj: Box, topology: Box) -> None:
+
+  if not 'vlans' in obj:                                            # No VLANs? Nothing to do
+    return
+
   obj_name = 'global VLANs' if obj is topology else obj.name
   obj_path = 'vlans' if obj is topology else f'nodes.{obj.name}.vlans'
   default_fwd_mode = obj.get('vlan.mode',None) or get_global_parameter(topology,'vlan.mode')
-
-  if not 'vlans' in obj:
-    return
 
   for vname in list(obj.vlans.keys()):
     if not obj.vlans[vname]:
       obj.vlans[vname] = data.get_empty_box()
 
     vdata = obj.vlans[vname]
-
     if not 'mode' in vdata:                                         # Do we have 'mode' set in the VLAN definition?
       if default_fwd_mode and obj is not topology:                  # Propagate default VLAN forwarding mode if needed
-        vdata.mode = default_fwd_mode
-
+        vdata.mode = default_fwd_mode                               # but only for node level VLANs, to avoid that
+                                                                    # a node level vlan.mode setting gets ignored
     if not 'id' in vdata:                                           # When VLAN ID is not defined
       vdata.id = _dataplane.get_next_id('vlan_id')                  # ... take the next free VLAN ID from the list
 
     if vdata.get('mode',default_fwd_mode or 'irb')=='route':        # Don't assign a prefix to routed VLANs
+      if 'prefix' in vdata:
+        log.error(f'Routed vlan {vname} in {obj_name} cannot be assigned a prefix ({vdata.prefix})',
+                  log.IncorrectAttr,'vlan',more_hints="use a pool instead")
       continue
 
     vlan_pool = [ vdata.pool ] if isinstance(vdata.get('pool',None),str) else []
@@ -1214,7 +1217,6 @@ def create_vlan_access_links(topology: Box) -> None:
       continue                                                              # ... no problem, skip it
     if not 'links' in vdata:                                                # No VLAN links?
       continue                                                              # ... no problem, move on
-
     for cnt,l in enumerate(vdata.links):                                    # So far so good, now iterate over the links
       link_data = links.adjust_link_object(                                 # Create link data from link definition
                     l=l,
@@ -1306,7 +1308,10 @@ class VLAN(_Module):
 
         # Set node-level VLAN forwarding mode and merge topology VLAN data into node VLAN data
         set_node_vlan_mode(node.vlans[vname],node,topology.vlans[vname],topology)
+
         node.vlans[vname] = topology.vlans[vname] + node.vlans[vname]
+        if node.vlans[vname].mode == 'route':
+          node.vlans[vname].pop('prefix',None)              # Remove any prefix for routed VLANs
       else:                                                 # Single node VLAN, set forwarding mode
         set_node_vlan_mode(node.vlans[vname],node,node.vlans[vname],topology)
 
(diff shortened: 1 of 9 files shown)


ipspace:
> Following the logic in #2023 Netlab shouldn't allocate a prefix to routed VLANs, only to each of the links that refers to that VLAN

This feels like one of those "rearranging deck chairs on Titanic" exercises, but I agree that prefix is unnecessary.

> This PR avoids allocating a (unused) prefix to routed VLANs, and fixes a lingering typo `topology.node[...]`

Yeah, at least that typo has to be fixed. One would guess we have no test that would catch that one, do we ;)

> Tested: `NETLAB_DEVICE=frr NETLAB_PROVIDER=clab ./device-module-test -v vlan`

That's commendable, but it only proves that the typical use cases we could think of work as expected. Also, for VLANs I would run FRR (router) and EOS (l3-switch) tests.

I will do some more test tomorrow before merging this.

netsim/modules/vlan.py
```diff
@@ -100,12 +100,13 @@ def validate_vlan_attributes(obj: Box, topology: Box) -> None:
     if not 'id' in vdata:                                           # When VLAN ID is not defined
       vdata.id = _dataplane.get_next_id('vlan_id')                  # ... take the next free VLAN ID from the list
 
-    vlan_pool = [ vdata.pool ] if isinstance(vdata.get('pool',None),str) else []
-    vlan_pool.extend(['vlan','lan'])
-    pfx_list = links.assign_link_prefix(vdata,vlan_pool,topology.pools,topology.nodes,f'{obj_path}.{vname}')
-    vdata.prefix = addressing.rebuild_prefix(pfx_list)
-    if not 'allocation' in vdata.prefix:
-      vdata.prefix.allocation = 'id_based'
+    if vdata.get('mode','irb')!='route':                            # Don't assign a prefix to routed VLANs
```
ipspace, line 103: This might work, but is conceptually wrong.

Either we KNOW the **mode** is set by this time (so there's no need for `vdata.get`, and if that's the case it would be nice to have a comment saying that for our future visits) or we MUST use the standard way to get the VLAN mode (the default might be set in topology).

The way I'm reading the code just above this, the same function sets the VLAN mode only for node VLANs.

Finally, as this is the last thing in the function, it would be nice to use the "left-aligned happy path" approach (= exit if mode = route)

I can make these changes when running the tests tomorrow.

ipspace: > This might work, but is conceptually wrong.

It doesn't even work. The VLAN gets a prefix in the following topology

```
---
provider: clab
defaults.device: frr
module: [ vlan ]

vlan.mode: route
vlans:
  red:

nodes: [ r1,r2 ]
links:
- interfaces: [ r1,r2 ]
  vlan.trunk: [ red ]
```

jbemmel: > Either we KNOW the **mode** is set by this time

From looking at the code, ```validate_vlan_attributes``` gets called for the topology first, and then for each node. We refrained from setting vlan.mode on a global vlan because otherwise a node level vlan.mode setting would get ignored

Have updated the code to:
```if vdata.get('mode',default_fwd_mode or 'irb')!='route':```

**Comment by jbemmel** [comment]
Added an error message and cleanup code for node vlans that override ```vlan.mode``` for a global VLAN that has a prefix assigned. 

Note that it doesn't catch this scenario:
```
---
module: [ vlan ]

vlans:
  v1:
    mode: 'irb'            # default
    prefix: 192.168.0.0/24
    links: [ n1-n2 ]

nodes:
  n1:
    vlan.mode: route
  n2:
    vlan.mode: route
```
In this case, n1 gets ```10.1.0.1/30``` and n2 gets ```10.1.0.2/30```

If one of the nodes uses ```vlan.mode: irb```, they get assigned IPs from the correct prefix

File: netsim/modules/vlan.py [hunk 2]
```diff
@@ -650,7 +656,7 @@ def create_vlan_links(link: Box, v_attr: Box, topology: Box) -> None:
         link_data.vlan.mode = 'route'
         for intf in link_data.interfaces:
           intf.vlan.mode = 'route'
-      else:
+      elif prefix is not None:
```
ipspace, line 659: Not sure we ever got here in the tests, but this effectively blocks using empty "prefix:" on a VLAN. Changing to a slightly safer approach.

ipspace:
> Note that the following scenario is counter intuitive:

Lots of things are counter-intuitive with VLANs ;)

> In this case, n1 gets `10.1.0.1/30` and n2 gets `10.1.0.2/30`

That is correct, because the segment terminates at N1/N2, so it needs a different prefix.

> If one of the nodes uses `vlan.mode: irb`, they get assigned IPs from the correct prefix.

Also correct, because the segment could be stretched beyond one of the devices.

ipspace (changes requested):
Show stopper: you effectively broke the "downgrade routed vlan to IRB/bridge on a node" scenario (test case rt-vlan-mode-link-route.yml). The two bridged segments (h-sw and h-br) must have the same prefix, while the fully routed segments must have different prefixes.

It looks like you decided some eggs must be broken, so you recreated the test case results and moved on, but we do have test cases for a reason ;)

We could decide to:

A. leave things as they are
B. implement the "no downgrading from routed VLANs" rule, which would be a breaking change, and require additional testing
C. Allocate a prefix to a routed VLAN if we find some nodes connected to it that use mode IRB or Bridge.

It seems to me like you started this quest because you wanted to generate the "no prefix on routed VLANs" error message. I'm not sure that's worth the effort.

On a more positive note, I do have an idea that might save the day.

ipspace:
Found something. Not what I thought, but might be good enough.

* _netlab_ rejects a mix of bridged/routed VLANs unless the `defaults.vlan.warnings.mixed_fwd_check` flag is set to False.
* The new code is probably safe under the default conditions.

We could either:

A) Enforce the 'no mixed fwd' rule
B) Allocate prefixes to routed VLANs if that flag is set to False.

(A) is a breaking change, (B) complicates the code a bit and has to be well-documented in the code so we'll know what we're doing a few months down the road. I would still prefer going with (B) considering how complex some people want their VLANs to be ;)

ipspace (approved):
I think this is now good to go. It could be cleaner, but not if we want to keep backward compatibility 🤷‍♂️

netsim/modules/vlan.py
```diff
@@ -80,26 +80,48 @@ def validate_global_node_match(vname: str, node: Box, topology: Box) -> bool:
...
+    """
+    We need to assign prefixes to:
+    
+    * VLANs in bridge/irb forwarding mode
+    * All VLANs if the mixed_fwd_check is disabled
+    * node-only VLANs
+    """
+    #
+    fwd_mode = vdata.get('mode',default_fwd_mode or 'irb')
+    need_pfx = fwd_mode != 'route' or (mixed_fwd_mode and obj is topology)
```
jbemmel, line 113: The code doesn't match the text - a 1:1 translation seems to be
```
need_pfx = fwd_mode != 'route' or mixed_fwd_mode or (obj is not topology)
```

netsim/modules/vlan.py
```diff
@@ -1303,7 +1324,10 @@ def node_pre_transform(self, node: Box, topology: Box) -> None:
 
         # Set node-level VLAN forwarding mode and merge topology VLAN data into node VLAN data
         set_node_vlan_mode(node.vlans[vname],node,topology.vlans[vname],topology)
+
         node.vlans[vname] = topology.vlans[vname] + node.vlans[vname]
+        if node.vlans[vname].mode == 'route':
```
jbemmel, line 1329: With the updated logic, if the intent is to keep assigning prefixes to node-only vlans then popping the prefix should only be done if this is a global vlan and the mixed_fwd_mode flag is not set

netsim/modules/vlan.py
```diff
@@ -80,26 +80,48 @@ def validate_global_node_match(vname: str, node: Box, topology: Box) -> bool:
...
+  if not 'vlans' in obj:                                            # No VLANs? Nothing to do
+    return
+
   obj_name = 'global VLANs' if obj is topology else obj.name
   obj_path = 'vlans' if obj is topology else f'nodes.{obj.name}.vlans'
   default_fwd_mode = obj.get('vlan.mode',None) or get_global_parameter(topology,'vlan.mode')
-
-  if not 'vlans' in obj:
-    return
+  mixed_fwd_mode = topology.get('defaults.vlan.warnings.mixed_fwd_check',None) is False
```
jbemmel, line 90: The name ```vlan.warnings.mixed_fwd_check``` sounds like something that would disable a warning, but not affect anything else (i.e. changing the way the vlan module allocates prefixes)

Perhaps something like ```vlan.enable_mixed_fwd``` or ```vlan.allocate_routed_prefixes``` would be clearer?

jbemmel:
> Show stopper: you effectively broke the "downgrade routed vlan to IRB/bridge on a node" scenario (test case rt-vlan-mode-link-route.yml). The two bridged segments (h-sw and h-br) must have the same prefix, while the fully routed segments must have different prefixes.

I think you're talking about the ```red``` vlan, which is globally routed but downgraded to ```bridge``` at the ```br(idge)``` and ```sw(itch)``` nodes?

If so, couldn't we require that such a scenario should be modeled with a global vlan in ```irb``` mode, downgraded to pure ```route``` mode or ```bridge``` mode at selected nodes if needed?

That way, we can still keep the simple "no prefix on routed vlans" rule while still enabling all scenarios

netsim/modules/vlan.py
ipspace: > The name `vlan.warnings.mixed_fwd_check` sounds like something that would disable a warning, but not affect anything else (i.e. changing the way the vlan module allocates prefixes)

The error message says "to disable this check" or some such.

> Perhaps something like `vlan.enable_mixed_fwd` or `vlan.allocate_routed_prefixes` would be clearer?

It might be, but I firmly believe one should treat settings (or other contracts with your customers) like AWS, not like Google or some networking vendors who feel changing API without versioning it is perfectly fine. Once we ship it (and we did ship this one), we stick with it unless there's a good reason not to.

This should be invisible to everyone but the poor souls who think combining routed VLANs with IRB nodes attached to them makes sense.

ipspace:
> I think you're talking about the `red` vlan, which is globally routed but downgraded to `bridge` at the `br(idge)` and `sw(itch)` nodes?

Yes.

> If so, couldn't we require that such a scenario should be modeled with a global vlan in `irb` mode, downgraded to pure `route` mode or `bridge` mode at selected nodes if needed?

That's the default behavior, but for whatever reason (honestly can't remember what it was) we kept the "warning" switch to turn it off.

> That way, we can still keep the simple "no prefix on routed vlans" rule while still enabling all scenarios

That is the default behavior.

netsim/modules/vlan.py
ipspace: Nope, this is OK (although maybe not necessary). We don't need an IRB prefix if the node operates in routed VLAN mode. For global VLANs, the IRB prefix is taken from the global VLAN data.

netsim/modules/vlan.py
jbemmel: > This should be invisible to everyone but the poor souls who think combining routed VLANs with IRB nodes attached to them makes sense.

Think of it as IRB VLANs with routed nodes attached instead?

jbemmel: > It might be, but I firmly believe one should treat settings (or other contracts with your customers) like AWS, not like Google or some networking vendors who feel changing API without versioning it is perfectly fine. Once we ship it (and we did ship this one), we stick with it unless there's a good reason not to.

I may be missing something, but aren't we introducing this 'vlan.warnings.mixed_fwd_check' here? It's not something we have shipped. I am talking about renaming the flag, not removing it

jbemmel:
I realize we're approaching the limits of what one might consider "reasonable effort" spent on issues like this, but I'd like to make one last argument in favor of deprecating the assignment of prefixes to routed vlans altogether, and instead suggest to use a global ```irb``` vlan with ```vlan.mode='route'``` override at nodes

netsim/modules/vlan.py
ipspace: > I may be missing something, but aren't we introducing this 'vlan.warnings.mixed_fwd_check' here? It's not something we have shipped. I am talking about renaming the flag, not removing it

1. Search the code for `mixed_fwd_check` (I found it when I tried to figure out why the transformation tests work while my simple topology is throwing errors). You'll find it mentioned in four different places.
2. Open https://github.com/ipspace/netlab/blob/2aa83191c9662ade154489c9ba5bc269971c8768/netsim/modules/vlan.py#L318
3. Use the "blame" functionality to find the relevant commit
4. Open the commit (https://github.com/ipspace/netlab/commit/f36289e04eec6d581e5937bc0d5c80fc008d5fbb). Somewhere in the header, you'll notice the last tag without it was 1.9.2-post1
5. FWIW, this functionality was introduced to fix https://github.com/ipspace/netlab/issues/1461 which was opened based on a discussion in #1457 which was opened by... ;)

ipspace:
> I realize we're approaching the limits of what one might consider "reasonable effort" spent on issues like this

We've been past those limits months ago, but some people keep exploring the rabbit trails 🤷‍♂️

> but I'd like to make one last argument in favor of deprecating the assignment of prefixes to routed vlans altogether, and instead suggest to use a global `irb` vlan with `vlan.mode='route'` override at nodes

That's the default behavior, and you have to make REAL efforts to change it.

netsim/modules/vlan.py
jbemmel: Ah, I missed that. ok, PR closed

PR #2033, opened 2025-03-17
Rewrite the default gateway processing

ipspace:
* Use the 'gateway' next hop when creating host default routes
* Create host default routes for IPv6 if the host does not use RA
* Add 'create GW of last resort' functionality to the static route processing
* Remove the 'add link/vlan gateway' code as this is now handled by the static route processing

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -454,34 +454,21 @@ def default_static_route_list(topology: Box) -> list:
     if ap_name in ['mgmt','router_id']:
       continue
 
-    if not isinstance(ap_data.get('ipv4',False),str):
-      continue
-
-    sr_list.append(data.get_box({ 'ipv4': ap_data.ipv4 }))
+    for af in ['ipv4','ipv6']:
+      if not isinstance(ap_data.get(af,False),str):
+        continue
+      sr_list.append(data.get_box({ af: ap_data[af], '_skip_missing': True, 'nexthop.gateway': True }))
 
   return sr_list
 
 '''
-Given the global list of static routes, create a host-specific list that
-contains the host default gateway
-'''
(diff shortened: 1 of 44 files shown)


netsim/augment/nodes.py
```diff
@@ -454,30 +454,13 @@ def default_static_route_list(topology: Box) -> list:
...
-  next_hop['intf'] = intf.ifname
-  next_hop['idx'] = 0
-
-  for sr_entry in sr_list:
-    host_list.append(sr_entry + { 'nexthop': next_hop })
-
-  return host_list
-
 '''
 For all hosts, create 'routing.static' table (unless it exists) that contains IPv4
```
jbemmel, line 465: IPv4 and/or IPv6

netsim/devices/linux.yml
```diff
@@ -15,6 +15,8 @@ features:
     passive: False
   routing:
     static: true
+  initial:
+    ipv6_use_ra: true
```
jbemmel, line 19: Also need to set this for FRR (at least); used as host in scenarios where VLANs are involved

netsim/utils/log.py
```diff
@@ -315,9 +315,13 @@ def warning(*,
   if flag_value is False:
     return
 
+  if 'category' not in kwargs:
```
jbemmel, line 318: This change seems unrelated to gateway processing... ;)

jbemmel (approved):
Minor comments

netsim/utils/log.py
ipspace: OK, I can make it into a separate commit ;)

netsim/devices/linux.yml
ipspace: Need to check how those devices behave first.

PR #2043, opened 2025-03-18
Skip creation of host static routes for DHCP clients

### LINKED ISSUE #2035: [BUG] DHCP clients do not need IPv4 static routes
The new approach to host routes (generate static routes in nodes.py) does not check whether a host is a DHCP client. DHCP clients do not need IPv4 static routes, they get the default GW from the DHCP server.

ipspace:
Fixes #2035

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -486,6 +486,19 @@ def add_host_static_routes(topology: Box) -> None:
     
     features = devices.get_device_features(n_data,topology.defaults)
     host_af_list = [ af for af in n_data.af if af != 'ipv6' or not features.initial.ipv6_use_ra ]
+    if 'dhcp' in topology.get('module',[]):       # If we use DHCP module, we have to scan for the DHCP clients
+      for af in list(host_af_list):          
+        if n_data.get('dhcp.server',False):       # ... unfortunately before the DHCP module post-transform cleanup
+          continue                                # ... so we have to skip known DHCP servers
+        for intf in n_data.interfaces:            
+          if intf.get(f'dhcp.server',None):       # ... as well as DHCP relays
+            continue
+          if intf.get(f'dhcp.client.{af}',None):  # Now check for the DHCP clients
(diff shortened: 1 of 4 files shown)


netsim/augment/nodes.py
```diff
@@ -486,6 +486,19 @@ def add_host_static_routes(topology: Box) -> None:
     
     features = devices.get_device_features(n_data,topology.defaults)
     host_af_list = [ af for af in n_data.af if af != 'ipv6' or not features.initial.ipv6_use_ra ]
+    if 'dhcp' in topology.get('module',[]):       # If we use DHCP module, we have to scan for the DHCP clients
```
jbemmel, line 489: Would it be possible to structure this such that the DHCP module performs this work? It breaks modularity

e.g. have the routing module create its static routes, but have the DHCP module remove them for clients

jbemmel (approved):
It should work, added a suggestion about a more modular structure

ipspace:
(Adding a top-level comment so this doesn't get hidden in the code comments)

> Would it be possible to structure this such that the DHCP module performs this work? It breaks modularity

The static route next hops are processed in the "routing" module post_transform hook, and the DHCP flags are cleaned up in the "dhcp" module post_transform hook. The only way to get in between them is to have another module that would be called between the two.

> e.g. have the routing module create its static routes, but have the DHCP module remove them for clients

These static routes do not belong to the routing module. Someone else is using the functionality from the routing module to implement host routing.

Also, DHCP cannot remove static routes because it does not know who added them and why. We could have a flag on the static routes saying "do not use with DHCP", but then the DHCP module would have to do all sorts of convoluted cleanups (for example, removing the "routing" module).

It might be possible to move the DHCP post_transform processing to post_link_transform phase, so the **nodes** code would get an already-cleaned-up DHCP state. If not, I'll merge this as-is.

netsim/augment/nodes.py
ipspace: See https://github.com/ipspace/netlab/pull/2043#issuecomment-2733652427

Issue #2051, opened 2025-03-20
[BUG] Plugin search paths


jbemmel:
When I run the following command in directory ```tests/integration``` it fails:
```
netlab up routing/07a-community-append.yml -d dellos10 -v
```
```
TASK [fail] **********************************************************************************************************************************************************************************
task path: /home/jeroen/Projects/netlab/netsim/ansible/tasks/deploy-custom-config.yml:31
redirecting (type: connection) ansible.builtin.network_cli to ansible.netcommon.network_cli
redirecting (type: connection) ansible.builtin.docker to community.docker.docker
fatal: [x1]: FAILED! => changed=false 
  msg: Cannot find template for device frr in custom config directory /home/jeroen/Projects/netlab/tests/integration/bgp.policy
redirecting (type: terminal) ansible.builtin.dellos10 to dellemc.os10.os10
redirecting (type: cliconf) ansible.builtin.dellos10 to dellemc.os10.os10
fatal: [dut]: FAILED! => changed=false 
  msg: Cannot find template for device dellos10 in custom config directory /home/jeroen/Projects/netlab/tests/integration/bgp.policy
```

It only works when I run it from ```tests```:
```
netlab up integration/routing/07a-community-append.yml -d dellos10 -v
```

The reason seems to be that a 'bgp.policy' directory exists under ```tests/integration```, which hides the plugin directory

PR #2052, opened 2025-03-20
Treat custom config requests with .j2 suffix differently

### LINKED ISSUE #2015: [BUG] MAX_NODE_ID_LENGTH not applied to group members

ipspace:
The file search path for custom config templates included the bare filename because the config parameter could include a .j2 suffix.

That can result in a matching directory where we really want to find the .j2 template within the directory (#2051 provides an excellent example).

This modification removes the bare config filename from the search path and modifies the search logic -- if the config filename includes the '.j2' suffix, then we're searching only for that filename, otherwise we're searching for all potential combinations of things.

Fixes #2015

### DIFF
--- a/netsim/ansible/tasks/deploy-custom-config.yml
+++ b/netsim/ansible/tasks/deploy-custom-config.yml
@@ -16,7 +16,7 @@
     node_provider: "{{ provider|default(netlab_provider) }}"
     params:
       paths: "{{ paths_custom.dirs }}"
-      files: "{{ paths_custom.files }}"
+      files: "{{ [ custom_config ] if custom_config.endswith('.j2') else paths_custom.files }}"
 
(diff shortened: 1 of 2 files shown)


jbemmel (approved):
May want to update https://github.com/ipspace/netlab/blob/0f97276618d9b2406847b83cc74454afb1f61d79/docs/custom-config-templates.md?plain=1#L27

jbemmel:
> Fixes #2015

#2051

PR #2053, opened 2025-03-20
Dell OS10 : Add routing support

jbemmel:
This PR adds support for most of the (supported) features of the ```routing``` module

* Reduce default BGP advertisement interval from 30s to 1s (required to pass integration tests involving communities)
* Set BGP timers to keep-alive=3, hold-time=9 (like frr)

Tested:
* ```NETLAB_DEVICE=dellos10 NETLAB_PROVIDER=libvirt ./device-module-test -v routing```
* ```NETLAB_DEVICE=dellos10 NETLAB_PROVIDER=libvirt ./device-module-test -v vrf```
* ```NETLAB_DEVICE=dellos10 NETLAB_PROVIDER=libvirt ./device-module-test -v bgp```

### DIFF
--- a/netsim/ansible/templates/bgp/dellos10.macro.j2
+++ b/netsim/ansible/templates/bgp/dellos10.macro.j2
@@ -66,7 +66,11 @@
  exit
 {%   endif %}
 {% endif %}
-  
+
+{# Override default timers of 60/180 #}
+  timers 3 9
+{# Reduce default advertisement interval of 30s to minimum 1s #}
+  advertisement-interval 1
   no shutdown
   exit
 {%- endmacro %}
--- a/netsim/extra/bgp.policy/dellos10.j2
+++ b/netsim/extra/bgp.policy/dellos10.j2
@@ -45,3 +45,6 @@ router bgp {{ bgp.as }} vrf {{ vname }}
 {%   endfor %}
 {% endif %}
 !
+
+! Clear BGP sessions to make policy changes take effect
+do clear ip bgp *
--- a/netsim/extra/bgp.policy/dellos10.j2
+++ b/netsim/extra/bgp.policy/dellos10.j2
@@ -1,0 +1,47 @@
+{#
+  Apply route map to a BGP neighbor
+#}
+{% macro apply_route_map(n,af,rm_per_af=False) -%}
+{%   if 'policy' in n %}
+  address-family {{ af }} unicast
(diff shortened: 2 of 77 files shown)


ipspace:
> Note that some community related tests work when executed individually, but fail in the `device-module-test`; I tried working around these seemingly timing related issues, but to no avail

Probably irrelevant, but some platforms don't resend updates if you change the route maps (or community lists). It might be worth clearing the BGP sessions after the config is done (some other platforms have to do that as well)

netsim/extra/bgp.policy/dellos10.j2
```diff
@@ -0,0 +1,50 @@
...
+{%         endif %}
+{{         ebgp_neighbor(n,af) }}
+{%       endfor %}
+{%     endfor %}
+{%   endfor %}
+{% endif %}
+!
+
+! Clear BGP sessions to make policy changes take effect
+do clear ip bgp *
```
jbemmel, line 50: Yes, I saw that for Cisco - it did fix some use cases for Dell OS10, but not the communities

jbemmel: Note that the community tests that fail, involve a community that gets added by the ```probe``` node, not ```DUT``` itself (iirc, haven't looked too deeply)

Will require further testing, just didn't want to let it block merging the 99% of working cases

ipspace:
You can always add device/test specific wait intervals to the integration tests and include those changes in the PR.

Just search the `tests/integration` directory tree for vptx or vjunos-router if you need an example ;)

netsim/ansible/templates/bgp/dellos10.macro.j2
```diff
@@ -66,7 +66,11 @@
  exit
 {%   endif %}
 {% endif %}
-  
+
+{# Override default timers of 60/180 #}
+  timers 3 9
```
ipspace, line 71: This was unnecessary, as the BGP timers impact only the session failure detection logic.

ipspace (commented):
Looks good to me, running integration tests

ipspace (approved):
Routing integration tests passed, merging.

netsim/ansible/templates/bgp/dellos10.macro.j2
jbemmel: Yes, and they are negotiated so when peering with FRR things work fine. However, between Dell OS10 devices they would use defaults less suited to data center environments

PR #2054, opened 2025-03-20
Add restart-policy to allowed clab attributes

alistairking:
Without this, we get an error when trying to use `clab.restart-policy`:
```
[ATTR]    groups: Incorrect node_group attribute 'restart-policy' in topology.groups.kagents.clab
```

### DIFF
--- a/netsim/providers/clab.yml
+++ b/netsim/providers/clab.yml
@@ -44,6 +44,7 @@ attributes:
       license: str
       runtime: str
       startup-delay: int
+      restart-policy: str
   interface:
     name: str
   link:


ipspace (approved):
I'm wondering where you found use for "restart-policy" (and would love to hear about it), but have no problems merging this.

Great job and welcome to the contributors club ;))

alistairking:
Hi! 
Yes sorry, I should have started with:
Thanks for this truly awesome piece of software. I've only been using it for 4 days now, but I'm in love.

As for how I'm using restart-policy... I have some linux containers running a Kentik telemetry agent that restarts as part of its normal authorization lifecycle. It initially starts unauthenticated, and then once it is authorized (via an out-of-band mechanism), it cleanly shuts down, expecting to be restarted.

ipspace:
> Hi! Yes sorry, I should have started with: Thanks for this truly awesome piece of software. I've only been using it for 4 days now, but I'm in love.

Thanks a million ;)

> As for how I'm using restart-policy... I have some linux containers running a Kentik telemetry agent that restarts as part of its normal authorization lifecycle. It initially starts unauthenticated, and then once it is authorized (via an out-of-band mechanism), it cleanly shuts down, expecting to be restarted.

Keep in mind that restarting a container loses all its links as **containerlab** handcrafts the vEth pairs. The only interface that is not impacted by the restart is the management interface.

Also, your PR just made it into the 1.9.6 release, so you can use the regular Python package.

PR #2056, opened 2025-03-21
Dell OS10: Add support for originated default routes in OSPF

jbemmel:
* Tweak OSPF timers to avoid instabilities in FRR neighbors during ```ospf/ospfv2/20-default.yml```
* Add quirk about limited attribute support

Testing: ```NETLAB_DEVICE=dellos10 NETLAB_PROVIDER=libvirt ./device-module-test -v ospf/ospfv2```

+ Test ```ospf/ospfv2/20-default.yml``` now passes with warnings (no support for cost on originated default route)
+ Test ```ospf/ospfv2/22-default-vrf.yml``` passes


ipspace (changes requested):
Running integration tests. In the meantime, you might want to add the new features to platform support tables (module/routing.md and module/ospf.md)

jbemmel:
> you might want to add the new features to platform support tables (module/routing.md and module/ospf.md)

Done. Re: "delete" for communities, Dell does not support targeted delete but it can remove all communities. Not sure if that's covered by current policies?

PR #2057, opened 2025-03-21
Allow custom bash scripts to run against cEOS containers

alistairking:
Hey, me again. 

I was trying to get sflow export working from cEOS and after much pain, realized that it doesn't support packet sampling. 

As a workaround, I figured I'd install hsflowd directly in the cEOS container, but then found I couldn't run custom bash scripts like I can for FRR containers.

I'm pretty new to ansible (and ultra new to netlab), so maybe there's a better way to to do this?
I'm open to suggestions.

### DIFF
--- a/netsim/ansible/tasks/deploy-config/eos.yml
+++ b/netsim/ansible/tasks/deploy-config/eos.yml
@@ -1,4 +1,24 @@
+- set_fact: deployed_config={{ lookup('template',config_template) }}
+- set_fact: bash_script={{ "#!/bin/bash" in deployed_config }}
+
+- template:
+    src: "{{ config_template }}"
+    dest: /tmp/config.sh
+  when: bash_script
+  vars:
+    ansible_user: root
+    ansible_connection: docker
+
+- name: "run /tmp/config.sh to deploy {{netsim_action}} config from {{ config_template }}"
+  command: bash /tmp/config.sh
+  when: not ansible_check_mode and bash_script
+  become: true
+  vars:
+    ansible_user: root
+    ansible_connection: docker
+
 - name: "eos_config: deploying {{ netsim_action }} from {{ config_template }}"
   eos_config:
     src: "{{ config_template }}"
+  when: not ansible_check_mode and not bash_script
   tags: [ print_action, always ]


ipspace:
Implementing this as a generic solution makes little sense. While I'm positive we could get something to work on vEOS VMs, this only works on cEOS containers. Also, the only place where this would be useful (at the moment) would be the custom config scripts.

I think it's much better to use a **eos-clab.yml** task list for a specific custom config template. Let me figure out the scaffolding that you could then use.

I also added a few nits you might find useful as code comments.

netsim/ansible/tasks/deploy-config/eos.yml
```diff
@@ -1,4 +1,24 @@
...
+    ansible_user: root
+    ansible_connection: docker
+
+- name: "run /tmp/config.sh to deploy {{netsim_action}} config from {{ config_template }}"
+  command: bash /tmp/config.sh
+  when: not ansible_check_mode and bash_script
+  become: true
+  vars:
+    ansible_user: root
+    ansible_connection: docker
```
ipspace, line 18: Because you're relying on Docker connection, it would be best to have this task list as "eos-clab.yml", so it's not used for a vEOS VM.

netsim/ansible/tasks/deploy-config/eos.yml
```diff
@@ -1,4 +1,24 @@
+- set_fact: deployed_config={{ lookup('template',config_template) }}
+- set_fact: bash_script={{ "#!/bin/bash" in deployed_config }}
+
+- template:
+    src: "{{ config_template }}"
+    dest: /tmp/config.sh
+  when: bash_script
```
ipspace, line 7: You could group multiple tasks with the same **when** condition into a block. Never tried what happens when a task inside a **block** has another **when** condition though (and it might be Ansible release dependent)

netsim/ansible/tasks/deploy-config/eos.yml
```diff
@@ -1,4 +1,24 @@
...
+  when: not ansible_check_mode and bash_script
+  become: true
+  vars:
+    ansible_user: root
+    ansible_connection: docker
+
 - name: "eos_config: deploying {{ netsim_action }} from {{ config_template }}"
   eos_config:
     src: "{{ config_template }}"
+  when: not ansible_check_mode and not bash_script
```
ipspace, line 23: Modules like **eos_config** use **ansible_check_mode** internally to deploy the config or just generate the diff. Using it in a **when** condition would effectively disable the dry run functionality.

netsim/ansible/tasks/deploy-config/eos.yml
ipspace: ... also, you might need tags on this task to display it within **netlab initial -q** (or whatever the "don't display all the clutter" flag is)

ipspace:
Here's a working example for Arista EOS containers and VMs:

https://github.com/ipspace/netlab-examples/tree/master/config/eos-bash

Thanks again for a wonderful challenge and a really good container solution.

alistairking:
Ah! That's a nifty solution. 

I hadn't thought of running custom Ansible tasks.

PR #2060, opened 2025-03-23
Add the 'activate' dictionary to VRF EBGP neighbors

ipspace:
This makes the data structures more consistent between VRF and global BGP neighbors and makes it easier to write the device configuration templates as we no longer have to check whether the 'activate' dictionary is defined.

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -225,6 +225,7 @@ def build_ebgp_sessions(node: Box, sessions: Box, topology: Box) -> None:
     if not af_list:                                                   # This interface has no usable address
       continue
 
+    intf_vrf = l.get('vrf',None)
     if node_as != node_local_as:
       if not features.bgp.local_as:
         log.error(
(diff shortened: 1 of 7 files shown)


netsim/modules/bgp.py
```diff
@@ -339,13 +340,15 @@ def build_ebgp_sessions(node: Box, sessions: Box, topology: Box) -> None:
...
-              node.vrfs[l.vrf].bgp.neighbors = []
-            node.vrfs[l.vrf].bgp.neighbors.append(ebgp_data)
-        else:                 # Global neighbor
-          node.bgp.neighbors.append(ebgp_data)
+        if intf_vrf :                                 # VRF neighbor
+          if node.vrfs[l.vrf].bgp is False:           # Is BGP disabled for this VRF?
+            continue                                  # ... yeah, move on
+          for af in log.AF_LIST:                      # Add the 'activate' dictionary to make neighbor data consistent
+            if af in ebgp_data:                       # ... as this parameter is not user-controllable
+              ebgp_data.activate[af] = True           # ... enable all default AFs present on the neighbor
```
jbemmel, line 348: What about IBGP neighbors (```build_ibgp_sessions```)?

tests/topology/expected/bgp-vrf-local-as.yml
```diff
@@ -234,7 +234,9 @@ nodes:
             connected:
               auto: true
           neighbors:
-          - as: 65000
+          - activate:
+              ipv4: true
```
jbemmel, line 238: None of the test cases has ```ipv6: true```

jbemmel (changes requested):
* Test coverage for ipv6 activation of VRF neighbors

tests/topology/expected/bgp-vrf-local-as.yml
ipspace: Fair point. Added.

PR #2073, opened 2025-03-24
Dell OS10: fix BGP router-id in vrf

### LINKED ISSUE #2069: [BUG] Dell OS10 VRF configuration template does not set per-VRF BGP router ID
13-multi-vrf-bgp-loopback:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/13-multi-vrf-bgp-loopback.yml)
* [Test results](https://tests.netlab.tools/dellos10/libvirt/vrf/13-multi-vrf-bgp-loopback.yml-validate.log)

ssasso:
Fix #2069


ipspace (approved):
If it worked with **bgp.router_id** before, it should work now as well ;)

PR #2066, opened 2025-03-23
SR Linux: Fix bgp af activation

jbemmel:
* Move EVPN activation to EVPN module
* Rewrite AF activation logic; disable neighbor when no address families are activated

Fixes https://github.com/ipspace/netlab/issues/2059

Tested:
* ```NETLAB_DEVICE=srlinux NETLAB_PROVIDER=clab ./device-module-test -v bgp```
* ```NETLAB_DEVICE=srlinux NETLAB_PROVIDER=clab ./device-module-test -v evpn``` (with latest EVPN tests)

Not all EVPN tests pass, but the ones that used to work continue to work

Note: Test case 03 would require https://documentation.nokia.com/srlinux/24-3/books/evpn-vxlan/evpn-vxlan-tunnels-layer-3.html#evpn-ifl-interoperability-with-evpn-iff to interop with FRR

PR #2074, opened 2025-03-24
Fix the SR Linux BGP AF activation

### LINKED ISSUE #2059: [BUG] SR Linux activates IPv4 AF over IPv6 BGP session and vice versa
**20-dual-stack-activate:**
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp/20-dual-stack-activate.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/bgp/20-dual-stack-activate.yml-validate.log)

**22-multi-vrf-bgp-ipv6:**
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vrf/22-multi-vrf-bgp-ipv6.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/vrf/22-multi-vrf-bgp-ipv6.yml-validate.log)

ipspace:
This is a minimal set of changes that ensures the BGP address families are activated as expected -- the address families that are not explicitely enabled are disabled.

This approach results in a situation where a BGP neighbor might not have any address families. That's rejected by SR Linux, so we need an extra device quirk that cleans up the neighbor transport IP addresses based on the AFs we'll need with the neighbor.

Fixes #2059

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -76,6 +76,9 @@
 
 {% macro bgp_families(neighbor,ipv4=True,ipv6=True,import=None,export=None) %}
 {%   set activate = neighbor.activate|default( {'ipv4': True,'ipv6': True } ) %}
+# neighbor: {{ neighbor }}
+# ipv4: {{ ipv4 }}
+# ipv6: {{ ipv6 }}
     afi-safi:
 {%   if ipv4 %}
     - afi-safi-name: ipv4-unicast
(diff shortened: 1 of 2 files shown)


jbemmel:
> This approach results in a situation where a BGP neighbor might not have any address families. That's rejected by SR Linux, so we need an extra device quirk that cleans up the neighbor transport IP addresses based on the AFs we'll need with the neighbor.

I think a better approach here, is to shutdown neighbors with no (IP) address families as I did in #2066:
```
{%   if not activate | dict2items | map(attribute='value') | select %}
    admin-state: disable
{%   else %}
```
The issue there is that 'else' is wrong - the address families still need to be configured as 'disabled'. Because later on, the EVPN module activates the EVPN family, and then IPv4 ends up as 'enabled' because it is not disabled.

I verified that SR Linux is ok with no address families being active if the neighbor is 'shutdown'

PR #2076, opened 2025-03-24
'loopback' attribute in global VRF definition must be a bool

### LINKED ISSUE #2068: [BUG] VRF loopback address should not be defined in the global VRF definition
It's perfectly OK to set the "I want to have a loopback in this VRF" flag in the global VRF definition (so it's applied to all nodes using the VRF) but it makes no sense to specify the IP address in the global VRF definition as that would be applied to all nodes using that VRF.

### To Reproduce

This is the config snippet that should trigger an error.

```
vrfs:
  red:
    loopback: 10.0.0.42/32
    links: [ dut-r1, dut-r2 ]
  blue:
    loopback: 10.0.0.43/32
    links: [ dut-r3, dut-r4 ]
```

ipspace:
Fixes #2068

### DIFF
--- a/tests/integration/vrf/04-multi-vrf-loopback-ipv6.yml
+++ b/tests/integration/vrf/04-multi-vrf-loopback-ipv6.yml
@@ -21,13 +21,15 @@ groups:
 
 vrfs:
   red:
-    loopback:
-      ipv6: 2001:db8:c001:cafe::/64
     links: [ dut-h1, dut-h2 ]
   blue:
-    loopback:
-      ipv6: 2001:db8:c001:cafe::/64
     links: [ dut-h3, dut-h4 ]
+
+nodes:
+  dut:
+    vrfs:
+      red.loopback.ipv6: 2001:db8:c001:cafe::/64
+      blue.loopback.ipv6: 2001:db8:c001:cafe::/64
 
 validate:
   red:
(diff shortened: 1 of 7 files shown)


tests/integration/vrf/04-multi-vrf-loopback-ipv6.yml
```diff
@@ -21,14 +21,16 @@ groups:
...
   blue:
-    loopback:
-      ipv6: 2001:db8:c001:cafe::/64
     links: [ dut-h3, dut-h4 ]
 
+nodes:
+  dut:
+    vrfs:
+      red.loopback.ipv6: 2001:db8:c001:cafe::/64
+      blue.loopback.ipv6: 2001:db8:c001:cafe::/64
```
jbemmel, line 32: It could be useful to allow ```vrf.loopback``` to be a pair of prefixes in addition to a bool, such that the user could do:

```
nodes:
  dut:
    vrf.loopback:
      ipv4: False
      ipv6: 2001:db8:c001:cafe::/64
```

Having the same loopback IP replicated in every VRF seems common?

ipspace: > It could be useful to allow `vrf.loopback` to be a pair of prefixes in addition to a bool, such that the user could do:

A) You can do that if you wish (or my .ipv6 stuff wouldn't work)
B) There's no need to set .ipv4 to False. If you specify a prefix for a node loopback, it's not merged with anything else.

> Having the same loopback IP replicated in every VRF seems common?

I wouldn't do that, but... 🤷‍♂️

PR #2082, opened 2025-03-25
Make the behavior of hosts more controlled

ipspace:
* Each device (apart from routers) must have its valid roles defined
* A node can become a host only if the device allows the 'host' role
* Static routing for hosts checks the 'mgmt_vrf' feature and uses either default route or specific routes for address pool prefixes.

Note:
* We cannot use 'mgmt_vrf' for FRR containers, as we don't know whether we'll manage to load the 'vrf' module during the initial configuration

### DIFF
--- a/docs/nodes.md
+++ b/docs/nodes.md
@@ -76,7 +76,7 @@ nodes:
 * **mgmt** -- management IPv4/IPv6 addresses. Used primarily with the [**external** provider](labs/external.md)
 * **mtu** -- sets device-wide (*system*) MTU. This MTU is applied to all interfaces that don't have an explicit MTU ([more details](links-mtu)).
 * **provider** -- virtualization provider used by this node (see [](labs/multi-provider.md) for more details).
-* **role** -- when set to **host**, the device does not get a loopback IP address and uses static routing toward the [default gateway](links.md#hosts-and-default-gateways). The only supported host device is *linux*, for which the host **role** is set in system device defaults.
+* **role** -- when set to **host**, the device does not get a loopback IP address and [uses static routing](node-router-host) toward the [default gateway](links-gateway) ([more details](node-router-host))
 * **unmanaged** -- when set to *True*, the node is not managed or configured by *netlab*. Use this parameter when integrating *netlab* topologies with additional external devices, which should not be configured by *netlab* ([more details](external-unmanaged)).
 
 ```{tip}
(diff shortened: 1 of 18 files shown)


netsim/augment/nodes.py
```diff
@@ -367,6 +367,18 @@ def augment_node_device_data(n: Box, defaults: Box) -> None:
...
+  role = n.get('role',None)
+  if role:
+    features = devices.get_device_features(n,defaults)
+    allowed_roles = features.initial.get('roles',['router'])
+    if role not in allowed_roles:
+      log.warning(
+        text=f"Node {n.name} (device {n.device}) cannot have role '{role}'",
+        more_hints=[ f'Allowed roles for this device type are: {",".join(allowed_roles) }' ],
+        flag='nodes.roles',
+        category=log.IncorrectType,
```
jbemmel, line 379: ```log.IncorrectValue```

netsim/augment/nodes.py
```diff
@@ -469,9 +481,16 @@ def default_static_route_list(topology: Box) -> list:
 * The IPv4 static routes are generated if the node uses IPv4 AF
 * The IPv6 static routes are generated if the node uses IPv6 AF and does not listen
   to Router Advertisements
+
+Hosts that have management VRF will get the default static routes, other hosts will
+get static routes configured in the 'routing.static.host' parameter or generated from
+the address pools.
 '''
 def add_host_static_routes(topology: Box) -> None:
   sr_list = topology.get('routing.static.host',None)
```
jbemmel, line 490: default ```[]```?

ipspace: You might want to have no static routes on the hosts, in which case `[]` might come handy as a valid value.

netsim/defaults/attributes.yml
```diff
@@ -92,7 +92,7 @@ node:
   id: { type: int, min_value: 1, max_value: 150 }
   config: list
   group: list
-  role: id
+  role: { type: str, valid_values: [ router, host, bridge, gateway ] }
```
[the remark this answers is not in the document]
ipspace: It was defined (forever) in Linux documentation. One could use it to implement a NAT device or a transparent proxy.

docs/nodes.md
```diff
@@ -76,15 +76,30 @@ nodes:
 * **mgmt** -- management IPv4/IPv6 addresses. Used primarily with the [**external** provider](labs/external.md)
 * **mtu** -- sets device-wide (*system*) MTU. This MTU is applied to all interfaces that don't have an explicit MTU ([more details](links-mtu)).
 * **provider** -- virtualization provider used by this node (see [](labs/multi-provider.md) for more details).
-* **role** -- when set to **host**, the device does not get a loopback IP address and uses static routing toward the [default gateway](links.md#hosts-and-default-gateways). The only supported host device is *linux*, for which the host **role** is set in system device defaults.
+* **role** -- when set to **host**, the device does not get a loopback IP address and [uses static routing](node-router-host) toward the [default gateway](links-gateway) ([more details](node-router-host))
```
jbemmel, line 79: Should document the list of valid values here?

ipspace: The default is "router" and "bridge" or "gateway" don't do anything yet. I will add the list of valid values once we add at least one more ;)

Issue #1909, opened 2025-02-08
lag module should model peerlink as regular VLAN


jbemmel:
Most MLAG platforms support things like OSPF on the peerlink between the MLAG router pair; the ```lag``` module should create a subinterface as appropriate, and allow these modules to do their job

For example, this should work as expected:
```
module: [ ospf, lag, vlan ]

links:
- lag:
    members: [ s1, s2 ]
    mlag.peergroup: True
```

Likewise:
```
module: [ ospf, lag, vlan ]

links:
- lag:
    members: [ s1, s2 ]
    mlag.peergroup: True
  ospf: False
```

ipspace:
As most platforms want to have some sort of VLANs on the peer link, what's stopping you from using a routed VLAN to run OSPF on it?

jbemmel:
Nothing, it needs to be added that’s all

Since different platforms use different VLANs, I’m thinking the lag module should generate the vlan trunk stanzas, and let the vlan module handle the details

ipspace:
@jbemmel -- still interested in doing this? if not, please close it.

jbemmel:
I still believe it would make sense, but closing as requested

Issue #1611, opened 2024-12-05
[BUG] Failure to allocate IP from P2P pool in case of access VLAN


jbemmel:
Consider the following topology:
```
vlans:
 v1:
  pool: p2p

nodes:
  r:
    module: [vlan]
  h1:
    device: linux
    provider: clab
    id: 5

links:
- r:
    vlan.access: v1   # Fails: Cannot assign host index 5 in ipv4 from prefix 10.1.0.0/30 to node h1
  h1:
- r:
  h1:
  pool: p2p           # Works fine (assigns from subnet 10.1.0.4/30 if vlan is still defined)
```

The ```p2p``` pool should imply a ```p2p``` allocation strategy, which works in the case without an access vlan. However, when the VLAN is added an ```id_based``` allocation strategy gets applied - which fails

Release: dev branch

jbemmel:
Adding ```allocation: p2p``` at https://github.com/ipspace/netlab/blob/9816fd9b301fb0d763d602884a043921167d5572/netsim/defaults/addressing.yml#L10

fixes it - but may hide an underlying issue

jbemmel:
The ```vlan``` module assumes an ```id_based``` allocation strategy if none is specified: https://github.com/ipspace/netlab/blob/9816fd9b301fb0d763d602884a043921167d5572/netsim/modules/vlan.py#L107

jbemmel:
In case of an access VLAN, a prefix get allocated due to the definition of the vlan. This prefix is assigned to the link, along with the default ```id_based``` allocation strategy, so ```assign_interface_addresses``` no longer calls ```get_prefix_IPAM_policy``` which has some more intelligent logic to select a strategy

Commenting out https://github.com/ipspace/netlab/blob/9816fd9b301fb0d763d602884a043921167d5572/netsim/modules/vlan.py#L107 leads to a ```sequential``` allocation policy which also works

jbemmel:
Another possible improvement is to introduce an ```id_modulo``` allocation policy:
```
def IPAM_id_based(link: Box, af: str, pfx: netaddr.IPNetwork, ndict: Box) -> None:
  for intf in link.interfaces:
    set_interface_address(intf,af,pfx,ndict[intf.node].id)

def IPAM_id_modulo(link: Box, af: str, pfx: netaddr.IPNetwork, ndict: Box) -> None:
  start = 1 if pfx.last != pfx.first + 1 else 0
[...]

PR #2084, opened 2025-03-25
lag module: Model peerlink as regular lag link

jbemmel:
Currently the link between MLAG peers is crafted internally within the lag module, using specific internal mechanisms to assign IP addresses and making it impossible to configure features such as OSPF or BGP between the peers.

This PR models the peerlink as a regular lag link with a VLAN trunk, enabling the full suite of Netlab features to be configured between the peers:
* Assign IP addressing using ```prefix``` or ```pool```
* Apply module processing (ospf, bgp, gateway to name a few)
* Configure the use of LACP

Fixes https://github.com/ipspace/netlab/issues/1909

Note: Arubacx templates are modified but will need further adjustments and testing (like defining the vlan that gets created for peering, similar to eos)

Testing: added a test for OSPF over peerlink
* ```NETLAB_DEVICE=cumulus_nvue NETLAB_PROVIDER=libvirt ./device-module-test -v lag```
* ```NETLAB_DEVICE=eos NETLAB_PROVIDER=clab ./device-module-test -v lag```

Breaking changes:
* Dell OS10 - the ```backup destination``` IP is now taken from ```intf.lag.mlag.peer_backup_ip``` instead of ```intf.lag.mlag.peer```

Implementation notes:
* If no prefix or pool is provided, the peerlink defaults to the IP configuration specified in the device features (for backwards compatibility, and to conform to the platform suggested defaults found in example snippets online)
* IP assignment to a routed vlan is achieved by creating a vlan in ```irb``` mode with a prefix, overriding the mode to ```route``` at the interface level
* On the platforms I checked, link-local IP addresses in the 169.254.0.0/16 range are distributed in OSPF, in spite of what one might expect due to their "link-local" nature. Using the same prefix at different mlag pairs works, because the locally connected route has preference over the OSPF route. Might deserve a note or caveat (and we could even opt to assign IPs from the P2P pool by default, instead of custom IP prefixes specified in the lag module device features)


ipspace (changes requested):
Let's start with the obvious one: is there a reason you couldn't add "allocation: p2p" to the global P2P pool?

It is supposed to be used for P2P links anyway, and uses /30 prefix, so there cannot be more than two nodes attached to it. Alternatively, one could define custom pool and use it for peer link VLANs.

Both options are better (IMO) than adding random global attributes.

jbemmel:
> Let's start with the obvious one: is there a reason you couldn't add "allocation: p2p" to the global P2P pool?

I suggested that in the past (#1611) but you disagreed at the time. Happy to revisit

ipspace:
> > Let's start with the obvious one: is there a reason you couldn't add "allocation: p2p" to the global P2P pool?
> 
> I suggested that in the past (#1611) but you disagreed at the time. Happy to revisit

Well, thanks for pointing that out ;) The problem was (and still is) that using P2P allocation strategy on a VLAN link masks the fact that an IRB VLAN has a single prefix. Not specifying the **p2p** allocation strategy crashes the allocation process, alerting the user to the fact that something weird is going on.

And now we uncovered another gotcha: you're doing too much tweaking. I never got that far because I stumbled on the "new global pool" showstopper.

ipspace:
> This PR models the peerlink as a regular lag link with a VLAN trunk, enabling the full suite of Netlab features to be configured between the peers:

Is there supposed to be a default VLAN to be used on that link for the global IP routing, or is it just a VLAN trunk like any other trunk?

> Note: Arubacx templates are modified but will need further adjustments and testing (like defining the vlan that gets created for peering, similar to eos)

There's a showstopper right there ;)

> IP assignment to a routed vlan is achieved by creating a vlan in `irb` mode with a prefix, overriding the mode to `route` at the interface level

There's absolutely no need for that. Peer links are established between devices of the same type anyway, and if a device does not support mixed trunks, they you use IRB for the global IP routing VLAN. Any further device-specific limitations can be addressed in device quirks.

Please (always) try to modify the minimum amount of data structures necessary. We're learned the hard way that doing anything else creates unpredictable effects in the future.

> * All vlans need to be allowed on the peerlink, I prefer to do this implicitly rather than explicitly list all vlans used in the topology because the latter may create "unsupported" scenarios (such as mixed trunks on Dell OS10).

I have no problem with this (just skip the VLAN assignment on the peer link), but it has to be documented.

> It would be nicer to have "allow all VLANs" syntax, such that we could avoid lag module specific modifications to the vlan template

I don't see a need for the "allow all" trunks from the topology perspective, but I have no problem with an internal flag like **vlan._all_vlans** that would be used in VLAN configuration templates instead of checking the peerlink interface name or type.

> * Global p2p pool could not be used due to the lack of a 'p2p' allocation strategy

As you can assign any pool to any VLAN, I fail to see why this is relevant ;)

jbemmel:
> > * Global p2p pool could not be used due to the lack of a 'p2p' allocation strategy
> 
> As you can assign any pool to any VLAN, I fail to see why this is relevant ;)

I did try it - it results in an address allocation failure for the ```topology/input/lag-mlag-m_to_m.yml``` test case

jbemmel:
> Please (always) try to modify the minimum amount of data structures necessary. We're learned the hard way that doing anything else creates unpredictable effects in the future.
> 

Ironically, this is exactly the reason why I opted for a separate ```mlag_p2p``` addressing pool, rather than add ```allocation: p2p``` to the existing global ```p2p``` one. It results in the minimum amount of change to the installed base

docs/module/vlan.md
```diff
@@ -148,6 +148,7 @@ To use a VLAN on a link, add **vlan** dictionary to a link or an interface on a
 
 * **access** -- the name of access VLAN configured on the link or interface
 * **trunk** -- a list or dictionary of VLANs configured on a trunk port -- see [](module-vlan-trunk-attributes) for more details.
+* **_allow_all** -- internal flag to mark a trunk for which all VLANs are allowed
```
ipspace, line 151: We're not documenting internal flags (for good reason -- we never promised the customers they can use them)

docs/module/lag.md
```diff
@@ -113,6 +113,12 @@ links:
     mlag.peergroup: True # (also) used to derive a unique MAC address for this group of MLAG peers
 ```
 
+### Peerlink configuration
```
ipspace, line 116: This left me a bit confused. I think it would be nice to have two short examples, one of them documenting how the peer link can be used as pure routed link (although we know there are other things going on in the background), another one describing how to use VLANs on the peer link (example use case: VRF Lite)

It's also worth mentioning that you chose that netlab configures "allow all VLANs" on the peer link. That is not always the case in real-life deployments.

jbemmel:
> You can set **prefix.allocation** on any link

I did try that first too, but it didn't seem to work. I'll try again

docs/module/vlan.md
jbemmel: removed

jbemmel:
Too many customizations / tweaks

PR #2085, opened 2025-03-26
Complete the support for multi-role devices acting as hosts (IOS, EOS)

ipspace:
* Modify IOS and EOS configuration templates to suppress IPv6 RA, listen to RA messages to generate IPv6 default route, and disable packet forwarding (when available)
* Add default role (router) and enable 'ipv6.use_ra' in EOS and IOS device parameters

Other changes:

* Add 'host role' integration test
* Rename 'ipv6_use_ra' feature to 'ipv6.use_ra'
* Display 'ipv6.use_ra' and 'roles' features in 'netlab show modules -m initial' support table
* Update transformation test results -- every instance of IOS or EOS device has 'role' attribute in node data

Documentation:

* Add 'role support' table to platform overview

Developer documentation:

* Move 'initial configuration features' section to 'device parameters'
* Add a section on device roles to 'initial device configuration' document

Off-topic but still relevant:

* Remove duplicate indices of 'implementation guidelines' in developer documentation

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -50,6 +50,17 @@
 | BIRD Internet Routing Daemon [❗](caveats-bird) | bird               |
 | dnsmasq DHCP server [❗](caveats-dnsmasq)       | dnsmasq            |
 
+(platform-host)=
+Most devices behave as routers (or layer-3 switches); the following devices can take multiple roles or behave as [IP hosts](node-router-host):
+
+| Device | router | host |
+|-----------------------|:--:|:--:|
+| Arista EOS            | ✅ | ✅ |
+| Bird                  | ✅ | ✅ |
+| Cisco IOS/IOS XE[^18v]| ✅ | ✅ |
+| dnsmasq               | ❌  | ✅ |
+| Generic Linux         | ❌  | ✅ |
+
 **Notes:**
 
 * Use the **[netlab show devices](netlab-show-devices)** command to display the list of supported devices and daemons.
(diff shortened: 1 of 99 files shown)


netsim/ansible/templates/initial/ios.j2
```diff
@@ -14,12 +14,16 @@ ip host {{ hname }} {{ addr_list|join (' ') }}
 !
 {% if af.ipv4|default(False) %}
```
jbemmel, line 15: ```and role != 'host'`` ?

ipspace:
> It would help if the bot would state what these test results are for...

The previous request ;) It's a single-runner FIFO system. But yeah, I could probably make that work without too much hassle.

netsim/ansible/templates/initial/ios.j2
ipspace: Yeah, that was the original idea.

However, IPv4 routing on IOS is special. If you turn off "ip routing", it becomes an IPv4 host from the 1980s. It does not display the IP routing table, and probably does not use the static routes (just the default gateway). I decided leaving IPv4 forwarding enabled is the lesser evil.

ipspace: ... added as a caveat

docs/platforms.md
```diff
@@ -50,6 +50,17 @@
 | BIRD Internet Routing Daemon [❗](caveats-bird) | bird               |
 | dnsmasq DHCP server [❗](caveats-dnsmasq)       | dnsmasq            |
 
+(platform-host)=
+Most devices behave as routers (or layer-3 switches); the following devices can take multiple roles or behave as [IP hosts](node-router-host):
+
+| Device | router | host |
+|-----------------------|:--:|:--:|
+| Arista EOS            | ✅ | ✅ |
```
jbemmel, line 58: Would be good to add FRR to this list

jbemmel (approved):

docs/platforms.md
ipspace: Currently it's failing the `initial/05-host` test for several reasons. We have to modify the configuration templates to make it work as a proper host.

Issue #2090, opened 2025-03-27
[BUG] Prefix allocation ignored on VLAN in trunk


jbemmel:
Lab:
```
---
module: [vlan]

groups:
  _auto_create: True
  nodes:
    members: [ n1, n2 ]

nodes:
  n1:
    id: 3

vlans:
  v1:
    pool: p2p
    prefix:
      allocation: p2p

links:
- n1:
  n2:
  vlan.trunk: [ v1 ]
```

Result:
```
[ERRORS]  Errors found in prefix_allocation.yml
[VALUE]   links: Address 10.1.0.3/30 on node n1/links[1].v1 is a subnet broadcast address
```

(because the default ```id_based``` allocation gets applied)

Bug is here: https://github.com/ipspace/netlab/blob/2e824422fbd26fb31496f09f7f5845972c55b78d/netsim/modules/vlan.py#L132

```links.assign_link_prefix``` copies the 'allocation' attribute to ```vdata.prefix```, but it gets overwritten

ipspace:
We had this discussion before, and I spent plenty of time trying to explain to you why the existing behavior makes sense in most cases. While I'm sorry to hear it interferes with your latest project, that doesn't mean we're going to change it.

PR #2091, opened 2025-03-28
Fix vlan trunk prefix 'allocation' attribute

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/2090

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -417,7 +417,6 @@ def assign_link_prefix(
       print(f'link {link_path} got prefix {pfx_data} from {link.prefix}')
     
     if link.prefix is False:                              # There should be no prefix on this link, get out
-      link.prefix = {}
       return
 
     if isinstance(link.prefix,str):                       # Is the prefix an IPv4 address?
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -409,7 +409,7 @@ def assign_link_prefix(
       pools: typing.List[str],
       addr_pools: Box,
       nodes: Box,
-      link_path: str = 'links') -> Box:
+      link_path: str = 'links') -> None:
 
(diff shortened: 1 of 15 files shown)


ipspace:
We had this discussion before, and I spent plenty of time trying to explain to you why the existing behavior makes sense in most cases.

While I'm sorry to hear it interferes with your latest project, that doesn't mean we're going to change it. I'm not willing to add more complexity and potential technical debt to address another niche case of yours.

ipspace:
On a more positive note:

A) have you tried using **ipv4: 1** (or 2) and **ipv6: 1** (or 2) on the two interfaces attached to that link?
B) Why do you keep insisting that the L3 link over LAG you're emulating with an IRB VLAN because of platform limitations must be a /31 (the root cause of our recent discussions)?

jbemmel:
> On a more positive note:
> 
> A) have you tried using **ipv4: 1** (or 2) and **ipv6: 1** (or 2) on the two interfaces attached to that link? B) Why do you keep insisting that the L3 link over LAG you're emulating with an IRB VLAN because of platform limitations must be a /31 (the root cause of our recent discussions)?

Where do you see a /31? All I’m trying to do is reuse the global ‘’’p2p’’’ pool, which happens to default to /30 but could be a /31 (which would break the suggested ip assignment)

As suggested, I am looking for the minimal set of changes that does what needs to be done and is acceptable to you

ipspace:
> Where do you see a /31?

OK, make it a /30 (carefully stepping around the elephant to focus on a minor detail 😜).

Still, it looks like things would work with no changes if you wouldn't insist on using the **p2p** pool. Am I right?

jbemmel:
> Still, it looks like things would work with no changes if you wouldn't insist on using the **p2p** pool. Am I right?

Yes, we could leave the peerlink VLAN without ```pool``` and it would get a ```lan``` pool and things would work

PR #2093, opened 2025-03-28
lag: Check for lag links with Linux bridges after removing virtual ones

### LINKED ISSUE #2092: [BUG] Superfluous warnings about non-existent links with Linux bridges
#2038 adds a warning for LAG links using a Linux bridge. However:
```
user@vm:~/Projects/netlab/tests$ netlab create integration/lag/10-mlag.yml -d eos
[WARNING] lag: The LAG link between s1 and s2 and x1 is using a Linux bridge. LACP will not work
[HINT]    See https://netlab.tools/module/lag/#lag-multi-provider for more details
          Set defaults.warnings.lag.bridge to False to hide this warning
```

LACP does work in this integration test; the link it is warning about gets removed from the topology

jbemmel:
Fixes #2092

### DIFF
--- a/netsim/modules/lag.py
+++ b/netsim/modules/lag.py
@@ -356,18 +356,12 @@ def populate_mlag_peer(node: Box, intf: Box, topology: Box) -> None:
   intf.pop('vlan',None)                                                     # Remove any VLANs provisioned on peerlinks
 
 """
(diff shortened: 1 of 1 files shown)


ipspace (commented):
Whatever works for you 🤷‍♂️

PR #2094, opened 2025-03-28
Implement 'host' functionality for FRRouting

ipspace:
* Enable SLAAC and RA-derived default route
* Disable IPv4 and IPv6 forwarding
* Disable RA generation in host mode
* Minor tweaks to the interface config process (ifdown at start, ifup at the end of interface sysctl modification)

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -14,9 +14,13 @@ echo
 echo "Use vtysh to connect to FRR daemon"
 echo
 SCRIPT
-{% set fwd_flag = '0' if role == 'host' else '1' %}
-sysctl -w net.ipv4.ip_forward={{ fwd_flag }}
-sysctl -w net.ipv6.conf.all.forwarding={{ fwd_flag }}
+#
+# This is an artifact of unknown provenance that should be removed in a year or two (= 2026/2027)
+#
+# FRR controls these parameters with 'ip forwarding' and 'ipv6 forwarding' commands
+#
+sysctl -w net.ipv4.ip_forward={{ '1' if 'ipv4' in af and role == 'router' else '0' }}
+sysctl -w net.ipv6.conf.all.forwarding={{ '1' if 'ipv6' in af and role == 'router' else '0' }}
 #
 {% if clab is not defined %}
 cat <<SCRIPT >.bash_profile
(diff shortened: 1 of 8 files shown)


netsim/ansible/templates/initial/frr.j2
```diff
@@ -14,6 +14,10 @@ echo
 echo "Use vtysh to connect to FRR daemon"
 echo
 SCRIPT
+{% set fwd_flag = '0' if role == 'host' else '1' %}
+sysctl -w net.ipv4.ip_forward={{ fwd_flag }}
+sysctl -w net.ipv6.conf.all.forwarding={{ fwd_flag }}
```
jbemmel, line 19: ``` = {{ '0' if role=='host' or not af.ipv6|default(False) else '1' }}```

netsim/ansible/templates/initial/frr.j2
```diff
@@ -93,17 +95,18 @@ fi
...
+ip link set {{ i.ifname }} down
 {%   if i.ipv6 is not defined %}
 sysctl -qw net.ipv6.conf.{{ i.ifname }}.disable_ipv6=1
-{%   elif role|default('router') != 'host' %}
-sysctl -qw net.ipv6.conf.{{ i.ifname }}.autoconf=0
-sysctl -qw net.ipv6.conf.{{ i.ifname }}.accept_ra=0
-ip link set {{ i.ifname }} down
-ip link set {{ i.ifname }} up
+{%   else %}
+{%     set ra_flag = '0' if role == 'router' else '1' %}
```
jbemmel, line 102: What about 'gateway'? ```'1' if role == 'host' else '0'```

netsim/ansible/templates/initial/frr.j2
```diff
@@ -144,7 +150,7 @@ interface {{ i.ifname }}
 {%  if i.ipv6 is string and i.ipv6|ipv6 %}
  ipv6 address {{ i.ipv6 }}
 {%  endif %}
-{%  if i.type != 'loopback' %}
+{%  if i.type != 'loopback' and role == 'router' %}
```
jbemmel, line 153: ```role != 'host'```?
```role in ['router','gateway']```?

jbemmel (changes requested):
So this is where having 4 allowed values for 'role' adds complexity...

ipspace:
> So this is where having 4 allowed values for 'role' adds complexity...

As we currently allow only two roles for FRR I fail to see the relevance of your point.

PR #2095, opened 2025-03-29
Refactor the role-specific node transformation into a separate subsystem

ipspace:
* Create 'roles' subsystem
* Initialize the subsystem as a set of role-specific plugins appended to the topology plugins
* Move loopback creation/addressing into 'router' role plugin
* Move default static route processing into 'host' role plugin

### DIFF
--- a/netsim/augment/__init__.py
+++ b/netsim/augment/__init__.py
@@ -1,13 +1,13 @@
 # Import the whole augmentation tree
 
+from . import devices
 from . import nodes
 from . import groups
 from . import links
-from . import devices
 from . import plugin
 from . import topology
-from . import main
 from . import config
 from . import components
 from . import tools
 from . import validate
+from . import main
--- a/netsim/roles/__init__.py
+++ b/netsim/roles/__init__.py
@@ -13,7 +13,7 @@ Return all nodes matching the specified role(s)
 def select_nodes_by_role(topology: Box, select: typing.Union[str,list]) -> typing.Generator:
   roles = select if isinstance(select,list) else [ select ]
   for node in topology.nodes.values():
-    if node.get('role','router') in select:
+    if node.get('role','router') in roles:
       yield node
   
 """
--- a/netsim/roles/__init__.py
+++ b/netsim/roles/__init__.py
@@ -1,0 +1,26 @@
+#
+# The "node roles" subsystem appends known (statically defined) node roles modules
+# to the topology Plugins (after the user-defined plugins)
+#
+import typing
+from box import Box
+
+from ..data import append_to_list
+
+"""
+Return all nodes matching the specified role(s)
+"""
+def select_nodes_by_role(topology: Box, select: typing.Union[str,list]) -> typing.Generator:
+  roles = select if isinstance(select,list) else [ select ]
+  for node in topology.nodes.values():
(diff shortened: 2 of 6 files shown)


ipspace:
This refactoring will make it cleaner to add new node roles (next one: bridge) and keep only the common functionality in **augment.nodes** (which was becoming a kitchen-sink of exceptions)

netsim/roles/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+from ..data import append_to_list
+
+"""
+Return all nodes matching the specified role(s)
+"""
+def select_nodes_by_role(topology: Box, select: typing.Union[str,list]) -> typing.Generator:
+  roles = select if isinstance(select,list) else [ select ]
+  for node in topology.nodes.values():
+    if node.get('role','router') in select:
```
jbemmel, line 16: ```if node.get('role','router') in roles```

Your variation happens to work due to substring matching, but line 14 suggests a different intent

netsim/roles/host.py
```diff
@@ -0,0 +1,87 @@
...
+from . import select_nodes_by_role
+
+'''
+Create the default static route list from pool prefixes
+'''
+def default_static_route_list(topology: Box) -> list:
+  sr_list = []
+
+  for ap_name,ap_data in topology.addressing.items():
+    if ap_name in ['mgmt','router_id']:
```
jbemmel, line 26: Perhaps this list should be defined in a YAML file, such that future plugins may modify it if needed?

For example, an ```mlag_p2p``` pool might not be intended to have static routes...

netsim/roles/router.py
```diff
@@ -0,0 +1,57 @@
+'''
+Create detailed node-level data structures from topology
+
+* Discover desired imagex (boxes)
```
jbemmel, line 4: images

netsim/roles/router.py
```diff
@@ -0,0 +1,57 @@
...
+      if af == 'ipv6':
+        if prefix_list[af].prefixlen == 128:
+          n.loopback[af] = str(prefix_list[af])
+        else:
+          n.loopback[af] = addressing.get_nth_ip_from_prefix(prefix_list[af],1)
+      else:
+        n.loopback[af] = str(prefix_list[af])
+      n.af[af] = True
+
+  for af in log.AF_LIST:
```
jbemmel, line 46: Not specific to this PR, but ```log``` is a bit of a weird place to put the definition for ```AF_LIST```
```addressing``` would make more sense to me

File: netsim/augment/__init__.py [hunk 5]
```diff
@@ -1,13 +1,13 @@
...
 from . import links
-from . import devices
 from . import plugin
 from . import topology
-from . import main
 from . import config
 from . import components
 from . import tools
 from . import validate
+from . import main
```
jbemmel, line 13: Talking about reordering chairs on a deck... :)

jbemmel (changes requested):
Couple of minor nits

netsim/augment/__init__.py
ipspace: I had a circular dependency issue and thought ordering things in approximate order of dependencies might solve that. Of course it didn't ;)

PR #2096, opened 2025-03-29
Accept 'int' and 'bool' values for management IP addresses

### LINKED ISSUE #1959: Treat management network like another link
Instead of handcrafting management IP addresses we should:

* Convert the management pool into a link-like data structure
* Use the "standard" interface address assignment logic to assign IPv4 or IPv6 addresses to management interfaces

This would allow management IP addresses specified as ints, address, or address/prefix. We would have to make an exception and reject True/False values though.

_Based on https://github.com/ipspace/netlab/pull/1955#pullrequestreview-2632910371 by @jbemmel_

ipspace:
Closes #1959

Note: it was too cumbersome to tweak the management IP logic (and all its exceptions) into something that could be used with regular interface-on-link IP assignment. Adding logic to handle bools and ints was easier ;)

### DIFF
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -17,7 +17,7 @@ from .. import providers
 from . import devices,addressing,links
 from ..data.validate import validate_attributes,get_object_attributes
 from ..data.types import must_be_int,must_be_string,must_be_id,must_be_device
-from ..data import global_vars
+from ..data import global_vars,is_true_int
 from ..modules._dataplane import extend_id_set,is_id_used,set_id_counter,get_next_id
 
 MAX_NODE_ID: typing.Final[int] = 250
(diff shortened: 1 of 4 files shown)


ipspace:
@jbemmel: You wanted to have ints as **mgmt.ipv4** values. Here they are ;)

netsim/augment/nodes.py
```diff
@@ -140,30 +143,37 @@ def augment_mgmt_if(node: Box, defaults: Box, addrs: typing.Optional[Box]) -> No
       'nodes')
     return
 
+  start = addrs.get('start',1)                                        # Get first address, skipping the default GW
   for af in 'ipv4','ipv6':                                            # Try to assign IPv4 or IPv6 management address
```
jbemmel, line 147: ```log.AF_LIST``` would be more consistent, but I do prefer the current code

jbemmel (approved):

PR #2097, opened 2025-03-29
'bridge' node role: a simple layer-2 bridge

ipspace:
This is the initial implementation of the 'bridge' node role:

* Bridge interfaces without a VLAN parameter are assigned to the default VLAN (name/id defined as a topology const)
* The default value of 'vlan.mode' for a bridge device is 'bridge' (not IRB)
* Multi-access link with a single attached bridge are expanded into a series of point-to-point links

Note: any device implementing the VLAN configuration module could potentially be a bridge. Whether that makes sense is a totally different question.

Future plans after this PR:

* Per-node default VLAN name/ID (can be used to implement multiple independent bridging domains)
* Integration test (at least on the platform level, it might not make sense to have it for individual devices)

### DIFF
--- a/docs/node-roles.md
+++ b/docs/node-roles.md
@@ -5,14 +5,14 @@ _netlab_ supports three types of nodes:
 
 * **[router](node-role-router)** -- a device performing a combination of layer-3 and (optional) layer-2 forwarding.
 * **[host](node-role-host)** -- an IP host, usually using static (default) routes instead of a routing protocol
-* **[bridge](node-role-bridge)** -- simple layer-3 bridges
+* **[bridge](node-role-bridge)** -- simple layer-2 switches (devices formerly known as *bridges*)
 
 Most _netlab_-supported devices act as *routers*; see the [platform support tables](platform-host) for a list of devices that can act as hosts or bridges.
 
 (node-role-router)=
 ## Routers and Layer-3 Switches
 
-The defining characteristics of routers are:
+The defining characteristics of devices with the **router** role   (the default device role) are:
 
 * At least one global [loopback interface](node-loopback) that can be used as the *router ID* and the control-plane endpoint
 * Layer-3 packet forwarding for the configured address families (IPv4/IPv6)
@@ -24,7 +24,7 @@ Routers usually run routing protocols but can also rely on static routing. When 
 
 Hosts do not have loopback interfaces (it's easiest if they have a single interface) and use static routes toward an adjacent [default gateway](links-gateway). On devices that don't have the management VRF, Vagrant or containerlab set up the default route, and _netlab_ adds static IPv4 routes for IPv4 prefixes defined in [address pools](address-pools).
 
-Hosts that have a management VRF (mostly network devices used as hosts) get two IPv4 default routes. Vagrant or containerlab sets up the IPv4 default route in the management VRF, and netlab adds a default route in the global VRF.
+Hosts that have a management VRF (mostly network devices used as hosts) get two IPv4 default routes. Vagrant or containerlab sets up the IPv4 default route in the management VRF, and netlab adds a default route toward an adjacent router in the global routing table.
 
 Most hosts listen to IPv6 RA messages to get the IPv6 default route. _netlab_ can add an IPv6 default route[^SRv6] on devices that do not listen to RA messages.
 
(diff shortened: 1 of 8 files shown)


ipspace:
This is mostly just a tweak in the data transformation logic using existing link/interface parameters from the **vlan** configuration module. Any device that supports VLANs should be able to work with this.

docs/node-roles.md
```diff
@@ -0,0 +1,106 @@
+(node-router-host)=
+# Routers, Host, and Bridges
+
+_netlab_ supports three types of nodes:
+
+* **[router](node-role-router)** -- a device performing a combination of layer-3 and (optional) layer-2 forwarding.
+* **[host](node-role-host)** -- an IP host, usually using static (default) routes instead of a routing protocol
+* **[bridge](node-role-bridge)** -- simple layer-3 bridges
```
jbemmel, line 8: ```layer-2+``` ?

docs/node-roles.md
```diff
@@ -0,0 +1,106 @@
...
+	h2:
+		device: linux
+	br:
+		device: ioll2
+		role: bridge
+
+links: [ rtr-br, h1-br, h2-br ]
+```
+
+In the above topology, *netlab* assigns an IP prefix from the **lan** pool to the VLAN segment connecting the four devices. You can change the **br.vlans.br_default** VLAN definition to change the parameters of the default bridge VLAN on node **br**.
```
jbemmel, line 63: Does this mean that my Dell OS10 PR https://github.com/ipspace/netlab/pull/1911 will now be required/accepted?

netsim/roles/bridge.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+Set the default VLAN mode on bridges to 'bridge' in the pre-transform hook
+to prevent VLAN module from setting it to some other value.
+
+Also, define 'br_default' node VLAN if it's not defined
+
+Return 'True' if we found at least one bridge (so we need further processing)
+"""
+def create_default_VLAN(topology: Box) -> bool:
+  BR_DEFAULT = global_vars.get_const('bridge.default_vlan.name','br_default')
```
jbemmel, line 26: I would expect the default VLAN name and ID to be part of the VLAN module, rather than this 'bridge' role

netsim/roles/bridge.py
```diff
@@ -0,0 +1,110 @@
...
+        link_cnt += 1
+        l_data._linkname = f'{link._linkname}.{link_cnt}'   # Create unique link and and linkindex
+        l_data.linkindex = links.get_next_linkindex(topology)
+        l_data.interfaces = [ get_box(br_intf), intf ]      # ... recreate P2P interfaces
+        topology.links.append(l_data)                       # ... and append the new P2P link to the links
+
+    topology.links.remove(link)                             # Finally, remove original link
+
+def pre_transform(topology: Box) -> None:
+  if not create_default_VLAN(topology):                     # Add default bridge VLAN
```
jbemmel, line 104: Most devices I'm familiar with - Dell OS10, EOS, FRR - would have an (implicit) default VLAN also in 'router' mode.

Perhaps default VLAN creation would fit better in the VLAN module?

jbemmel (commented):
IMHO some key changes would fit better in the vlan module, rather than this new 'bridge' role/module

netsim/roles/bridge.py
ipspace: This is not a VLAN default. This is the default used for the devices with **bridge** role.

netsim/roles/bridge.py
ipspace: > Most devices I'm familiar with - Dell OS10, EOS, FRR - would have an (implicit) default VLAN also in 'router' mode.

Layer-3 switches have a default VLAN.

> Perhaps default VLAN creation would fit better in the VLAN module?

This is not adding the "default VLAN" functionality to the **vlan** module. This is an implementation of a very simple bridge using the **vlan** module functionality, and as that one needs **vlans**, we have to use them.

ipspace:
> IMHO some key changes would fit better in the vlan module, rather than this new 'bridge' role/module

You missed (some of) the point. Will make it clearer in the documentation.

Thanks a million for the feedback!

PR #2098, opened 2025-03-31
Junos MAC-VRF EVPN implementation

ipspace:


ipspace:
@ssasso Would love to hear your thoughts on the MAC-VRF configuration.

It might make sense to retain the VLAN-aware configuration and use a device-specific flag to differentiate between the two (a simple include of one file or the other should do the trick).

I'll write the documentation after we hammer out the details ;)

ssasso:
The real problem is not defining the MAC-VRF (as you correctly did), but the implication that you must also assign the interface to the MAC VRF itself (as you did as well).

Everything goes well if you have only interfaces with access VLANs, but if you have a trunk, with multiple VLANS (which must belong to multiple MAC-VRF instances, in case of vlan-aware), you cannot assign the same interface (with family ethernet-switching) to multiple mac-vrfs...

That's why in case of a trunk you are forced to use SP-Style, with multiple sub-interfaces (for the different 8021q tags), and then you can directly add the subinterfaces to the different mac-vrf instances.

(i.e., https://www.reddit.com/r/Juniper/comments/1ef72fg/evpn_macvrf_using_vlanbased_service_and_lags/ )

netsim/ansible/templates/evpn/vjunos-switch.j2
```diff
@@ -32,29 +32,48 @@ protocols {
...
+}
+
+routing-instances {
+{%   for vname,v in vlans.items() if v.evpn.evi is defined and v.evpn.bundle is not defined %}
+  vlan_{{ vname }} {
+    instance-type mac-vrf;
+    service-type vlan-based;
+    vtep-source-interface {{ vxlan.vtep_interface }};
+    route-distinguisher {{ v.evpn.rd }};
+    vrf-target import {% for t in v.evpn.import %}target:{{ t }} {% endfor -%};
```
ssasso, line 49: I'm not 100% sure you can have multiple import and export targets. If stuff has not changed, you need to use import and export policies - See https://github.com/ipspace/netlab/blob/dev/netsim/ansible/templates/initial/junos.vrf.j2#L61

ipspace:
Here's a Cunning Plan (hoping you watched the Black Adder 😎):

* We add **evpn._vlan_aware** node-level flag (it won't be checked as it starts with an underscore)
* Without that flag, we use the new (VLAN-based) config and add a device quirk that checks whether the EVPN-backed VLANs are used only on access links (the most common use case)
* If someone wants to use EVPN-backed VLANs on VLAN trunks, they have to set that flag, but then they'll get the other EVPN config with potential interoperability implications.

When we get bored, we could refactor the Junos interface/VLAN configuration to use the **family ethernet-switching** config if the **features.vlan.model** is l3-switch and the SP-style VLAN subinterfaces if the **features.vlan.model** is router. That would require exporting device features into device group variables (so they can be checked in the templates) but that wouldn't hurt anyway.

How does this sound?

ipspace:
> Got integration test request `/integration clab vjunos-switch evpn`

BTW, I stopped the integration test run. The VM on which we can run the tests has spinning rust storage, and the vjunos-switch seems to fail to get ready within the wait-for-SSH timeout period. The log events were 15 minutes apart, which means that it probably failed to start in 15 minutes. Useless :(

ipspace:
FWIW, all VLAN, VXLAN and EVPN integration tests passed.

netsim/ansible/templates/evpn/vjunos-switch.j2
ipspace: It's confusing. Multiple targets are accepted, but only one is used :(

ipspace: ... added. Now I have to write an integration test to check it out :(

ipspace:
> Weird: the multi/RT logic seems to work (the export communities are attached, remote MACs are imported), but the traffic does not flow.

Forget it, my bad :( It works ;)

ssasso:
(quick test of two switches, one with *default_macvrf* and the other with new mac-vrf code works well)

ipspace:
> I think I am missing some bytes.

Then they are probably missing ;)

> The device quirk for this has still to be added, right?

Yes. I wanted to hear what you think about the whole idea (and my throwing-spaghetti-at-the-wall approach)

ssasso:
> > (quick test of two switches, one with _default_macvrf_ and the other with new mac-vrf code works well for both L2 and L3 VNI)
> 
> Why am I not surprised. The SIP of Networking keeps giving 🤦‍♂️

Maybe I explained it wrongly... It's all *working good* so we can proceed with your idea and adding the quirk to set the use of default macvrf in case a trunk is present

ipspace:
> Maybe I explained it wrongly...

No, I got it. One of these days I'll write another blog post about the beauties of having three kinds of EVPN VLAN service, some of them semi-compatible with each other depending on the implementation

> It's all **working good** so we can proceed with your idea and adding the quirk to set the use of default macvrf in case a trunk is present.

Perfect. Will add the quirk in a day or two and then ask for your final review.

Thank you! Ivan

ipspace:
Added the quirk and caveats. Hope I haven't missed something else.

ssasso (approved):
looks all good!! thanks!!

PR #2103, opened 2025-04-01
VLAN support on Linux containers

ipspace:
Features:
* VLAN access and trunk links
* SVI (IRB) interfaces
* Works on containers (with 'vanilla' configuration) only

Does not work:
* IRB on containers (Linux cannot be a router)
* The weird VLAN 1 trunk scenario


ipspace:
The root cause for VLAN 1 integration test failure: the static routes cannot be configured pointing to VLAN interfaces as the VLAN interfaces don't exist at that point yet.

It looks like we'll have to create the VLAN interfaces in the initial configuration process and then check whether they exist in the VLAN interface creation code (which is shared with FRR).

jbemmel:
> The root cause for VLAN 1 integration test failure: the static routes cannot be configured pointing to VLAN interfaces as the VLAN interfaces don't exist at that point yet.
> 
> It looks like we'll have to create the VLAN interfaces in the initial configuration process and then check whether they exist in the VLAN interface creation code (which is shared with FRR).
> 
> I can add that to this PR or merge this and create another PR.

I imagine there may be a reason why we couldn’t add ‘’’config_after:[ vlan ]’’’ to the routing module?

netsim/ansible/templates/initial/linux/vanilla-ifconfig.j2
```diff
@@ -0,0 +1,18 @@
+{% macro ifconfig(intf) %}
+{% if intf.ipv4 is defined and intf.ipv4 is string %}
+set +e
+ip addr del {{ intf.ipv4 }} dev {{ intf.ifname }} 2>/dev/null
+set -e
+ip addr add {{ intf.ipv4 }} dev {{ intf.ifname }}
+{% endif %}
+{% if intf.ipv6 is defined and intf.ipv6 is string %}
+sysctl -w net.ipv6.conf.{{ intf.ifname }}.disable_ipv6=0
```
jbemmel, line 9: We may want to do this also if ```intf.ipv6 == True```?

ipspace:
> I imagine there may be a reason why we couldn’t add ‘’’config_after:[ vlan ]’’’ to the routing module?

Thanks for the kick in the right direction ;) Looking at the code I couldn't see why it wouldn't work (but there might be some circular dependency that I didn't find somewhere).

> And then move the routing config bits to https://github.com/ipspace/netlab/blob/dev/netsim/ansible/templates/routing/linux.j2?

Yeah, that's the root cause of the problem, and there's no good reason (apart from a slightly faster deployment playbook) not to implement **routing** module on Linux in the traditional way. Plan to work on fixing it tomorrow.

Thanks again!

ipspace:
All VLAN/LAG/routing integration tests pass (as far as the functionality does not require a "router" note).

jbemmel (approved):
Looks ok to me

PR #2105, opened 2025-04-01
Allow interface name to be set on a per-interface basis

alistairking:
For example:
```
links:
  - router01:
    router02:
      name: 'My Description'
```

Also set interface descriptions on parent interfaces on SR Linux.
For example:
```
--{ + running }--[  ]--
A:router01# info interface ethernet-1/1
    interface ethernet-1/1 {
        description "My Description"
        admin-state enable
        subinterface 0 {
            description "My Description"
            ipv4 {
                admin-state enable
                address 172.17.0.33/30 {
                    primary
                }
            }
        }
    }
```

I'm happy to break up this PR if you'd prefer.

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -805,7 +805,12 @@ def set_interface_name(ifdata: Box, link: Box, ifcnt: int) -> None:
     ifdata.name = link.name
     return
 
-  node_name = link.interfaces[ifcnt].node
+  link_iface = link.interfaces[ifcnt]
+  if 'name' in link_iface:
+    ifdata.name = link_iface.name
+    return
+
+  node_name = link_iface.node
   n_list = [ link.interfaces[i].node for i in range(0,len(link.interfaces)) if i != ifcnt ]
   ifdata.name = create_ifname(node_name,n_list)
 
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -10,6 +10,7 @@ from box import Box
 # Related modules
 from ..utils import log,strings
 from .. import data
+from ..modules import get_effective_module_attribute
 from ..data.validate import validate_attributes,get_object_attributes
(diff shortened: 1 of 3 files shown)


alistairking:
Just saw that tests have failed. Will check those.

jbemmel:
Since it is a description, perhaps ‘’’description’’’ or ‘’’desc’’’ would be more clear?

alistairking:
> Since it is a description, perhaps ‘’’description’’’ or ‘’’desc’’’ would be more clear?

Where?

jbemmel:
> > Since it is a description, perhaps ‘’’description’’’ or ‘’’desc’’’ would be more clear?
> 
> Where?

‘’’
links:
- r1:
  r2:
    desc: “My description”

alistairking:
Oh, but this is already existing syntax for setting the description at the link level, it just didn't get applied at the interface level prior to my change.

netsim/augment/links.py
```diff
@@ -805,7 +805,12 @@ def set_interface_name(ifdata: Box, link: Box, ifcnt: int) -> None:
     ifdata.name = link.name
     return
 
-  node_name = link.interfaces[ifcnt].node
+  link_iface = link.interfaces[ifcnt]
```
ipspace, line 805: * The intent was absolutely spot-on
* The logic was incorrect -- interface **name** should take precedence over link **name**
* We already have a function that fetches an attribute from a hierarch of objects. Replaced the explicit logic with the call to that function.

ipspace (approved):

ipspace:
Thanks a million for spotting and fixing this one ;)

netsim/augment/links.py
alistairking: Aha, right. Thanks for this!

Issue #2102, opened 2025-04-01
[BUG] bird is failing "initial interface configuration" integration test


ipspace:
It looks like RA is not running on Bird daemon

```
[ping]    IPv4 ping H1 => H2 [ node(s): h1 ]
[PASS]    h1: Ping to h2 (172.16.1.2) succeeded
[PASS]    Test succeeded in 0.1 seconds

[ra]      Check RA-generated default route [ node(s): h1,h2 ]
[WAITING] Waiting for RA message to generate the default route (retrying for 30 seconds)
[WAITING] Waiting for RA message to generate the default route (14 seconds left)
[FAIL]    Node h1: IPv6 default route is missing
[FAIL]    Node h2: IPv6 default route is missing

[FAIL]    Mandatory test failed, validation stopped
```

jbemmel:
In daemons/bird/bird.j2 add:
```
{% for intf in interfaces|default([]) if 'ipv6' in intf %}
{%   if loop.first %}
protocol radv {
{%   endif %}
        interface "{{ intf.ifname }}" {
                max ra interval 5;      # Fast failover with more routers
        };
{%   if loop.last %}
}
{%   endif %}
{% endfor %}
```

PR #2109, opened 2025-04-03
Bird: Add IPv6 RA config

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/2102

Also sets link MTU in RA messages; test ```initial/04-mtu``` passes too


ipspace (approved):
Lovely. Thank you!

PR #2110, opened 2025-04-03
Bird: Fix test case bgp/08-ibgp-localas next hop processing

jbemmel:
Tested: ```netlab up integration/bgp/08-ibgp-localas.yml -d bird -p clab``` -> all 11 cases pass


ipspace (approved):
It's interesting that you don't need to set the "rr client" on the local-as IBGP neighbor, the "next-hop-self always" thing is spot-on.

Thank you! Ivan

PR #2114, opened 2025-04-04
Fix: Create extra loopback interfaces on Linux-based daemons

### LINKED ISSUE #2112: [BUG] Extra loopback interfaces are not created on bird daemon
02-loopback:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/initial/02-loopback.yml)
* [Test results](https://tests.netlab.tools/bird/clab/initial/02-loopback.yml-validate.log)

ipspace:
Fixes #2112

### DIFF
--- a/netsim/ansible/templates/initial/linux/vanilla.j2
+++ b/netsim/ansible/templates/initial/linux/vanilla.j2
@@ -14,7 +14,7 @@ sysctl -w net.ipv6.conf.all.forwarding={{ pkt_fwd }}
 #
 # Interface addressing, create any bond devices
 #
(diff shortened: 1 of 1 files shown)


ipspace:
The change is made where it is because bird-clab.j2 includes linux-clab.j2 which includes linux/vanilla.j2

jbemmel (approved):
Found https://github.com/ipspace/netlab/blob/5c5f7b75b8a999beec880c37a7cffad82b295f51/netsim/ansible/templates/initial/cumulus.j2#L120 which also lacks 'loopback'

PR #2117, opened 2025-04-04
Linux: Add support for ipv4 unnumbered peering

jbemmel:
Replaces https://github.com/ipspace/netlab/pull/2111

Linux supports a peer construct:
```
ip addr add {{ loopback ip }} peer {{ peer loopback }} dev {{ intf.ifname }}
```
This PR uses that to add support for a limited form of IPv4 unnumbered interfaces, restricted to cases where there is a single direct peer that also uses IPv4 unnumbered (in theory the above could also support multiple peers with regular ipv4 addresses, but that would require each peer to have an interface route to the unnumbered loopback ip )

* use ```features.initial.ipv4.unnumbered: peer``` to signal support
* add ```_unnumbered_peer``` attribute to suitable ipv4 interfaces
* enables OSPFv2 unnumbered support for Bird (test case ospfv2/05-unnumbered validated)

Note that unnumbered interfaces in VRFs are broken on FRR (loopback not added to VRF); will open separate issue


jbemmel:
Note: it may be possible to do generic ipv4 peering, with peers discovering the unnumbered ip through OSPF. Let me know if that would be of interest

jbemmel:
In case of unnumbered inside a VRF, the peer loopback IP gets updated by the vrf module quite late in the process

Instead of setting _parent_ipv4 early and then overwriting it in the vrf module, I'm wondering if we couldn't postpone setting _parent_ipv4 until after the VRF module has had a chance to create its loopbacks

ipspace:
Wrong button, sorry :(

ipspace:
> In case of unnumbered inside a VRF, the peer loopback IP gets updated by the vrf module quite late in the process

OK.

> Instead of setting _parent_ipv4 early and then overwriting it in the vrf module, I'm wondering if we couldn't postpone setting _parent_ipv4 until after the VRF module has had a chance to create its loopbacks

Why would that matter? We solve the simple case first, and if needed, add a better (second order 😜) solution later.

> 4. `augment.links.cleanup` (too late - other processing needs to see the correct unnumbered interfaces)

What other processing? The only time we need _parent_ipv4 is in Linux config template. I would add the "set _parent_ipv4 in neighbors list" code to `augment.links.cleanup`

jbemmel:
This is the 'minimal change' version - it would be possible to merge ```links.set_parent_interface``` with ```links.set_unnumbered_peers``` and remove ```vrf.fix_vrf_unnumbered```

netsim/augment/links.py
```diff
@@ -1035,13 +1035,21 @@ def interface_feature_check(nodes: Box, defaults: Box) -> None:
...
-            module='interfaces')
+        if isinstance(ifdata.ipv4,bool) and ifdata.ipv4:
+          if not features.initial.ipv4.unnumbered:
+            log.error(
+              text=f'Device {ndata.device} does not support unnumbered IPv4 interfaces',
+              more_hints=[ f'Used on node {node} interface {ifdata.ifname} (link {ifdata.name})' ],
+              category=log.IncorrectValue,
+              module='interfaces')
+          elif features.initial.ipv4.unnumbered == 'peer':
+            if len(ifdata.neighbors)!=1 or ifdata.neighbors[0].get('ipv4',False) is not True:
```
ipspace, line 1046: From the pure packet forwarding perspective, does the remote end have to be an unnumbered link? Admittedly, OSPF wouldn't work.

netsim/ansible/templates/initial/linux/vanilla-ifconfig.j2
```diff
@@ -1,9 +1,13 @@
 {% macro ifconfig(intf) %}
-{% if intf.ipv4 is defined and intf.ipv4 is string %}
+{% if intf.ipv4 is defined %}
+{%   if intf.ipv4 is string %}
 set +e
 ip addr del {{ intf.ipv4 }} dev {{ intf.ifname }} 2>/dev/null
 set -e
 ip addr add {{ intf.ipv4 }} dev {{ intf.ifname }}
+{%   elif '_unnumbered_peer' in intf %}
+ip addr add {{ intf._parent_ipv4 }} peer {{ intf._unnumbered_peer }} dev {{ intf.ifname }}
```
ipspace, line 9: This would probably fail if you run "netlab initial" twice, but I'll fix that in another PR.

ipspace (approved):

netsim/augment/links.py
jbemmel: On Linux the remote can be a regular ipv4 prefix, but I decided that the semantics for ```peer``` should be that both sides are unnumbered to simplify things.

Otherwise, the remote would need a route to the unnumbered source loopback address, defeating the purpose of having an unnumbered interface for sake of simplicity

netsim/ansible/templates/initial/linux/vanilla-ifconfig.j2
ipspace: I'll use **ip addr flush _intf_ scope global** instead.

PR #2118, opened 2025-04-04
Cumulus: Enable ipv6 on extra loopbacks too

jbemmel:
Triggered by https://github.com/ipspace/netlab/pull/2114

May not be strictly needed, but just to be consistent


ipspace (approved):
It's most probably irrelevant as you don't want to forward traffic on the loopback interface (and the VM interfaces were always IPv6-enabled anyway), but yeah, it's better to be consistent.

Issue #2119, opened 2025-04-04
[BUG] unnumbered interfaces in VRF fail to check for presence of loopback


jbemmel:
Edited version of ```ospf/ospfv2/05-unnumbered```:
```YAML
---
message: |
  This lab tests OSPF running over unnumbered IPv4 interfaces.

module: [ ospf ]

groups:
  probes:
    device: frr
    provider: clab
    members: [ x1 ]
    module: [ ospf, vrf ]

nodes:
  dut:
  x1:

vrfs:
  X:

links:
- dut:
  x1:
    vrf: X
  prefix.ipv4: True
  mtu: 1500
```

The OSPF neighborship fails to establish, because the loopback interface supporting the unnumbered interface on x1 isn't present inside the vrf X

PR #2120, opened 2025-04-04
VRF: Issue error when unnumbered OSPF interfaces are used without having a loopback in the VRF

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/2119

### DIFF
--- a/netsim/modules/vrf.py
+++ b/netsim/modules/vrf.py
@@ -314,7 +314,7 @@ def fix_vrf_unnumbered(node: Box, vrfname: str, lbdata: Box) -> None:
 def vrf_loopbacks(node : Box, topology: Box) -> None:
   loopback_name = devices.get_device_attribute(node,'loopback_interface_name',topology.defaults)
 
-  if not loopback_name:                                                        # pragma: no cover -- hope we got device settings right ;)
+  if not loopback_name:                                         # pragma: no cover -- hope we got device settings right ;)
     log.print_verbose(f'Device {node.device} used by {node.name} does not support VRF loopback interfaces - skipping assignment.')
     return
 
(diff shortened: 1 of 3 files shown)


ipspace:
This might be an overreaction. Have to check out a few platforms, but BGP (v4-over-v6) might still work. OSPF doesn't because the unicast packets are sent to the loopback IP address which is not in the VRF.

jbemmel:
> This might be an overreaction. Have to check out a few platforms, but BGP (v4-over-v6) might still work. OSPF doesn't because the unicast packets are sent to the loopback IP address which is not in the VRF.

Right, ```vrf/14-multi-vrf-unnumbered``` wouldn't pass if missing loopbacks were an issue

ipspace:
> Right, `vrf/14-multi-vrf-unnumbered` wouldn't pass if missing loopbacks were an issue

And we do an end-to-end ping in the validation, which means that even next hops somehow work out.

jbemmel:
> > Right, `vrf/14-multi-vrf-unnumbered` wouldn't pass if missing loopbacks were an issue
> 
> And we do an end-to-end ping in the validation, which means that even next hops somehow work out.

`vrf/14-multi-vrf-unnumbered` works because the BGP module enables ipv6 LLA when it sees IPv4 unnumbered. It caught me a bit by surprise, because nothing in that topology suggests ipv6

Perhaps we should add a quirk (warning) that ipv6 lla is being enabled to support BGP ipv4 unnumbered, suggesting users to make it explicit to make the warning go away. Then it's more clear what is happening

Back to the PR - we can limit the warning to cases where OSPF is being used inside the VRF

ipspace:
Your code gets the job done, but I don't like where it sits. Let me think about this a bit more.

Also: we support four IGPs. Are you sure the other three work?

jbemmel:
> Also: we support four IGPs. Are you sure the other three work?

I added a note to self about IS-IS, but you're right that a better solution would be to call a function ```has_igp_enabled(link)``` or something similar. Do we have that?

ipspace:
> I added a note to self about IS-IS, but you're right that a better solution would be to call a function `has_igp_enabled(link)` or something similar. Do we have that?

See #2123 -- I found just the right function that is executed per-protocol and scans all node interfaces for their VRF membership.

> ( I am not familiar with any other IGP besides OSPF and - very little - IS-IS, so I'd have to defer to others to tell me if those other protocols would even work )

From a purely Ivory Tower perspective 😜:

* IS-IS should establish adjacencies and exchange routing information, but I'm not sure the next hops would work out
* RIPv2 might work as it does not send unicast packets.
* EIGRP probably wouldn't work for the same reason OSPF doesn't

It's an interesting challenge for a blog post or two though ;)

PR #2121, opened 2025-04-04
BGP: Add warning when ipv6 is being enabled to support ipv4 unnumbered

jbemmel:
Context: https://github.com/ipspace/netlab/pull/2120

Test: ```netlab create integration/vrf/14-multi-vrf-unnumbered.yml -d frr -p clab -v```

Nothing in that topology says that ipv6 will be enabled, but it is required to support ipv4 unnumbered

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -298,6 +298,10 @@ def build_ebgp_sessions(node: Box, sessions: Box, topology: Box) -> None:
       if ipv4_unnum and not ipv6_num:                 # Unnumbered IPv4 without numbered IPv6
         ipv6_lla = True                               # ... requires IPv6 LLA session
         if not 'ipv6' in l:                           # ... and IPv6 enabled on the interface
+          log.error(f"BGP module is enabling IPv6 LLA on node {node.name} interface {l.ifname} to support IPv4 unnumbered",
+                    more_data="Set 'ipv6' to True to make this warning go away",
+                    category=Warning,
+                    module='bgp')
           l.ipv6 = True
 
       if ipv6_lla:                                    # IPv6 LLA session...


ipspace:
There are a few things wrong with this idea:

* You really don't need box-wide IPv6 for BGP unnumbered to work. It's just per-interface LLA.
* That per-interface LLA does not impact anything else in any way. You cannot route IPv6, you're not running IPv6 routing protocols, you can't even forward packets between interfaces.

Apart from security considerations, enabling IPv6 LLA on an interface never hurt anyone.

From the usability standpoint, if you know how BGP unnumbered works, you'll be annoyed by the warning, if you don't, you'll be confused ;)

Finally, the advice is wrong. If you set **ipv6: True** on an interface (or two), it will also enable the per-box AF which will potentially result in IPv6 routing protocols being configured for no good reason.

jbemmel:
> There are a few things wrong with this idea:
> 
> * You really don't need box-wide IPv6 for BGP unnumbered to work. It's just per-interface LLA.
> * That per-interface LLA does not impact anything else in any way. You cannot route IPv6, you're not running IPv6 routing protocols, you can't even forward packets between interfaces.
> 
> Apart from security considerations, enabling IPv6 LLA on an interface never hurt anyone.
> 
> From the usability standpoint, if you know how BGP unnumbered works, you'll be annoyed by the warning, if you don't, you'll be confused ;)
> 
> Finally, the advice is wrong. If you set **ipv6: True** on an interface (or two), it will also enable the per-box AF which will potentially result in IPv6 routing protocols being configured for no good reason.

From the viewpoint of education, it would IMHO still be nice to have a way to explain what is going on, at least to students who are trying to learn about networking.

Perhaps the advice is wrong, but could we have something akin to an "explain what's going on" flag that would - in this case - note that IPv6 LLA is being enabled by the BGP module to support IPv4 unnumbered, and point to a web URL with the explanation you've given here?

ipspace:
> From the viewpoint of education, it would IMHO still be nice to have a way to explain what is going on, at least to students who are trying to learn about networking.

Feel free to add a link to https://bgplabs.net/basic/d-interface/ into the BGP module documentation ;)

> Perhaps the advice is wrong, but could we have something akin to an "explain what's going on" flag

The amount of noise netlab & friends (Vagrant/Containerlab/Ansible) generate probably drowns any useful hint we might generate. I'd be surprised if anyone looks at all that except when things don't work.

PR #2122, opened 2025-04-05
Fix: Move VRF loopback creation to post_link_transform hook

ipspace:
This change exposes the VRF loopback interfaces to all post-transform hooks. The only visible side effect is the addition of 'bgp.advertise' flag as the BGP post-transform hook sees a new interface. That would cause a new 'network' statement in the device configurations, but no extra routes as most devices already do 'redistribute connected' into VRF BGP.

### DIFF
--- a/netsim/modules/vrf.py
+++ b/netsim/modules/vrf.py
@@ -308,8 +308,9 @@ def fix_vrf_unnumbered(node: Box, vrfname: str, lbdata: Box) -> None:
     if intf.get('ipv4',None) is not True:                       # Skip interfaces that are not unnumbered
       continue
 
-    intf._parent_intf = lbdata.ifname                           # Make in-VRF loopback the parent interface for in-VRF
-    intf._parent_ipv4 = lbdata.ipv4                             # ... IPv4 unnumbereds
+    intf._parent_intf = lbdata.ifname                           # Make in-VRF loopback the parent interface for
+    intf._parent_ipv4 = lbdata.ipv4                             # ... in-VRF IPv4 unnumbereds
+    intf._parent_vrf  = vrfname                                 # ... and remember where the data came from
 
 def vrf_loopbacks(node : Box, topology: Box) -> None:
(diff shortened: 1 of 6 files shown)


tests/topology/expected/vrf-loopback.yml
```diff
@@ -8,6 +8,7 @@ nodes:
   r1:
     af:
       ipv4: true
+      vpnv4: true
```
jbemmel, line 11: Doesn't this point at a missing test case? The extra loopbacks should get distributed using VPNv4, if it wasn't doing that before then that's a bug (?)

netsim/modules/vrf.py
```diff
@@ -308,8 +308,9 @@ def fix_vrf_unnumbered(node: Box, vrfname: str, lbdata: Box) -> None:
     if intf.get('ipv4',None) is not True:                       # Skip interfaces that are not unnumbered
       continue
 
-    intf._parent_intf = lbdata.ifname                           # Make in-VRF loopback the parent interface for in-VRF
-    intf._parent_ipv4 = lbdata.ipv4                             # ... IPv4 unnumbereds
+    intf._parent_intf = lbdata.ifname                           # Make in-VRF loopback the parent interface for
+    intf._parent_ipv4 = lbdata.ipv4                             # ... in-VRF IPv4 unnumbereds
+    intf._parent_vrf  = vrfname                                 # ... and remember where the data came from
```
jbemmel, line 313: Where is (will?) this get used?

ipspace: In #2123 ;)

tests/topology/expected/vrf-loopback.yml
ipspace: This was an edge case -- a VRF that contained only a loopback interface -- and the difference between before and now is that now the loopback is created before BGP post_transform hook, so the extra AF is set.

In theory, that could have caused a problem in a pretty contrived MPLS/VPN scenario, but now we fixed it anyway ;)

As for missing test case, you might claim we were sloppy inspecting the test results of that test case 🤷‍♂️

PR #2123, opened 2025-04-05
Check source VRF of IPv4 addresses used on unnumbered VRF interfaces

### LINKED ISSUE #2119: [BUG] unnumbered interfaces in VRF fail to check for presence of loopback

ipspace:
While it's possible to run BGP over unnumbered VRF interfaces borrowing
an IPv4 address over a global loopback (because no IPv4 addresses are
involved), that does not work with most IGPs.

This change (based on work by @jbemmel in #2119/#2120) adds a check
for the source of the unnumbered IPv4 address to VRF-specific IGP
processing.

Replaces #2120, closes #2119

### DIFF
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -300,10 +300,8 @@ def build_vrf_interface_list(
 
   # Time to cleanup IGP data
   for vname,vdata in node.get('vrfs',{}).items():                     # ... iterate over the list of VRFs
-    try:
-      proto_active = vdata.get(f'{proto}.active',False)               # Get the IGP data for the VRF
-    except:                                                           # ... assume 'not active' if get fails
-      proto_active = False
+    proto_active = isinstance(vdata[proto],Box) \
+                   and vdata[proto].get(f'active',False)              # Get the IGP active status for the VRF
     if not proto_active:                                              # If there's no record of active IGP neighbors
       remove_vrf_imports(node,vname,vdata,proto)                      # Remove all mentions of the IGP imports
       vdata.pop(proto,None)                                           # ... remove the VRF IGP instance
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -261,35 +261,61 @@ def remove_vrf_interfaces(node: Box, proto: str) -> None:
 #
 # build_vrf_interface_list -- copy VRF interfaces into VRF definition
 #
-def build_vrf_interface_list(node: Box, proto: str, topology: Box) -> None:
+def build_vrf_interface_list(
+      node: Box,
+      proto: str,
+      topology: Box,
+      vrf_unnumbered_check: bool = True) -> None:                     # Check the source IP VRF for IPv4 unnumbereds
+  
+  unnum_err_list = data.get_empty_box()
+
   for l in node.interfaces:
-    if proto in l and 'vrf' in l:
-      if node.vrfs[l.vrf][proto] is True:                                   # Handle 'force' the protocol by setting it to True
-        node.vrfs[l.vrf][proto] = { 'active': True }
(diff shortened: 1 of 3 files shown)


ipspace:
Includes a commit from #2122, has to be merged after that PR.

netsim/modules/_routing.py
```diff
@@ -261,35 +261,61 @@ def remove_vrf_interfaces(node: Box, proto: str) -> None:
...
+    if vrf_unnumbered_check and \
+        l.get('ipv4',None) is True and \
+        l.get('_parent_vrf',None) != l.vrf:                           # Have to check the loopback source VRF?
+      data.append_to_list(unnum_err_list,l.vrf,l.ifname)              # ... remember we failed
+
+    l.pop(proto,None)                                                 # Finally, remove global IGP parameters from interface
+                                                                      # Next we need to find if the VRF instance of IGP matters
+    for neighbor in l.neighbors:                                      # ... iterate over the list of neighbors
+      n_data = topology.nodes[neighbor.node]
+      if proto in n_data.get('module',[]):                            # ... and check if at least one of them uses the IGP
```
jbemmel, line 298: In theory, it could be that 'proto' is disabled on that specific interface on the neighbor side

ipspace: In theory, you might be correct ;) In practice, I don't feel like checking that ;))

netsim/modules/_routing.py
```diff
@@ -261,35 +261,61 @@ def remove_vrf_interfaces(node: Box, proto: str) -> None:
...
+        node.vrfs[l.vrf][proto].active = True
+
+  # Time to cleanup IGP data
+  for vname,vdata in node.get('vrfs',{}).items():                     # ... iterate over the list of VRFs
     try:
-      proto_active = vdata.get(f'{proto}.active',False)                     # Get the IGP data for the VRF
-    except:                                                                 # ... assume 'not active' if get fails
+      proto_active = vdata.get(f'{proto}.active',False)               # Get the IGP data for the VRF
+    except:                                                           # ... assume 'not active' if get fails
```
jbemmel, line 305: Does this ever happen? if this is to catch ```vdata``` being ```None```, then the code should be:
```
proto_active = vdata is not None and vdata.get(f'{proto}.active',False)
```

ipspace: It shouldn't (particularly now that we type-check everything), but do we want to change it and see what happens?

jbemmel (changes requested):
The try/catch thing isn't added by this PR, but should be cleaned up IMHO
(and I'm wondering if 'catch' ever happens, maybe the data would be sanitized before we get here?)

netsim/modules/_routing.py
jbemmel: > It shouldn't (particularly now that we type-check everything), but do we want to change it and see what happens?

I do - but it's easier for me to say that ;)

ipspace: > Does this ever happen? if this is to catch `vdata` being `None`, then the code should be:

The `vdata` is guaranteed to be set (at least in the **vrf** module), but `vdata[proto]` could be **False** (for example: `ospf: False`).

Rewrote the test to check for vdata[proto] type first and then use a get instead of a try/except.

PR #2125, opened 2025-04-05
Bird: Do not send IPv6 RA in 'host' role

jbemmel:
Fixes https://tests.netlab.tools/bird/clab/initial/05-host.yml-validate.log


ipspace (approved):
Thank you!

PR #2127, opened 2025-04-05
Bird: Implement OSPF timers, priority, password (simple, OSPFv2 only)

jbemmel:
Modified OSPFv2 tests (removing VRFs) pass


ipspace (approved):
Thank you! I also tweaked the integration tests so that they pass even for devices that do not support VRFs

PR #2128, opened 2025-04-05
Bird: Filter BGP communities per neighbor type

jbemmel:
Fixes https://tests.netlab.tools/bird/clab/bgp/05-community.yml-validate.log


ipspace (approved):
Thank you! You're becoming a veritable bird whisperer ;))

PR #2129, opened 2025-04-05
Bird: Prefer BGP routes over OSPF routes

jbemmel:
Passes all ```bgp/03-ibgp-rr tests```

Source that put me on the right track: https://bird.network.cz/pipermail/bird-users/2018-April/012175.html

Bird has no 'administrative distance', but *preference* can be used to the same effect. By default, OSPF has preference 150 and BGP has 100; by setting BGP to 160, BGP routes are preferred

PR also adds route policy debugging printouts that can be useful: 
```
birdcl
debug bgp_r2_ipv4 all
echo all
restart bgp_r2_ipv4
```
This prints out the routes exported to r2, evaluated against the policy

### DIFF
--- a/netsim/daemons/bird/bgp.j2
+++ b/netsim/daemons/bird/bgp.j2
@@ -15,17 +15,17 @@ protocol static bgp_ipv4 {
 
 filter bgp_prefixes {
   if source ~ [ RTS_STATIC, RTS_BGP ]
-    then accept;
+    then accept "Static or BGP route:", net;
 
 {% if bgp.advertise_loopback|default(False) %}
   if source ~ [ RTS_DEVICE ] && ifname = "lo"
-    then accept;
+    then accept "advertise loopback:", net;
 {% endif %}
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
I love the custom debugging messages bit. However, the "solution" is wrong.

IBGP routes should never be better than OSPF routes because we use OSPF to get the BGP next hops. If you make IBGP better (like you did), then a BGP route will be used to reach the BGP next hop, resulting in recursive routing, eventual loss of BGP session, and persistent oscillation.

OTOH, EBGP routes should be preferred over OSPF routes, but we're not testing that as it's the default behavior on most platforms and it's not exactly relevant in sane scenarios.

It looks like Bird is not willing to advertise a BGP route if it uses an IGP route for the same prefix (usually called "an inactive BGP route"), and there's no good way around that.

Please resubmit the PR, but just with the debugging bits.

jbemmel:
What if we set the preference only for EBGP neighbors?

PR #2131, opened 2025-04-06
Bird: Add route filter debugging messages

jbemmel:
PR adds route policy debugging printouts that can be useful:
```
birdcl
debug bgp_r2_ipv4 all
echo all
restart bgp_r2_ipv4
```
This prints out the routes exported to r2, evaluated against the policy

Replaces #2129


ipspace (approved):
Thank you!

PR #2133, opened 2025-04-07
Initial implementation of firewall.zonebased plugin

ssasso:
Starting point for zone-based firewall, for now supporting default zone-to-zone policies (with `permit`/`deny` actions)

The plugin has been put with `firewall.` as prefix, as in the future we may want to add additional "subplugins" (i.e., `firewall.nat`).

Created basic integration test as well.

Supported on:
* vSRX
* VyOS

### DIFF
--- a/netsim/extra/firewall.zonebased/defaults.yml
+++ b/netsim/extra/firewall.zonebased/defaults.yml
@@ -6,6 +6,8 @@ devices:
     features.firewall.zonebased: True
   vyos:
     features.firewall.zonebased: True
+  none:
+    features.firewall.zonebased: True
 
 attributes:
   node:
--- a/netsim/extra/firewall.zonebased/defaults.yml
+++ b/netsim/extra/firewall.zonebased/defaults.yml
@@ -1,0 +1,28 @@
+# firewall.zonebased
+#
+---
+devices:
+  vsrx:
+    features.firewall.zonebased: True
+  vyos:
+    features.firewall.zonebased: True
+
+attributes:
+  node:
+    firewall.zonebased:
+      default_rules: { type: list, _subtype: _fw_rule_entry }
+  interface:
+    firewall.zone: { type: str }
+
+  # specific attributes
+  _fw_rule_entry:
+    action:
+      type: str
+      _required: true
+      valid_values: [ permit, deny ]
+    from_zone:
+      type: str
+      _required: true
+    to_zone:
+      type: str
+      _required: true
--- a/netsim/extra/firewall.zonebased/defaults.yml
+++ b/netsim/extra/firewall.zonebased/defaults.yml
@@ -12,7 +12,7 @@ attributes:
     firewall.zonebased:
       default_rules: { type: list, _subtype: _fw_rule_entry }
   interface:
-    firewall.zone: { type: str }
+    firewall.zone: { type: id }
 
   # specific attributes
   _fw_rule_entry:
@@ -21,8 +21,8 @@ attributes:
       _required: true
       valid_values: [ permit, deny ]
     from_zone:
-      type: str
+      type: id
       _required: true
     to_zone:
-      type: str
+      type: id
       _required: true
--- a/netsim/extra/firewall.zonebased/plugin.py
+++ b/netsim/extra/firewall.zonebased/plugin.py
@@ -17,21 +17,21 @@ def post_transform(topology: Box) -> None:
         for intf in node.get('interfaces',[]):
             # if link has a firewall zone, apply extra config to this node
             fw_zone = intf.get('firewall.zone', '')
-            if fw_zone:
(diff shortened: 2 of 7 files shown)


netsim/extra/firewall.zonebased/defaults.yml
```diff
@@ -0,0 +1,28 @@
...
+    features.firewall.zonebased: True
+  vyos:
+    features.firewall.zonebased: True
+
+attributes:
+  node:
+    firewall.zonebased:
+      default_rules: { type: list, _subtype: _fw_rule_entry }
+  interface:
+    firewall.zone: { type: str }
```
ipspace, line 15: Is it OK to allow any string value or would it be better to insist on IDs (`type: id`)?

netsim/extra/firewall.zonebased/defaults.yml
```diff
@@ -0,0 +1,28 @@
...
+
+  # specific attributes
+  _fw_rule_entry:
+    action:
+      type: str
+      _required: true
+      valid_values: [ permit, deny ]
+    from_zone:
+      type: str
```
ipspace, line 24: Same as above: should this be `type: id`?

netsim/extra/firewall.zonebased/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+import netaddr
+
+_config_name = 'firewall.zonebased'
+
+def post_transform(topology: Box) -> None:
+    global _config_name
+    for node in topology.nodes.values():
+        features = devices.get_device_features(node,topology.defaults)
+        zbf_supported = 'firewall.zonebased' in features
+        if zbf_supported:
```
ipspace, line 17: This would create "firewall" and "firewall.zonebased" on every device that has this feature. Does that make sense? Wouldn't it be better to rely on lab topology to define these data structures and leave them missing if they are not defined?

netsim/extra/firewall.zonebased/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+        for intf in node.get('interfaces',[]):
+            # if link has a firewall zone, apply extra config to this node
+            fw_zone = intf.get('firewall.zone', '')
+            if fw_zone:
+                if not zbf_supported:
+                    log.error( f"Node {node.name}({node.device}) does not support 'firewall.zone' used on {intf.name}",
+                            category=log.IncorrectAttr,module=_config_name)
+                else:
+                    # add zone to list of zones
+                    node.firewall.zonebased._zones[fw_zone] = {}
```
ipspace, line 32: This will automatically create all the necessary boxes (the beautiful part of working with boxes), there's no need for the above initialization.

netsim/extra/firewall.zonebased/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+def post_transform(topology: Box) -> None:
+    global _config_name
+    for node in topology.nodes.values():
+        features = devices.get_device_features(node,topology.defaults)
+        zbf_supported = 'firewall.zonebased' in features
+        if zbf_supported:
+            # Create basic structures if not exists
+            if 'firewall' not in node:
+                node.firewall = {}
+            if 'firewall.zonebased' not in node:
```
ipspace, line 21: I would turn this check around -- if ZBF is not supported, but firewall.zonebased exists, we would report an error.

netsim/extra/firewall.zonebased/plugin.py
```diff
@@ -0,0 +1,43 @@
...
+                            category=log.IncorrectAttr,module=_config_name)
+                else:
+                    # add zone to list of zones
+                    node.firewall.zonebased._zones[fw_zone] = {}
+                    # add config template
+                    api.node_config(node,_config_name)
+        # now check that zones defined in rules are defined
+        rule_idx = 0
+        for def_rule in node.get('firewall.zonebased.default_rules', []):
+            for z in [ def_rule.from_zone, def_rule.to_zone ]:
```
ipspace, line 38: You can iterate over [ 'from_zone', 'to_zone' ] and then use that as a lookup key. That will give you the ability to report exactly what's wrong in the error message.

ipspace (changes requested):
Nice job. Suggested a few changes, but found only the "always creating firewall.zonebased data structures in advance" one annoying.

ssasso:
@ipspace should I add the "none" device to the supported ones to have the automatic CI/CD testing to pass?

ipspace:
> @ipspace should I add the "none" device to the supported ones to have the automatic CI/CD testing to pass?

Please do, they are a nice check that we haven't messed up something that matters enough to be tested ;)

ipspace (approved):
I polished the code a bit (hope you don't mind) and added the "none" device.

Issue #2136, opened 2025-04-08
JunOS Policy Framework


ssasso:
Following up #2134

JunOS has a "different" way for controlling policies, route redistribution, and other "basic attributes" (i.e., next hop).

Idea is to study and build a "policy framework", most probably implemented as python code on the device quirk file 
(and avoid shitting up jinja templates), to translate netlab policies and attributes in JunOS policies.

Basic required use cases:
* announce loopbacks, other connected subnets (based on interface/link) and *originate* routes;
* control next-hop attribute, especially on iBGP sessions;
* handle default route origination;
* interact with custom policies.

Route Target handling, even if controlled by policies on the config itself, can stay where it is (at least, for now).

ipspace:
@ssasso does #2137 close this or is there more work to do?

ssasso:
This can be closed :)

PR #2137, opened 2025-04-09
JunOS: routing policy improvements

ssasso:
Rework of JunOS routing policies features, as per #2136 study.

Most of the caveats are now removed due to nested and chained policies.

The remaining one is for *as-path* regex, which is different from other vendors. Created a simple converter with few basic use cases, which also made the integration tests to pass.

Still missing: 
* ~~proper handling of filters in case of *vrf*~~
* support for `<protocol>.import` (which was missing in any case)

Tested with integration tests `bgp/`, `vrf/`, `bgp.session/` and `routing/` for full compatibility with what we had before.


netsim/extra/bgp.session/junos.j2
```diff
@@ -80,7 +86,20 @@ routing-instances {
   remove-private;
 {%   endif %}
 {%   if n.default_originate|default(false) %}
-  export [ default-route {{k}}-export ];
+{# replace export policy with updated one, which includes default route policy #}
+  delete: export;
+  export [ 
+{%     if k == 'ibgp' and bgp.next_hop_self|default(false) %}
```
jbemmel, line 92: What about ```localas_ibgp``` neighbors?

ssasso: They are treated exactly like before as ebgp https://github.com/ipspace/netlab/blob/54f17a5132443fadd627e8f236c0f378e0736a75/netsim/ansible/templates/bgp/junos.j2#L98

ipspace:
> Added two integration tests for testing routing policy together with VRFs.

IIRC, there's nothing VRF-specific in the routing policies. You just want to check whether they are applied to the VRF BGP neighbors, right?

If that's the case, then I'd prefer having the VRF version of the test bundled with the original test, and use a plugin (like I did for OSPFv2/OSPFv3) to disable the VRF part of the test for devices that do not support VRFs.

Running a single test for a resource hog like Junos (times three, as I'm testing three Junos platforms 🤦‍♂️, plus we have NX-OS) takes way too long when I run the full set of tests for all platforms, so I'm trying to minimize the number of tests.

ipspace:
> @ipspace would it be possible to start your run of (no regression) integration tests as well (including the new ones), before I start adding the `<protocol>.import` **new feature**? -- thanks!!

I started them in the morning, and then you pushed a few more commits ;) Rerunning them right now.

ssasso:
> IIRC, there's nothing VRF-specific in the routing policies. You just want to check whether they are applied to the VRF BGP neighbors, right?

correct.

> If that's the case, then I'd prefer having the VRF version of the test bundled with the original test, and use a plugin (like I did for OSPFv2/OSPFv3) to disable the VRF part of the test for devices that do not support VRFs.

ok, will have a look and try to extend the existing ones.

ipspace:
> ok, will have a look and try to extend the existing ones.

No worries. I can do that.

ipspace:
Tests completed. There seems to be something wrong with BGP communities (in the routing module) plus a weird VXLAN error that trips most of the EVPN tests. Looks like I'll have to run the full suite of tests on vjunos-switch once this is done.

jbemmel:
> There seems to be something wrong with BGP communities

Could it be that it needs a mapping like https://github.com/ipspace/netlab/blob/6eca13be6e63b135cde2be059a4fd239525c9f71/netsim/devices/eos.yml#L36?

ssasso:
> There seems to be something wrong with BGP communities (in the routing module)

I definitely broke some stuff with the latest commits, since the community tests was working before :( :(

ipspace:
> I definitely broke some stuff with the latest commits

That's why we have tests ;)

ipspace (approved):
Looks reasonable ;), tests are passing ;))

Thanks a million for unravelling this spaghetti mess.

PR #2145, opened 2025-04-14
ArubaOS-CX: clear all bgp session (soft) after policy config

### LINKED ISSUE #2139: [BUG] Aruba routing policies are not applied to VRF BGP peers
03-med-out:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/03-med-out.yml)
* [Test results](https://tests.netlab.tools/arubacx/libvirt/routing/03-med-out.yml-validate.log)

10-match-prefix:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/10-match-prefix.yml)
* [Test results](https://tests.netlab.tools/arubacx/libvirt/routing/10-match-prefix.yml-validate.log)

ssasso:
Fix #2139

### DIFF
--- a/netsim/extra/bgp.policy/arubacx.j2
+++ b/netsim/extra/bgp.policy/arubacx.j2
@@ -47,6 +47,9 @@ router bgp {{ bgp.as }}
 {% endfor %}
 {% endif %}
 !
-{% for ngb in bgp._session_clear|default([]) %}
-do clear bgp {{ ngb }}
+do clear bgp all * soft out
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Nice solution, I like the soft in/out clear instead of a hard reset of the BGP sessions. However, I had to increase all the wait times in `routing/*yml` integration tests because ArubaCX takes "forever" to send the changed updates.

PR #2146, opened 2025-04-14
VyOS: routing policy shall consider only ipv4 and ipv6 af

### LINKED ISSUE #2143: [BUG] VyOS: Prefix filter does not work on VRF BGP sessions
10-match-prefix:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/10-match-prefix.yml)
* [Test results](https://tests.netlab.tools/vyos/clab/routing/10-match-prefix.yml-validate.log)

10-match-prefix:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/10-match-prefix.yml)
* [Test results](https://tests.netlab.tools/vyos/libvirt/routing/10-match-prefix.yml-validate.log)

We probably have to clear the VRF BGP session (similar to #2141)

ssasso:
Fix #2143

### DIFF
--- a/netsim/ansible/templates/routing/_route_policy_vyos.j2
+++ b/netsim/ansible/templates/routing/_route_policy_vyos.j2
@@ -83,7 +83,7 @@ top
 #}
 {% macro create_route_maps(rp_dict) -%}
 {%   for rp_name,rp_list in rp_dict.items() %}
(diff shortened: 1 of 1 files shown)


ipspace (approved):

PR #2149, opened 2025-04-14
VyOS: ansible check for config failure

### LINKED ISSUE #2147: [BUG] VyOS ansible config does not fail if the config template is not applied
In case the VyOS config change is not applied because of a config error (or template logical error), the ansible config task **shall** fail as well (and this is not happening now).

ssasso:
(proposal, based on stderr error message match - I know it's not "elegant", but it seems to work, since I found a template error with this)

Fix #2147

Fixes also a template error discovered with this "validation".


ipspace (approved):
I guess it's better than the alternatives ;)

ipspace:
> (already running also on my side a batch run of integration tests to check if other templates have errors)

I'll merge this one. If you find something, please submit it as a separate PR so we have a log of what's been fixed.

ssasso:
all looks good on my tests (apart from the existing caveats for ospfv3 and vrrp).

PR #2159, opened 2025-04-16
Set arp_announce to 2 on FRR and Linux

### LINKED ISSUE #2157: [not-really-a-BUG] frr (and others) is passing test 22-static-inter-vrf.yml only because loose arp filters
FRR (and other platforms) is passing test *routing/22-static-inter-vrf.yml* only because of its ARP default behavior.

Or, better, because it is answering to a "wrong" ARP request coming from nodes *x*.
(we already encountered this)

See the first ARP Request arriving at *dut* here:
```
09:53:17.327298 ARP, Request who-has 172.16.1.1 tell 172.16.42.3, length 28
09:53:17.327314 ARP, Reply 172.16.1.1 is-at aa:c1:ab:85:e9:4e, length 28
09:53:17.327317 IP 172.16.42.3 > 172.16.41.2: ICMP echo request, id 34, seq 0, length 64
09:53:17.327351 ARP, Request who-has 172.16.1.3 tell 172.16.1.1, length 28
09:53:17.327353 ARP, Reply 172.16.1.3 is-at aa:c1:ab:5c:55:87, length 28
09:53:17.327353 IP 172.16.41.2 > 172.16.42.3: ICMP echo reply, id 34, seq 0, length 64
```

Linux (and most probably also other platforms) by default are answering to that ARP Request (`sysctl net.ipv4.conf.xxxxx.arp_filter=0`).

Some others are not, or have a different setting for `arp_filter`.

In a specific case, VyOS has a more restrictive arp_filter, for which it's not answering to the initial ARP Request - and so it's failing the test for, let's say, "no reason".

This could be solved by pinging *dut* local interface from *x* nodes before trying the inter-vrf ping.

ipspace:
Closes #2157

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -80,6 +80,12 @@ fi
 /usr/lib/frr/frrinit.sh restart
 {% endif %}
 
+# Set system defaults
+#
+# Send ARP requests from a sane source IP address
+#
+sysctl -w net.ipv4.conf.all.arp_announce=2
+
 #
 # Create loopbacks, stub and lag/bond devices
 #
--- a/netsim/ansible/templates/initial/linux/ubuntu.j2
+++ b/netsim/ansible/templates/initial/linux/ubuntu.j2
@@ -82,6 +82,7 @@ systemctl restart lldpd
 #
 {% set pkt_fwd = "1" if role|default("host") == "router" else "0" %}
 cat <<SCRIPT > /etc/sysctl.d/10-netsim.conf
+net.ipv4.conf.all.arp_announce=2
 net.ipv4.ip_forward={{ pkt_fwd }}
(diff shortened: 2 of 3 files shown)


ssasso (approved):

jbemmel:
There seems to be no direct ipv6 equivalent, but perhaps https://sysctl-explorer.net/net/ipv6/use_oif_addrs_only/ applies too?

```net.ipv6.conf.interface.use_oif_addrs_only=1```

ipspace:
> There seems to be no direct ipv6 equivalent, but perhaps https://sysctl-explorer.net/net/ipv6/use_oif_addrs_only/ applies too?
> 
> `net.ipv6.conf.interface.use_oif_addrs_only=1`

The way I read the documentation, this might break IBGP (and anything else running off loopback interface)

PR #2162, opened 2025-04-16
FRR: Support deleting of communities

jbemmel:
FRR supports community deletion, but only using a list. This PR adds a quirk to configure such lists

Source: https://github.com/FRRouting/frr/issues/4692

Notes:
* Post suggests to enable ```bgp route-reflector allow-outbound-policy``` to fix deletion on reflected routes, PR adds this flag since I see no downside to having it enabled

Tested: ```NETLAB_DEVICE=frr NETLAB_PROVIDER=clab ./device-module-test routing```

### DIFF
--- a/docs/module/routing.md
+++ b/docs/module/routing.md
@@ -136,7 +136,7 @@ The **set.community** attribute can be used to set these BGP communities on supp
 | Cumulus Linux       | ✅ | ✅ | ✅ | ✅ | ❌  |
 | Dell OS10           | ✅ | ❌ | ✅ | ✅ | ❌  |
 | Junos               | ✅ | ✅ | ✅ | ✅ | ✅ |
-| FRR                 | ✅ | ✅ | ✅ | ✅ | ❌  |
+| FRR                 | ✅ | ✅ | ✅ | ✅ | ✅  |
 | VyOS                | ✅ | ✅ | ✅ | ✅ | ❌  |
 
 ### Shortcut Routing Policy Definitions
--- a/netsim/ansible/templates/routing/frr.j2
+++ b/netsim/ansible/templates/routing/frr.j2
@@ -64,7 +64,9 @@ vrf {{ r_vrf }}
 {%   endfor %}
 {% endif %}
 
+{% if bgp is defined %}
 router bgp {{ bgp.as }}
   ! Allow outbound policies for reflected iBGP routes, if any
   bgp route-reflector allow-outbound-policy
 !
+{% endif %}
--- a/netsim/ansible/templates/routing/frr.j2
+++ b/netsim/ansible/templates/routing/frr.j2
@@ -7,7 +7,11 @@
   set extcommunity bandwidth {{ p_entry.set.bandwidth }}
 {%       endif %}
 {%     endif %}
-{{     routemap.common_route_map_entry(p_entry,af_list) }}
+{%     if p_entry.set.community.delete is defined %}
+  set comm-list {{ p_entry.set.community._frr_list }} delete
+{%     else %}
+{{       routemap.common_route_map_entry(p_entry,af_list) }}
+{%     endif %}
 {%-   endcall %}
 {% endif %}
 {% if routing.prefix|default({}) %}
@@ -59,3 +63,8 @@ vrf {{ r_vrf }}
 {%     endfor %}
 {%   endfor %}
 {% endif %}
+
+router bgp {{ bgp.as }}
+  ! Allow outbound policies for reflected iBGP routes, if any
+  bgp route-reflector allow-outbound-policy
+!
--- a/netsim/devices/frr.py
+++ b/netsim/devices/frr.py
@@ -24,6 +24,24 @@ def check_stp_on_trunks(node: Box, topology: Box) -> None:
       more_data=err_data,
       node=node)
 
+"""
+FRR supports the deleting of communities, but only through community lists.
+This method creates such named lists for any policies that delete communities
+"""
+def create_bgp_community_delete_lists(node: Box) -> None:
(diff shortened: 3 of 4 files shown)


jbemmel:
Note: ```integration/routing/24-static-indirect-inter-vrf.yml``` failed when I ran it as part of ```device-module-test```, but passes when run manually

ipspace:
> Increasing the wait from 5 to 10 seconds for `ospfv3_x3` seems to fix it

Thank you! One of these days, I'll have to define the minimum waiting times for specific operations, and go through all integration tests to ensure the wait times are consistent.

ipspace:
IIRC (based on https://netlab.tools/module/routing/#id2) multiple platforms use the same approach as FRR (most probably, FRR got "inspired" by what Cisco IOS is doing ;), so it would make sense to make that code part of common "routing" module functionality.

However, I think I'll have to take the time to split that blob of complexity into smaller bits first. A single file contains stuff that implements four loosely-related things, and it's getting harder to find what needs to be fixed and where it is.

jbemmel:
> IIRC (based on https://netlab.tools/module/routing/#id2) multiple platforms use the same approach as FRR (most probably, FRR got "inspired" by what Cisco IOS is doing ;), so it would make sense to make that code part of common "routing" module functionality.
>

I had the same thought, but I figured we could start here and "promote" the code to more generic usage later

> However, I think I'll have to take the time to split that blob of complexity into smaller bits first. A single file contains stuff that implements four loosely-related things, and it's getting harder to find what needs to be fixed and where it is.

That, plus the dependencies vary. Having community policies without the BGP module being used makes limited sense, so I initially didn't think I needed to check for ```bgp.as``` being defined

I would suggest to split things (at least) along the lines of "requires BGP" versus not (e.g. static routes)

ipspace:
> I had the same thought, but I figured we could start here and "promote" the code to more generic usage later

A centralized approach would bypass the problem I mentioned as a review comment, as the functionality would reside within `_route_map_ios.j2`

> That, plus the dependencies vary. Having community policies without the BGP module being used makes limited sense, so I initially didn't think I needed to check for `bgp.as` being defined

Well, I missed that one :( Changing BGP parameters in another unrelated module just because you think they should be changed is somewhat "suboptimal" and makes for an interesting troubleshooting exercise.

Also, why do you think you need that? IIRC we don't allow routing policies to be set on IBGP neighbors

jbemmel:
> Well, I missed that one :( Changing BGP parameters in another unrelated module just because you think they should be changed is somewhat "suboptimal" and makes for an interesting troubleshooting exercise.
>

I first added it to the ```bgp``` script, but then changed my mind and put it in the ```routing``` script since it doesn't need to be there for sure if no routing policies are used. Re: "unrelated", that's my point: The routemap parts of the ```routing``` module are very much related to the ```bgp``` module

I think some restructuring between the core bgp module, the routing module and the bgp.X plugins would be useful.
Perhaps convert the 'routing' module into a static-routing plugin, moving the policy related parts to the bgp.policy plugin

> Also, why do you think you need that? IIRC we don't allow routing policies to be set on IBGP neighbors

Based on the documentation and https://github.com/ipspace/netlab/blob/84ef77a2d2e3071243cf9752784286c4550b5a67/netsim/extra/bgp.policy/plugin.py#L297 one would think so, but when I change R2 in https://github.com/ipspace/netlab/blob/dev/docs/plugins/bgp.policy.md#applying-routing-policies-to-ebgp-neighbors to also have bgp.as=65000 there is no warning and route policies get applied on the interfaces

I'll admit that I didn't spend a whole lot of time investigating this, but I figured it wouldn't hurt to enable the flag - anticipating use-cases for policies on ibgp neighbors ( e.g. in route reflector scenarios )

ssasso:
> The routemap parts of the `routing` module are very much related to the `bgp` module

IIRC, they can be used also in other modules for controlling route redistribution (`XXX.import`).

Potentially in the future they can be used also by additional plugins (i.e., pbr, acl, ...).

So I would keep the routemap part in `routing` - not strictly related to bgp.

jbemmel:
> > The routemap parts of the `routing` module are very much related to the `bgp` module
> 
> IIRC, they can be used also in other modules for controlling route redistribution (`XXX.import`).
> 
> Potentially in the future they can be used also by additional plugins (i.e., pbr, acl, ...).
> 
> So I would keep the routemap part in `routing` - not strictly related to bgp.

Then perhaps rename 'routing' to 'routemap' and move the static routing bits somewhere else?

ipspace:
> I first added it to the `bgp` script

I still wonder why you'd feel like you have to tweak that nerd knob. What problem were you trying to solve?

> The routemap parts of the `routing` module are very much related to the `bgp` module

Absolutely not. As @ssasso pointed out, they are used for other purposes, and can also be used to define generic route maps that the user can then use for whatever with custom config templates.

> I'll admit that I didn't spend a whole lot of time investigating this, but I figured it wouldn't hurt to enable the flag - anticipating use-cases for policies on ibgp neighbors ( e.g. in route reflector scenarios )

1. Using routing policies on IBGP neighbors is generally a bad idea (not that this argument ever persuaded some people). We are forced to use them on Arista EOS because they messed up the **next-hop-self** code.
2. Unless I'm getting dumber faster than I thought, we have no good mechanism to specify BGP behavior toward IBGP neighbors anyway, so how do you expect to apply route policies on them?

ipspace:
> Then perhaps rename 'routing' to 'routemap' and move the static routing bits somewhere else?

Prefix lists can also be used elsewhere, and in any case, this is turning into a bikeshedding exercise. We're not renaming a module that's been around for several releases.

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/2240

Issue #2154, opened 2025-04-15
[BUG] Unable to declare routing.community in groups


jbemmel:
Lab:
```
---
groups:
  test:
    members: [ x ]
    module: [ routing ]
    routing:
     community:
      c1: '65233:101'
     policy:
      set_pref:
      - match.community: c1
        set.locpref: 120

nodes:
  x:
   routing:
     community:
      c2: '65233:102'
     policy:
      set_pref:
      - match.community: c2
        set.locpref: 130
```

Result:
```
user@vm:~/$ netlab create /tmp/community_in_group.yml
[ERRORS]  Errors found in community_in_group.yml
[TYPE]    routing: attribute 'topology.groups.test.routing.community.c1[1]' must be a dictionary, found str
[HINT]    use 'netlab show attributes --module routing' to display valid attributes
[FATAL]   Cannot proceed beyond this point due to errors, exiting
```

Commenting out the 'groups' part, things work fine at the node level

ipspace:
Ah, I have to normalize the group data as well. Thank you, will fix.

ipspace:
I tried to solve this, but hit way too many dependencies:

* The group data is validated before module pre_default is called, so it's impossible to do data normalization in **routing.pre_default** for group data
* I tried to move the group data validation to a later phase, but it has to be done before the data is copied into nodes, otherwise we get node-level error messages instead of group-level ones
* Without copying group data into nodes, we don't get **device** and **module** information into nodes
* Without **module** information, the **pre_default** hook crashes or is not executed.

The only sane way out of this would be to document the caveat -- group **routing** objects cannot use the shortcuts.

Alternatively, I could try another insane solution: add a module "init" hook and move the data normalization there. Also, I would wrap the "init" data processing into try/except blocks and report errors when something crashes to avoid dealing with all sorts of crazy things that could happen due to invalid data types.

So, which one should it be?

ipspace:
> So, which one should it be?

Of course I decided to go for the insane one after one too many coffee breaks 🤦‍♂️ The first PR is already ready for review.

PR #2166, opened 2025-04-17
Implement static (defaults-driven) module hooks

ipspace:
This PR adds another way of calling module hooks: the "hooks" list in the module definition. The primary use case are hooks that have to be called before the topology data is validated and the "modules" list is finalized.

The "normalize" execute_module_hooks function is called before the data validation starts, giving the modules using it (planned: routing) the ability to do pre-validation data transformation.

See the discussion in #2154 for more details (hint: this implements the 'insane' option described in that discussion)

While implementing that functionality, I also refactored the whole module hook calling subsystem and removed duplicate code.

### DIFF
--- a/netsim/modules/__init__.py
+++ b/netsim/modules/__init__.py
@@ -18,7 +18,7 @@ from ..augment import devices
 # List of attributes we don't want propagated from defaults to global/node
 #
 no_propagate_list = [
-  "attributes", "extra_attributes", "features",
+  "attributes", "extra_attributes", "features", "hooks",
   "requires", "supported_on", "no_propagate",
   "config_after", "transform_after", "warnings" ]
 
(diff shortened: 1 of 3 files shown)


netsim/modules/__init__.py
```diff
@@ -458,38 +461,41 @@ def get_effective_module_attribute(
...
-  for m in topology.get('module',[]) + topology.get('_extra_module',[]):
-    if not mod_load.get(m):
-      mod_load[m] = _Module.load(m,topology.get(m))
-    if log.debug_active('modules'):
-      if hasattr(mod_load[m],f"module_{method}"):
-        print(f'Calling module {m} module_{method}')
-    mod_load[m].call("module_"+method,topology)
+  m_list = topology.get('module',[])
+  m_list += [ m for m in topology.get('_extra_module',[]) if m not in m_list ]
+  for m in m_list:
```
jbemmel, line 487: This assumes there is no need to check dependency order for 'hooks'?

jbemmel (approved):

netsim/modules/__init__.py
ipspace: Exactly. However, the `_extra_module` ones are added to the module list in `merge_global_module_params`(before pre_transform hooks are called), so we should be good.

This change was focused more on ensuring a module listed in `_extra_module` (current use case: **bgp.policy** plugin adding **routing** module) is not called twice.

PR #2168, opened 2025-04-17
Expand routing object shortcut definitions with the 'normalize' hook

### LINKED ISSUE #2154: [BUG] Unable to declare routing.community in groups

ipspace:
The 'normalize' hook is the only hook that can be called before the group data is validated, so it has to be used if we want to support shortcut routing object definitions in group data.

Closes #2154 (see that issue for more details)

### DIFF
--- a/netsim/modules/routing.py
+++ b/netsim/modules/routing.py
@@ -444,19 +444,26 @@ normalize_routing_data: execute the normalization functions for all routing obje
 def normalize_routing_data(r_object: Box, topo_object: bool = False, o_name: str = 'topology') -> None:
   global normalize_dispatch
 
-  for dp in normalize_dispatch.values():
-    if 'transform' in dp:
-      if dp['namespace'] in r_object:
-        xform = r_object[dp['namespace']]
-        xform = dp['transform'](xform,o_type=dp['object'],topo_object=topo_object,o_name=o_name)
-        if xform is not None:
-          r_object[dp['namespace']] = xform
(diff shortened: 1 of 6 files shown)


jbemmel (approved):
Starting to get pretty complex to understand the order in which things get called.
Perhaps we could have sections in source code files, or even separate files:
```
modules/routing/
 normalize.py
 transform.py
 ... etc ...
```

ipspace:
> Starting to get pretty complex to understand the order in which things get called.

Yeah ;)

> Perhaps we could have sections in source code files, or even separate files:

I was aiming for `policy.py`, `community.py`, `static.py` etc (you get the idea) in `modules/routing` with `__init__.py` containing the top-level dispatch code.

Issue #1481, opened 2024-11-03
Collecting netlab usage data


ipspace:
It would be great to know how people use _netlab_; currently, we can only guess as we get little feedback and zero hard data.

The proposal to implement the usage data collection and eventual upload is in `docs/roadmaps/usage.md`. Feedback or PRs against that file are most welcome.

ssasso:
Some very draft ideas:

Receiving and storing data can be achieved using cloudflare workers+kv or D1 storage, or with AWS Lambda+DynamoDB (plus putting some limits on it).

To demonstrate that we do not collect sensible data we could also show the collected data and some reporting?

DanPartelly:
Ivan's proposed collection mechanism is in plain-text yml dictionary , so any users can actually see the data collected, and the upload is user triggered, so I guess this covers the issue. 

I personally would be interested with what host OSes Netlab is used as well. 

> 
> To demonstrate that we do not collect sensible data we could also show the collected data and some reporting?
>

ipspace:
> I personally would be interested with what host OSes Netlab is used as well.

Agree. Would you please document how we could collect that in a way that would work on most Linux distros while still providing reasonably easy-to-interpret results?

For example, `uname -a` produces a printout that someone might be able to deduce Ubuntu release from, but it's way beyond my capabilities. Is `lsb_release` available everywhere?

ipspace:
> Receiving and storing data can be achieved using cloudflare workers+kv or D1 storage, or with AWS Lambda+DynamoDB (plus putting some limits on it).

These days I would definitely go with CF workers + KV/D1/R2

> To demonstrate that we do not collect sensible data we could also show the collected data and some reporting?

"The user could inspect the usage data with **netlab usage show**" ;) https://github.com/ipspace/netlab/blob/dev/docs/roadmap/usage.md?plain=1#L19

ssasso:
Ok for the inspection of collected data, but seeing some "reporting stats" could be interesting imho

jbemmel:
> Ok for the inspection of collected data, but seeing some "reporting stats" could be interesting imho

That's why I was thinking a GitHub repo might be a nice option - it puts the (anonymized) reported stats in a public place that people can go look at if they want to - not hidden in some backend database

DanPartelly:
I like this, conceptually. There is nothing like letting the user watch the data. 

[...]

PR #2172, opened 2025-04-18
Core usage-statistics-gathering functionality

ipspace:
* Extract some useful data from lab topology (called at 'netlab up')
* Track which netlab commands were used

Implements the data-gathering part of #1481

### DIFF
--- a/netsim/cli/__init__.py
+++ b/netsim/cli/__init__.py
@@ -16,7 +16,7 @@ from box import Box
 from . import usage
 from .. import augment
 from .. import __version__
-from ..utils import log, strings, status as _status, read as _read
+from ..utils import log, strings, status as _status, read as _read, stats
 from ..data import global_vars
 
 DRY_RUN: bool = False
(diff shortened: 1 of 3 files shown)


ipspace:
@ssasso @jbemmel Ideas for further collection points are most welcome ;)

@ddutt I would appreciate if you could take a look at the averaging function (around https://github.com/ipspace/netlab/pull/2172/files#diff-e893498c6040e698ee576846255d50f1e557208055eec4ec48bbc5995bfe5211R114)

jbemmel:
IMHO this type of change should (at least) be prominently mentioned in the user documentation, as statistics collection is potentially seen as intrusive

Furthermore, there should be a way to opt-out of stats collection (default: collect is fine)

ipspace:
> IMHO this type of change should (at least) be prominently mentioned in the user documentation

Agreed.

> as statistics collection is potentially seen as intrusive

Please note that I have no plans to upload the statistics without an explicit user action (see #1481 for details).

> Furthermore, there should be a way to opt-out of stats collection (default: collect is fine)

Also agree. Obviously we need **netlab usage stop** and **netlab usage start** together with **netlab usage clear** and **netlab usage show** (plus **netlab usage upload** in the future).

PR #2177, opened 2025-04-22
Implement provider-specific module/role support

ipspace:
This change extends the "features" dictionary to device/provider data, allowing us to specify platform-specific
module support (for example, VLANs run only on Linux containers) or provider-specific functionality
(for example, Arista EOS cannot run DHCP clients in containers).


jbemmel (approved):
No comments

PR #2186, opened 2025-04-24
SR Linux: Need to explicitly specify type as link-local for ipv6 addresses that are... link-local (duh)

jbemmel:
One would think that the NOS could derive the correct intended type...but no, not this one

Considered adding a test case to ```integration/initial/01-interfaces``` but I don't think any other platform suffers from this particular stupidity

### DIFF
--- a/netsim/ansible/templates/initial/srlinux.j2
+++ b/netsim/ansible/templates/initial/srlinux.j2
@@ -23,6 +23,7 @@
 {%   if intf.ipv6 is string %}
     address:
     - ip-prefix: "{{ intf.ipv6 }}"
+      type: "{{ 'link-local' if intf.ipv6|ipaddr('link-local') else 'global' }}-unicast"
 {%   endif %}
 {% if ipv6_ra %}
     neighbor-discovery:


ipspace:
> One would think that the NOS could derive the correct intended type...but no, not this one

IIRC Cisco IOS requires a similar hack. You wouldn't want someone to configure LLA IPv6 on an interface (because they've seen such addresses on Stack Overflow) and then open a ticket when it doesn't work.

> Considered adding a test case to `integration/initial/01-interfaces` but I don't think any other platform suffers from this particular stupidity

I'm a bit worried about supporting static LLAs. Next thing you know, someone will ask for two IPv6 addresses (GUA and static LLA).

Also, I'm guessing the reason for this particular PR is the "running BGP between static LLAs" idea you got from my recent blog post. Our current configuration templates would break on most implementations I'm familiar with because LLA has interface scope, so you have to specify the LLA and the interface name in the BGP neighbor. Yet again, not sure that's worth supporting.

ipspace (approved):
The IPv6 interface configuration still works, so I'm merging this. However, see the caveats...

PR #2187, opened 2025-04-25
Dell OS10 bgp features

jbemmel:
* BFD timers and BFD on BGP sessions
* BGP passwords, timers, allow AS in, remove private AS, default originate

Tested: ```device-module-test bgp.session -d dellos10 -p libvirt -v```

### DIFF
--- a/netsim/extra/bgp.session/dellos10.j2
+++ b/netsim/extra/bgp.session/dellos10.j2
@@ -1,0 +1,47 @@
+{% macro ebgp_neighbor(n,af) -%}
+{%   set peer = n[af] if n[af] is string else 'interface ' + n.local_if|default('?') %}
+  neighbor {{ peer }}
+{%   if n.bfd|default(False) %}
+   bfd
+{% endif %}
+{%   if n.remove_private_as|default([]) %}
+   remove-private-as
+{%   endif %}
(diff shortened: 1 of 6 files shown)


netsim/extra/bgp.session/dellos10.j2
```diff
@@ -0,0 +1,47 @@
+{% macro ebgp_neighbor(n,af) -%}
+{%   set peer = n[af] if n[af] is string else 'interface ' + n.local_if|default('?') %}
+  neighbor {{ peer }}
+{%   if n.bfd|default(False) %}
+   bfd
```
ssasso, line 5: IIRC BFD has to be enabled also globally and on all the interfaces where the bfd session will be established.

Maybe you forgot to commit the bfd module template for os10? (I don't find it in the repo)

ipspace (changes requested):
The BFD test is failing because the BFD configuration template is missing, everything else works great.

netsim/extra/bgp.session/dellos10.j2
jbemmel: Yep, forgot to commit the new file. Added now

ipspace (approved):
Everything works now ;) Thank you!

PR #2188, opened 2025-04-25
Bird: Enable RFC8950 support

jbemmel:
Initially hit unrelated issues with inherited roles, fixed by https://github.com/ipspace/netlab/commit/a4c43a5f6b73cf4188abf6970e78d6f7fee89dae

This PR adds support for
* RFC8950 extended next hops

PR also updates outdated caveats


ipspace:
> Unfortunately bird currently cannot pass `integration/bgp/22-dual-stack-rfc8950.yml` because it only supports `host` mode

That's an unrelated error. Bird should support mode "router" or it wouldn't pass most IBGP tests (where I had to add "role: router" to make Bird pass the test)

> and IP forwarding is disabled

It wouldn't be if the DUT **role** would be set to router.

> After `sysctl -w net.ipv4/6.conf.all.forwarding=1` it works

Please don't rush and tweak the nerd knobs to make your stuff work. It's much better (in the long term) to figure out why things don't work the way they should.

netsim/daemons/bird/bgp.j2
```diff
@@ -49,12 +49,15 @@ filter bgp_export_{{ ntype }} {
 {% endfor %}
 
 {% for n in bgp.neighbors %}
-{%   for af in [ 'ipv4','ipv6' ] if af in n %}
+{%   for af in [ 'ipv4','ipv6' ] if af in n and n[af] is string %}
```
ipspace, line 52: I knew this wouldn't end well before I even started the tests. How do you configure unnumbered BGP neighbors?

netsim/daemons/bird.yml
```diff
@@ -20,6 +20,8 @@ virtualbox:
 features:
   bgp:
     activate_af: true
+    ipv6_lla: true
```
ipspace, line 23: BGP over IPv6-LLA does not work. IPv4 AF over GUA IPv6 EBGP session (RFC 8950) works

ipspace (changes requested):
It's not "Unfortunately bird currently cannot pass integration/bgp/22-dual-stack-rfc8950.yml", it's "bird does not pass any unnumbered tests". Please don't rush, and don't test just the one scenario you're interested in.

The test results are @ https://tests.netlab.tools/_html/bird-clab-bgp

Also, the "x99-unnumbered-static" test is (A) irrelevant (because the static LLA is configured only on one end) and (B) overbloated. It might be an interesting scenario for the "netlab examples", though.

jbemmel:
> It wouldn't be if the DUT **role** would be set to router.

I did modify the test to set the role to ```router```, I was reporting what Netlab told me

```
ERRORS]  Errors found in 22-dual-stack-rfc8950.yml
[TYPE]    nodes: Node dut (device bird/provider clab) cannot have role 'router'
[HINT]    Allowed roles for this device type are: host,bridge
          Set defaults.warnings.nodes.roles to False to disable this check
```

I'll check into it

jbemmel:
Update: the ```role: router``` part works fine on the ```master``` branch. I suspect it's https://github.com/ipspace/netlab/pull/2177/files; the roles attributed to bird seem to come from the ```linux``` clab specific features

netsim/daemons/bird/bgp.j2
```diff
@@ -49,12 +49,14 @@ filter bgp_export_{{ ntype }} {
...
-protocol bgp bgp_{{ n.name }}_{{ af }} {
+{%   for af in [ 'ipv4','ipv6' ] if af in n and (n[af] is string 
+                                                 or (af=='ipv6' and n.ipv4|default(False) is sameas True)) %}
+{%     set dual_unnum = n.ipv4|default(False) is sameas True and n.ipv6|default(False) is sameas True %}
+protocol bgp bgp_{{ n.name }}_{{ 'dual_unnumbered' if dual_unnum else af }} {
 {%     set loopback = loopback|default({}) %}
 {%     set local_ip = loopback[af]|default('') %}
   local {{ local_ip.split('/')[0] if n.type == 'ibgp' else '' }} as {{ n.local_as|default(bgp.as) }};
-  neighbor {{ n[af] }} as {{ n['as'] }};
+  neighbor {{ n[af] if n[af] is string else 'range ::/0' }} as {{ n['as'] }};
```
ipspace, line 59: This hack is too ugly to merge. Looks like we'll stay at RFC 8950 support with no unnumbered EBGP.

netsim/daemons/bird/bgp.j2
jbemmel: Removed BGP IPv6 LLA support

netsim/daemons/bird/bgp.j2
jbemmel: ok, removed

netsim/daemons/bird/bgp.j2
```diff
@@ -70,15 +73,21 @@ protocol bgp bgp_{{ n.name }}_{{ af }} {
   rr cluster id {{ bgp.rr_cluster_id }};
 {%       endif %}
 {%     endif %}
-{%     if af in n.activate and n.activate[af] %}
-  {{ af }} {
+{%     for _af in [ 'ipv4','ipv6' ] if _af==af or (_af=='ipv4' and n.ipv4_rfc8950|default(False)) %}
+{%      if _af in n.activate and n.activate[_af] %}
```
ipspace, line 77: * I would prefer consistent indentation instead of squeezing another if block in between existing indentations.
* `n.activate[_af]|default(False)` might be cleaner than what you're using
* Finaly, it would be possible to combine everything into one **for ... if** block (and you can break it across multiple lines to make it easier to read).

ipspace (approved):
Integration tests pass. Thank you!

PR #2190, opened 2025-04-25
Bird: Fix inherited roles from Linux parent

### LINKED ISSUE #2189: Daemons inherit linux provider-specific roles
Turns out there is a problem here: Daemons by default inherit from the ```linux``` parent. Since ```clab.features.initial.roles``` is not overridden, Bird on clab is now restricted to ```[ host, bridge ]```

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/pull/2177#discussion_r2060158994_

jbemmel:
A recent PR #2177 introduced support for per-provider features, updating the Linux device features for clab. This PR fixes the resulting inherited feature flags for bird, including 'router' (and omitting 'bridge')

DNSmasq seems fine, however its description should be updated

Fixes #2189

### DIFF
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -13,6 +13,11 @@ clab:
     docker_shell: bash -il
   image: netlab/bird:latest
   build: 'https://netlab.tools/netlab/clab/#netlab-clab-build'
+  features:
+    initial:
+      ipv4:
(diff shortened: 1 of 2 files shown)


jbemmel:
Our PRs crossed...

Checking on the missing ```unnumbered: peer``` flag

Issue #2193, opened 2025-04-25
[BUG] BGP sessions over unnumbered IPv4 interfaces use IPv6 but top level 'af' only lists ipv4


jbemmel:
Lab: ```netlab create integration/bgp/06-unnumbered.yml -d bird -p clab```

Result:
```
# Ansible inventory created from ['/home/jeroen/Projects/netlab/tests/integration/bgp/06-unnumbered.yml', '/home/jeroen/.netlab.yml', 'package:topology-defaults.yml']
#

_daemon: true
_daemon_config:
  bgp: /etc/bird/bgp.mod.conf
  bird: /etc/bird/bird.conf
_daemon_parent: linux
af:
  ipv4: true   <!-- only ipv4 here
bgp:

... [ cut ] ...

hostname: clab-bgp-dut
interfaces:
- _parent_intf: lo
  _parent_ipv4: 10.0.0.1/32
  _unnumbered_peer: 10.0.0.2/32
  ifindex: 1
  ifname: eth1
  ipv4: true
  ipv6: true
```

Consequently, bird doesn't enable router advertisements because that depends on ```af.ipv6``` being true

Likely happens because https://github.com/ipspace/netlab/blob/170ac871b8e6f87dd490a31b21727fee5fdbee05/netsim/modules/bgp.py#L301 enables ipv6 without updating the global ```af```

It's kind of moot for now if we remove IPv6 LLA support - but fyi

ipspace:
We had various riffs on this discussion so many times in the past 🤦‍♂️ and we can't even blame the retrograde Mercury this time 😜

It makes no sense to enable an address family on the whole device just because you need interface-scope transport on an interface. That would trigger (for example) needless OSPFv3 configuration.

AFAIK, we successfully implemented the same functionality on a half-dozen platforms without changing the data model, so the data model might not be the problem you're looking for ;)

jbemmel:
> AFAIK, we successfully implemented the same functionality on a half-dozen platforms without changing the data model, so the data model might not be the problem you're looking for ;)

If ```af.ipv6``` is not set when ipv6 is used on some interface, then it is not the correct condition to check for enabling router advertisements on Bird here: https://github.com/ipspace/netlab/blob/0b210b1fe8cff48e0cafcec6ebc99685259c234e/netsim/daemons/bird/bird.j2#L4

ipspace:
[...]

PR #2194, opened 2025-04-26
Bird: Remove check on af.ipv6 for enabling Router Advertisements

jbemmel:
af.ipv6 may be false even when some interfaces use ipv6 (#2193)


ipspace:
I'm sorry but this seems to be completely irrelevant. Would you like me to explain why?

PR #2195, opened 2025-04-26
Add support for building multiple different versions of daemons, starting with Bird v3

jbemmel:
At some point we might decide to make v3 the default, and rename current ```Dockerfile``` to ```Dockerfile.v2```


jbemmel:
> I just don't like how you copied the whole Dockerfile just to change the Bird version. Give me a day or two to think this over.

We could consider a 'base' Dockerfile with daemons deriving from that. Would also take care of keeping the Ubuntu version consistent

PR #2201, opened 2025-04-26
Draft proposal: Postpone loading of override _top attributes, and do it only for modules actually used

jbemmel:
Case in point: the vlan.vni attribute really belongs in the VXLAN module, and should not be included when swapping vxlan for MPLS

This is incomplete; for example, an exception would have to be made for 'netlab show attributes vlan' to list all attributes from all modules

### DIFF
--- a/netsim/modules/__init__.py
+++ b/netsim/modules/__init__.py
@@ -307,20 +307,39 @@ def merge_global_module_params(topology: Box) -> None:
   reorder_node_modules(topology,'transform_after')
 
 '''
+merge_top_attributes: Complete the merging of module-defined modifications to global data structures,
+                      postponed from read_yaml such that only attributes defined by modules *in use* are
+                      applied and validated
+'''
+def merge_top_attributes(topology: Box) -> None:
+  data = topology.defaults
+  for m in topology.module:
+    if not '_top' in data.get(m,{}):                                   # Do we have to modify parent defaults outside of include scope?
+      continue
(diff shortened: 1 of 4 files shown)


ipspace:
> Case in point: the vlan.vni attribute really belongs in the VXLAN module, and should not be included when swapping vxlan for MPLS

And why exactly is this bothering you?

jbemmel:
> And why exactly is this bothering you?

I was converting an EVPN VXLAN lab to use MPLS instead, but forgot to remove the VNIs. Netlab still generated some EVPN configuration based on those, even though there was no vxlan module anymore 

It makes no sense to define vnis without using vxlan, so I figured I’d move the definition to the vxlan module and that would take care of it. Except it doesn’t - it makes no difference where _top attributes are defined, all are included 

IMHO it would be nice if it did make a difference

ipspace:
> I was converting an EVPN VXLAN lab to use MPLS instead, but forgot to remove the VNIs. Netlab still generated some EVPN configuration based on those, even though there was no vxlan module anymore

Yeah, that's a bummer. Blessed with the Perfect Hindsight&trade; I can see how it would have been better to name that attribute **vxlan.vni**. Alas, we don't break promises made to our users just because we find that convenient, so that ship has sailed.

> IMHO it would be nice if it did make a difference

You're solving the wrong problem, particularly considering that we identified (so far) exactly one problematic attribute. Check for **vlan.vni** in EVPN-enabled VLANs when the transport is set to MPLS.

PR #2202, opened 2025-04-27
Add CLI interface to usage statistics

ipspace:
Also: implement _disabled flag in stats.json to stop collection


ipspace:
@jbemmel: added the capability to disable collection of the statistics as per your suggestion. Also, if you feel we should point out somewhere else that we're collecting local statistics, please let me know.

jbemmel (approved):
Looks good to me

PR #2205, opened 2025-04-27
EVPN: Error when VNIs are used with mpls transport

jbemmel:
Context: https://github.com/ipspace/netlab/pull/2201

* Document the setting, correct valid values
* If evpn.transport is set by user, check if module is active (but not if default is selected; e.g. BGP EVPN RR use case doesn't require any transport module)
* Issue error when VNIs are used with (user selected) mpls transport

### DIFF
--- a/docs/module/evpn.md
+++ b/docs/module/evpn.md
@@ -108,6 +108,7 @@ Most EVPN/VXLAN implementations support only IPv4 VXLAN transport; some can run 
 
 EVPN module supports these default/global/node parameters:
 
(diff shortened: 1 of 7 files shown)


ipspace (changes requested):
Thanks for documenting the **evpn.transport** parameter. It would be a bit hard for someone to use EVPN over MPLS with the old documentation :(

However, I don't like the idea of setting the transport to MPLS based on the modules used in the topology. I would suggest we keep "vxlan" as the default value, and it probably does make sense to check whether the corresponding transport module is active in the lab topology.

ipspace (approved):

PR #2207, opened 2025-04-28
Make 'localas_ibgp' a valid bgp.activate value

### LINKED ISSUE #2204: Implied same treatment of localas_ibgp and ibgp doesn't work as intended
https://github.com/ipspace/netlab/blob/05c0c1485ec795eddda4e2a97819cb67bb94b7d4/netsim/modules/bgp.py#L450

```activate``` keys are address families 'ipv4' and 'ipv6', so ```'ibgp' in activate``` is always False and the ```append``` (fortunately) never gets executed

Suggested fix:
```
  for af in ['ipv4','ipv6']:
    if 'ibgp' in activate[af] and 'localas_ibgp' not in activate[af]:           # Make 'ibgp' imply 'localas_ibgp'
      activate[af].append('localas_ibgp')
```

(and note the lack of a test case covering this line)

ipspace:
Closes #2204


jbemmel (approved):
works for me

PR #2211, opened 2025-04-29
Fix: prune IGP interfaces based on IGP/interface AF match

### LINKED ISSUE #2203: [BUG] FRR templates stumble over inconsistent IS-IS interface settings
Lab (may not make complete sense as such, just to highlight and reproduce the issue)
```
---
message: Issue with interfaces having no overlapping address families enabled

addressing:
  p2p:
    ipv4: False       # ipv6-only
    ipv6: 2001:1::/48

  loopback:
    ipv4: False       # No ipv4 loopbacks
    ipv6: 2001::/48

isis.af:
  ipv4: True          # IPv4-only

groups:
  _auto_create: True
  hosts:
    members: [ h1, h2 ]
    device: linux
    provider: clab
  p:
    members: [ dut, p2 ]
    module: [ isis ]
    device: frr
    provider: clab

links:
- h1:
  dut:
    isis.passive: True
- dut-p2
- p2:
    isis.passive: True
  h2:
```

Result for p2:
```
% Configuration failed.
  
    Error type: validation
    Error description: YANG error(s):
     Path: Data location "/frr-interface:lib/interface[name='eth1']/frr-isisd:isis".
     Error: Mandatory node "area-tag" instance does not exist.
  
    The following commands were dynamically grouped into the same transaction and rejected:
    -  isis network point-to-point
```

The issue is that IS-IS only has AF ipv4 enabled, while the loopback and dut-p2 interface only have ipv6:
```YAML
{% for l in interfaces|default([]) if 'isis' in l %}
interface {{ l.ifname }}
! {{ l.name|default("") }}
{%   if 'ipv4' in l and 'ipv4' in isis.af %}
 ip router isis {{ isis.instance }}
{%   endif %}
{%   if 'ipv6' in l and 'ipv6' in isis.af  %}
 ipv6 router isis {{ isis.instance }}
{%   endif %}
{%   if l.isis.network_type is defined %}
 isis network {{ l.isis.network_type }}
{%   endif %}
{%   if l.isis.cost is defined or l.isis.metric is defined %}
 isis metric {{ l.isis.metric|default(l.isis.cost) }}
{%   endif %}
{%   if l.isis.passive %}
 isis passive
{%   endif %}
!
{% endfor %}
```
The above loop configures isis parameters without including either ```ip``` or ```ipv6``` router isis X

Similar issue may exist with other IGPs; if there is no overlap between the address families enabled on an interface and those enabled for the IGP, the IGP config should be removed from the interface

ipspace:
Fixes #2203

### DIFF
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -69,6 +69,29 @@ def routing_af(node: Box, proto: str, features: typing.Optional[Box] = None) -> 
         log.IncorrectValue,
         proto)
 
+# After parsing/building the IGP AF structure an offloading VRF interfaces into VRF data
+# structure, recheck the interface status. If there's no match between interface AF and IGP AF,
+# remove IGP from the interface
+
+def check_interface_af(node: Box, proto: str) -> None:
+  for vrf_proto in routing_protocol_data(node,proto):
+    interfaces = vrf_proto.get('interfaces',node.interfaces)                # Get VRF IGP interfaces or global intf
+    if 'af' not in vrf_proto:                                               # Is this IGP instance limited by AFs?
+      continue
+    AF_mismatch = False
+    for intf in interfaces:
+      if proto not in intf:                                                 # IGP not active on this interface
+        continue
+      if any((intf.get(af,False) or intf.get(f'dhcp.client.{af}',False))
+              and vrf_proto.af.get(af,False)
+                for af in log.AF_LIST):                                     # Do we have at least one matching AF?
+        continue
+      intf.pop(proto,None)                                                  # No, remove IGP from the interface
+      AF_mismatch = True                                                    # And remember we did that
+
+    if AF_mismatch and 'interfaces' in vrf_proto:                           # Do we have to rebuild VRF intf list?
+      vrf_proto.interfaces = [ intf for intf in vrf_proto.interfaces if proto in intf ]
+
 # Set network type for an interface:
 #
 # * If the network type is specified, validate it against a list of allowed network types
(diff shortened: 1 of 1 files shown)


netsim/modules/_routing.py
```diff
@@ -69,6 +69,29 @@ def routing_af(node: Box, proto: str, features: typing.Optional[Box] = None) ->
         log.IncorrectValue,
         proto)
 
+# After parsing/building the IGP AF structure an offloading VRF interfaces into VRF data
```
jbemmel, line 72: and offloading

netsim/modules/_routing.py
```diff
@@ -69,6 +69,29 @@ def routing_af(node: Box, proto: str, features: typing.Optional[Box] = None) ->
...
+def check_interface_af(node: Box, proto: str) -> None:
+  for vrf_proto in routing_protocol_data(node,proto):
+    interfaces = vrf_proto.get('interfaces',node.interfaces)                # Get VRF IGP interfaces or global intf
+    if 'af' not in vrf_proto:                                               # Is this IGP instance limited by AFs?
+      continue
+    AF_mismatch = False
+    for intf in interfaces:
+      if proto not in intf:                                                 # IGP not active on this interface
+        continue
+      if any((intf.get(af,False) or intf.get(f'dhcp.client.{af}',False))
```
jbemmel, line 85: We don't use it anywhere afaik, but note that ```intf.get(af)``` returns ```None``` if the key doesn't exist; the ```False``` could be omitted

netsim/modules/_routing.py
```diff
@@ -339,7 +362,7 @@ def remove_unused_igp(node: Box, proto: str, remove_module: bool = True) -> bool
     return False                                                            # ... OK, we're good
 
   for vdata in node.get('vrfs',{}).values():                                # Is protocol active in at least one VRF?
-    if proto in vdata:
+    if proto in vdata and vdata[proto].interfaces:                          # ... and at least on one interface?
```
jbemmel, line 365: ```vdata[proto].get('interfaces')``` to avoid creating ```{}```?

jbemmel (approved):

netsim/modules/_routing.py
ipspace: I don't think we'd get a protocol data structure in a VRF with no relevant interfaces, but OK

netsim/modules/_routing.py
ipspace: You're absolutely correct, thank you! However, IIRC, behavior changed across various versions of Box, so I learned (the hard way) to be on the safe side.

PR #2224, opened 2025-05-01
SRv6: Reworked, adding FRR support with IS-IS and BGP L3VPN

jbemmel:
* Adds FRR support for SRv6 over IS-IS, in combination with BGP L3VPN v4/v6
* Implements required BGP changes - such as enabling IPv4 over IPv6 with extended next hops
* OSPFv3 is added as a possible IGP for future extension; none of the platforms I have access to support it today

The module now implements its own BGP configuration separate from the core bgp module, avoiding changes that might  re-interpret existing attributes.

Notes:
* The FRR template currently used hard-coded ```usid``` behavior, this could be made configurable (and won't interop with SR OS as-is)


docs/module/srv6.md
```diff
@@ -1,18 +1,61 @@
...
 The module currently depends on IS-IS module and will trigger a configuration error if the **isis** module is not enabled in the network topology.
 
-## Parameters
+## Platform Support
+The following table describes per-platform support of individual router-level SRv6 features:
 
-* Locator: an IPv6 address range to allocate to SRv6 nodes; each node is assigned a unique /64 prefix from this range
+| Operating system         | IS-IS | OSPFv3 | BGP v4/v6 |
+| ------------------------ |:-----:|:------:|:---------:|
+| FRR                      |   ✅  ||  ❌   |     ✅    |
```
ipspace, line 19: Extra vertical bar, results in empty OSPFv3 column and unsupported BGP v4/v6 column

docs/module/srv6.md
```diff
@@ -1,18 +1,61 @@
...
+| FRR                   |   ✅  |   ✅  |
+| Nokia SR OS           |   ❌  |   ❌  |
+
+**Notes**
+* VPNv4 and VPNv6 address families are enabled on IPv6 IBGP/EBGP sessions
+
+## Configurable Global and Node Parameters
+
+* **addressing.srv6_locator** -- global address pool for allocation of SRv6 locator prefixes, default prefix defined by `topology.defaults.srv6.locator_pool`
+* **srv6.af**  -- IP address families to enable in the overlay, default IPv4 + IPv6
```
ipspace, line 36: What does "enable address family in overlay" mean?

docs/module/srv6.md
```diff
@@ -1,18 +1,61 @@
...
+* **srv6.bgp** -- enable SRv6 for BGP when module is available, default `True`
```
ipspace, line 37: What does "enable SRv6 for BGP" mean?

docs/module/srv6.md
```diff
@@ -28,26 +71,23 @@ We want to use unnumbered P2P interfaces and dual-stack deployment with IPv6 con
 ```
 addressing:
   p2p:
-    unnumbered: true
+    ipv6: true
```
ipspace, line 74: This will result in numbered IPv4 interfaces. Is that what you want? What's the point of SRv6 in this example?

netsim/ansible/templates/bgp/frr.j2
```diff
@@ -56,12 +56,10 @@ router bgp {{ bgp.as }}
   network {{ pfx|ipaddr('0') }}
 {%   endfor %}
 !
-{%   for n in bgp.neighbors
-       if n[af] is defined 
-         and (n.activate[af] is defined and n.activate[af]) %}
-{%     set peer = n[af] if n[af] is string 
+{%   for n in bgp.neighbors if n.activate.get(af) %}
+{%     set peer = n[af] if n.get(af) is string
```
ipspace, line 60: It's still not a good idea to keep insisting on using an attribute for something it was NOT designed to do, but whatever.

docs/module/srv6.md
```diff
@@ -1,18 +1,61 @@
...
+| ----------------------| :---: | :---: |
+| FRR                   |   ✅  |   ✅  |
+| Nokia SR OS           |   ❌  |   ❌  |
+
+**Notes**
+* VPNv4 and VPNv6 address families are enabled on IPv6 IBGP/EBGP sessions
+
+## Configurable Global and Node Parameters
+
+* **addressing.srv6_locator** -- global address pool for allocation of SRv6 locator prefixes, default prefix defined by `topology.defaults.srv6.locator_pool`
```
ipspace, line 35: Any reason why you won't say what it is?

netsim/modules/srv6.py
```diff
@@ -1,29 +1,101 @@
...
 import ipaddress
 
-class SRV6(_Module):
+DEFAULT_VPN_AF: typing.Final[dict] = {
+  'ipv4': [ 'ibgp' ],
+  'ipv6': [ 'ibgp' ]
+}
 
-  def node_post_transform(self, node: Box, topology: Box) -> None:
+POOL_NAME = "srv6_locator"
```
ipspace, line 19: You could use `data.global_vars.get_const` instead of a hardcoded value (but probably within a function, I'm not sure the topology is already set up yet when the module is loaded)

netsim/modules/srv6.py
```diff
@@ -1,29 +1,101 @@
...
+        if node.af.get(vpn_af):    # Check if the VPN AF is enabled
+          nb[vpn_af] = nb.ipv6     # ...and enable it over IPv6 (only)
+
+class SRV6(_Module):
+  """
+  module_pre_default - create the 'srv6_locator' address pool
+  """
+  def module_pre_default(self, topology: Box) -> None:
+    # Defining this as _top addressing includes it in *every* topology
+    if POOL_NAME not in topology.defaults.addressing:
```
ipspace, line 44: I would usually add `global POOL_NAME` just to be on the safe side.

netsim/modules/srv6.py
```diff
@@ -1,29 +1,101 @@
...
+  def node_pre_transform(self, node: Box, topology: Box) -> None:
+    mods = node.get('module',[])
+    d_features = devices.get_device_features(node,topology.defaults)
+    if node.srv6.get('bgp'):
+      if not d_features.srv6.get('bgp') and 'bgp' in mods:
+        log.error(
+          f"Node {node.name} does not support BGP with SRv6",
+          module='srv6')
+      elif 'bgp' not in mods:
+        node.srv6.bgp = False
```
ipspace, line 63: I would expect a warning when you turn off functionality the user requested.

netsim/modules/srv6.py
```diff
@@ -1,29 +1,101 @@
...
+"""
+Configures BGP address families for neighbors, including extended nexthop where needed
+"""
+def configure_bgp_for_srv6(node: Box, topology: Box) -> None:
+  srv6_af = node.get('srv6.af',{})
+  srv6_vpn = node.get('srv6.vpn')
+  for nb in node.get('bgp.neighbors',[]):
+    if 'ipv6' not in nb:           # Skip IPv4-only neighbors
+      continue
+    nb.ipv4_rfc8950 = True         # Enable extended next hops by default
```
ipspace, line 30: * I don't think your implementation supports inter-AS SRv6 SIDs, so maybe you should limit this to IBGP neighbors
* It does not make sense to enable extended next hops if we have an IPv4 address for the neighbor or if the activate flag will not be set.

netsim/modules/srv6.yml
```diff
@@ -1,21 +1,52 @@
...
-    locator: str
-    address_families:
-    bgp: bool
-    isis: bool
+    af:
+    bgp:
+    igp:
+    locator: { type: ipv6, use: subnet_prefix }
+    vpn:
     transit_only: bool
```
ipspace, line 37: This seems to be implemented only on SR-OS. Any plans for similar functionality on FRR?

ipspace (changes requested):
A bunch of minor details, otherwise it looks good to me.

netsim/modules/srv6.yml
jbemmel: I can add a support column, I looked at https://www.segment-routing.net/images/20250311-srv6-usid-frr-netdev-0x19.pdf but the sid ```show``` commands and static sids aren't available in the current FRR version and I figured we could leave that FFS.

The flag helps optimize resource usage on SR OS, since FRR is pure software it doesn't make much difference currently

docs/module/srv6.md
jbemmel: "enable SRv6 support for BGP"

docs/module/srv6.md
jbemmel: Changed the example to a simplified version of test case 01

netsim/modules/srv6.py
jbemmel: added

netsim/modules/srv6.py
jbemmel: I added an EBGP test case, I may be missing some of the technical nuances but the pings seem to work. 
Made extended next hops dependent on having ipv4 AF without ipv4 transport

netsim/modules/srv6.py
jbemmel: ok

docs/module/srv6.md
jbemmel: it's a surprise ;)

docs/module/srv6.md
ipspace: So it's effectively "AF to transport over SRv6" (I got as much from the code), but then you need to propagate the reachability information across the SRv6 core somehow, and if I got it right, BGP is the only way to do it, so why do we have two settings?

docs/module/srv6.md
ipspace: And what does that mean? I'm guessing you're trying to say "use SRv6 SIDs for BGP next hops", but see also "why do we need srv6.af and srv.bgp"

ipspace:
I'm still not sure the configuration data model is OK. In every overlay solution you need some data-plane transport (SRv6, SIDs exchanged with IS-IS) and some control-plane protocol to exchange reachability information.

It looks to me like BGP is the only control-plane protocol to exchange prefixes with SRv6 SIDs that you implemented so far, so I don't understand why you want to configure SRv6 AF separately from SRv6 BGP. Looking at the config templates, that seems to be the way SRv6 is configured on SR-OS, but that does not mean their data model is fit for our use.

Also, I'm guessing one of my comments was lost: I don't see how you plan to extend SRv6 beyond a single AS, so I don't understand why you're enabling it on EBGP sessions.

docs/module/srv6.md
jbemmel: Because I was overthinking things :) We can simplify it

jbemmel:
> 
> Also, I'm guessing one of my comments was lost: I don't see how you plan to extend SRv6 beyond a single AS, so I don't understand why you're enabling it on EBGP sessions.

I'll admit I haven't spent much thought on this and I'm fine with limiting things to iBGP for now, but eventually something like https://www.ietf.org/id/draft-mishra-spring-inter-domain-routing-arch-01.html should be possible

image

ipspace:
> > Also, I'm guessing one of my comments was lost: I don't see how you plan to extend SRv6 beyond a single AS, so I don't understand why you're enabling it on EBGP sessions.
> 
> I'll admit I haven't spent much thought on this and I'm fine with limiting things to iBGP for now

Yet again: if you implement **srv6.bgp** in the same way we did **mpls.bgp**, you'll solve the (data model) problem for good. Today the default would be:

````
srv6.bgp:
  ipv4: [ ibgp ]
  ipv6: [ ibgp ]
````

> but eventually something like https://www.ietf.org/id/draft-mishra-spring-inter-domain-routing-arch-01.html should be possible

And when this unicorn fart gets implemented in FRR, you change the values of **srv6.bgp** attribute to allow **ebgp** and you're done ;)

**Comment by jbemmel** [comment]
I reverted the changes to the core bgp templates, and made it such that the ```srv6``` module configures the ipv6 neighbors itself, rather than relying on the ```bgp``` module. This avoids any impact to users that are not including this module (the majority)

I experimented with ebgp multihop sessions, but I figured we can leave that to a future PR ;)

**Comment by ipspace** [comment]
This looks pretty good to me. When you're ready to have it merged, please let me know.

**Comment by jbemmel** [comment]
> This looks pretty good to me. When you're ready to have it merged, please let me know.

I'd like for the module to promote - or at least simplify - adhering to best practices regarding address allocation:

* I found https://www.cisco.com/c/en/us/support/docs/ip/ipv6-routing/220485-configure-design-and-migration-best-prac.html#toc-hId--872789322 which says "'It is a recommended best practice that SRv6 SIDs for locators are allocated from the private IPv6 Unique Local Address (ULA) address range which starts with FC00:". It also recommends allocating loopbacks within the locator space

* Last year the IETF published https://datatracker.ietf.org/doc/rfc9602/ which reserves 5f00::/16 for SRv6 SID ranges

I've added a flag to (re)allocate ipv6 loopbacks from the locator range, and found that - in case of FRR - a "network" statement doesn't work (/48 route doesn't get advertised) but putting the locator-based loopback address as a /48 does. Any objections to making this behavior the default? Or does it feel too "hacky" to you?

**Comment by ipspace** [comment]
> Cisco: recommended best practice that SRv6 SIDs for locators are allocated from the private IPv6 Unique Local Address (ULA) address range which starts with FC00:

Which would imply an intra-AS-only use case. I would hope ULAs are not advertised with BGP ;)

> It also recommends allocating loopbacks within the locator space

Why?

> Last year the IETF published https://datatracker.ietf.org/doc/rfc9602/ which reserves 5f00::/16 for SRv6 SID ranges

Use this.

> I've added a flag to (re)allocate ipv6 loopbacks from the locator range, and found that - in case of FRR - a "network" statement doesn't work (/48 route doesn't get advertised)

What network statement? You had a working FRR example published as a blog post. Why are you complicating things?

File: netsim/modules/srv6.py [hunk 12]
```diff
@@ -82,30 +82,36 @@ def node_pre_transform(self, node: Box, topology: Box) -> None:
...
     locator_net = ipaddress.IPv6Network(locator)
-    if 'ipv6' not in node.loopback:
-      log.error(
-        f"SRv6 requires an ipv6 loopback address on node {node.name}",
-        category=log.MissingValue,
-        module='srv6')
-    elif ipaddress.IPv6Interface(node.loopback.ipv6).network.subnet_of(locator_net):
+    if 'ipv6' not in node.loopback or topology.defaults.srv6.allocate_loopback:
+      if topology.defaults.srv6.allocate_loopback:          # Auto-assign a loopback from locator range
+        node.loopback.ipv6 = locator.split("/")[0] + "1/48" # Use /48 to advertise the full range, not mere /64
```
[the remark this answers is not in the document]
jbemmel: I did struggle with that - how does one get to an ```IPv6Interface``` object from an ```IPv6Network``` object?

```next(locator_net.subnets(new_prefix=48)).network_address + "1/48"``` avoids the 'split', but I didn't see a way to avoid the string concatenation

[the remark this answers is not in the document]
Comment (ipspace) [reply netsim/modules/srv6.py:95]: Now I'm even more confused. You want to advertise the whole node locator range as /48 from the loopback? Why?

File: netsim/ansible/templates/srv6/frr.bgp.j2 [hunk 13]
```diff
@@ -13,6 +13,9 @@ router bgp {{ bgp.as }}
   neighbor {{ peer }} send-community both
 {%   if n.next_hop_unchanged is defined %}
   neighbor {{ peer }} attribute-unchanged next-hop
+{%     if af == 'ipv6' %}
```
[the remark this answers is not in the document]
jbemmel: I am testing beyond the intra-AS use case, ```next_hop_unchanged``` only gets set for EBGP neighbors so this does not impact the intra-AS use case.

I am not sure if this will be needed, we can remove it for the time being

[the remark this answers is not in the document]
ipspace: > I am testing beyond the intra-AS use case

Maybe it would be better if you'd first focus on something that is merge-worthy instead of chasing random unicorn farts?

> `next_hop_unchanged` only gets set for EBGP neighbors so this does not impact the intra-AS use case.

And what happens when you want to implement global-IPv6-over-SRv6 when running BGP with customers?

netsim/modules/srv6.py
```diff
@@ -82,30 +82,36 @@ def node_pre_transform(self, node: Box, topology: Box) -> None:
...
-        f"Node {node.name} ipv6 loopback address {node.loopback.ipv6} overlaps with locator {locator}",
+        f"Node {node.name} ipv6 loopback address {node.loopback.ipv6} is not part of locator {locator}",
+        hint="Having the loopback as part of the locator space simplifies prefix distribution, and is recommended",
         category=Warning,
         module='srv6')
+
+
+  def node_post_transform(self, node: Box, topology: Box) -> None:
+    mods = node.get('module',[])
+    for igp in node.get('srv6.igp',[]):                   # Check if the IGP module is still active, it may have been removed
```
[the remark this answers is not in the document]
jbemmel: It won't. As I was creating various topologies, I'd change BGP AS'es and at times ended up removing the last internal link. The module checks for isis being present, but it may get removed - usually due to user error - so I figured I'd catch that as a courtesy to users (and myself)

docs/module/srv6.md
```diff
@@ -82,12 +84,12 @@ addressing:
...
-  loopback:
-    ipv4: False
-    ipv6: 2001:db8:cafe::/48
   srv6_locator:
-    ipv6: 2001:aa::/40       # Must not overlap with interfaces
+    ipv6: 5F00::/40          # Must not overlap with interfaces, this is from the IANA reserved range
     prefix6: 48              # Assign a /48 to each node
+  loopback:
+    ipv4: False
+    ipv6: FC00::/48          # Gets replaced by an address from the per-node locator range
```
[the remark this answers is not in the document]
jbemmel: To support loopback allocation from each node's unique locator prefix range, we cannot specify a single global prefix. Nodes that aren't running SRv6 may still need an ipv6 loopback, and even in cases where all nodes run SRv6 - like here - we cannot omit ```ipv6``` while also setting ```ipv4``` to `False`

I agree it's confusing, but I didn't see a cleaner way without modifying core functionality - so it's the proverbial rock and a hard place

netsim/ansible/templates/srv6/frr.bgp.j2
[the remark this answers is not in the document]
jbemmel: > And what happens when you want to implement global-IPv6-over-SRv6 when running BGP with customers?

You should probably create an issue, and we can take a look :)

I am focusing, I explore the boundaries of what's working and what not, and try to understand what's missing. There's always going to be some rope to hang yourself with, using 'ebgp' with the srv6 module is one example

ipspace (changes requested):
I made a slight addition to the IPv4 BGP integration test: I added a P-router (or whatever it's called in SRv6) in the middle, and the whole thing falls apart because the IPv6 NH for an IPv4 route is a link-local address, not a SRv6 SID.

There is no need to argue whether a 3-router network is a realistic network or not ;) I hope you'll figure out how to make this work without enabling BGP on the P-router, or srv6.bgp goes out of the PR -- I doubt it's supported by the current FRR release anyway.

jbemmel:
> OTOH, the VPNv4-over-SRv6 works flawlessly with a P-router in the middle. One more hint that maybe FRR does not support what you tried to do?

So this is what led me to experiment with disabling IPv6 LLA, unmodified next hops etc.

I can get things to work with IBGP and no IGP by adding static v6 routes to the loopbacks. That same setup even works with EBGP  and the same static routes, but when I try to replace the static route with BGP advertised ones I start getting the IPv6 LLA addresses as next hops and things stop working

ipspace:
> So this is what led me to experiment with disabling IPv6 LLA, unmodified next hops etc.

Throwing random spaghetti at the walls hoping they will stick is not a viable long-term strategy, unless you work for a networking vendor (see also https://archive.psg.com/051000.ccr-ivtf.pdf)

> I can get things to work with IBGP and no IGP by adding static v6 routes to the loopbacks.

Lovely. What could possibly go wrong? Next step: add an SDN controller to calculate the static routes?

> That same setup even works with EBGP and the same static routes, but when I try to replace the static route with BGP advertised ones I start getting the IPv6 LLA addresses as next hops and things stop working

Maybe it's time to admit FRR does not support all your ideas yet and clean up the PR to limit it to what works with FRR. The global IPv4-over-IPv6 you got "working" has nothing to do with SRv6 but with the way Zebra handles ARP tables. Even if your ideas might work on SR-OS I'm not going to merge something that works on a single device that's hard to get.

jbemmel:
> Lovely. What could possibly go wrong?

I am not suggesting that as a "solution" to implement, just to help understand where support/features are missing and things go wrong. In the use-case with P router the IS-IS routes are preferred over IBGP routes, and static routes are preferred over both.

jbemmel:
OK, it does look like regular BGP over SRv6 isn't supported by FRR - feature removed.

I guess we don't have a device to test those use cases available - shall we remove them, and only leave a VPN test case?

ipspace:
Just pushed my final tweaks.

The FRRouting "BGP services over SRv6" implementation is suboptimal (I would prefer using a 4-letter NSFW word):

* The only thing that works well is IPv4 L3VPN.
* IPv6 L3VPN seems to work only for connected subnets.
* IPv6 L3VPN breaks when run in parallel with the regular IPv6 AF.

I'm guessing that's why you disabled the regular IPv4/IPv6 AF in SRv6 module, which would (I'm pretty sure) kill your SR-OS implementation, so I commented that out and added relevant caveats and settings in the L3VPN integration test (which still fails for FRR, but maybe we could make that a warning).

Anyway, if you're OK with these tweaks I can merge this as-is.

Long story short: years later, SRv6 is still a half-baked idea (see also https://meckanix.substack.com/p/mpls-wc-2025-and-the-learnings-from)

jbemmel:
> Anyway, if you're OK with these tweaks I can merge this as-is.

I'm ok with it, thank you

ipspace (approved):
Let's merge this stuff and fix the edge cases later.

PR #2226, opened 2025-05-01
Make VirtualBox and direct Windows/macOS installations obsolete

### LINKED ISSUE #1629: [DOC] Declaring bare-bones Windows unsupported
There might be people running netlab (and Ansible) on Windows with VirtualBox. Should we declare that combination unsupported and ask people using Windows to move to WSL?

### LINKED ISSUE #1788: Remove VirtualBox support in release 2.0
I can't find a good reason to keep VirtualBox in netlab:

* I haven't seen a request to implement something for VirtualBox in years, and we only support it for a few devices.
* VirtualBox on MacOS no longer makes sense. The last x86 product was discontinued almost two years ago.
* WSL runs on Windows 10 Home edition for ~ 4 years (https://www.reddit.com/r/bashonubuntuonwindows/comments/j5c0xd/is_wsl_2_available_on_windows_10_home/)

Obviously, one could still use VirtualBox with netlab release 1.9.x (and I have no problems backporting critical fixes into a -postX release), but it would be gone from 2.0

ipspace:
WSL and multipass Ubuntu VM are much better alternatives.

Closes #1788, closes #1629

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 
 *[netlab](https://netlab.tools)* is bringing infrastructure-as-code concepts to networking labs. You'll describe your high-level network topology and routing design in a YAML file, and the tools in this repository will
 
-* Create *Vagrantfile* configuration file for *libvirt* or *VirtualBox* environment
-* Create *containerlab* configuration file
+* Create *Vagrantfile* configuration file for *libvirt/KVM* environment
+* Create *containerlab* configuration file to run Docker containers
 * Create Ansible inventory and configuration file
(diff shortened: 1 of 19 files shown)


ipspace:
@ssasso @jbemmel: I don't think anyone should be mightily upset with this change ;), would appreciate feedback on the wording.

jbemmel:
> @ssasso @jbemmel: I don't think anyone should be mightily upset with this change ;), would appreciate feedback on the wording.

I would suggest wording along the lines of "deprecated", adding that it will be completely removed in the next major release.

ipspace:
> I would suggest wording along the lines of "deprecated", adding that it will be completely removed in the next major release.

Thank you. Added.

PR #2228, opened 2025-05-02
Simplified Aruba CX build recipe (works with release 10.15)

ipspace:


ipspace:
I had a hunch that the "abrupt" shutdown of the VM might be causing the problems I experienced when building AOS CX 10.15 box, so I searched for a more deterministic VM shutdown, and it works like a charm.

Also: do we need the "netlab" user at all or could it be removed from the build recipe?

PR #2231, opened 2025-05-04
Check reserved ranges during IPv4/IPv6 address/prefix validation

### LINKED ISSUE #2229: [BUG] must_be_ipv4/subnet_prefix should check for reserved IP addresses
Lab:
```
---
addressing:
  loopback:
    ipv4: 0.0.0.0/24

groups:
  _auto_create: True
  routers:
    members: [ n1, n2 ]
    device: frr
    provider: clab

links:
- n1-n2
```

A corner case for sure, but it highlights several issues: ```netlab create lab.yml```
* Invalid range accepted for loopback.ipv4
* 'kind' is set to 'frr' in the generated clab.yml file; the clab provider setting of 'linux' doesn't get applied
* the invalid loopback 0.0.0.1/32 generates an error, but the Ansible script does not report it. The interface IP address does not get applied

ipspace:
Also:
* Split the address validation error test in prefix- and link/addr tests
* Fix the error messages in augment/links to use more_* arguments
* Change the static route prefix attribute type from 'subnet_prefix' to 'prefix'

Fixes #2229

### DIFF
--- a/netsim/data/types.py
+++ b/netsim/data/types.py
@@ -650,7 +650,7 @@ id            |          |    OK     |      | OK  |
 Furthermore, we can use 'named' prefixes in some scenarios.
 """
 
-RESERVED_PREFIX: typing.Dict[str,dict] = {
+RESERVED_PREFIXES: typing.Dict[str,dict] = {
   'IPv4': {
     'local':     ipaddress.IPv4Network('0.0.0.0/8'),
     'loopback':  ipaddress.IPv4Network('127.0.0.0/8'),
@@ -672,7 +672,7 @@ def common_addr_parse(
       xform_int: typing.Optional[typing.Callable] = None,
       xform_pfx: typing.Optional[typing.Callable] = None) -> dict:
 
-  global RESERVED_PREFIX
+  global RESERVED_PREFIXES
 
   def check_int_value(value: int, xform_int: typing.Optional[typing.Callable]) -> dict:
     if value < 0 or value > 2**32-1:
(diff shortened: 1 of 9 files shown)


netsim/data/types.py
```diff
@@ -669,6 +683,15 @@ def check_int_value(value: int, xform_int: typing.Optional[typing.Callable]) ->
       else:
         return { '_valid': True }
 
+  def check_reserved_range(
+        ranges: dict,
+        p_addr: typing.Any) -> typing.Optional[str]:
+    for k,v in ranges.items():
```
jbemmel, line 689: Suggestion to remove 'ranges' as a parameter and simply do:
```
def check_reserved_range( p_addr: typing.Any) -> typing.Optional[str]:
    for k,v in RESERVED_PREFIX[af].items():
```

jbemmel (approved):
Some minor tweaks possible

netsim/data/types.py
ipspace: Call me old-fashioned, but I don't like references to variables in a hard-to-figure-out scope (like outer functions) 🤷‍♂️

jbemmel: > Call me old-fashioned, but I don't like references to variables in a hard-to-figure-out scope (like outer functions) 🤷‍♂️

I took the use of nested functions as you getting less old-fashioned by the PR ;)

Alternatively, we could define a function near the RESERVED_PREFIXES:
```
def check_reserved_range(af: str, p_addr: typing.Any) -> typing.Optional[str]:
    for k,v in RESERVED_PREFIXES[af].items():
```

(I didn't like the 'global RESERVED_PREFIXES' at the top, only to be used way down)

PR #2233, opened 2025-05-04
Use shared functionality to computed BGP session attributes

### LINKED ISSUE #2198: [BUG] Unable to disable bgp.gtsm on a link when set at node level
Lab (based on ```integration/bgp.session/05-gtsm.yml```)
```
plugin: [ bgp.session ]
module: [ bgp ]

groups:
  probes:
    device: frr
    provider: clab
    members: [ x1, x2 ]

defaults.bgp.as: 65000

nodes:
  x1:
    bgp.as: 65100

  x2:
    bgp.gtsm: 2  # Test GTSM on 2-hop iBGP session, iBGP peer not part of this lab

links:
- x1: # For 2-hop iBGP session
  x2:
  bgp.gtsm: False
```

Result:
```
  File "/home/jeroen/Projects/netlab/netsim/data/validate.py", line 677, in validate_attributes
    fixed_data = validate_attributes(
                 ^^^^^^^^^^^^^^^^^^^^
  File "/home/jeroen/Projects/netlab/netsim/data/validate.py", line 649, in validate_attributes
    for k in data.keys():
RuntimeError: dictionary changed size during iteration
```

While this can easily be fixed using ```for k in list(data.keys()):``` the issue remains that the ```bgp.gtsm: 2``` setting gets applied to the EBGP session, even though the intent was to disable it

Note that `False` is not an int nor allowed by the datatype; using ```bgp.gtsm: 0``` is caught (disallowed)

ipspace:
* Instead of 'get from interface or node' code, use shared 'get_effective_module_attribute' functionality which now includes the 'removed_attribute' support.
* Define 'description' attribute (to be implemented in another PR) to simplify tracking of test result data ;)
* Add a transformation test checking out all inheritance combinations

Fixes #2198

### DIFF
--- a/netsim/extra/bgp.session/defaults.yml
+++ b/netsim/extra/bgp.session/defaults.yml
@@ -35,6 +35,7 @@ devices:
     as_override: True
     bfd: True
     default_originate: True
+    description: True
     gtsm: True
     passive: True
     password: True
@@ -75,6 +76,8 @@ devices:
     password: True
     remove_private_as: True
     timers: True
+  none:
+    copy: frr
   junos.features.bgp:
     allowas_in: True
     as_override: True
@@ -152,7 +155,7 @@ bgp:
     session:                  # All bgp.session attributes
       attr: [
         as_override, allowas_in, default_originate, password, passive,
-        gtsm, timers, tcp_ao, bfd, remove_private_as, rs ]
+        gtsm, timers, tcp_ao, bfd, remove_private_as, rs, description ]
     global:
       bfd:
         type: bool
@@ -205,6 +208,7 @@ bgp:
         max_value: 10
       as_override: bool
       default_originate: bool
+      description: str
       passive: bool
       password: str
       tcp_ao:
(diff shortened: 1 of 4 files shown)


netsim/extra/bgp.session/defaults.yml
```diff
@@ -205,6 +208,7 @@ bgp:
         max_value: 10
       as_override: bool
       default_originate: bool
+      description: str
```
jbemmel, line 211: Could consider to put this ```description``` attribute on the interface. Could be useful for other modules besides bgp?

I've sometimes used the ```name``` attribute of links to document test cases, but there is nothing yet for interfaces

ipspace: > Could consider to put this `description` attribute on the interface. Could be useful for other modules besides bgp?

I will probably make this a standard BGP attribute. It's the stuff that's configured on the BGP neighbor statement, and is not necessarily the same thing you'd want to see on an interface.

> I've sometimes used the `name` attribute of links to document test cases, but there is nothing yet for interfaces

Shall we make the "name" an interface attribute as well and check whether it exists before recreating it?

jbemmel (approved):
Wondering if we should go for a more generic ```description``` - but otherwise ok

PR #2234, opened 2025-05-04
Experimental: implement generic 'false -> remove' validation rule

ipspace:
Apart from the data types that have an explicit 'False' value (bool, IP addresses), this change removes all other 'x: False' attributes and uses '_removed_attr' functionality to make sure they are not overwritten.

Interestingly, all transformation tests pass with no changes.


ipspace:
I got a crazy idea: now that we have the `_removed_attr` functionality, let's try to extend the "False means remove" concept to everything that cannot have a legal False value (exceptions: mostly bools and addresses).

Unbelievably, transformation tests pass, I had to tweak a few error tests (because "x: False" no longer produced an error), and the integration tests all passed.

@jbemmel: Can you try to break it? ;)

PR #2235, opened 2025-05-04
The 'prefix6' attribute should be treated just like the 'prefix' attribute

jbemmel:
This removes nonsensical values such as 'True' and string-type '127'


ipspace:
> This removes nonsensical values such as 'True' and string-type '127'

No, it doesn't. All you did is prevent prefix6 to be copied into the node data (which is a good thing).

ipspace (approved):

PR #2247, opened 2025-05-07
Ubuntu: Postpone disabling ipv6 on interfaces until after netplan has been applied

### LINKED ISSUE #2245: Ubuntu initial script references 'bond' interfaces before they are created
https://github.com/ipspace/netlab/blob/f57b25bb693df1eeb57fb9f2c8bd034e5155cfcc/netsim/ansible/templates/initial/linux/ubuntu.j2#L93

In case of a 'bond' interface, the device doesn't exist until the Netplan config gets applied. The sysctl settings should be moved after that

Lab:
```YAML
---

groups:
  _auto_create: True
  nodes:
    members: [ n1, n2 ]
    device: linux
    provider: libvirt
    module: [ lag ]

links:
- lag:
    members: [ n1-n2 ]
```

jbemmel:
Fixes #2245

### DIFF
--- a/netsim/ansible/templates/initial/linux/ubuntu.j2
+++ b/netsim/ansible/templates/initial/linux/ubuntu.j2
@@ -78,24 +78,6 @@ systemctl enable lldpd
 systemctl restart lldpd
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Thank you!

PR #2248, opened 2025-05-07
LAG: Accept user specified lag.ifindex, also when listed in descending order

### LINKED ISSUE #2246: [BUG] lag.ifindex ignored when descending order is used
Lab:
```
---
groups:
  _auto_create: True
  nodes:
    members: [ n1, n2 ]
    module: [ lag ]

links:
- lag:
    ifindex: 2
    members:
    - n1:
      n2:
- lag:
    ifindex: 1
    members:
    - n1:
      n2:
```

An FRR device gets ```bond2``` and ```bond3```, not ```bond1``` as specified

jbemmel:
Fixes #2246

Tested: ```./device-module-test -d eos -p clab lag```

### DIFF
--- a/netsim/modules/lag.py
+++ b/netsim/modules/lag.py
@@ -123,16 +123,12 @@ def set_lag_ifindex(bond: Box, intf: Box, topology: Box) -> bool:
       if is_mside:                                             # For M: side we need matching lag.ifindex
         if next_ifindex is None:                               # Do we have a local preference?
           next_ifindex = _dataplane.get_next_id(ID_SET)        # No -> allocate globally
-        bond.lag.ifindex = link_lag_ifindex = next_ifindex  # Put on the link for the other M side to match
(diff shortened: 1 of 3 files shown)


ipspace:
This one looks a bit fishy. Have to check what happens if you have three LAGs and you number then "2, 1, none".

Will merge after 2.0.0 goes out.

jbemmel:
> This one looks a bit fishy. Have to check what happens if you have three LAGs and you number then "2, 1, none".
> 
That would assign overlapping ifindex values. The code is not currently trying to be overly sophisticated about auto-assigning missing ifindex values - it would mean keeping track of a set per node. Could be added

We could also change it such that the user has to specify all ```ifindex``` values if they decide to do one. I manually assign values to model what's configured, but in general I'd expect most users to leave them auto-assigned

ipspace (approved):
I'm not happy with the changes as they give the users too much rope to hang themselves with, but the integration tests pass, so I'm YOLO-merging this.

Issue #2251, opened 2025-05-08
Replace BGP RR/NHS logic with neighbor attributes


ipspace:
Instead of the current logic which goes along the lines of: is it an IBGP neighbor, and am I a RR, and is he not a RR... we should have per-neighbor **rr_client** (true) and **next_hop_self** (true/all) attributes that would be computed in the BGP module.

These attributes would then be used to in BGP templates to set neighbor options.

The logic should be backward-compatible (apart from EOS/FRR/IOS where it would also replace the fixed IBGP local-as settings), so the change wouldn't force us to redo/retest all templates, but I definitely hope we'll find it convenient to remove the old mess (OK, maybe not from the SONiC/IOS XR/RouterOS templates 😜)

ipspace:
Core functionality in #2575, template changes coming in separate PRs.

ipspace:
Device configuration templates in #2588

PR #2252, opened 2025-05-08
BGP: use neighbor attributes instead of per-template logic (FRR Proof-of-concept)

jbemmel:
Addresses #2251

Thoughts:
* Is ```rr_client``` relevant to MPLS VPN neighbors? Could create a macro that sets both flags, and include that everywhere for consistency (I've confirmed it works for EOS, but there may be other platforms where it's configured per AF?)
* None of the topology/expected test cases has ```next_hop_self: all```

### DIFF
--- a/netsim/ansible/templates/bgp/frr.j2
+++ b/netsim/ansible/templates/bgp/frr.j2
@@ -56,26 +56,18 @@ router bgp {{ bgp.as }}
   network {{ pfx|ipaddr('0') }}
 {%   endfor %}
 !
-{%   for n in bgp.neighbors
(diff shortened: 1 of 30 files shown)


ipspace:
Thanks for the effort, but I think I'll do it myself. I don't want to go into arguments like "why does next-hop-self make sense on RRs and what should it be applied to"

jbemmel:
> Thanks for the effort, but I think I'll do it myself. I don't want to go into arguments like "why does next-hop-self make sense on RRs and what should it be applied to"

as you prefer

PR #2253, opened 2025-05-09
Bird bgp features

jbemmel:
* timers
* BFD for bgp
* originate_default
* passive peering

Note that Bird could support per-interface BFD timers, but the current template only supports global timers to start with

I had to tweak the startup hold time to pass the 05-gtsm test more consistently (default = 240 seconds)


ipspace (approved):
Integration tests passed

Issue #2254, opened 2025-05-09
[BUG] Simple VXLAN broken on Cumulus Linux


ipspace:
#1840 and #1851 broke simple VXLAN on Cumulus Linux

01-vxlan-bridging:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vxlan/01-vxlan-bridging.yml)
* [Test results](https://tests.netlab.tools/cumulus/libvirt/vxlan/01-vxlan-bridging.yml-validate.log)

02-vxlan-bridging-multinode:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vxlan/02-vxlan-bridging-multinode.yml)
* [Test results](https://tests.netlab.tools/cumulus/libvirt/vxlan/02-vxlan-bridging-multinode.yml-validate.log)

03-vxlan-irb:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vxlan/03-vxlan-irb.yml)
* [Test results](https://tests.netlab.tools/cumulus/libvirt/vxlan/03-vxlan-irb.yml-validate.log)

@jbemmel: I have no idea why your changes don't work, but if I remove them, VXLAN works. The test results will be visible once the current batch of test results is uploaded, you can reproduce them from the **dev** branch. Please fix thus stuff.

ipspace:
FWIW, these are the only configuration changes between release 1.9.3 (where VXLAN works) and the current release (where it breaks):

```
+
+echo "INIT: Configuring system wide default MTU policy"
+cat > /etc/network/ifupdown2/policy.d/mtu.json <<CONFIG
+{ "address": { "defaults": { "mtu": "1500" } } }
+CONFIG
+#
 echo "INIT: creating loopback interface"
 #
 # Create loopback interface entry
@@ -65,6 +71,8 @@
   address 10.1.0.2/30
 auto swp2
 auto swp3
+auto vlan1000
+auto vlan1001
 CONFIG
```

jbemmel:
Unable to reproduce :(

At first I launched ```netlab up integration/vxlan/01-vxlan-bridging.yml -d cumulus``` and it failed validation. Then I played around with different image versions, didn't work, then commented out the system wide MTU file and pings started working. Then I put them back in, and pings still worked

ipspace:
> Unable to reproduce :(

It's random, but it's at least a 90% chance you'll hit the bug.

[...]

PR #2257, opened 2025-05-09
VXLAN test: Test DUT against reference implementation (FRR) instead of against itself

jbemmel:
This allows for the introduction of an OSPF check, clarifying delays a bit. It also guards against a theoretical possibility of devices that only interop with themselves

Proposal in the context of #2254

### DIFF
--- a/tests/integration/vxlan/01-vxlan-bridging.yml
+++ b/tests/integration/vxlan/01-vxlan-bridging.yml
@@ -54,8 +54,8 @@ validate:
 
   ping_red:
     description: Ping-based reachability test in VLAN red
-    wait_msg: Waiting for STP to converge
-    wait: 20
+    wait_msg: Waiting for STP to start forwarding
+    wait: stp_forwarding
     nodes: [ h1 ]
     plugin: ping('h2')
   ping_blue:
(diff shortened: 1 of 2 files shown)


ipspace (changes requested):
Makes perfect sense. Will merge once the integration tests I'm running finish.

The once change you should make is to increase the STP timer to 30+ seconds (35 should do the trick). It takes 30 seconds for STP to go from "learning" to "forwarding" and sometimes OSPF starts surprisingly fast.

Also, it would be nice to add the STP time to the wait_times (for example, as **stp_forwarding**)

ipspace:
OK, I'm back ;) and would like to discuss the bigger picture before merging a variant of this (adding @ssasso to the discussion):

A) The same change (one DUT, one or more FRR nodes) should be applied to all VXLAN tests (no worries, I'll do that)
B) That changes the meaning of "VXLAN works" (as in: between identical devices) to "VXLAN works with FRR". I think our FRR VXLAN implementation is solid enough to be a reference implementation (and of course I'll run the full set of integration tests before merging whatever we agree upon).
C) The "multivendor" test becomes obsolete.

Anything else I've missed?

jbemmel:
> B) That changes the meaning of "VXLAN works"

Since the EVPN tests are already structured like this (for the most part), the above is limited to "static VXLAN works" (and shouldn't be a big issue given that the EVPN tests pass)

Some EVPN tests (like 20-vxlan-irb-ospf.yml) could be converted too

PR #2259, opened 2025-05-09
Cumulus Linux: Change STP parameters on bridge and VXLAN interface

ipspace:
This seems to be the fix for #2254 -- I have no idea why things worked in the past, stopped working after changes mentioned in #2254, and behaved randomly afterwards. The root cause seems to be access ports stuck in **learning** phase forever (so clearly a bug somewhere in Linux networking code, but who's counting).

The only other significant thing that changed is b02bc24, but that should not have impacted interface configuration.

Anyway, let's wait until I can run the full suite of integration tests.


jbemmel:
I suspect things became "intermittent" after https://github.com/ipspace/netlab/commit/82ef4ad1f2555cea000e7687e812fd1b47951a48

Toggling the STP state on the bridge at just the wrong moment could leave ports hanging

Issue #2261, opened 2025-05-13
[BUG] Ansible crashes with 'loopback' attributes on a Linux host


sdargoeuves:
### Describe the bug

Not really a bug, as it's described in the [doc](https://netlab.tools/labs/linux/#loopback-interface), so it's expected that it doesn't work.

> netlab does not configure a global loopback IP address on Linux nodes with role set to host (the default netlab setting).
A loopback IP address is [allocated](https://netlab.tools/example/addressing-tutorial/#loopback-addresses) to a Linux node if you set role to any other value, and the initial configuration script configures an additional IP address on the lo interface, for example:

But maybe it could fail nicely by informing during the creation phase.

### To Reproduce

I was trying to create a loopback with an IPv4 to a linux host, which was working prior to `2.0.0` (I've just tested on `1.9.5`)

### Expected behavior

It would be nice to fail faster, during the `netlab create` phase

### Lab topology

```
  h1:
    device: linux
    provider: clab
    role: bridge
    loopback:
      ipv4: 172.16.0.50/32
```

### Output

```
TASK [Create container setup script from /home/sa/code/quick-netlab-lab/netlab/netsim/ansible/templates/initial/linux-clab.j2] *********************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'dict object' has no attribute 'ifname'
fatal: [h1 -> localhost]: FAILED! => changed=false 
  msg: 'AnsibleUndefinedVariable: ''dict object'' has no attribute ''ifname'''
changed: [h2 -> localhost]
```

### Version

`2.0.0`

jbemmel:
```
# Ansible inventory created from ['/tmp/lb.yml', '/home/jeroen/.netlab.yml', 'package:topology-defaults.yml']
#

af: {}
box: python:3.13-alpine
clab:
  binds:
  - clab_files/h1/hosts:/etc/hosts
  config_templates:
  - hosts:/etc/hosts
  kind: linux
hostname: clab-tmp-h1
interfaces: []
loopback:
  ipv4: 172.16.0.50/32
mgmt:
  ifname: eth0
  ipv4: 192.168.121.101
  mac: 08:4f:a9:01:00:00
module:
- vlan
mtu: 1500
role: bridge
vlan:
  mode: bridge
vlans:
  br_default:
    id: 1
    mode: bridge
    prefix:
      allocation: id_based
      ipv4: 172.16.0.0/24
```

[...]

PR #2262, opened 2025-05-14
Linux loopback fix

jbemmel:
Fixes https://github.com/ipspace/netlab/issues/2261

https://netlab.tools/labs/linux/#loopback-interface says Netlab will allocate a loopback IP to a Linux node with a role other than 'host'. This implies that the ```loopback_interface_name``` needs to be defined for Linux devices

* Move 'set_node_af' from links to nodes, and invoke it also when there are no links
* Move 'loopback_interface' from the router role to the node augmentation logic, and invoke it also when the user specifies a loopback interface
* Note the ```break``` instead of ```continue``` in ```set_node_af```

### DIFF
--- a/netsim/augment/devices.py
+++ b/netsim/augment/devices.py
@@ -67,8 +67,8 @@ def get_device_features(node: Box, defaults: Box) -> Box:
 """
 Get device loopback name (built-in loopback if ifindex == 0 else an additional loopback)
 """
-def get_loopback_name(node: Box, defaults: Box, ifindex: int = 0) -> typing.Optional[str]:
-  lbname = get_device_attribute(node,'loopback_interface_name',defaults)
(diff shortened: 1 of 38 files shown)


ipspace:
I think you were too aggressive (did I get it right that you disabled extra loopbacks on Linux?), while the other things (like setting **af** even when there are no links) makes perfect sense.

Let's first agree on what we want to do in #2261 and move from there. I might have to adjust this PR or start from scratch.

jbemmel:
> I think you were too aggressive (did I get it right that you disabled extra loopbacks on Linux?), while the other things (like setting **af** even when there are no links) makes perfect sense.
> 
Too conservative perhaps - I added the if-logic for host roles to keep the test result of ```link-loopback``` the same

jbemmel:
Replaced by https://github.com/ipspace/netlab/pull/2267

Issue #2266, opened 2025-05-14
[BUG] Every device MUST have "loopback_interface_name"


ipspace:
It's time to simplify things and fix old stuff: we have to make sure that every device has **loopback_interface_name** attribute if it wants to be a router. Not having that attribute should result in a fatal error (as part of the fix of #2261)

ipspace:
Fixed in fba0dedc84ec1289741e5dcf98661cc62c4a3631 (Linux), 44a5557d714022fffef87f7e59ea08bf92cff337 (IOS XR) and 9bbb46e2dbe4ef3784687f761637376e47e0ff30 (RouterOS6).

ASAv cannot have a loopback interface, we'll need further hacks in #2267

PR #2267, opened 2025-05-15
Allow loopback interfaces on all node types when specified by the user

ipspace:
This is the main logic for fixing #2261 and #2266. @jbemmel: would appreciate if you could double-check it.

Next, I have to create test cases to validate the edge conditions (**loopback.ipv4** on host, **loopback: True** and **loopback: False** work correctly), and finally run all known devices through these test cases to verify that the configuration templates don't break.

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -1077,15 +1077,11 @@ def copy_link_gateway(link: Box, nodes: Box) -> None:
         intf.gateway[af] = link.gateway[af]
 
 """
-Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family configured
+Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family configured on its interfaces
 """
 def set_node_af(nodes: Box) -> None:
   for n in nodes.values():
     for af in ['ipv4','ipv6']:
-      if af in n.get('loopback',{}):
-        n.af[af] = True
-        continue
-
       for l in n.get('interfaces',[]):
         if af in l:
           n.af[af] = True
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -1082,6 +1082,8 @@ Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family 
 def set_node_af(nodes: Box) -> None:
   for n in nodes.values():
     for af in ['ipv4','ipv6']:
+      if n.af.get(af) is True:
+        continue
       for l in n.get('interfaces',[]):
         if af in l:
           n.af[af] = True
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -199,6 +199,7 @@ Set up the default loopback interface
 """
 def augment_loopback_interface(n: Box, pools: Box, topology: Box) -> None:
   lb_value = n.get('loopback',None)
+  lb_ifname = devices.get_loopback_name(n,topology)
 
   if lb_value is False:                           # If 'loopback' is set to false, the user doesn't want it
     n.pop('loopback',None)                        # ... so remove any loopback data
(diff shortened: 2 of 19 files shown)


netsim/roles/router.py
```diff
@@ -12,44 +12,5 @@
 
 from . import select_nodes_by_role
```
jbemmel, line 13: No longer used :)

netsim/augment/nodes.py
```diff
@@ -194,6 +194,61 @@ def check_duplicate_mgmt_addr(topology: Box) -> None:
...
+  lb_value = n.get('loopback',None)
+
+  if lb_value is False:                           # If 'loopback' is set to false, the user doesn't want it
+    n.pop('loopback',None)                        # ... so remove any loopback data
+    return                                        # ... and get out
+  elif lb_value is True:                          # Or maybe the user is adamant he wants a loopback?
+    n.loopback = {}                               # ... we can do that, will start with no extra data
+  elif lb_value is None:                          # Oh, the user doesn't care
+    if n.get('role','router') != 'router':        # ... so they will not get a loopback for a non-router device
+      return
```
jbemmel, line 210: What about ```loopback: bla```?

netsim/augment/links.py
```diff
@@ -1077,15 +1077,11 @@ def copy_link_gateway(link: Box, nodes: Box) -> None:
         intf.gateway[af] = link.gateway[af]
 
 """
-Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family configured
+Set node.af flags to indicate that the node has IPv4 and/or IPv6 address family configured on its interfaces
 """
 def set_node_af(nodes: Box) -> None:
   for n in nodes.values():
     for af in ['ipv4','ipv6']:
```
jbemmel, line 1084: ```for af in ['ipv4','ipv6'] if af not in n.af```

jbemmel (changes requested):
Couple of corner cases

netsim/augment/nodes.py
```diff
@@ -194,6 +194,61 @@ def check_duplicate_mgmt_addr(topology: Box) -> None:
...
+  pool = n.get('loopback.pool','loopback')
+  prefix_list = addressing.get(pools,[ pool ],n.id)
+
+  for af in prefix_list:
+    if prefix_list[af] is True:
+      log.error(
+        f"Address pool {pool} cannot contain unnumbered/LLA addresses",
+        category=log.IncorrectType,
+        module='nodes')
+    elif not n.loopback[af] and not (prefix_list[af] is False):
```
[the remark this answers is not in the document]
ipspace: Thank you, will check and add to a test case

netsim/augment/links.py
ipspace: Yeah, we could do that optimization. Won't buy us much, but it's nice to have ;)

ipspace:
@jbemmel: Addressed all your corner cases (thank you), made loopback parameter processing as flexible as I possibly could, documented the changes, fixed the templates that broke without the loopback interface...

You might want to look at this again before I merge it.

tests/topology/expected/dual-stack.yml
```diff
@@ -238,7 +238,6 @@ nodes:
     loopback:
       ifindex: 0
       ifname: Loopback0
-      ipv4: 10.0.0.1/32
```
jbemmel, line 241: Is this expected? Topology has dual-stack loopback pool

tests/topology/expected/unnumbered.yml
```diff
@@ -36,12 +36,10 @@ links:
   - ifindex: 2
     ifname: Ethernet1/2
     ipv4: true
-    ipv6: true
```
jbemmel, line 39: Expected?

jbemmel (approved):
looks ok to me

Issue #2274, opened 2025-05-22
[BUG] Invalid CL NVUE OSPF configuration when running OSPF on VRF loopback


ipspace:
CL NVUE OSPF VRF configuration template generates device "lo1" (which is invalid) when a device has OSPF running on a VRF loopback. The error is here:

https://github.com/ipspace/netlab/blob/dev/netsim/ansible/templates/ospf/cumulus_nvue.j2#L29

I could "fix" that replacing l.ifname with 'loopback' but that would then crash if there's more than one loopback in the VRF. Either the configuration template has to be completely redesigned, or we need a device quirk that will throw an error when a VRF has more than one loopback on CL NVUE

### Lab topology

```
---
vrfs:
  tenant:
    loopback: True
    links: [ dut ]
    ospf.active: True

nodes:
  dut:
    device: cumulus_nvue
    module: [ vrf, ospf ]
```

jbemmel:
At first glance, I would think that ```ospf.active: True``` inside a VRF should cause ospf to be enabled for any loopbacks inside that VRF?

ipspace:
> However, the `lo1` loopback does not have OSPF enabled in the topology file:

VRF OSPF interfaces are in vrfs.x.ospf.interfaces, and the configuration template loops over those

ipspace:
> This may be true in general? i.e. a loopback inside a VRF can only belong to the area for that VRF, not any custom area?

Like the regular loopback, the default VRF loopback belongs to the VRF **ospf.area**. Additional loopbacks created with **links** can belong to any area.

PR #2277, opened 2025-05-22
Use VRF loopbacks in VRF OSPFv2/v3 integration tests

ipspace:
Triggered by #2274. Who knows what else will pop up

### DIFF
--- a/tests/integration/vrf/11-multi-vrf-ospf.yml
+++ b/tests/integration/vrf/11-multi-vrf-ospf.yml
@@ -12,26 +12,41 @@ message: |
 defaults.interfaces.mtu: 1500
 defaults.sources.extra: [ ../wait_times.yml ]
 
+module: [ ospf ]
 groups:
-  _auto_create: True
   ce:
     members: [ r1, r2, r3, r4 ]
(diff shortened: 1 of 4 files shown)


tests/integration/vrf/11-multi-vrf-ospf.yml
```diff
@@ -26,6 +26,7 @@ groups:
 vrfs:
   red:
     links: [ dut-r1, dut-r2 ]
+    loopback: True
```
jbemmel, line 29: I couldn't find where this is being tested? i.e. confirm that dut redistributes this loopback IP via OSPF correctly

ipspace: This effectively tests the configuration template (it should take IP addresses from **loopback_address** not from **loopback**)

ipspace:
> Do we want to add a test case using a different OSPF area within a VRF? See #2278 - Cumulus NVUE would fail that test

Good idea. Will add. Thank you.

ipspace:
The multi-VRF OSPF tests now use OSPF areas

PR #2280, opened 2025-05-22
Fix: implement stricter checks on node names on links

### LINKED ISSUE #2273: [BUG] Netlab crashes with invalid node name in a VLAN link
### Lab topology

```
---
defaults.device: eos
provider: clab

groups:
  _auto_create: True
  hosts:
    members: [ h3, h4 ]
    device: linux
    provider: clab
  switches:
    members: [ dut, s2 ]
    module: [ vlan, vxlan, ospf ]

vlans:
  blue:
    role: external
    links: [ dut-h3, dut-h3a, s2-h4 ]
```

### Output

```
[FATAL]   nodes: Internal error: node does not have a name {'id': 5, 'name': {}}
```

### Diagnosis

It looks like **h3a** is referenced without checking whether it exists. The topology has to be big enough to trigger the error (the above topology is the minimum one I could reproduce the bug with)

ipspace:
The code introduced with netlab components accepted any name in the link string definition as long as a valid node name was its prefix. That check was upgraded to 'node name + _ but only for components'

Furthermore, the content of 'links' list is rechecked after the component expansion to ensure the node names are correct (it's impossible to do the final check before components are expanded).

Fixes #2273

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -99,9 +99,12 @@ def adjust_link_object(l: typing.Any, linkname: str, nodes: Box) -> typing.Optio
     for n in l.split('-'):                # ... split it into a list of nodes
       n = n.strip()                       # ... strip leading and trailing spaces (fixing #816)
       valid_node = n in nodes
-      if not valid_node:
-        valid_node = len([ x for x in nodes if n.startswith(x) ]) > 0
-
+      if not valid_node:                  # ... maybe its a link to a node within a component
+        valid_node = len([ 
+          x for x in nodes                # ... so check the node names with a '_' suffix
+            if n.startswith(x+"_")        # ... but only for components
+              and 'include' in nodes[x]]) > 0
+                                          # ... note this is not a final check, it's only a viability check
       if valid_node:                      # If the node name is valid
         link_intf.append({ 'node': n })   # ... append it to the list of interfaces
       else:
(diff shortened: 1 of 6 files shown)


netsim/augment/links.py
```diff
@@ -99,9 +99,12 @@ def adjust_link_object(l: typing.Any, linkname: str, nodes: Box) -> typing.Optio
...
-        valid_node = len([ x for x in nodes if n.startswith(x) ]) > 0
-
+      if not valid_node:                  # ... maybe its a link to a node within a component
+        valid_node = len([ 
+          x for x in nodes                # ... so check the node names with a '_' suffix
+            if n.startswith(x+"_")        # ... but only for components
+              and 'include' in nodes[x]]) > 0
+                                          # ... note this is not a final check, it's only a viability check
       if valid_node:                      # If the node name is valid
         link_intf.append({ 'node': n })   # ... append it to the list of interfaces
```
jbemmel, line 109: Could opt to use a different attribute name, e.g. ```_unvalidated_node``` and then rewrite that in ```validate_link_nodenames``` as a way to check that all code paths properly validate node names, and no logic uses unvalidated node names

jbemmel (approved):
Optional suggestion to add some robustness (and complexity)

ipspace:
> Optional suggestion to add some robustness (and complexity)

+1 on complexity ;)

PR #2281, opened 2025-05-22
VXLAN integration tests use FRR as the probe device

ipspace:
Also:
* VNI is set to 5000/5001 to deal with ArubaCX FUBAR
* Validation tests check individual components (OSPF, STP...) before an end-to-end ping, hopefully resulting in easier troubleshooting
* Test use symbolic wait times for consistency
* Removed unnecessary device from alt-vtep test
* Removed multi-vendor test (all tests are now multi-vendor)
* Remove OSPF-in-VRF from router-on-stick test to work around a CL NVUE bug
* Use test.fixup plugin to change probe device to EOS when testing ArubaCX

Replaces #2257


tests/integration/vxlan/01-vxlan-bridging.yml
```diff
@@ -11,48 +11,77 @@ message: |
...
+    members: [ dut, s2 ]
     module: [ vlan, vxlan, ospf ]
+  probes:
+    members: [ s2 ]
+    device: frr
+    provider: clab
 
 vlans:
   red:
+    vni: 5000
```
jbemmel, line 32: Do we have/need a device quirk that warns about this?

jbemmel (approved):
Some minor side issues

tests/integration/vxlan/01-vxlan-bridging.yml
ipspace: We don't have one, but we could definitely use it. Thank you!

ipspace: Implemented in #2283

Issue #2278, opened 2025-05-22
Cumulus NVUE template does not configure VRF OSPF area


jbemmel:
https://github.com/ipspace/netlab/blob/a8e84884c3116faee9057699ec751c38a2655420/netsim/ansible/templates/ospf/cumulus_nvue.j2#L15

Triggered by #2274, there seems to be a more general issue here: Cumulus NVUE implements a subset of the full spectrum of configuration options for FRR; while FRR allows one to configure a different OSPF area for a loopback interface inside a VRF, the Cumulus NVUE abstraction layer does not

Need to fix this, making sure that https://github.com/ipspace/netlab/pull/2277 catches the omission (if not already)

ipspace:
Fixed in #2282

PR #2282, opened 2025-05-22
Cumulus NVUE vrf loopback OSPF fixes

### LINKED ISSUE #2274: [BUG] Invalid CL NVUE OSPF configuration when running OSPF on VRF loopback

jbemmel:
Cumulus NVUE has some limitations due to the way VRF loopbacks are represented in its data model. It will redistribute loopback IP addresses inside a VRF, but cannot configure OSPF attributes for those interfaces that are different from the parent VRF

* Configure the per-VRF OSPF area (which could differ)
* Skip VRF loopback interfaces when configuring OSPF interfaces
* Don't second-guess the loopback ospf configuration, use what's provided

Fixes #2274 and #2278

### DIFF
--- a/netsim/ansible/templates/vrf/cumulus_nvue.j2
+++ b/netsim/ansible/templates/vrf/cumulus_nvue.j2
@@ -24,15 +24,18 @@
 
 {# Workaround for lack of NVUE support for VRF route leaking - note can only create 1 snippet for frr.conf globally #}
 {% set import_lengths = vrfs.values()|map(attribute='import',default=[])|map('length')|list %}
-{% if import_lengths|max > 1 %}
+{# Workaround for lack of control over VRF loopbacks and OSPF area config #}
+{% set ospf_loopbacks = vrfs.values()|map(attribute='ospf.interfaces',default=[])|flatten|selectattr('type','eq','loopback')|list %}
+{% if import_lengths|max > 1 or ospf_loopbacks %}
 - set:
     system:
       config:
         snippet:
           frr.conf: |
-{%   for vname,vdata in vrfs.items() %}
+{%  if import_lengths|max > 1 %}
+{%    for vname,vdata in vrfs.items() %}
             router bgp {{ bgp.as }} vrf {{ vname }}
-{%     for _af in ['ipv4','ipv6'] if _af in vdata.af %}
+{%      for _af in ['ipv4','ipv6'] if _af in vdata.af %}
               address-family {{ _af }} unicast
                 label vpn export auto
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
VRF IPv6 loopbacks result in configuration errors because you're trying to configure ::/128 on the loopback and the modified OSPFv2 multi-VRF test fails validation because the loopback interface is not in area 0 and so DUT cannot become an ABR in blue VRF

Details: https://tests.netlab.tools/_html/cumulus_nvue-libvirt-vrf

jbemmel:
> trying to configure ::/128 on the loopback

image

I apologize, copied without testing.

It looks like Cumulus NVUE may not be able to support scenario's with different OSPF areas in VRFs - will test further

ipspace:
> `netlab up integration/vrf/11-multi-vrf-ospf.yml -d cumulus_nvue -v` passes on my system

Did you try the new test with OSPF areas (part of #2277, now in dev branch)?

jbemmel:
> > `netlab up integration/vrf/11-multi-vrf-ospf.yml -d cumulus_nvue -v` passes on my system
> 
> Did you try the new test with OSPF areas (part of #2277, now in dev branch)?

I did: ```[FAIL]        12 tests completed, 3 tests failed```

I modified the templates to configure the OSPF area per VRF, but NVUE doesn't seem to do anything with that information. VRF loopbacks are not added to the correct OSPF area

I can manually modify the FRR config and add the OSPF config to the vrf interfaces, but I was hoping the NVUE templates would do this for me.

Looks like we may have to add a quirk checking for VRF loopbacks icm OSPF and declare defeat

jbemmel:
Nvm - it works like this :)

ipspace (approved):
All integration tests pass now. Thanks a million -- that was a nasty can of worms.

PR #2283, opened 2025-05-22
Add device quirk for out-of-range VNI on ArubaCX

ipspace:


ssasso (changes requested):
Thank you!!

Could you please add a check for warn only if not using 'external' provider?

This is a limitation only on the virtual version.

ssasso (approved):
Thanks!

PR #2288, opened 2025-05-23
Fix DNS resolution failures on Debian nodes with '_' in node name

### LINKED ISSUE #2286: [BUG] FRR Debian nodes with underscore in name are incredibly slow
FRR Debian nodes cannot set hostname with '_' and omit the underscore from the hostname, resulting in
DNS lookup failures and incredibly slow **vtysh** (and other things).

To fix that, we have to:

* replace '_' with '-' (similar to many other devices) in FRR initial config template
* Add entries with '-' instead of '_' to /etc/hosts

The final /etc/hosts would thus something like:

````
10.0.0.1 s1-p s1_p
10.1.0.1 eth1.s1_p
10.1.0.5 eth2.s1_p
10.1.0.9 eth3.s1_p
10.0.0.2 s2-p s2_p
10.1.0.2 eth1.s2_p
10.1.0.13 eth2.s2_p
10.1.0.17 eth3.s2_p
````

ipspace:
* Replace the '_' with '-' in hostnamectl call to get the same hostname as on Linux (affects only FRR VMs)
* Add entries with '-' for every host with '_' in its name to the /etc/hosts file (affects all Linux-based nodes)

Fixes #2286

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -36,7 +36,7 @@ SCRIPT
 #
 # Configure system defaults on Ubuntu
 #
-hostnamectl set-hostname {{ inventory_hostname }}
+hostnamectl set-hostname {{ inventory_hostname.replace('_','-') }}
 {% include 'linux/packages.j2' %}
 #
 # Install FRR on a Ubuntu VM if needed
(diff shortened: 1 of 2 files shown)


jbemmel:
Reminds me of #1619 …

ipspace:
> Reminds me of #1619 …

🤷‍♂️ Neverending story ;))

PR #2306, opened 2025-05-27
Simplified SR OS configuration deployment

ipspace:
* Move gNMI checks into readiness tasklist
* Cleanup the configuration deployment tasklist to use Jinja2 lookup with gNMI-only deployment
* Remove MD-CLI templates
* Generate complete interface name in device data
* Generate 'loopback.X' or 'system' loopback name in device data
* Treat system interface like any other interface in initial configuration deployment

Minor changes:
* Add VLAN to interface description only if VLAN != 0
* Set loopback interface description to "loopback"

### DIFF
--- a/netsim/ansible/templates/initial/sros.j2
+++ b/netsim/ansible/templates/initial/sros.j2
@@ -78,12 +78,12 @@
 {%  set v6 = l.ipv6|default(False) %}
 {%  set in_base = l.vrf is not defined and (l.vlan is not defined or l.vlan.mode|default('irb')=='route') %}
 
-{%  if l.type == "loopback" %}
+{%  if l.ifname == "system" %}
+{%   set desc = "System interface" %}
+{%  elif l.type == "loopback" %}
 {%   set desc = "Loopback" %}
 {%  elif l.name is defined %}
 {%   set desc = l.name|replace('->','~') + " (" + l.role|default('') + ")" %}
-{%  elif l.ifname == "system" %}
-{%   set desc = "System interface" %}
 {%  elif l.type == "stub" %}
 {%   set desc = "Stub interface" %}
 {%  endif %}


netsim/ansible/templates/initial/sros.j2
```diff
@@ -80,13 +78,18 @@
...
 {%   set desc = l.name|replace('->','~') + " (" + l.role|default('') + ")" %}
-{%  elif l.type|default("") == "stub" %}
+{%  elif l.ifname == "system" %}
+{%   set desc = "System interface" %}
+{%  elif l.type == "stub" %}
 {%   set desc = "Stub interface" %}
 {%  endif %}
 
 {%  set portname = l.parent_ifname if l.type=='vlan_member' else l.ifname %}
+{%  set portname = portname.replace('eth-','') %}
```
jbemmel, line 92: Not sure I follow why this is necessary - why not remove 'eth-' from ```interface_name```?

netsim/ansible/templates/initial/sros.j2
```diff
@@ -80,13 +78,18 @@
...
 {%  set in_base = l.vrf is not defined and (l.vlan is not defined or l.vlan.mode|default('irb')=='route') %}
 
-{%  if l.name is defined %}
+{%  if l.type == "loopback" %}
+{%   set desc = "Loopback" %}
+{%  elif l.name is defined %}
 {%   set desc = l.name|replace('->','~') + " (" + l.role|default('') + ")" %}
-{%  elif l.type|default("") == "stub" %}
+{%  elif l.ifname == "system" %}
+{%   set desc = "System interface" %}
```
jbemmel, line 86: Does this ever get reached? System is a loopback, so I think this check should come first

netsim/ansible/templates/initial/sros.j2
ipspace: Your solution created interface names (starting with "i" or "stub") in Jinja2 template. That causes a discrepancy between the device printouts and "netlab report" printouts, and might confuse people using "netlab capture".

This approach makes the interface names consistent, but we have to reverse-engineer portname from interface name.

PR #2309, opened 2025-05-30
Update MANIFEST.in to include netsim/daemons sub directories

jdpoland:
Enable the inclusion of sub directories under netsim/daemons in the distribution artifacts. This includes the  Dockerfiles and jinja2 templates for bird and dnsmasq.


ipspace:
Thanks a million. I have to admit I believed in "fake it till you make it" approach to Python packages, and obviously I faked and failed :(

Will push out a new release in a few days (I have a few other things I'd like to fix).

ipspace (approved):

PR #2324, opened 2025-06-04
JunOS: support for LAG module

ssasso:
Tested:

* 01-l3-lag.yml
* 02-lag-vlan-trunk.yml
* 03-l3-lag-passive.yml
* 04-lag-vlan-routed-trunk.yml

With:

* vjunos-switch
* vptx
* vsrx in *packet mode* (except for 02-lag-vlan-trunk and 03-l3-lag-passive.yml which use vlans - is not supported)

**NOTE**: vjunos-router does not support LAG:

```
 ##
 ## Warning: configuration block ignored: unsupported platform (vmx)
 ##
```


ipspace (approved):
Integration tests pass (on vJunos-switch), so I guess we're good to go ;)

PR #2325, opened 2025-06-04
JunOS: remove MPLS bgp-igp-both-ribs

ssasso:
Following @nmodena's advice, removing `bgp-igp-both-ribs` from default MPLS config.


ipspace:
This is probably the first case of indirect PR in this repo ;)

PR #2329, opened 2025-06-04
SR-OS: Configure SR-MPLS for IPv6

ipspace:
* SR-MPLS must be enabled in IS-IS MT2 topology
* Resolution of BGP IPv6 next hops over SR tunnels must be configured


ipspace:
SR-MPLS for IPv6 was obviously added to the config template (**ipv6-node-sid**) but never tested. I'm no longer surprised.

PR #2332, opened 2025-06-06
Draft: Enable running of vPTX on AMD CPUs

jbemmel:
Draft PR for https://github.com/ipspace/netlab/discussions/2331 - to be tested

### DIFF
--- a/netsim/cli/libvirt_actions/package.py
+++ b/netsim/cli/libvirt_actions/package.py
@@ -12,6 +12,7 @@ import pathlib
 import shutil
 import tarfile
 import re
+import platform
 
 from box import Box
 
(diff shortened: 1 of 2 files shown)


ipspace:
Python 3.12 returns generic CPU info for the Ryzen CPU:

```
$ lscpu|grep -A 3 Vendor
Vendor ID:                            AuthenticAMD
Model name:                           AMD Ryzen 9 8945HS w/ Radeon 780M Graphics
CPU family:                           25
Model:                                117

$ python3
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.processor()
'x86_64'
```

It could be that the CPU is too new for Python shipping with Ubuntu 24.04. In any case, it looks like we'll have to check /proc/cpuinfo

FWIW, that info is funny as well. I get "GenuineIntel" and "AuthenticAMD" in Vendor ID 🤦‍♂️

ipspace:
Replaced by a7ddfe9d2684955893f72e196facb180cf9e04a1 -- I had to use a different method to find the CPU type, and the fixes had to be applied in two places (VM XML definition and Vagrantfile)

PR #2338, opened 2025-06-08
Import (redistribute) static routes into OSPF on EOS/FRR/IOS

ipspace:
The existing configuration templates work, the only change is the extra keyword in the 'ospf.import' feature flag.

The main modifications of this PR are the changes to the integration tests that use a new 'static_import' plugin to adjust the test based on whether the DUT supports static routes and redistribution of static routes into OSPF.

### DIFF
--- a/tests/integration/plugin/static_import/plugin.py
+++ b/tests/integration/plugin/static_import/plugin.py
@@ -1,0 +1,74 @@
+#
+# Check whether the device under test supports VRFs and removes
+# VRF links and related tests if needed
+#
+
+import typing
+from box import Box
+from netsim.augment.devices import get_device_features
+from netsim.utils import log
+from netsim.data import get_box
+
+def remove_validation(topology: Box, warning: str) -> None:
+  for tname,tdata in get_box(topology.validate).items():
+    if 'static' in tname:
+      log.info(f'Removing validation test {tname}')
+      topology.validate.pop(tname,None)
+
+  topology.validate.r_warning = {
+    'wait': 0,
+    'level': 'warning',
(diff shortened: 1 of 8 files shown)


tests/integration/plugin/static_import/plugin.py
```diff
@@ -0,0 +1,74 @@
+#
+# Check whether the device under test supports VRFs and removes
```
jbemmel: Description needs to be updated

tests/integration/plugin/static_import/plugin.py
```diff
@@ -0,0 +1,74 @@
...
+    'wait': 0,
+    'level': 'warning',
+    'fail': warning }
+
+def is_static_import(attr: typing.Any) -> bool:
+  if not isinstance(attr,Box):                        # Current node attribute is not a box, move on
+    return False
+  if 'import' not in attr:                            # ... or it does not have 'import' key
+    return False
+  if not isinstance(attr['import'],Box):              # ... or the 'import' key is not a dictionary
```
jbemmel, line 28: Could it be a List?

jbemmel:
Cumulus_NVUE template should also work, but I guess we can update other devices in a next PR

jbemmel (changes requested):
Just a minor copy-paste documentation issue

ipspace:
> Cumulus_NVUE template should also work, but I guess we can update other devices in a next PR

I will enable feature flags in other devices and run the tests to see what works. However, it seems that Cumulus NVUE does not support static routes yet.

PR #2339, opened 2025-06-09
fortios - mtu and docs

sdargoeuves:
After playing with the FortiGate, I've successfully reproduced the creation of the Vagrant box using the current latest versions (`7.4.8` and `7.6.3`)
I have noticed a weird thing, on `7.4.8`, the MTU on the link between the fw and the device in front, was set to 9500, but it is 1500 on the `7.6.3`. I've only noticed when realising OSPF was configured but not established.

I've added a small fix for this in the yml file, so the MTU will be defaulted to 1500 on `fortios`, unless set in the interface parameter (not in node).

This is the PR I've created, with some documentation updates (caveats and platforms). Please let me know if I haven't followed the PR process correctly, or anything else I may have done incorrectly, I'm still trying to figure out how to correctly work with fork...

I have also documented the whole process to add a fortigate to *netlab* [here](https://noodleops.space/2025-06/add-a-fortigate-in-your-virtual-lab-from-qcow-to-netlab-by-creating-a-vagrant-box/)

### DIFF
--- a/netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
+++ b/netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
@@ -54,6 +54,8 @@
       lldp_reception: "enable"
       lldp_transmission: "enable"
       mode: "static"
+      mtu: "{{ interface.mtu | default(1500) }}"
+      mtu_override: "enable"
       name: "{{ interface.ifname }}"
       macaddr: "52:dc:ca:fe:{{ id }}:{{ interface.ifindex }}"
       type: "physical"
--- a/netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
+++ b/netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
@@ -54,7 +54,7 @@
       lldp_reception: "enable"
       lldp_transmission: "enable"
       mode: "static"
-      mtu: "{{ interface.mtu | default(1500) }}"
+      mtu: "{{ interface.mtu | default(mtu) }}"
       mtu_override: "enable"
       name: "{{ interface.ifname }}"
       macaddr: "52:dc:ca:fe:{{ id }}:{{ interface.ifindex }}"
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -279,6 +279,14 @@ diag debug application httpsd -1
 * We're not testing Fortinet implementation as part of the regular integration tests; the configuration scripts might be outdated. If you encounter a problem, please open an issue.
 ```
 
+## Fortinet FortiOS 7.4.x/7.6.x
+
+* Starting from FortiOS 7.2, FortiGate devices do not come with a license out of the box. Users can link *one* device with a permanent evaluation license to an account on the Fortinet support portal.
+* The license needs to be added before creating the Vagrant box.
+* There are restrictions associated with the evaluation license, including a maximum of three interfaces, firewall policies, and routes... For more detailed information, refer to the [evaluation license restrictions](https://docs.fortinet.com/document/fortigate/7.6.3/administration-guide/441460).
+* The license is linked to the serial number of the device and the UUID. To ensure that the serial number remains consistent each time you start the lab, set the `libvirt.uuid` node parameter to the appropriate value.
+* MTU can be defined on the interface level, default is forced to 1500 bytes due to a different behaviour between `7.4.8` and `7.6.3` releases.
+
 ### OSPF Caveats
 
 * Fortinet implementation of OSPF configuration module does not implement per-interface OSPF areas. All interfaces belong to the OSPF area defined in the node data.
--- a/docs/labs/fortios.md
+++ b/docs/labs/fortios.md
@@ -9,7 +9,10 @@ FortiOS (FortiGate) is supported by the **netlab libvirt package** command. To b
 * Execute **netlab libvirt package fortios _qcow-file-name_** and follow the instructions
 
 ```{warning}
-* _netlab_ supports FortiGate devices that use username/password to authenticate API calls. The last software releases known to work are 7.0.x and 7.2.0.
+* _netlab_ supports FortiGate devices that use username/password to authenticate API calls.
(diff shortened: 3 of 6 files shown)


ssasso:
Hello @sdargoeuves that looks great!

If you can, I would suggest to separate the management traffic (on the first interface) from the rest, like we do on other devices with the management VRF. This will allow to avoid route conflicts between the management interface and the rest of the topology.

The permanent trial mode for FortiGate-VM (according to the documentation) there is support for a maximum of two virtual domains (VDOM). When using multi-VDOM mode, the root VDOM must be an admin type and the other can be a traffic VDOM. The admin vdom, practically speaking, is the management vrf.

I think it could be great if you can link the post with the instructions directly in the documentation, like we do here (as an example): https://github.com/ipspace/netlab/blob/dev/docs/labs/routeros7.md

thanks!

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
```diff
@@ -54,6 +54,8 @@
       lldp_reception: "enable"
       lldp_transmission: "enable"
       mode: "static"
+      mtu: "{{ interface.mtu | default(1500) }}"
```
ipspace: I try to avoid Jinja2 defaults if possible. I think it might be better to set the **mtu** to 1500 in the device data (similar to **dellos10**)

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
```diff
@@ -54,6 +54,8 @@
...
+      mtu_override: "enable"
```
ipspace, line 58: It would be nice to use the {% if interface.mtu is defined %} block around these two commands.

ipspace (commented):
Nice job, thank you. I would use a slightly different approach to the MTU defaults (see comments).

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
ssasso: this is an ansible play->task, I am not so sure you can use "if" here. But you should be able to use `|default(omit)`

ipspace:
Update: Ignore my MTU suggestions, but please do add either the installation recipe or a link to it into `docs/labs/fortios.md`

What @ssasso suggested also makes sense, but I have no idea what the overall impact of that change would be -- we'd probably have to change all configuration tasks to use a different VDOM, so I would definitely suggest to do that in a separate PR.

ssasso:
> but I have no idea what the overall impact of that change would be -- we'd probably have to change all configuration tasks to use a different VDOM, so I would definitely suggest to do that in a separate PR.

Only need to change it here: https://github.com/ipspace/netlab/blob/7f8866d4c136bbd57fce1ab60007d5afc2d5271b/netsim/devices/fortios.yml#L18

Playbooks only use a reference to that variable.

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
sdargoeuves: based on your latest comment, are you happy if I leave it like this, or should I look at adding the default mtu in the device data?
I can see where I can add this default value in the `netsim/devices/fortios.yml` file, I'm just not sure how to access this variable in this `initial.yml` file 🤔 
The logic for `fortios` is quite different from `dellos` or other platforms, as it's not using the `j2`.

If it makes sense, I could investigate how to change this and apply similar logic as the `dellos` to the `fortios`. But if it's just an odd use case, and you are ok with the current implementation, we could keep it as it is.

ssasso: @sdargoeuves you can take as an example the main vdom name: it's defined as device "variable" on the device file, and then it's used on the ansible tasks.

sdargoeuves: Thanks, I'm starting to wrap my head around it! I'll play with it later

sdargoeuves:
(sorry for the mess, I was wondering how to update this PR with the latest update, but it seems to be doing it automagically!)

I've updated the docs, hopefully accordingly to your suggestions!

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
sdargoeuves: I've tested this `default(mtu)` after adding the mtu in the `devices/fortios.yml` and it now works as expected.

sdargoeuves:
> > but I have no idea what the overall impact of that change would be -- we'd probably have to change all configuration tasks to use a different VDOM, so I would definitely suggest to do that in a separate PR.
> 
> Only need to change it here:
> 
> https://github.com/ipspace/netlab/blob/7f8866d4c136bbd57fce1ab60007d5afc2d5271b/netsim/devices/fortios.yml#L18
> 
> Playbooks only use a reference to that variable.


It doesn't seem that straightforward, at least with my current understanding. Should I create another vdom for the normal traffic and use the root vdom for the management (interface port1)? If that's the case, I would create this in a different PR, as the current playbook would have to be updated to take into account the correct vdoms for the initialisation.

ssasso:
> It doesn't seem that straightforward, at least with my current understanding. Should I create another vdom for the normal traffic and use the root vdom for the management (interface port1)? If that's the case, I would create this in a different PR, as the current playbook would have to be updated to take into account the correct vdoms for the initialisation.

In other images, we configure the management vrf and related interfaces during the VM/Image creation.

See i.e., https://github.com/ipspace/netlab/blob/4e21bc58cb8472bf965457e18c73ae4eb62c17db/netsim/install/libvirt/eos.txt#L35

I was thinking the same could be done for FortiOS.

ipspace:
I tried to test the changes, but it seems like my license expired (hooray for an Ansible module that throws a screen-long error message when that happens), so I'm YOLO-merging this ;)

sdargoeuves:
thank you so much!

PR #2344, opened 2025-06-09
JunOS: support for discard static route

ssasso:


ssasso:
Side note: 
Test `ping_h2_v6` fails at the first time, while succeeds at the second run.

Investigating.

ipspace:
> Side note: Test `ping_h2_v6` fails at the first time, while succeeds at the second run.
> 
> Investigating.

Probably late arrival of IPv6 RA. Will add that step to the integration test

ssasso:
Update:

Adding
```
  ra:
    description: Check RA-generated default route
    wait: ra_send
    wait_msg: Waiting for RA message to generate the default route
    nodes: [ h1, h2 ]
    plugin: default6()
    stop_on_error: True
```
like we do on `initial/07-router.yml` before the ping (and including wait_times) seems to make it work.

ssasso:
> Probably late arrival of IPv6 RA. Will add that step to the integration test

yeah, we wrote the same at the same time ;)

ipspace:
> Probably late arrival of IPv6 RA. Will add that step to the integration test

Added in e56d8f15dc96becce040faa9413518d03557cd45

ipspace (approved):
Looks good to me and the integration tests pass ;) Thank you!

PR #2350, opened 2025-06-09
FortiOS: increase required RAM to 2G

### LINKED ISSUE #2347: [BUG] Fortigate wrong RAM value in Vagrant domain template
It seems the latest versions of FortiOS is not able to boot with 1G of RAM - they require 2G.

However, on the Vagrant template, 1024M is set.

https://github.com/ipspace/netlab/blob/e53969d08cf9e839d6709ea103d7d8989b343d04/netsim/templates/provider/libvirt/fortios-domain.j2#L7

@sdargoeuves on your testing did you modify this as well, or were you able to run with only 1G?

I tried both 7.4.8 and 7.6.3 and they both get stuck with 1G ram only.

ssasso:
Fix #2347


ipspace (approved):
Thank you!

PR #2353, opened 2025-06-09
Configure BGP after IGP protocols

ipspace:


ipspace:
Looking at the code, I couldn't see why this wouldn't work (the famous last words), so I'll YOLO-merge it and hope for the best. We'll run a few integration tests between now and the next release, so the unexpected quirks should surface sooner rather than later.

PR #2354, opened 2025-06-09
Dell OS10: bgp route redistribution from ospf, connected, static, vrf

jbemmel:
Depends on the PR to configure OSPF before BGP

Tested:
* bgp/30,31
* vrf/15


ipspace:
Please rebase on dev. Thank you

ipspace (approved):
Looks good and the integration tests passed ;)

PR #2360, opened 2025-06-10
Draft: OSPF stub plugin

jbemmel:
For discussion - needs more documentation and validation tests, but I'd first like to agree on the general approach before putting more effort

### DIFF
--- a/netsim/extra/ospf.stub/plugin.py
+++ b/netsim/extra/ospf.stub/plugin.py
@@ -3,6 +3,7 @@ from box import Box
 from netsim import api
 from netsim.augment import devices
 from netsim.utils import log
+from ipaddress import IPv4Address
 
 _config_name = "ospf.stub"
 _require = [ "ospf" ]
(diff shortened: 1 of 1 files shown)


docs/plugins/ospf.stub.md
```diff
@@ -0,0 +1,42 @@
...
+This enables support for ospf.areas:
+
+```
+module: [ ospf ]
+
+ospf.areas:
+  1:
+    kind: stub
+    no_summarize: True
+```
```
ssasso, line 42: Suggest to add also the param `range` as an array of subnets for area range config - to be applied only on ABR

ipspace:
Yeah, I had that on the backburner for ages (https://github.com/ipspace/netlab/tree/ospf.areas), maybe it's time to finally get it done (after pushing out 25.06). However, it needs to address all types of OSPF areas, not just the stub areas. So we need:

* kind - stub/nssa (or regular if kind is missing)
* summary: False (default: true)
* default: False (default: true) -- for stub/nssa areas
* range: list of summarization ranges
* filter: suppressed prefixes

I had the data model (almost) worked out in https://github.com/ipspace/netlab/commit/711f49ec5ba064c04b24a884f68aab9324fe6fbc, the real show-stopper (why this isn't already out) was the dictionary. I wanted to use both area numbers as well as area addresses ("0.0.0.0") as the dictionary entries, but of course "0.0.0.0" gets exploded by Box and needs to be recreated.

Looks like the only way to get that solved would be to have a list of areas, with every element having **area** key, and the plugin reverse-engineering the list from the dict if needed.

Give me a few days ;)

jbemmel:
> 
> Give me a few days ;)

I ran into the ipv4 thing as well, perhaps https://github.com/cdgriffith/Box/pull/297 could help (if accepted)

jbemmel:
Replaced by #2363

PR #2361, opened 2025-06-11
Release notes for release 25.06

ipspace:


jbemmel:
Could mention the new version scheme

ipspace:
> Could mention the new version scheme

You're right. Will do that in the blog post announcing the release.

sdktr:
> Could mention the new version scheme

I thought I just woke up from a long winter sleep😅

PR #2363, opened 2025-06-12
OSPF area parameters

ipspace:
Supported parameters:

* Stub/NSSA areas
* Default route cost for OSPFv2
* Suppression of inter-area routes (optional)
* Inter-area and NSSA area ranges and suppressed area ranges

Everything option apart from suppression of inter-area routes is part of the RFCs and must be implemented on all devices.

Already done:

* Transformation plugin
* Reference implementation on FRR
* Integration tests for OSPFv2 and OSPFv3

Missing:

* NSSA range/filter
* VRF integration test
* Documentation


ipspace:
I resurrected that old commit, changed the data structure, and created the transformation plugin and reference implementation. That should be enough to get you started @jbemmel @ssasso (you can find the data structure in the integration tests)

ipspace:
I might add the ability to specify areas as dict later on, the final data structure will remain unchanged. You can use either **area** (to specify area as IP address) or **_area_int** (to specify area as integer) in the configuration templates.

ipspace:
Oh, and the structure of range/filter/external_range/external_filter lists will change to include **ipv4** and **ipv6** keys so we'll be able to use the same config template for everything.

netsim/extra/ospf.areas/frr.j2
```diff
@@ -0,0 +1,41 @@
+{% macro area_config(adata,af,abr) %}
+{%   if adata.kind == 'stub' %}
```
jbemmel, line 2: would suggest to ```{% set kind = adata.kind|default('regular') %}``` at the start, or remove the ```|default``` below as the rest of the template assumes it's defined

netsim/extra/ospf.areas/plugin.py
```diff
@@ -0,0 +1,153 @@
...
+      abr_config = True
+    else:
+      o_data.pop('areas',None)
+
+  return abr_config
+
+'''
+post_transform hook
+
+As we're applying interface attributes to BGP sessions, we have to copy
```
jbemmel: update description

tests/integration/ospf/ospfv3/40-area-parameters.yml
```diff
@@ -0,0 +1,84 @@
...
+  ospf.area: 12
+- interfaces: [ dut, r21 ]
+  ospf.area: 21
+- interfaces: [ dut, r22 ]
+  ospf.area: 22
+
+validate:
+  adj:
+    description: Check OSPF adjacencies
+    wait: ospfv2_adj_p2p
```
jbemmel: ospfv3_... (currently has same value)

netsim/extra/ospf.areas/frr.j2
ipspace: It's defined in https://github.com/ipspace/netlab/blob/c47f17f8f8289cbcc0425ace25a95795827e9179/netsim/extra/ospf.areas/plugin.py#L47, so no need for the default value. Thank you.

jbemmel (approved):
Nice job, couple of minor tweaks likely due to copy & paste

PR #2364, opened 2025-06-12
ospf.areas support for Dell OS10 & Cumulus NVUE

jbemmel:
Rebased after #2363 got merged

Notes:
* Dell OS10 does not support NSSA areas for OSPFv3
* Cumulus NVUE only supports OSPFv2

Dell OS10 test fails due to lack of OSPF import support in vrfs (hardcoded ```redistribute connected```) - will fix in separate PR

PR #2366, opened 2025-06-12
Dellos10 ospf import

jbemmel:
Required for #2364 to pass newly added VRF test


ipspace (approved):
Integration tests passed

Issue #2370, opened 2025-06-13
SR Linux template uses multiple different OSPF instances


jbemmel:
https://github.com/ipspace/netlab/blame/c99b4799e81254b8158f4a5cd31578cc0ca51beb/netsim/ansible/templates/ospf/srlinux.j2#L8

In hindsight and looking at the EOS template, it may be better to simply use the same pid for both ipv4 and ipv6.
Apparently there is no test that differentiates between having a single OSPF process versus 2?

ipspace:
> In hindsight and looking at the EOS template, it may be better to simply use the same pid for both ipv4 and ipv6.

Try it out?

> Apparently there is no test that differentiates between having a single OSPF process versus 2?

There's a single person I know that was bothered by this and used this approach on (hopefully) a single platform, so why would we need a test for it?

PR #2371, opened 2025-06-13
SR Linux: OSPF areas support

jbemmel:
Turns out SR Linux cannot configure a default metric for NSSA areas

Template mimics current OSPF instance naming, may need adjusting if #2370 is implemented


ipspace (approved):
Integration tests pass with the documented caveats (to be addressed in a separate PR)

PR #2372, opened 2025-06-13
Cumulus NVUE: Add OSPF default route support

jbemmel:


ipspace (approved):
Integration test passed

PR #2379, opened 2025-06-14
SR Linux: Use different process id for OSPFv3 inside VRF too

jbemmel:
Context: #2370 

While adding support for OSPF areas, I noticed that a single hard-coded OSPF process id of 0 was used inside VRFs. 
I thought it might be possible to simplify the base case too, but turns out we need 2 different ids for OSPFv2 and OSPFv3 in a configuration that uses both simultaneously within the same routing instance

The latter is probably an edge case, we could remove the test if you prefer - but it serves to illustrate what goes wrong without this PR


ipspace:
It works, nobody complained, and I don't think anyone cares. I see no reason to change it.

As for dual-stack case, it's true we don't have a dedicated dual-stack OSPF test, but we use dual-stack OSPF in several BGP test cases.

jbemmel:
> It works, nobody complained, and I don't think anyone cares. I see no reason to change it.

It doesn't work for the case of dual stack OSPF within a VRF, which is not covered by any test afaik

ipspace:
> It doesn't work for the case of dual stack OSPF within a VRF, which is not covered by any test afaik

How about you start with this fact the next time? Will add dual stack to some VRF test.

PR #2385, opened 2025-06-15
Move check for ipv6 128 length into 'get_nth_ip_from_prefix'

### LINKED ISSUE #2382: [BUG] netlab crashes with IPv6 /128 prefix on VRF loopback
Using the following lab topology:

```
nodes:
  rtr:
    device: frr
    module: [ vrf ]
    vrfs:
      tenant:
        loopback.ipv6: 2001:db8::1/128
```

... results in a crash in **netlab create**

jbemmel:
Fixes #2382


ipspace:
I'm guessing that this happily generates duplicate IP addresses from a /128 prefix. Probably not on a link as we check the prefix size there, but I'm not OK with having a function called "get_nth_ip_from_prefix" that ignores N.

Issue #2376, opened 2025-06-14
[BUG] Defaults ansible ssh user name / password wrong for device iosv , provider clab


DanPartelly:
### Describe the bug
Using Roman's  vrnetlab, git head
netlab provided default passwords / user names for device iosv, provider clab are wrong. Current good values are "admin" "admin" , not the old "vrnetlab" ones.  Its conceivable that situation is the same for iosvl2 under the same conditions.  

readiness checks do not pass.

### To Reproduce

 netlab up on any topology using iosv with clab.

### Expected behavior

readiness check should be passed

DanPartelly:
Workarround with topology-defaults.yml 

```
devices.iosv:
  clab.group_vars.ansible_ssh_pass: admin
  clab.group_vars.ansible_user: admin
```

ipspace:
@hellt -- as much as I love vrnetlab, this is a boring perpetual whack-a-mole game.

What would one have to do to persuade you guys to stop changing default usernames/passwords?

ipspace:
> Workarround with topology-defaults.yml
> 
> ```
> devices.iosv:
>   clab.group_vars.ansible_ssh_pass: admin
>   clab.group_vars.ansible_user: admin
> ```

You know what: we could set the container environment USERNAME and PASSWORD to **ansible_user** and **ansible_ssh_pass** ;)

hellt:
I read it twice and I don't understand what problem we are discussing

DanPartelly:
When ipspace did the initial integration for iosv / iosvl2 running under containerlab control 
pass = VR-netlab9
user = vrnetlab

worked. (or something like that, memory is sketchy)

Sometimes between then and now, the device became inaccessible using those credentials   and admin/admin is now the default. 


> I read it twice and I don't understand what problem we are discussing

hellt:
Mmm, that is weird. In containerlab we never use the vrnetlab password and always reset  it to either admin:admin if the system does not have password policy in place

DanPartelly:
@hellt  Ok, past is past and undebugable. No complains, gonna get changed. Then, as a rule now, admin/admin is right, for all devices, and likely to stay so for the future ?


> Mmm, that is weird. In containerlab we never use the vrnetlab password and always reset it to either admin:admin if the system does not have password policy in place

hellt:
All default credentials are listed in the corresponding kind documentation

[...]

PR #2386, opened 2025-06-15
Document the dependencies between netlab and vrnetlab usernames/pwds

ipspace:
Triggered by #2376


ipspace:
I decided to document the dependencies between vrnetlab and netlab usernames/passwords.

Any comments on the updated documentation @hellt / @DanPartelly ?

DanPartelly (approved):
As close as we can get, without testing all day long, as we depend on the upstream.

ipspace:
> As close as we can get, without testing all day long, as we depend on the upstream.

Way too many upstreams :( That's why I had to create "platform integration tests".

PR #2388, opened 2025-06-16
Forti vdom

sdargoeuves:


sdargoeuves:
Related to this [discussion](https://github.com/ipspace/netlab/discussions/2271#discussioncomment-13477479)

ipspace:
@ssasso -- I don't have a working FortiOS VM, so it's all up to you (no pressure ;)

Also, someone needs to write a decent commit message when merging.

sdargoeuves:
@ssasso I've just seen the zone-based firewall plugin, unless we can find a way to specify `vdom` in the topology file instead of `netlab_vdom`, we would need to update the file `netsim/extra/firewall.zonebased/deploy.fortinet.fortios.fortios.yml` to use `{{ netlab_vdom }}` instead of `{{ vdom }}`.
I've just ran the topology up, and it works, I'll push the update a bit later.
~~I've just had an idea to potentially improve what I did, so we don't have to update the plugin, or other future code, ideally, I should make sure {{ vdom }} is going to be the traffic vdom, rather than using {{ netlab_vdom }}
I'll test my idea, and hopefully update this PR soon~~ I'm starting to be even more confused 😕 I'll keep as it is currently, and will push the update to the zonebased file.

netsim/ansible/tasks/fortinet.fortios.fortios/initial.yml
```diff
@@ -1,7 +1,57 @@
...
+      management_vdom: "{{ vdom }}"
+      vdom_mode: multi-vdom
+      hostname: '{{ inventory_hostname.replace("_","-") }}'
+  register: vdom_mode_result
+  when: netlab_vdom_is_enabled
+
+- name: Ensure FortiGate is ready after VDOM mode change
+  block:
+    - name: Wait 60 seconds after VDOM mode change
+      ansible.builtin.wait_for:
```
ssasso, line 19: `wait_for` requires to explicitly specify the target host, otherwise it defaults to *127.0.0.1*.

See: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/wait_for_module.html#parameter-host

This should become:

```
[...]
ansible.builtin.wait_for:
  host: "{{ansible_host}}"
  port: 443
[...]
```

With the current code, this is the failure message:
```
TASK [Wait 60 seconds after VDOM mode change] *****
fatal: [fw1]: FAILED! => changed=false
  elapsed: 180
  msg: Timeout when waiting for 127.0.0.1:443
```

With the above proposed change, everything works fine.

ssasso (changes requested):
required change in `wait_for`

ssasso (approved):

ipspace:
@ssasso Is this ready to merge?

ssasso:
> @ssasso Is this ready to merge?

yes!

-- I updated the initial description of the PR, so you can use it as commit message.

ipspace:
> -- I updated the initial description of the PR, so you can use it as commit message.

@ssasso: Thank you!

@sdargoeuves: Great job ;) Thank you!

sdargoeuves:
@ipspace @ssasso thank you for your help, it was definitely a great learning experience!
I'll be back 😅

PR #2389, opened 2025-06-16
ArubaCX: support ospf.areas

ssasso:


ipspace (approved):
Nice job, integration tests pass.

ipspace:
As a side note (for future PRs), please don't mix two unrelated things (new features, new SW release) in the same PR. It makes it harder to write release notes 🤷‍♂️

ssasso:
> As a side note (for future PRs), please don't mix two unrelated things (new features, new SW release) in the same PR. It makes it harder to write release notes 🤷‍♂️

my bad - it was unintentional, I noticed it was on the PR only now (because I ran the tests with 10.13, then decided to run them also with 10.14 and changed the device, and then forgot about it)

PR #2392, opened 2025-06-16
Explain the state of individual platforms

ipspace:


ipspace:
I added the "support level" to the list of netlab-supported platforms to document what we test extensively, what we do on best-effort basis, and what we probably won't be doing (much).

Comments highly appreciated @ssasso @jbemmel

jbemmel:
Seems fair to me, no concerns

ssasso:
good for me as well :)

PR #2394, opened 2025-06-16
Add cost/metric parameter to ospf.areas summarization ranges

ipspace:


tests/integration/ospf/ospfv3/40-area-parameters.yml
```diff
@@ -30,6 +38,9 @@ ospf.areas:
   default.cost: 15
   external_range:
   - 2001:db8:cafe:1::/64
+  - ipv6: 2001:db8:cafe:3::/64
+    cost: 42
+    type: e2
```
jbemmel, line 43: Wondering where this is tested? Would something fail if this is changed to ```e1```?

jbemmel (approved):
May want to add a test for 'e1' versus 'e2' being correctly configured (or add a note on where this happens, if already)

tests/integration/ospf/ospfv3/40-area-parameters.yml
ipspace: Yes, the total cost would be a sum of E1 and backbone cost.

PR #2397, opened 2025-06-17
Dell OS10 : Add support for deleting a standard community list

jbemmel:
Test: routing/07b-community-delete and 07c-community-list-delete

Alternatively, could duplicate the delete_list logic from routing/ios.j2 (or could remove the latter)

### DIFF
--- a/netsim/ansible/templates/routing/_route_map_ios.j2
+++ b/netsim/ansible/templates/routing/_route_map_ios.j2
@@ -34,11 +34,11 @@ that use "industry standard" CLI (IOS, IOS-XE, FRR, Cumulus, Arista)
 {%     endif %}
 {%     if 'community' in p_entry.set %}
 {%       set cset = p_entry.set.community %}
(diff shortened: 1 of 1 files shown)


ipspace:
You modified a file shared by multiple configuration templates (= multiple devices). Did you run thorough tests on all those devices to check whether your changes have broken anything?

jbemmel:
> You modified a file shared by multiple configuration templates (= multiple devices). Did you run thorough tests on all those devices to check whether your changes have broken anything?

No, but I did look at it and considered the alternative mentioned (i.e. duplicate the delete_list logic from ios.j2)

I do not have access to those other devices, I think the change might work but I did not test it. I am submitting this PR as a suggestion - you could accept it (after running tests), or you could say "I prefer that you copy what is done for IOS" and that would be fine too

ipspace:
> I do not have access to those other devices, I think the change might work but I did not test it.

A cursory look at the code using that shared template would tell you it would probably break the EOS code.

> I am submitting this PR as a suggestion - you could accept it (after running tests)

Oh, it was just a suggestion, and I should do the testing? Good to know.

PR #2399, opened 2025-06-18
Prevent a configuration error during intial config when using unnumbered interfaces. Device family  ios.

DanPartelly:


ipspace:
All IOS XE platforms work with the current setup (see https://tests.netlab.tools/_html/coverage.initial), and IOSv (the last time I checked) did not support **ip unnumbered** at all, so what problem are you trying to solve?

ipspace:
On a more generic note, we should always start with "what problem are we trying to solve?" That is ideally documented in an issue, and once we agree there is an issue, a PR implements the fix.

Alternatively, you could describe the problem in a PR, either in the initial description, or in a later comment.

In any case, a PR that drops out of the blue sky with no explanation is not helpful.

DanPartelly:
It results in a configuration error when applied to tunnel interfaces, on my IOL images.  But of course, you are right about what you said.

ipspace:
> It results in a configuration error when applied to tunnel interfaces, on my IOL images. But of course, you are right about what you said.

Ah, that's a totally different story. Tunnel interface are P2P interfaces, and could probably be unnumbered even with the IOSv Classic.

That needs to be fixed, but the check has to be made on interface type.

PR #2400, opened 2025-06-18
Implement generic 'optional device features' check

### LINKED ISSUE #2373: [BUG] Make NSSA range summarization an optional feature
Most devices for which we implemented the ospf.areas configuration template cannot do summarization
of type-7 NSSA routes into summary type-5 routes.

It's time to give up and admit defeat:

* The 'external_range' and 'external_filter' attributes have to be optional and checked in the common code
* That would remove most of device quirks
* We could either generate a warning if a device does not support NSSA ranges, have a separate integration test that would test the NSSA ranges, or simply not test them.

ipspace:
Instead of writing a custom 'check optional features' routine in different modules and plugins, this PR implements a generic 'optional device features' check using either whitelisting or blacklisting (we might need further parameters in the future).

The new functionality is used to check ospf.areas features and implemented for Arista EOS NSSA ranges.

A subsequent PR will implement the same functionality for all other devices that support ospf.areas (and clean up a few device quirks ;)

Closes #2373

### DIFF
--- a/netsim/augment/devices.py
+++ b/netsim/augment/devices.py
@@ -119,6 +119,7 @@ def check_optional_features(
       category=Warning if category is None else category,
       more_data=f'Used in {path}',
       module=module)
+    return False
 
   if isinstance(data,Box) and isinstance(features,Box):
     OK = True
--- a/netsim/extra/ospf.areas/plugin.py
+++ b/netsim/extra/ospf.areas/plugin.py
@@ -132,7 +132,7 @@ def check_node_support(ndata: Box,topology: Box) -> bool:
     for a_entry in o_data.areas:
       OK = OK and devices.check_optional_features(
                     data=a_entry,
-                    path=path+f'.areas[area={a_entry.area}]',
+                    path=path+f'.ospf.areas[area={a_entry.area}]',
                     node=ndata,
                     topology=topology,
                     attribute='ospf.areas',
@@ -166,5 +166,6 @@ def post_transform(topology: Box) -> None:
           category=log.IncorrectAttr,
           module=_config_name)
         return
-      if check_node_support(ndata,topology):
-        api.node_config(ndata,_config_name)                 # ... and remember if we have to do extra configuration
+
+      check_node_support(ndata,topology)                    # Report warnings if a device does not support optional features
+      api.node_config(ndata,_config_name)                   # ... and remember that we have to do extra configuration


netsim/extra/ospf.areas/plugin.py
```diff
@@ -120,6 +120,26 @@ def prune_ospf_areas(ndata: Box) -> bool:
...
+'''
+Check whether the node supports all ospf.areas features used in area definitions
+'''
+def check_node_support(ndata: Box,topology: Box) -> bool:
+  OK = True
+  for (o_data,_,vname) in _ospf.rp_data(ndata,'ospf'):
+    if 'areas' not in o_data:
+      continue
+    path = f'nodes.{ndata.name}' + (f'.vrfs.{vname}' if vname else '')
+    for a_entry in o_data.areas:
```
jbemmel, line 132: Checking a dict of things for feature support could be a future generalization

netsim/augment/devices.py
```diff
@@ -64,6 +65,78 @@ def get_device_features(node: Box, defaults: Box) -> Box:
...
+  if features is False:             # The device has no idea about the required feature, get out of here
+    optional_features_error(node,attribute,path,module,category)
+    return False
+
+  if isinstance(features,str):      # Custom error message
+    log.error(
+      features,
+      category=Warning if category is None else category,
+      more_data=f'Used in {path}',
+      module=module)
```
jbemmel, line 121: return False

netsim/extra/ospf.areas/plugin.py
```diff
@@ -145,5 +165,6 @@ def post_transform(topology: Box) -> None:
           f'Device {ndata.device} (node {ndata.name}) does not support OSPF area parameters',
           category=log.IncorrectAttr,
           module=_config_name)
-      else:
+        return
+      if check_node_support(ndata,topology):
         api.node_config(ndata,_config_name)                 # ... and remember if we have to do extra configuration
```
jbemmel: If the device does not support an optional feature, wouldn't we still want to invoke the config script to provision the bits that it does support?

ipspace: You're correct. In this case (as we're checking optional features, and the main check is already done), we should just do the checks and move on anyway.

netsim/extra/ospf.areas/plugin.py
ipspace: The "check_optional_features" already does recursive checks, so there's a non-zero chance that checking for "default.cost" would also work ;)

ipspace:
> Some questions around the implications of "optional"

Fixing...

> I think this could use some explicit test cases

As we're (currently) generating only warnings, we have no mechanism that could be used to test this functionality (apart from the actual integration tests). The transformation process is not affected (because we configure the stuff anyway) and we need a hard error to be able to check the error messages generated during the transformation process.

PR #2403, opened 2025-06-18
Bugfix: Set 'linkindex' for VTEP loopback link

jbemmel:
Else a call to ```get_next_linkindex``` from ```vlan.create_loopback_vlan_links``` results in an error


jbemmel:
Replaced by #2406

PR #2406, opened 2025-06-18
Mlag vtep plugin enhancements

jbemmel:
* Set 'linkindex' on created stub link
* Adjust test to check that VTEP loopback gets advertised in OSPF
* Use `loopback.pool` if specified, to enable per-site loopback ranges

### DIFF
--- a/netsim/extra/mlag.vtep/plugin.py
+++ b/netsim/extra/mlag.vtep/plugin.py
@@ -28,19 +28,38 @@ def pre_link_transform(topology: Box) -> None:
       continue                                      # Nope - carry on
 
     # Allocate a shared IP for both peers
-    vtep_a = addressing.get(topology.pools, [POOL_NAME, 'vrf_loopback'])['ipv4']
+    vtep_a = None
     peers = [ i.node for i in l.interfaces ]
     for node_name in peers:
       node = topology.nodes[ node_name ]
       if node.get('lag.mlag.vtep',None) is False:
         continue                                    # Skip nodes for which the plugin is disabled
       if 'vxlan' in node.module:
-          vtep_loopback = data.get_empty_box()
-          vtep_loopback.type = 'loopback'           # Assign same static IP to both nodes
-          vtep_loopback.interfaces = [ { 'node': node_name, 'ipv4': str(vtep_a.network_address)+"/32" } ]
-          vtep_loopback._linkname = f"MLAG VTEP VXLAN interface shared between {' - '.join(peers)}"
-          vtep_loopback.vxlan.vtep = True
-          topology.links.append(vtep_loopback)
+        if not vtep_a:
+          pool = node.get('loopback.pool',POOL_NAME)
+          vtep_a = addressing.get(topology.pools, [pool, 'vrf_loopback'])['ipv4']
+        vtep_loopback = data.get_empty_box()
+        vtep_loopback.type = 'loopback'             # Assign same static IP to both nodes
+        vtep_loopback.interfaces = [ { 'node': node_name, 'ipv4': str(vtep_a.network_address)+"/32" } ]
+        vtep_loopback._linkname = f"MLAG VTEP VXLAN interface shared between {' - '.join(peers)}"
+        vtep_loopback.vxlan.vtep = True
+        vtep_loopback.linkindex = len(topology.links)+1
+        topology.links.append(vtep_loopback)
 
       if log.debug_active('links'):                 # pragma: no cover (debugging)
         print(f'\nmlag.vtep Create VTEP loopback link for {node_name}: {vtep_loopback}')
+
+#
+# post_transform: Copy IGP configuration from primary loopbacks to VTEP loopbacks
+#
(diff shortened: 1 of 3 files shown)


ipspace (changes requested):
I would suggest you create links with **type: loopback** instead of interfaces and let IGP modules handle the rest.

Also, did you check that you get the loopback IP addresses beyond the highest node ID (so you won't get duplicate IPs)? Looking at the code, it should work, but nonetheless...

jbemmel:
> I would suggest you create links with **type: loopback** instead of interfaces and let IGP modules handle the rest.
> 

The code *does* create links of type loopback, but for some reason the IGP attributes don't get applied. I'll investigate why - you're probably right that the custom copy logic shouldn't be necessary

> Also, did you check that you get the loopback IP addresses beyond the highest node ID (so you won't get duplicate IPs)? Looking at the code, it should work, but nonetheless...

I'll see if I can test for that

ipspace:
> I'll see if I can test for that

Just set some high IDs on the nodes and see what the extra loopbacks get

ipspace:
> > I would suggest you create links with **type: loopback** instead of interfaces and let IGP modules handle the rest.
> 
> The code _does_ create links of type loopback

Dang, I missed that 🤦‍♂️ I'm getting too old (or too tired)

ipspace:
I started the `mlag.vtep/01-vxlan-bridging.yml` with Arista EOS containers and I see OSPF area 0 configured on Loopback1 (the VXLAN source interface), so I'd say everything works as expected?

jbemmel:
> I started the `mlag.vtep/01-vxlan-bridging.yml` with Arista EOS containers and I see OSPF area 0 configured on Loopback1 (the VXLAN source interface), so I'd say everything works as expected?

You're not going to like this, but one word: "Cumulus"

I'll take a look and rework things as appropriate - thanks for taking the time to comment

ipspace:
> You're not going to like this, but one word: "Cumulus"

🤢🤷‍♂️

The data model produced from the `01-vxlan-bridging.yml` example seems OK, so the problem is probably in the Cumulus templates.

jbemmel:
> > You're not going to like this, but one word: "Cumulus"
> 
> 🤢🤷‍♂️
> 
> The data model produced from the `01-vxlan-bridging.yml` example seems OK, so the problem is probably in the Cumulus templates.

Issue is here: https://github.com/ipspace/netlab/pull/2411

Code is trying to merge OSPF settings if "compatible", but it wasn't covering all combinations. I'll remove the code from this PR

jbemmel:
> > I'll see if I can test for that
> 
> Just set some high IDs on the nodes and see what the extra loopbacks get

I put ids 10 and 20, the extra VTEP loopback gets 21

netsim/extra/mlag.vtep/plugin.py
```diff
@@ -28,19 +28,23 @@ def pre_link_transform(topology: Box) -> None:
...
-          topology.links.append(vtep_loopback)
+        if not vtep_a:
+          pool = node.get('loopback.pool',POOL_NAME)
+          vtep_a = addressing.get(topology.pools, [pool, 'vrf_loopback'])['ipv4']
+        vtep_loopback = data.get_empty_box()
+        vtep_loopback.type = 'loopback'             # Assign same static IP to both nodes
+        vtep_loopback.interfaces = [ { 'node': node_name, 'ipv4': str(vtep_a.network_address)+"/32" } ]
+        vtep_loopback._linkname = f"MLAG VTEP VXLAN interface shared between {' - '.join(peers)}"
+        vtep_loopback.vxlan.vtep = True
+        vtep_loopback.linkindex = len(topology.links)+1
```
ipspace: We have 'links.get_next_linkindex' for this

netsim/extra/mlag.vtep/plugin.py
```diff
@@ -28,19 +28,23 @@ def pre_link_transform(topology: Box) -> None:
...
-          vtep_loopback.interfaces = [ { 'node': node_name, 'ipv4': str(vtep_a.network_address)+"/32" } ]
-          vtep_loopback._linkname = f"MLAG VTEP VXLAN interface shared between {' - '.join(peers)}"
-          vtep_loopback.vxlan.vtep = True
-          topology.links.append(vtep_loopback)
+        if not vtep_a:
+          pool = node.get('loopback.pool',POOL_NAME)
+          vtep_a = addressing.get(topology.pools, [pool, 'vrf_loopback'])['ipv4']
+        vtep_loopback = data.get_empty_box()
+        vtep_loopback.type = 'loopback'             # Assign same static IP to both nodes
+        vtep_loopback.interfaces = [ { 'node': node_name, 'ipv4': str(vtep_a.network_address)+"/32" } ]
```
ipspace: Why don't you leave the prefix intact? If it's assigned from a correctly-configured pool, it would be /32 anyway

ipspace (changes requested):
A minor tweak and a suggestion

PR #2411, opened 2025-06-19
Update cumulus_nvue.py

jbemmel:
Fix merging of OSPF settings on loopbacks, code wasn't covering all combinations of passive: True/False


jbemmel:
Less needed due to #2412 but still better for consistency

PR #2412, opened 2025-06-19
Set loopback IGP passive based on global default, if defined

jbemmel:
Apply #1612 to loopback interfaces too

Plays a role in #2406 - the Cumulus quirk complains about inconsistent IGP settings on the different loopback interfaces

### DIFF
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -206,7 +206,8 @@ def add_loopback_igp(node: Box, proto: str, topology: Box) -> None:
   node.loopback[proto] = lb_data + node.loopback[proto]     # Merge device LB info with whatever is already on LB
                                                             # Note that an empty box is created on first reference
   if 'passive' not in node.loopback[proto]:                 # Finally, many templates expect 'passive' to be present
-    node.loopback[proto].passive = False                    # ... so add a bogus 'not passive' flag if needed
+    _passive = topology.get(f"{proto}.passive",False)       # Check the global default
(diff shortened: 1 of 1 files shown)


ipspace:
We had this "discussion" too many times (and no, the current attempt won't get merged either). Please think for a moment what "passive" interface means for a routing protocol, and how that applies to loopbacks.

jbemmel:
> We had this "discussion" too many times (and no, the current attempt won't get merged either). Please think for a moment what "passive" interface means for a routing protocol, and how that applies to loopbacks.

Whatever it means, it needs to be applied consistently across all loopback interfaces

I'm ok to ignore the global setting for secondary loopbacks too, instead of applying it consistently to the primary loopback interface as is done in this PR

Do you not agree that the ```ospf.passive``` setting generated by Netlab should be the same for all loopback interfaces, regardless whether they're "primary" or not? (barring user input that says ```loopback.ospf.passive: True```)

ipspace:
OK, so you prefer arguing instead of figuring things out. I'll keep that in mind.

"passive" interface means nothing more than "do not send or receive hello packets" which results in "no IGP adjacency on that interface". How relevant is that for loopbacks?

The only reason we have "passive" flag on loopbacks is to make the templates a tiny bit easier (so they don't have to check whether the attribute is defined). If a device has a problem with "passive" loopbacks, then the template is free to ignore that attribute for link type == 'loopback'

I understand that some of your code has some issues, but the solution is not to push them somewhere else. Just ignore the "passive" flag.

PR #2413, opened 2025-06-20
FRR: Don't complain about STP when it's disabled

jbemmel:


ipspace (approved):
Seems reasonable. You did run the tests, right?

jbemmel:
> Seems reasonable. You did run the tests, right?

In honesty, no - I encountered the issue, deployment failed, then I added the patch, now it works.

I didn't re-run all the tests, but I assume there is no change

PR #2416, opened 2025-06-21
Fix inital configuration template for ios family - tunnel interfaces

DanPartelly:
Fix inital configuration template for ios family.
tunnel interfaces use "ip unnumbered..." on IOS XE, not not "ip unnumbered poll..".
Correct configuration was generated in both 03_unnumbered integration test, as well as a custom test using tunnel type interfaces.

### DIFF
--- a/netsim/ansible/templates/initial/ios.j2
+++ b/netsim/ansible/templates/initial/ios.j2
@@ -79,7 +79,11 @@ interface {{ l.ifname }}
 #}
 {% if 'ipv4' in l %}
 {%   if l.ipv4 == True %}
+{%     if l.type is defined and l.type == "tunnel" %}
+ ip unnumbered {{ l._parent_intf }}
+{%     else %}
  ip unnumbered {{ l._parent_intf }} poll
+{%     endif %}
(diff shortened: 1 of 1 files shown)


netsim/ansible/templates/initial/ios.j2
```diff
@@ -79,7 +79,11 @@ interface {{ l.ifname }}
 #}
 {% if 'ipv4' in l %}
 {%   if l.ipv4 == True %}
+{%     if l.type is defined and l.type == "tunnel" %}
```
ipspace: By now, the l.type is defined for every interface (including the system loopback), so it's safe to use just `if l.type == "tunnel"` (similarly to what's done in some other parts of the same config template)

One of these days I have to clean up the old references, but there's always something better to do.

ipspace (approved):
Looks good to me, just a minor suggestion

netsim/ansible/templates/initial/ios.j2
DanPartelly: Superfluous  check dropped.

ipspace (approved):
Thank you!

Issue #2418, opened 2025-06-21
Netlab connect on multiple devices


DanPartelly:
### Discussed in https://github.com/ipspace/netlab/discussions/2377

<div type='discussions-op-text'>

<sup>Originally posted by **DanPartelly** June 14, 2025</sup>
As title says. A way to connect on multiple devices at once, preferably using globs similar to netlab exec. It would need a config options for the preferred user terminal command to be spawned in execution before ssh connection is open. Saves shitloads of time.

Opinions ?</div>

ipspace:
Thinking out loud as this still seems to be a very particular request (no problem with that ;) that needs just the right environment to work

What if you'd have a **netlab** command that would print the list of device matching a glob? Then you could use that list in a bash script to do whatever.

DanPartelly:
I believe that having this new report solves the problem better than the initial proposal. It is also a more flexible solution, the script can easily be customized to cover a large range of needs. Thanks for coming up with this.

PR #2420, opened 2025-06-23
New report: names of selected nodes (to be used in bash scripts)

ipspace:
This is a potential solution for #2418:

* netlab report nodes --node glob will print out the names of the selected nodes
* The names of the selected nodes can be used in a bash script, for example:

```
#!/bin/bash
for n in $(netlab report --node "${1:-*}" nodes); do
  echo $n
done
```


DanPartelly (approved):
This works great and solves the issue. Node globs, groups.

PR #2425, opened 2025-06-24
ESI-LAG: Initial draft implementation

ssasso:
Initial draft implementation of ESI-LAG in LAG module.

I tried to implement this as a plugin first, but it was really a mess to have to interact with the lag module (especially for mlag checks), so I decided to "embed" the code directly in the module itself.

JunOS implementation works, it should be straightforward to add support for FRR and EOS - and I can do it.

But first I'd like your opinion on this @ipspace @jbemmel , then I can also update the documentation as well.

With the current LAG module "syntax", ESI-LAG must be specified in this way:

```
links:
  sw1_edge:
  - lag:
      members:
      - sw1a:
          vlan.access: rosso
          lag:
            lacp_system_id: 1
            esi: true
        rosso1:
      - sw1b:
          vlan.access: rosso
          lag:
            lacp_system_id: 1
            esi: true
        rosso1:

```


jbemmel:
For the data model, I would consider the following:
```
ethernet_segments:
  seg1:
    id: lacp
  seg2:
    id: aa:bb:cc:dd:ee:ff:00:11:22:33

links:
- lag:
      members:
      - sw1a:
          vlan.access: rosso
          lag:
            lacp_system_id: 1
            esi: seg1
        rosso1:
      - sw1b:
          vlan.access: rosso
          lag:
            lacp_system_id: 1
            esi: seg1
        rosso1:
```

I'm not sure if this is best implemented as a separate module, or as part of the evpn module (since esi is closely linked with evpn, e.g. on SR Linux it would require evpn). The ```lacp_system_id``` obviously belongs with the ```lag``` module, but ```esi``` could be a separate interface attribute not necessarily under ```lag```

ssasso:
> I'm not sure if this is best implemented as a separate module, or as part of the evpn module (since esi is closely linked with evpn, e.g. on SR Linux it would require evpn). The `lacp_system_id` obviously belongs with the `lag` module, but `esi` could be a separate interface attribute not necessarily under `lag`

But this is explicitly addressing **ESI-LAG**, where the esi attribute must stay *under* lag config (at least both in junos, frr, eos). And IIRC in FRR esi mh attributes are supported only under bond interfaces (while other vendors support esi also outside the lag).

However the big question is: how can we separate them but at the same time keep them tied in some cases? that is what I was not able to easily achieve with a plugin (**especially on the "validation" of "mlag"** - i.e., not requiring peerlink & co...).

jbemmel:
> > I'm not sure if this is best implemented as a separate module, or as part of the evpn module (since esi is closely linked with evpn, e.g. on SR Linux it would require evpn). The `lacp_system_id` obviously belongs with the `lag` module, but `esi` could be a separate interface attribute not necessarily under `lag`
> 
> But this is explicitly addressing **ESI-LAG**, where the esi attribute must stay _under_ lag config (at least both in junos, frr, eos). And IIRC in FRR esi mh attributes are supported only under bond interfaces (while other vendors support esi also outside the lag).
> 
> However the big question is: how can we separate them but at the same time keep them tied in some cases? that is what I was not able to easily achieve with a plugin.

```
{% for intf in interfaces if intf.type == 'lag' and intf.lag.esi is defined %}
interface {{ intf.ifname }}
{%   if intf.lag._esi_auto_derive|default(false) %}
    evpn mh es-id {{intf.lag._esi_manual}}
{%   else %}
    evpn mh es-id {{intf.lag.esi}}
{%   endif %}
!
{% endfor %}
```

Taking an example from your FRR config - this is evpn configuration, not lag. If FRR has the limitation that ESI can only be applied to lags, a device quirk could enforce that (or a feature flag inside the ```esi``` module/plugin, e.g. ```modes: [ lag, evpn ]``` where FRR might have ```modes: [ lag ]``` and SR Linux would have ```modes: [ evpn ]``` )

In my example, if ```id``` is set to ```lacp``` the ESI logic would check that the interface to which the ESI is applied is also part of a lag

My main concern, is that the data model imho shouldn't be based on the (limited) way in which some particular platform supports the feature. In my mind the feature is "ethernet segments", with implementations under EVPN and (here) LAG

ssasso:
> My main concern, is that the data model imho shouldn't be based on the (limited) way in which some particular platform supports the feature. In my mind the feature is "ethernet segments", with implementations under EVPN and (here) LAG

totally agree on this.

but how we can change the LAG validation?

i.e., if I apply an ethernet segment to a (m)lag interface, then in some way we need to disable mlag peer link validation (or device not supporting "strict" mlag) - so evpn and lag context will be mixed here.

jbemmel:
> but how we can change the LAG validation?
> 
> i.e., if I apply an ethernet segment to a (m)lag interface, then in some way we need to disable mlag peer link validation (or device not supporting "strict" mlag) - so evpn and lag context will be mixed here.

I can see how it's hard to get a "clean" separation / modularity here

I would say it'd be ok to modify the logic in the lag module to check for an 'esi' interface attribute (and perhaps a ```lag.mlag_esi``` feature flag), and skip the mlag-not-supported error in that case. However, the provisioning scripts and global ```ethernet_segments``` etc. would be better placed in a separate module/plugin imho

ssasso:
@jbemmel I really need your help on this topic:

I moved the implementation to an external plugin, and I was successfully able to add additional verification to 'lag' module to avoid errors for not supporting mclag.

However, after passing throught the lag plugin, the vlan module starts complaining:

Topology snippet:

```
links:
# EVPN/VXLAN Link
- s1:
  s2:
  mtu: 1600
# ESI-LAG
- lag:
    members:
    - s1:
        lag.lacp_system_id: 1
        evpn.es: seg_1
      x1:
    - s2:
        lag.lacp_system_id: 1
        evpn.es: seg_1
      x1:
  vlan.access: red
```

Error:
```
stefano@donkey:~/GIT_H/netlab/tests/integration/evpn.es$ netlab create -p clab -d vjunos-switch 01-esi-lag.yml
[ERRORS]  Errors found in 01-esi-lag.yml
[VALUE]   vlan: You must use 'nodes.s1.vlans.red' dictionary to set VLAN interface attributes,
[DATA]    Link links[2], node s1, vlan red
          Interface VLAN forwarding mode: bridge, unexpected attributes evpn
          {'node': 's1', 'lag': {'lacp_system_id': 1, '_mlag': True, 'ifindex': 1}, 'evpn': {'es': 'seg_1'},
          'vlan': {'access': 'red'}, '_vlan_mode': 'bridge', 'ipv4': False, 'ipv6': False, 'type':
          'lag'}
[VALUE]   vlan: You must use 'nodes.s2.vlans.red' dictionary to set VLAN interface attributes,
[DATA]    Link links[2], node s2, vlan red
          Interface VLAN forwarding mode: bridge, unexpected attributes evpn
          {'node': 's2', 'lag': {'lacp_system_id': 1, '_mlag': True, 'ifindex': 1}, 'evpn': {'es': 'seg_1'},
          'vlan': {'access': 'red'}, '_vlan_mode': 'bridge', 'ipv4': False, 'ipv6': False, 'type':
          'lag'}
```

I guess this is related on how the lag module copies attributes to additional links? any hint? 

How did you make vlan not complaining for the 'lag' attributes?

thanks :-)

ssasso:
> How did you make vlan not complaining for the 'lag' attributes?

nevermind, I think I got it. it's in the vlan module itself.

netsim/extra/evpn.es/plugin.py
```diff
@@ -12,27 +12,70 @@
...
         es_supported = 'evpn.es' in features
         if not es_supported: continue
+        # Load Ethernet Segments data and expand it
+        es_data = node.get('evpn.ethernet_segments', {})
+        for intf in node.get('interfaces',[]):
+            intf_es = intf.get('evpn.es', '')
+            if not intf_es: continue
+            if intf_es and intf.type not in _es_supported_on:
+                log.error(
+                    f'Node {node.name}({node.device}) does not support EVPN Ethernet Segment on {intf.type} interfaces (found on: {intf.ifname})',
```
jbemmel: This is not specific to the device/node, it's the plugin that doesn't support it (on any device)

netsim/extra/evpn.es/plugin.py
```diff
@@ -12,27 +12,70 @@
...
+            if intf.type in list(set(_es_supported_on) - set(['lag'])) and not features.get('evpn.es.interface', False):
+                log.error(
+                    f'Node {node.name}({node.device}) does not support EVPN Ethernet Segment on "physical" interfaces (found on: {intf.ifname})',
+                    category=log.IncorrectAttr,
+                    module='evpn.es')
+                return
+            if intf_es not in es_data:
+                log.error(
+                    f'Node {node.name}({node.device}) invalid EVPN Ethernet Segment configured on interface {intf.ifname}',
+                    category=log.IncorrectAttr,
```
jbemmel: IncorrectValue

netsim/extra/evpn.es/plugin.py
ssasso: Well, I may be wrong, but I don't think ES - in general - is supported on i.e. irb interface or tunnel interface

ssasso: (but I can definitely change the error message)

netsim/data/types.py
```diff
@@ -831,6 +831,28 @@ def must_be_mac(value: typing.Any) -> dict:
 
   return { '_valid': True }
 
+@type_test()
+def must_be_esi_value(value: typing.Any) -> dict:
+  _value_msg = '10-byte ESI Value in format 00:XX:XX:XX:XX:XX:XX:XX:XX:XX'
+  if not isinstance(value,str):
+    return { '_type': '10-byte ESI Value' }
+
+  try:
```
jbemmel: Perhaps easier to do a regular expression match: r"00(:[0-9a-fA-F]{2}){9}" (or something like that)

Note that RFC7432 defines other types besides "00" that could conceivably be provisioned manually by users

netsim/extra/evpn.es/defaults.yml
```diff
@@ -0,0 +1,49 @@
...
+
+# data model:
+# -- global: evpn.ethernet_segments
+# -- on interface: es: xxx (must be a valid ethernet_segment reference)
+evpn:
+  attributes:
+    global:
+      ethernet_segments:
+        type: dict
+        _keytype: str
```
jbemmel: _keytype: id

netsim/modules/lag.yml
```diff
@@ -34,14 +34,16 @@ attributes:
         min_value: 1
         _alt_types: [ bool ]
       mac: mac
-
-  # interface:     # Internal, don't declare as official attribute
-    # _mlag: bool
+  interface:
+    lacp_system_id:
```
jbemmel, line 38: I think this is a typical candidate for auto-generation; users will probably care even less about this value, than the IP addresses being used

It's good to allow it to be specified, but I think the typical case should be auto-generated (e.g. based on node ID, or peergroup number).

Similarly, for manual specification we would probably want to allow node level - and then copy it to all the lag interfaces. It would typically be the same consistent value everywhere, no?

tests/integration/evpn.es/01-esi-lag.yml
```diff
@@ -0,0 +1,94 @@
...
+links:
+# EVPN/VXLAN Link
+- s1:
+  s2:
+  mtu: 1600
+# ESI-LAG
+- lag:
+    members:
+    - s1:
+        lag.lacp_system_id: 1
```
jbemmel: As commented above, I think it would be great if the plugin could implement logic to auto-generate the lacp_system_id:
When a lag interface references an ES with 'auto: true', if there is no manual override generate a lacp_system_id based on the minimum node ID of all the nodes attached to that segment (could be >2)

tests/topology/expected/lag-l2.yml
```diff
@@ -117,6 +117,8 @@ nodes:
       name: r1 -> r2
       neighbors:
       - ifname: port-channel1
+        lag:
```
jbemmel: It is odd that this changes - could point at a bug somewhere (not necessarily in your additions)

netsim/modules/lag.yml
ssasso: I'm scared that auto-generation for lacp_system_id can open a moltitude of can of worms. I'd rather prefer to focus on the basic stuff for now, - auto generation can be done on a subsequent PR.

ssasso:
Side note: *frr* seems to have some problems, under investigation.

Control plane looks good, looks like the official example found here: https://github.com/FRRouting/frr/tree/master/tests/topotests/bgp_evpn_mh

But test is not passing.

Included test is passing for vjunos-switch and eos.

ssasso:
@ipspace I'd love to hear also your feedback on this plugin - but no rush.

docs/plugins/evpn.es.md
```diff
@@ -0,0 +1,84 @@
+(plugin-evpn-es)=
+# EVPN Ethernet Segment (EVPN Multihoming)
```
jbemmel: ```evpn.mh``` might be a better name for the plugin - I suspect that's what users would be more likely to look for

netsim/extra/evpn.es/defaults.yml
```diff
@@ -0,0 +1,55 @@
...
+  attributes:
+    global:
+      ethernet_segments:
+        type: dict
+        _keytype: id
+        _subtype: _es_item
+    node:
+      ethernet_segments: { copy: global }
+    interface:
+      es: str
```
jbemmel: id

netsim/extra/evpn.es/frr.j2
```diff
@@ -0,0 +1,18 @@
...
+
+{### STILL NOT WORKING - to be better investigated ###}
+
+# ESI-LAG config in FRR
+cat >/tmp/esi_lag_config <<CONFIG
+evpn mh startup-delay 1
+!
+{% for intf in interfaces if intf.type == 'lag' and intf.evpn._esi is defined %}
+interface {{ intf.ifname }}
+  evpn mh es-id {{intf.evpn._esi.id}}
```
jbemmel: leaf01(config)# interface bond1
leaf01(config-if)# evpn mh es-df-pref 50000
leaf01(config-if)# evpn mh es-id 1
leaf01(config-if)# evpn mh es-sys-mac 44:38:39:FF:00:AA
leaf01(config-if)# exit

Source: NVidia Cumulus EVPN MH guide

jbemmel: or 
leaf01(config)# interface bond1
leaf01(config-if)# evpn mh es-df-pref 50000
leaf01(config-if)# evpn mh es-id 00:44:38:39:FF:00:AA:00:00:01
leaf01(config-if)# exit

ssasso: `evpn mh es-df-pref` is optional. as you can see, the rest is identical.
control plane messages are exchanged correcly, two frr istances see each other, DF is elected.

jbemmel: You can configure
```
evpn mh uplink
```
on uplinks. Docs suggest to enable ARP/ND suppression on VXLAN side

docs/plugins/evpn.es.md
ipspace: I would go for "evpn.multihoming", it's not like the storage is still expensive ;))

docs/plugins/evpn.es.md
```diff
@@ -0,0 +1,84 @@
...
+## Using the Plugin
+
+* Add `plugin: [ evpn.es ]` to the lab topology.
+* Define a set of *ethernet segments* on the topology top-level
+* Include the **evpn.es** attribute in the device interface
+
+### Supported attributes
+
+The plugin adds the following attributes defined at topology level:
+* **evpn.ethernet_segments** (dict) -- Key is the **ethernet segment** name. Each item is a *dict* with the following attributes:
```
ipspace: I'm guessing that not too many people care about the exact ESI values, so it might be nice to be able to say "hey, I need a segment" and get everything else working. **null** value for a segment should thus auto-generate **id**.

docs/plugins/evpn.es.md
```diff
@@ -0,0 +1,84 @@
...
+    * **id** (esi_id, mandatory) -- ESI in format `00:XX:XX:XX:XX:XX:XX:XX:XX:XX` (only Type-0 ESI is supported for now)
+    * **auto** (bool) -- Use ESI auto generation based on LACP System ID. If both `id` and `auto` are specified, explicit `id` takes over.
+
+Interface level attributes:
+* **evpn.es** (str) -- ethernet segment name as defined on `evpn.ethernet_segments`.
```
ipspace: A nice option would be to be able to specify the segment name here without defining it in the **evpn.ethernet_segments**, and have it magically appear in the segments with auto-generated ID.

I would suggest you do a count of how many times a segment is used in lab topology at the end and generate warnings for segments with a single interface. That would also catch typos in segment names if you auto-generate them from **evpn.es** attribute.

netsim/modules/lag.yml
ipspace: See the generic comment I'm about to make in the PR -- this thing is not as simple as it looks.

netsim/extra/evpn.es/frr.j2
ipspace: > Docs suggest to enable ARP/ND suppression on VXLAN side

There's an extremely good reason for that, see my generic comment

ipspace:
OK, now for the elephant in the room: replacing MC-LAG with EVPN MH is not as trivial as Juniper wants it to sound (a few people in the DC BU managed to do an incredible amount of damage with they cowboy tricks). At the very minimum, the two (or more) nodes of the MH cluster must have their LACP ID and ARP tables synchronized (see https://blog.ipspace.net/2022/11/mlag-vxlan-evpn/ and subsequent blog posts in that same series for more details).

I don't know how Arista and Junos do it, but I'm guessing you need to configure a static LACP system ID for things to work (like what you did in the integration test). Should that be the case, the checks in the plugin and LAG module have to reflect that. Ideally, we'd make it simple for the users and auto-generate ESI ID and LACP system ID for the (not)MLAG peers, but I'm also perfectly OK if we require static values. In any case, these requirements also have to be well-documented in the plugin documentation.

ssasso:
> I don't know how Arista and Junos do it, but I'm guessing you need to configure a static LACP system ID for things to work (like what you did in the integration test). Should that be the case, the checks in the plugin and LAG module have to reflect that. Ideally, we'd make it simple for the users and auto-generate ESI ID and LACP system ID for the (not)MLAG peers, but I'm also perfectly OK if we require static values. In any case, these requirements also have to be well-documented in the plugin documentation.

That's correct, you need to explicitly configure a static LACP system ID (same LACP ID per segment) - and that's why I modified the LAG module to include that parameter.

On the plugin itself, there is a specific validation that an ESI-LAG interface **must** have the attribute `lag.lacp_system_id` defined.

```
[...]
if not intf.get('lag.lacp_system_id',False):
    log.error(
        f'Node {node.name}({node.device}) cannot use auto ESI on LAG interface without "lacp_system_id" ({intf.ifname})',
        category=log.IncorrectAttr,
        module=_config_name)
[...]
```

I will definitely need to add more about this on the documentation - and will do it shortly.

As for auto LACP system ID generation, I think also ESI ID generation is a good idea. However, I wanted to focus on "basic" functionality first (i.e., make it work with user-defined values), and once that's ready and working I can expand it with more abstraction.

ssasso:
(oh wait the check must be performed for every ESI-LAG, not only where ESI is auto generated - the check is there, but in wrong place :) - moving it)

ipspace:
> (oh wait the check must be performed for every ESI-LAG, not only where ESI is auto generated - the check is there, but in wrong place :) - moving it)

Yeah 😉

ssasso:
Interesting part:

*cumulus_nvue* seems to work fine with the config templates I'm writing, and generated *frr* and *bond* config are **exactly** the same as the one configured by the *frr device* template.

I've read in some github issue/post that `It appears that in cumulus VX, SPH / non-DF filtering is implemented with TC filters handled by a separate cumulus daemon` - and I'm starting to think that this is true.

```
[lag]     Check LAG ports on X1

[ping]    End-to-end connectivity test [ node(s): h2,h3 ]
[PASS]    h2: Ping to h1 (172.16.0.4) succeeded
[PASS]    h3: Ping to h1 (172.16.0.4) succeeded
[PASS]    Test succeeded in 0.3 seconds
```

```
cumulus@s1:mgmt:~$ nv show evpn multihoming esi

ESInterface - Local interface, NHG - Nexthop group ID, DFPref - Designated
forwarder preference, VNICnt - ESI EVPN instances, MacCnt - Mac entries using
this ES as destination, RemoteVTEPs - Remote tunnel Endpoint, Flags - * - ready-
for_bgp, s - oper-up, x - non-designated-forward, r - remote, l - local, f -
designated-forward, b - bridge-port, A - active

ESI                            ESInterface  NHG        DFPref  VNICnt  MacCnt  RemoteVTEPs  Flags
-----------------------------  -----------  ---------  ------  ------  ------  -----------  -------
00:11:22:33:44:55:66:77:88:99  bond1        536870913  32767   1       3       10.0.0.2     lrf*bsA

cumulus@s1:mgmt:~$ nv show evpn multihoming esi 00:11:22:33:44:55:66:77:88:99
                        operational
----------------------  -----------
local-interface         bond1
df-preference           32767
mac-count               4
vni-count               1
nexthop-group-id        536870913
flags-string            lrf*bsA
flags
  local                 on
  remote                on
  ready-for-bgp         on
  bridge-port           on
  oper-up               on
  nexthop-group-active  on
  designated-forward    on
[remote-vtep]           10.0.0.2
```

jbemmel:
> I've read in some github issue/post that `It appears that in cumulus VX, SPH / non-DF filtering is implemented with TC filters handled by a separate cumulus daemon` - and I'm starting to think that this is true.

https://github.com/FRRouting/frr/discussions/11487#discussioncomment-3223654
```As I understand it, the SPH filters are applied as ingress TC rules to any interface configured in FRR with evpn mh uplink```

Cumulus docs also mention that the TC rules only apply to interfaces marked as 'evpn mh uplink'
```
nv set interface swp51-54 evpn multihoming uplink on
```

**Comment by ipspace** [comment]
> I've read in some github issue/post that `It appears that in cumulus VX, SPH / non-DF filtering is implemented with TC filters handled by a separate cumulus daemon` - and I'm starting to think that this is true.

According to https://containerlab.dev/manual/kinds/cvx/, Cumulus VX includes some kernel-mode support for MLAG. @ddutt might remember the details ;)

**Comment by ssasso** [comment]
> > I've read in some github issue/post that `It appears that in cumulus VX, SPH / non-DF filtering is implemented with TC filters handled by a separate cumulus daemon` - and I'm starting to think that this is true.
> 
> [FRRouting/frr#11487 (reply in thread)](https://github.com/FRRouting/frr/discussions/11487#discussioncomment-3223654) `As I understand it, the SPH filters are applied as ingress TC rules to any interface configured in FRR with evpn mh uplink`

but the point is that (on cumulus_nvue) everything works fine even **without configuring mh uplink**! So not specifing them is definitely not an issue, imho.

Anyway, unless some of you want to do more testing on frr itself, I will stop banging my head there for now.

**Comment by jbemmel** [comment]
> So not specifing them is definitely not an issue, imho.

Agree, was just pointing out that the TC rules likely aren't in play since you did not configure any uplinks.

I would suspect the difference between FRR and Cumulus NVUE is in sysctl settings - the nice thing of having it work on Cumulus means that you can inspect and replicate. It would be nice to have it working, but if you prefer to let it be for now I can certainly see why

**Comment by ssasso** [comment]
> I would suspect the difference between FRR and Cumulus NVUE is in sysctl settings

Another big candidate is different kernel... FRR runs as container on my host, while NVUE runs as KVM VM with its own "full stack". I will try (in best effort) so see if I spot some major difference on settings.

**Comment by ssasso** [comment]
@ipspace @jbemmel added code (and related doc update) for automatic definition of ESI + LACP Sys ID value

File: tests/topology/expected/lag-l2.yml (hunk 7 above)
Comment (ipspace) [reply tests/topology/expected/lag-l2.yml]: Agree -- this should be investigated. Also, this is definitely triggered by the additions (not saying the bug is not somewhere else), or we would have already seen it in dev branch.

File: netsim/devices/eos.py [hunk 13]
```diff
@@ -101,6 +101,13 @@ def passive_stub_interfaces(node: Box, topology: Box) -> None:
 
     intf.ospf.network_type = 'point-to-point'
 
+def esi_identifier_format(node: Box, topology: Box) -> None:
+  # Arista EOS, in EVPN ESI, wants ESI in format: 0000:0000:0000:0000:0000
+  for intf in node.interfaces:
+    if intf.get('evpn._esi.id'):
+      esi = intf.evpn._esi.id.replace(":", "")
+      intf.evpn._esi._eos_id = ':'.join(esi[i:i+4] for i in range(0, len(esi), 4))
```
ipspace, line 109: This is a neat trick. I have to remember it ;)

netsim/extra/evpn.multihoming/plugin.py
```diff
@@ -0,0 +1,206 @@
...
+        '_lacp_system_id': unique_ids[1],
+    }
+    return _auto_segments[esname]
+
+def _update_esi_stats(node: Box, interface: Box, esi_name: str, esi_data: dict) -> None:
+    if log.debug_active('plugin'):
+        print(f"evpn.multihoming: CALLED ESI STATS {node.name} {interface.ifname} {esi_name} {esi_data}")
+    # initialize struct if required
+    global _esi_stats
+    if esi_name not in _esi_stats.esi_reference: _esi_stats.esi_reference[esi_name] = []
```
ipspace: No need to change the code (comment for future reference only): it might be easier to have the _esi_stats indexed by esi_name, and then every entry be a data structure (= Box) with 'reference','lacp_id' and 'lacp_id_info'

netsim/extra/evpn.multihoming/plugin.py
```diff
@@ -0,0 +1,206 @@
...
+def post_transform(topology: Box) -> None:
+    global _config_name
+    global _auto_segments
+    global _esi_stats
+    _init_es_id_generator(topology)
+    for node in topology.nodes.values():
+        if not 'evpn' in node.module: continue
+        features = devices.get_device_features(node,topology.defaults)
+        es_supported = 'evpn.multihoming' in features
+        if not es_supported: continue
```
ipspace: You might want to check that the unsupported device does not use **evpn.es** anywhere on its interfaces?

netsim/extra/evpn.multihoming/plugin.py
```diff
@@ -0,0 +1,206 @@
...
+                intf_es_data = _generate_es_struct(intf_es)
+                if log.debug_active('plugin'):
+                    print(f'evpn.multihoming: Found interface ESI definition without reference: '
+                          f'{intf_es} ({node.name}->{intf.ifname}): creating structure. [{intf_es_data}]')
+                es_data.update({ intf_es: intf_es_data })
+            # If LAG, auto generated ESI-ID and LACP-System-ID are not present, use also auto generated LACP ID
+            if intf.type == 'lag' and not intf.get('lag.lacp_system_id',False) and '_lacp_system_id' in es_data[intf_es]:
+                intf.lag.lacp_system_id = es_data[intf_es]._lacp_system_id
+            # Error Validation checks
+            if intf_es and intf.type not in _es_supported_on:
```
ipspace: Again, for the future: I would put the validation checks into a separate function. This thing is becoming hard to read, particularly sorting the "what's really going on" part from the validation part.

netsim/extra/evpn.multihoming/plugin.py
```diff
@@ -0,0 +1,206 @@
...
+    global _auto_segments
+    global _esi_stats
+    _init_es_id_generator(topology)
+    for node in topology.nodes.values():
+        if not 'evpn' in node.module: continue
+        features = devices.get_device_features(node,topology.defaults)
+        es_supported = 'evpn.multihoming' in features
+        if not es_supported: continue
+        # Load Ethernet Segments data and expand it with auto segments
+        es_data = node.get('evpn.ethernet_segments', {})
```
ipspace: Do you want to update this data structure in isolation or in-place? This will probably give you a pointer to **node.evpn.ethernet_segments** if it exists and an isolated dictionary otherwise.

I'd go with simple **es_data = node.evpn.ethernet_segments** if you want to do an update of the node data structure. This would create an empty dictionary within the node data and return a pointer to it (which would then be stored in es_data)

ipspace (changes requested):
I would definitely check for **evpn.es** on interfaces of unsupported devices, or we would get confused users (as in "why does this not work?"). Everything else is at most a suggestion.

docs/plugins/evpn.multihoming.md
```diff
@@ -0,0 +1,143 @@
...
+links:
+# EVPN/VXLAN Switch to Switch Link
+- s1:
+  s2:
+  mtu: 1600
+# ESI-LAG
+- lag:
+    members:
+    - s1:
+        evpn.es: seg_1
```
jbemmel, line 89: I think we should require this to be declared explicitly. If not, a typo in the name would lead to 2 different ESIs being generated

netsim/devices/eos.py
jbemmel:
```
intf.evpn._esi._eos_id = ''.join(intf.evpn._esi.id[i] if (i-2)%6!=0 else '' for i in range(0,len(intf.evpn._esi.id)))
```

docs/plugins/evpn.multihoming.md
ipspace: ... which would trigger "only one node per segment" error

netsim/devices/eos.py
ipspace: This looks more and more like "turning Python into Perl" contest 🤪 I was staring at your code for a few minutes, and still only guess that it works.

docs/plugins/evpn.multihoming.md
jbemmel: ... but would still be confusing, as the root cause - a typo in the segment id - wouldn't be obvious from that error message

tests/topology/expected/lag-l2.yml
jbemmel: It's triggered by the addition of the ```lacp_system_id``` interface attribute

jbemmel: Add
```
intf_to_neighbor: False       # By default, do not include lag attributes in neighbors
```
to fix it

netsim/modules/lag.yml
```diff
@@ -34,14 +34,16 @@ attributes:
         min_value: 1
         _alt_types: [ bool ]
       mac: mac
-
-  # interface:     # Internal, don't declare as official attribute
-    # _mlag: bool
+  interface:
```
jbemmel, line 37: Add ```intf_to_neighbor: False       # By default, do not include lag attributes in neighbors``` here

jbemmel (changes requested):
* (IMHO) ethernet segments should be similar to vlans, in that they always need to be declared - even if all IDs are auto-generated. It helps users avoid mistakes by enforcing consistency / referential integrity
* should add ```intf_to_neighbor: False``` to lag module to avoid changes to neighbors

jbemmel:
Note that Cumulus NVUE fails to pass the test; after disabling the link, the ping from h2 to h1 fails

docs/plugins/evpn.multihoming.md
ssasso: I could add a "suggestion" in that warning message to allow the user to quickly identify the "root cause"

netsim/extra/evpn.multihoming/plugin.py
ssasso: remove data and generate a warning, or stop the execution with an error? which approach do you think it's best?

ipspace: Usually we'd stop the executing with an error. The few exceptions are things like lack of NSSA ranges -- things still sort-of work but suboptimally.

tests/integration/evpn.multihoming/01-esi-lag.yml
```diff
@@ -0,0 +1,97 @@
...
+    nodes: [ x1 ]
+    config:
+      template: ifdown
+      variable.state: down
+
+  failover:
+    description: End-to-end connectivity after a LAG member failure
+    nodes: [ h2, h3 ]
+    wait_msg: Waiting for ESI-LAG convergence
+    level: warning
```
jbemmel, line 87: Should this be a mere warning? Supporting failover scenario's is the whole point of doing ES Multihoming, Cumulus NVUE currently fails this test - so it basically doesn't support MH properly

ssasso: I copied (and adapted) your M-LAG tests. If this should not be a warning here, it should not be a warning here as well: https://github.com/ipspace/netlab/blob/110bf8ae45ff0777a2cfbb502fde6d7a2934829b/tests/integration/lag/10-mlag.yml#L60

ssasso:
* Updated lag to avoid attribute propagation on neigh
* @ipspace I decided to move the validation on a separate function as per your suggestion to keep the plugin flow "clean" as much as possible
* @jbemmel will investigate further NVUE - but if you have some spare time to have a look at it as well, it's appreciated

ipspace:
> @ipspace I decided to move the validation on a separate function as per your suggestion to keep the plugin flow "clean" as much as possible

Thank you!

> @jbemmel will investigate further NVUE - but if you have some spare time to have a look at it as well, it's appreciated

Based on how Nvidia rolls, NVUE is dead to me and I plan to declare it obsolete (together with CL 4.x) in the next release notes (did I mention https://www.youtube.com/shorts/tQIdxbWhHSM already?).

ipspace:
I lost track 🤦‍♂️ Is this ready for another review or a merge?

ssasso:
> I lost track 🤦‍♂️ Is this ready for another review or a merge?

Rebased, ready for review and merge if you think it's ok.

I won't investigate further on cumulus convergence on link failure.

ssasso:
> Looks good to me (and EOS implementation works ;). I streamlined the _esi_stats code a bit; if you're OK with those changes, let's merge this one.

all good, you can proceed, thank you!!!

Issue #2429, opened 2025-06-24
Make NUTS a _netlab_ external tool


ipspace:
It probably wouldn't be too hard to add NUTS (https://github.com/network-unit-testing-system/nuts) as an external tool.

/cc @ubaumann

ubaumann:
Yes this should be simple. I created a draft PR and created the files I think it needs but I will need one or two hours to test and finish the PR :)

Is it okay to execute the pytest directly with `netlab connect nuts` or would it be better to open a bash?

DanPartelly:
Do you guys maintain a docker container of NUTS, and would nuts be usable easily with it? It would save a lot of pain in the long run in using nuts.  A lot of distors do not pack nornir, napalm and the likes. 

EDIT: seems that it's in docker. Great news!

ubaumann:
@DanPartelly Yes, we have a latest container but with all source files in /app. I  will rework it in the near future to have a container for simple usage and a way to install custom dependencies. 
As a workaround for now, I use the `-w /nuts` to change the working directory to a clean directory.

ipspace:
> Is it okay to execute the pytest directly with `netlab connect nuts` or would it be better to open a bash?

I guess that's the most common way to start the tool, so it would be nice to have that happen automatically with **netlab connect**. Someone who wants to start bash can always do `netlab status` to find the name of the container, and then use `docker exec -it _container_ bash` to start bash.

ipspace:
Implemeneted in #2433

PR #2433, opened 2025-06-24
WIP: NUTS tool integration

ubaumann:
Add NUTS (https://github.com/network-unit-testing-system/nuts) as a tool

### DIFF
--- a/netsim/tools/nuts.yml
+++ b/netsim/tools/nuts.yml
@@ -1,0 +1,28 @@
+#
+# Network Unit Testing System (NUTS) tool
+#
+runtime: docker     # Default: start in a Docker container
+docker:
+  up:
+    docker run --rm -itd --name '{name}_nuts'
+      {sys.docker_net}
+      -v './nuts':/nuts
+      -w /nuts
+      ghcr.io/network-unit-testing-system/nuts
+  message:
+    Use 'netlab connect nuts' to start nuts CLI
+  connect:
+    docker exec -it '{name}_nuts' pytest
+  down:
+    docker kill '{name}_nuts'
+  cleanup:
+    docker volume rm '{name}_nuts'
+config:
+- dest: nr-config.yaml
+  template: nr-config.yaml
+- dest: inventory/groups.yaml
+  template: inventory/groups.yaml
+- dest: inventory/hosts.yaml
+  template: inventory/hosts.j2
+- dest: tests/test_interfaces.yaml
+  template: tests/test_interfaces.yaml
--- a/netsim/tools/nuts/inventory/hosts.j2
+++ b/netsim/tools/nuts/inventory/hosts.j2
@@ -1,0 +1,16 @@
+{% for nn,nd in nodes.items()
+     if nd.ansible_connection|default('') in ['ssh','paramiko','network_cli']
+       and ('ipv4' in nd.mgmt or 'ipv6' in nd.mgmt) %}
+{{ nn }}:
+  hostname: "{{ nd.mgmt.ipv4|default(nd.mgmt.ipv6) }}"
+  username: "{{ nd.ansible_user }}"
+  password: "{{ nd.ansible_ssh_pass }}"
+  connection_options:
+    netmiko:
+      extras: &id001
+        session_log: {{ nn }}.log
+        session_log_file_mode: append
+    napalm:
+      extras:
+        optional_args: *id001
+{% endfor %}
--- a/netsim/tools/nuts/tests/test_interfaces.j2
+++ b/netsim/tools/nuts/tests/test_interfaces.j2
@@ -1,0 +1,12 @@
+- test_class: TestNapalmInterfaces
+  test_data:
+{% for nn,nd in nodes.items()
+     if nd.ansible_connection|default('') in ['ssh','paramiko','network_cli']
+       and ('ipv4' in nd.mgmt or 'ipv6' in nd.mgmt) %}
+    - host: {{ nn }}
+      name: ToDo
+      is_enabled: true
+      is_up: true
+      # mtu: 9232
+      # speed: 10.0
+{% endfor %}
--- a/netsim/tools/nuts/inventory/groups.yaml
+++ b/netsim/tools/nuts/inventory/groups.yaml
@@ -1,0 +1,2 @@
+---
+# We could use groups to handle credentials for each device type
--- a/netsim/tools/nuts/nr-config.yaml
+++ b/netsim/tools/nuts/nr-config.yaml
@@ -1,0 +1,11 @@
+inventory:
+  plugin: SimpleInventory
+  options:
+    host_file: inventory/hosts.yaml
+    group_file: inventory/groups.yaml
+runner:
+  plugin: threaded
+  options:
+    num_workers: 100
+logging:
+  enabled: false


ubaumann:
Information and discussion: #2429

docs/extool/nuts.md
```diff
@@ -0,0 +1,17 @@
+(extool-nuts)=
+# Network Unit Testing System (NUTS)
+
+NUTS is a Pytest plugin enabling network testing using YAML files.
+
+* Add the following lines to the lab topology file to use NUTS with _netlab_:
+
+```
+tools:
+  suzieq:
```
ssasso: I guess this should be:
```
tools:
  nuts:
```

docs/extool/nuts.md
```diff
@@ -0,0 +1,17 @@
...
+```
+
+* Use the **netlab connect nuts** command to run pytest or **netlab connect suzieq bash** to open a shell in the docker container.
```
ipspace: s/suzieq/nuts/gi

netsim/tools/nuts/inventory/hosts.j2
```diff
@@ -0,0 +1,16 @@
+{% for nn,nd in nodes.items()
+     if nd.ansible_connection|default('') in ['ssh','paramiko','network_cli']
+       and ('ipv4' in nd.mgmt or 'ipv6' in nd.mgmt) %}
+{{ nn }}:
+  hostname: "{{ nd.mgmt.ipv4|default(nd.mgmt.ipv6) }}"
+  username: "{{ nd.ansible_user }}"
+  password: "{{ nd.ansible_ssh_pass }}"
```
ipspace, line 5: There's also the option to use Vagrant-generated SSH key on FRR VMs (not sure whether that's relevant). The path to the key (when it exists) is in **ansible_ssh_private_key_file** and it's relative to the current (lab) directory, so starting with `.vagrant/`

docs/extool/nuts.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Notes:**
+
+* All necessary files are created in the folder nuts and mounted to the docker container.
```
ipspace: You should add some information on adding new tests, and maybe a pointer to NUTS examples/tutorial/...

netsim/tools/nuts.yml
```diff
@@ -0,0 +1,28 @@
...
+runtime: docker     # Default: start in a Docker container
+docker:
+  up:
+    docker run --rm -itd --name '{name}_nuts'
+      {sys.docker_net}
+      -v './nuts':/nuts
+      -w /nuts
+      ghcr.io/network-unit-testing-system/nuts
+  message:
+    Use 'netlab connect nuts' to start nuts CLI
```
ipspace: "nuts CLI" or "nuts tests"?

ipspace (commented):
Very nice (and extremely fast) job. Please let me know once it's ready to merge ;)

docs/extool/nuts.md
ubaumann: Yes! copy/paste error. Thanks

docs/extool/nuts.md
ubaumann: Yes, a copy/past error. I guess I was tired last night. xD

ubaumann:
I have to add some more tests (OSPF, BGP, ...) and improve the documentation.
And do some testing. Hope I have it ready in 1 or 2 days

netsim/outputs/tools.py
```diff
@@ -36,6 +36,7 @@ def create_tool_config(tool: str, topology: Box) -> None:
       log.error(f'No destination file specified for tool configuration\n... tool {tool}\n... config {config}')
       continue
     fname = f'{tool}/{config.dest}'
+    Path(fname).parent.mkdir(exist_ok=True, parents=True)
```
ubaumann, line 39: This allows to use folders in the destination path. Would you need some tests?

ipspace: > This allows to use folders in the destination path.

Yeah I guessed that ;)

> Would you need some tests?

Maybe we can add a platform integration test (which wouldn't hurt for most tools anyway). Don't worry about this, I can handle it once you're done with this PR.

netsim/tools/nuts/inventory/hosts.j2
```diff
@@ -0,0 +1,52 @@
...
+  'cat8000c': 'ios',
+  'iol': 'ios',
+  'ioll2': 'ios',
+  'iosv': 'ios',
+  'iosvl2': 'ios',
+  'iosxr': 'iosxr',
+  'vmx': 'junos',
+  'vptx': 'junos',
+  'vsrx': 'junos',
+  'vjunos-switch': 'junos',
```
ssasso: \+ vjunos-router

ubaumann:
I had to change the connection behaviour to open a bash shell because I was unable to pass or change the command when using `netlab connect nuts`.
I would have liked to change the command executed when opening the container or pass a command to perform in the bash, but I was somehow unable to do so. 

The First Idea was to have something like:
`netlab connect nuts` - does a `docker exec -it '{name)_nuts' pytest`
`netlab connect nuts bash - does a `docker exec -it '{name)_nuts' bash`

So I just open the bash shell and the user has to type `pytest` but maybe that is even better because now we can also use pytest arguments.

ubaumann:
> Very nice (and extremely fast) job. Please let me know once it's ready to merge ;)

I think for a first experimental feature, it should be okayish. I have some issues to fix on the NUTS side.

netsim/tools/nuts/tests/test_bgp.j2
```diff
@@ -0,0 +1,43 @@
+{% set ns = namespace(found=false) %}
+- test_class: TestNapalmBgpNeighbors
+  test_data:
+{% for nn,nd in nodes.items() %}
+{%   for neighbor in nd.bgp.neighbors %}
```
ipspace: Just FYI (no need to change): this works only because the underlying data structure is a Box, creating nd.bgp and nd.bgp.neighbors on demand even if they did not exist before.

Even `nd.bgp is defined` would create nd.bgp, the only safe test is `"bgp" in nd`. However, that doesn't matter as _netlab_ creates a copy of the topology before creating a tool configuration file.

ipspace:
> I had to change the connection behaviour to open a bash shell because I was unable to pass or change the command when using `netlab connect nuts`.

Use **netlab connect -vv** to see exactly what's sent to the container.

> I would have liked to change the command executed when opening the container or pass a command to perform in the bash, but I was somehow unable to do so.

If you set the command to be executed as "pytest" then everything after "netlab connect nuts" is sent to pytest. However, in that case you have to do "docker exec -it bash" to start bash.

If you want to go down this path, then we'd have to modify the "netlab/up.py" code to treat the tool message as an f-string to generate the container name.

> The First Idea was to have something like: `netlab connect nuts` - does a `docker exec -it '{name)_nuts' pytest` `netlab connect nuts bash - does a `docker exec -it '{name)_nuts' bash`

The command to execute is an f-string, so in principle that would work if we would pass the args to the f-string in https://github.com/ipspace/netlab/blob/dev/netsim/cli/connect.py#L213, for example setting **topology.connect.args** to `rest`

> So I just open the bash shell and the user has to type `pytest` but maybe that is even better because now we can also use pytest arguments.

You could pass pytest arguments with **netlab connect nuts args**.

Anyway, I'm perfectly OK with having to run "pytest" in container. FWIW, **netlab connect nuts -c pytest** also works ;)

docs/extool/nuts.md
```diff
@@ -0,0 +1,40 @@
+(extool-nuts)=
+# Network Unit Testing System (NUTS)
+
+_This integration is experimental. Please report any issues you encounter on the [NUTS GitHub repository](https://github.com/network-unit-testing-system/nuts/issues)._
```
ipspace: I will do a bit of editing before merging the PR. For example, we have "warning" markup.

netsim/tools/nuts.yml
```diff
@@ -0,0 +1,32 @@
...
+#
+runtime: docker     # Default: start in a Docker container
+docker:
+  up:
+    docker run --rm -itd --name '{name}_nuts'
+      {sys.docker_net}
+      -v './nuts':/nuts
+      -w /nuts
+      ghcr.io/network-unit-testing-system/nuts
+  message:
```
ipspace, line 12: You can use `message: |` for multi-line message

netsim/tools/nuts/inventory/hosts.j2
```diff
@@ -0,0 +1,49 @@
+{% set napalm_platform = {
```
ipspace: You could move this to `inventory.napalm` tool parameter (like all defaults, the tool parameters are not schema-checked). It would allow people to add new devices via system defaults and you could check whether a device is supported when generating tests.

netsim/tools/nuts/inventory/hosts.j2
```diff
@@ -0,0 +1,49 @@
...
+  'vptx': 'junos',
+  'vsrx': 'junos',
+  'vjunos-switch': 'junos',
+  'vjunos-router': 'junos',
+  'srlinux': 'srl',
+  'routeros': 'ros',
+  'routeros7': 'ros',
+  'arubacx': 'aoscx',
+} %}
+{% set netmiko_platform = {
```
ipspace: Same as above: inventory.platform tool parameter

netsim/tools/nuts/inventory/hosts.j2
```diff
@@ -0,0 +1,49 @@
...
+     if nd.ansible_connection|default('') in ['ssh','paramiko','network_cli']
+       and ('ipv4' in nd.mgmt or 'ipv6' in nd.mgmt) %}
+{{ nn }}:
+  hostname: "{{ nd.mgmt.ipv4|default(nd.mgmt.ipv6) }}"
+  username: "{{ nd.ansible_user }}"
+  password: "{{ nd.ansible_ssh_pass }}"
+  platform: "{% if nd.device in napalm_platform %}{{ napalm_platform[nd.device] }}{% else %}{{ nd.device }}{% endif %}"
+  connection_options:
+    netmiko:
+{%  if nd.device in netmiko_platform %}
```
ipspace: I think this would become `defaults.tools.nuts.inventory.netmiko`

netsim/tools/nuts/tests/test_bgp.j2
```diff
@@ -0,0 +1,43 @@
+{% set ns = namespace(found=false) %}
+- test_class: TestNapalmBgpNeighbors
+  test_data:
+{% for nn,nd in nodes.items() %}
```
ipspace: If a topology has an FRR container, the tests crash because the the host is in the tests, but not in the inventory. You should check whether `nd.device` is supported (which would be easy if you move the inventory mappings into tool data)

netsim/tools/nuts/tests/test_bgp.j2
```diff
@@ -0,0 +1,43 @@
...
+{%   for neighbor in nd.bgp.neighbors %}
+{%     if "ipv4" in neighbor %}
+{%     set ns.found = true %}
```
ipspace: If we assume that every neighbor has at least one AF (and I think that's a pretty safe assumption, otherwise I'll prune the neighbor list in the BGP module), then you could change the top loop to `for nn,nd in nodes.items() if nd.bgp.neighbors` and then use the `else` clause of the FOR loop to generate `[]`. No pressure though, this is OK.

netsim/tools/nuts/tests/test_bgp.j2
ipspace: FWIW, you check whether the connection is reasonable in the interface checks, but not here.

netsim/tools/nuts/tests/test_interfaces.j2
```diff
@@ -0,0 +1,14 @@
+- test_class: TestNapalmInterfaces
+  test_data:
+{% for nn,nd in nodes.items()
+     if nd.ansible_connection|default('') in ['ssh','paramiko','network_cli']
```
ipspace: You could replace this with the checks for `node.device` belonging to one of the inventory schema. Also, _netlab_ is pretty unhappy if a node has no management address, so it's safe to assume it has one.

netsim/tools/nuts/tests/test_ospf.j2
```diff
@@ -0,0 +1,51 @@
+{% set ns = namespace(found=false) %}
+{% set neighbors_by_host = {} %}
+- test_class: TestNetmikoOspfNeighbors
+  test_data:
+{% for nn,nd in nodes.items()
```
ipspace, line 5: Yet again, you probably have to check whether the device is in the inventory

netsim/tools/nuts/tests/test_bgp.j2
ubaumann: I faced an issue that `{% if neighor.ipv4 %}` broke the bgp deployment. I have now tried to make the bgp and neighbor test more stable.

docs/extool/nuts.md
ubaumann: I updated it to use the warning block

netsim/tools/nuts.yml
ubaumann: When using `message: |` I am getting:

```
  File "/home/urs/projects/netlab/netsim/utils/strings.py", line 124, in eval_format
    return str(eval(ex,dict(data)))      # An awful hack to use f-string specified in a string variable
  File "<string>", line 1
    f'Use \'netlab connect nuts\' to connect to the NUTS container.
```

I am doing now just a new line to keep the yaml nicer
```yaml
  message:
    Use 'netlab connect nuts' to connect to the NUTS container.
    Run 'pytest' in the container to execute tests.
```

netsim/tools/nuts/inventory/hosts.j2
ubaumann: Thanks for this tip. Did not know it was so easy. This made a lot of stuff easier and nicer

ubaumann: is now `defaults.tools.nuts.inventory.napalm`

netsim/tools/nuts/inventory/hosts.j2
ubaumann: is now `defaults.tools.nuts.inventory.netmiko`

netsim/tools/nuts.yml
ipspace: >   File "/home/urs/projects/netlab/netsim/utils/strings.py", line 124, in eval_format
>     return str(eval(ex,dict(data)))      # An awful hack to use f-string specified in a string variable

Oh, ****, that... Sorry :(

netsim/tools/nuts/tests/test_bgp.j2
ubaumann: I now check the connection option, if bgp is active on this device and if the platform is in `defaults.tools.nuts.test_creation.bgp` (all NAPALM platforms I could find for now, but probably many community drivers have a bad get_bgp_neighbor implementation)

netsim/tools/nuts/tests/test_bgp.j2
ubaumann: I use the namespace `ns.found` because I need to know across the entire topology if any device has been found and created some tests. (all tests are in the same list so Nornir can collect the data in parallel)

I have some duplicated code I could probably get rid of, however, looping over `[neighbor.ipv4, neighbor,ipv6]` would need some more testing, and I'm not sure if one more loop makes it more readable. I can take a look at it

ipspace: Yeah, I know why you're using ns.found. I'm just saying that if **nd.bgp.neighbors** has some usable value you'll get some usable data in the test (which is why I said you could use the for/else loop).

Never mind, it's perfectly OK the way it is

ipspace:
The checks have passed. Ready to merge @ubaumann?

ubaumann:
@ipspace I just realised I forgot the logic to create the interface tests only for NAPALM platforms, but I hope everything is now fine.
Ready to merge if the checks still pass.

PR #2437, opened 2025-06-25
Create a vagrant pin file in apt preferences to stop potential upgrades

### LINKED ISSUE #2436: [BUG] Pin vagrant version
Vagrant is getting worse with every release; the last usable one is 2.4.3-1 (the one we're using right now). However, even though we install the version that works, the next **apt-get upgrade** will overwrite it, so we have to pin the version in **apt** preferences.

ipspace:
Fixes #2436

### DIFF
--- a/netsim/install/libvirt.sh
+++ b/netsim/install/libvirt.sh
@@ -53,6 +53,13 @@ $SUDO sh -c 'echo "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_rel
 #
 # Pin vagrant version to one we know works
 cat <<FILE | $SUDO tee /etc/apt/preferences.d/vagrant
+#
+# We have to pin the vagrant version because the newer versions contain bugs that
+# can completely ruin the virtualization environment, leaving stale VMs running
+#
+# See https://github.com/ipspace/netlab/issues/2185 and
+# https://github.com/ipspace/netlab/issues/2436 for details
+#
 Package: vagrant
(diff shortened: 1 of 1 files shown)


ipspace:
Would love to hear your feedback @ssasso @jbemmel

Anyway, it looks like in a few years we'll be left only with containerlab 🤷‍♂️

netsim/install/libvirt.sh
```diff
@@ -50,6 +50,13 @@ set -e
 # changed to new method - ghostinthenet - 20220417
 curl -fsSL https://apt.releases.hashicorp.com/gpg | $SUDO gpg --dearmor -o /etc/apt/trusted.gpg.d/hashicorp-security.gpg
 $SUDO sh -c 'echo "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" > /etc/apt/sources.list.d/vagrant.list'
+#
+# Pin vagrant version to one we know works
+cat <<FILE | $SUDO tee /etc/apt/preferences.d/vagrant
+Package: vagrant
+Pin: version 2.4.3-1
```
jbemmel, line 64: Can we add a comment that this is being pinned by Netlab install script, with explanation that later versions were found to have issues (on 6/25/2025)

ipspace: Sure, makes perfect sense. Thanks a million for the suggestion!

jbemmel (approved):

PR #2438, opened 2025-06-26
Add 'remove box' functionality to 'netlab libvirt' command

ipspace:
While it's easy to remove boxes with 'vagrant box', the plugin is too lazy (or does not get called) to cleanup the related libvirt volume.

The 'netlab libvirt remove' command finds the relevant box, removes the box, and deletes the associated libvirt volume(s).


ipspace:
@ssasso: If you still use vagrant (and have to build the boxes), you might find this command useful. Would love to hear your feedback.

PR #2441, opened 2025-06-26
Popos install support

zalekz:
The install scripts **ubuntu.sh** and **containerlab.sh** were modified to also allow installation on Pop!_OS. The modifications are minimal and supposedly do not break the logic of installing the required and nice-to-have packages on a vanilla Ubuntu system. After the modification, the installation procedure was tested thoroughly (but in "manual" mode) on virtual machines running Pop!_OS 22.04 as well as Ubuntu Server 22.04 LTS and 24.04 LTS. One hint unrelated to the script modifications — in general, it is best to perform the entire **netlab** installation from within Python **venv** on Ubuntu 22.04 or 24.04, because of the "scared Python" configuration in these distros.

### DIFF
--- a/netsim/install/containerlab.sh
+++ b/netsim/install/containerlab.sh
@@ -8,7 +8,7 @@ Docker/Containerlab Installation Script
 =====================================================================
 This script installs Docker and containerlab on a Debian or Ubuntu
 system. The script was tested on Debian 12 (bookworm) and Ubuntu 20.04,
-22.04, and 24.04.
+22.04, and 24.04, as well as Pop!_OS 22.04 LTS.
 
 NOTE: the script is set to abort on first error. If the installation
 completed you're probably OK even though you might have seen errors
@@ -43,9 +43,10 @@ echo "Install Docker GPG key and set up Docker repository"
 # Begin code to identify distribution and populate DISTRIBUTION variable - ghostinthenet - 20220417
 if [ -f /etc/debian_version ]; then
  if [ -f /etc/lsb-release ]; then
-  if [[ $(grep DISTRIB_ID /etc/lsb-release | awk -F'=' '{print $2;}') == 'Ubuntu' ]]; then
+  # Make sure both Pop!_OS and Ubuntu are supported, but identify the DISTRIBUTION as 'ubuntu' for Docker downloads
+  if [[ $(awk -F'=' '$1~/DISTRIB_ID/ && $2~/(Pop|Ubuntu)/ {print "MATCH";}' /etc/lsb-release) == 'MATCH' ]]; then
    DISTRIBUTION='ubuntu'
-  # Exit if lsb-release distribution ID isn't Ubuntu - ghostinthenet 20220418
+  # Exit if lsb-release distribution ID isn't Pop or Ubuntu - ghostinthenet 20220418
   else
    echo "Installed distribution is an untested Ubuntu derivative..."
    exit 1
@@ -70,6 +71,14 @@ curl -fsSL https://download.docker.com/linux/$DISTRIBUTION/gpg | $SUDO gpg --dea
 echo "deb [arch=$(dpkg --print-architecture)] https://download.docker.com/linux/$DISTRIBUTION $(lsb_release -cs) stable" | $SUDO tee /etc/apt/sources.list.d/docker.list > /dev/null
 #
 echo "Install Docker Engine"
+
+# Make sure no other process is blocking the apt-get update -- wait long enough
+# for any GUI-based system triggered interference to complete (e.g. packagekitd with Pop!_OS Desktop)
+echo "...If any triggered process is blocking the apt update, wait for it to complete."
+echo "Adjust the waiting time as necessary. Currently set to  11 seconds..."
+sleep 11s
+echo "Continuing.."
+
 $SUDO apt-get update
 $SUDO apt-get install -y $FLAG_QUIET docker-ce docker-ce-cli containerd.io
 echo "Install containerlab version $CONTAINERLAB_VERSION"
--- a/netsim/install/ubuntu.sh
+++ b/netsim/install/ubuntu.sh
@@ -4,7 +4,8 @@ Ubuntu Packages Installation Script
 =====================================================================
 This script updates your system, installs additional APT packages,
 and nice-to-have tools like git, jq... The script was tested on Debian
-12 (bookworm) and Ubuntu 20.04, 22.04, and 24.04.
+12 (bookworm) and Ubuntu 20.04, 22.04, and 24.04,
+as well as Pop!_OS 22.04 LTS.
 
 NOTE: the script is set to abort on first error. If the installation
 completed you're probably OK even though you might have seen errors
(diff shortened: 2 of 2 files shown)


netsim/install/containerlab.sh
```diff
@@ -70,6 +71,14 @@ curl -fsSL https://download.docker.com/linux/$DISTRIBUTION/gpg | $SUDO gpg --dea
 echo "deb [arch=$(dpkg --print-architecture)] https://download.docker.com/linux/$DISTRIBUTION $(lsb_release -cs) stable" | $SUDO tee /etc/apt/sources.list.d/docker.list > /dev/null
 #
 echo "Install Docker Engine"
+
+# Make sure no other process is blocking the apt-get update -- wait long enough
+# for any GUI-based system triggered interference to complete (e.g. packagekitd with Pop!_OS Desktop)
+echo "...If any triggered process is blocking the apt update, wait for it to complete."
+echo "Adjust the waiting time as necessary. Currently set to  11 seconds..."
+sleep 11s
```
jbemmel, line 79: Can we make this conditional on the host being Pop!_OS? (or a process named 'packagekitd' existing perhaps?) It is not needed for most common users, so this would make the majority wait uselessly

zalekz: Yup, I/we can make it conditional, but I think I must note here that running another apt update too soon after a previous one (as in this script), in my experience, might be an issue in Ubuntu as well.

E.g. when testing on my generic **Ubuntu** Server VMs, the oneliner of "**netlab install -y ubuntu ansible libvirt containerlab**" would fail for a similar reason (though something different from packagekitd would get in the way) in the beginning of the **libvirt** script. But splitting the procedure into two netlab install "stages" — "**netlab install -y ubuntu ansible**" and then "**netlab install -y libvirt containerlab**" would work. That is why I decided that the better workaround, at least, for the containerlab.sh script would be just to wait long enough, also because I'm afraid it might be the same issue on any Desktop-flavor of GNOME/Ubuntu-based system.

ipspace:
Looking at this challenge with a less-tired set of eyes ;)

* It's time to refactor the installation process. While I can do the tests on various Ubuntu distros and Debian Bookworm, I would appreciate if @zalekz would do the tests on Pop!_OS
* #2443 and #2444 address the things I didn't like in this PR. That still leaves the "missing package".

@zalekz: Could you please start with why the existing mechanism to identify a missing package does not work on Pop!_OS? **apt-cache show** returns an error if the package does not exist on Ubuntu, so there should be no problem if Pop!_OS does not have it.

ipspace:
Finally, @zalekz -- would you know what the best Vagrant box for Pop!_OS would be? I got a few hits on Vagrant Public Registry but with dubious history and provenance.

zalekz:
> Looking at this challenge with a less-tired set of eyes ;)
> 
> • It's time to refactor the installation process. While I can do the tests on various Ubuntu distros and Debian Bookworm, I would appreciate if @zalekz would do the tests on Pop!_OS
> 
I am happy to help as much as I can! :100: 

> • [Identify Linux distro in installation process #2443](https://github.com/ipspace/netlab/issues/2443) and [Check for package manager locks before starting the installation process #2444](https://github.com/ipspace/netlab/issues/2444) address the things I didn't like in this PR. That still leaves the "missing package".
> 
> 
> @zalekz: Could you please start with why the existing mechanism to identify a missing package does not work on Pop!_OS? **apt-cache show** returns an error if the package does not exist on Ubuntu, so there should be no problem if Pop!_OS does not have it.

@ipspace: It looks like for Pop!_OS, the package sort of "exists", but as a "purely virtual" one so the **apt-cache show** returns no errors, yet the package cannot be found and installed:

```
┌─[netlab@pop-os:~]$ 
└─〉 apt-cache show linux-modules-extra-$(uname -r)
N: Can't select versions from package 'linux-modules-extra-6.12.10-76061203-generic' as it is purely virtual
N: No packages found
┌─[netlab@pop-os:~]$ 
└─〉 echo $?
0

```

For comparison, when I try a totally bogus package name, the **apt-cache show** exits with (an expected) error:

```
┌─[netlab@pop-os:~]$ 
└─〉 apt-cache show linux-modules-extra-$(uname -r)-NONEXIST
N: Unable to locate package linux-modules-extra-6.12.10-76061203-generic-NONEXIST
N: Couldn't find any package by glob 'linux-modules-extra-6.12.10-76061203-generic-NONEXIST'
N: Couldn't find any package by regex 'linux-modules-extra-6.12.10-76061203-generic-NONEXIST'
E: No packages found
┌─[netlab@pop-os:~]$ 
└─〉 echo $?
100

```
Shouldn't parsing the **apt-cache** output be more reliable then checking its exit code in such situation?..

zalekz:
> Finally, @zalekz -- would you know what the best Vagrant box for Pop!_OS would be? I got a few hits on Vagrant Public Registry but with dubious history and provenance.

@ipspace : Unfortunately, I am totally new to Vagrant — sorry to say that, but I hope to learn more also by digging deeper into netlab internals ;-) .. I can try to build a box out of one of my Pop!_OS VMs in its initial state and share it somehow if that helps

ipspace:
> I can try to build a box out of one of my Pop!_OS VMs in its initial state and share it somehow if that helps

No worries, I'll take one of the dubious boxes and try it out. What could possible go wrong 🤪

ipspace:
> @ipspace: It looks like for Pop!_OS, the package sort of "exists", but as a "purely virtual" one so the **apt-cache show** returns no errors, yet the package cannot be found and installed:

Bummer 🤦‍♂️ What if we go ahead and try to install it, but ignore the error code?

> Shouldn't parsing the **apt-cache** output be more reliable then checking its exit code in such situation?..

Why do you think we have exit codes in Unix? That's the only sane mechanism to signal something went wrong, otherwise we could write expect scripts to do screen scraping till cows come home.

zalekz:
> > @ipspace: It looks like for Pop!_OS, the package sort of "exists", but as a "purely virtual" one so the **apt-cache show** returns no errors, yet the package cannot be found and installed:
> 
> Bummer 🤦‍♂️ What if we go ahead and try to install it, but ignore the error code?
> 

@ipspace: I didn't manage force-install it, seems that it is indeed not present, that is why I'm referring to this package as to "a mysterious one" (to me). By the way, what specific kernel modules from that linux-modules-extra collection do we need to run netlab simulations flawlessly? I think it makes sense for me to check if I actually have those modules in my  test instances of Pop!_OS or not. 


```
┌─[netlab@pop-os:~]$ 
└─〉 sudo apt install linux-modules-extra-$(uname -r)
[sudo] password for netlab: 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package linux-modules-extra-6.12.10-76061203-generic is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'linux-modules-extra-6.12.10-76061203-generic' has no installation candidate

```


> > Shouldn't parsing the **apt-cache** output be more reliable then checking its exit code in such situation?..
> 
> Why do you think we have exit codes in Unix? That's the only sane mechanism to signal something went wrong, otherwise we could write expect scripts to do screen scraping till cows come home.

In this specific case we seem to be in a grey area of "the item does not exist, but that's okay", where the exit code cannot be trusted, unfortunately...
```
┌─[netlab@pop-os:~]$ 
└─〉 apt-cache show linux-modules-extra-$(uname -r)
N: Can't select versions from package 'linux-modules-extra-6.12.10-76061203-generic' as it is purely virtual
N: No packages found

┌─[netlab@pop-os:~]$ 
└─〉 echo $?
0

```

ipspace:
Replaced by #2470 #2473 #2479 #2507

PR #2445, opened 2025-06-27
Dell OS10: Support mixed VLAN with OSPF / VXLAN topologies

jbemmel:
I realize this feature has been proposed (and rejected) in previous discussions, but I’m submitting this PR with a fresh implementation and clearer rationale in the hope that it can be reconsidered.

Why I’m proposing this again
In my world I am dealing with an enterprise network including Dell switches that run OSPF. It does not include VXLAN today, but we are considering to add it. I have made been making private patches to various templates to make that work, but maintaining this is cumbersome and something I'd like to avoid going forward

What this PR adds
This PR limits the use of virtual-network SVI interfaces to scenario's where VXLAN is used (i.e. VLANs having a VNI), selectively renaming those SVI interfaces. This allows for hybrid topologies in which some VLANs use OSPF, while others run VXLAN

All integration tests that previously worked still pass, and additional ones (such as ospf/ospfv2/06-vlan-mtu, ospf/ospfv3/07-vlan-mtu and gateway/03-vrrp-vlan.yml) now work too

Why it matters
I believe this feature adds value for users who use Netlab for real-world networks based on Dell switches. I’m happy to adjust the implementation if you have suggestions.

Thanks again for your time and for maintaining this project. I appreciate your consideration.

### DIFF
--- a/netsim/devices/dellos10.py
+++ b/netsim/devices/dellos10.py
@@ -7,16 +7,21 @@ from . import _Quirks,need_ansible_collection,report_quirk
 from ..utils import log, routing as _rp_utils
 from ..augment import devices
 
-def check_ospf_on_virtual_networks(node: Box) -> None:
+def check_vlan_ospf(node: Box, iflist: BoxList, vname: str) -> None:
+  name = node.name
   err_data = []
-  for vname in node.get('vxlan.vlans',[]):
-    vlan = node.get(f"vlans.{vname}",{})
-    if 'ospf' in vlan:
-      err_data.append(f'VXLAN VLAN {vname} (VNI {vlan.vni})')
+  for intf in iflist:
+    if 'ospf' not in intf or intf.type != 'svi':              # ospf.interfaces is not renamed
+      continue
+    if 'vlan' in intf:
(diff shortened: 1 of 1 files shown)


ssasso (approved):
Great rework, looks good to me.

I think it could be great if you can add a caveat or specific quirk to warn the user (or change to virtual-network mode) if MLAG is used.

MLAG with "plain" vlan interface is... "unstable" (to use nice wording)

ssasso:
(oh, and please describe the new behavior in the related documentation pages - tnx :) )

jbemmel:
> I think it could be great if you can add a caveat or specific quirk to warn the user (or change to virtual-network mode) if VLT (MLAG) is used.
> 
> VLT with "plain" vlan interface is... "unstable" (to use nice wording)

Thanks for the tip, taking a closer look

ipspace:
It looks like the "no OSPF over VXLAN VLANs" quirk got lost in translation, the evpn.20-vxlan-irb-ospf test starts but then fails to validate.

Apart from that, the integration tests pass. If you feel OK @ssasso, go ahead and merge it (please clean up the commit message that will be messy due to many fixes added after the PR was opened).

Issue #2414, opened 2025-06-21
[BUG] Static routes are created for fake ebgp.multihop "interface"


ipspace:
EBGP multihop plugin creates a fake "interface" that is used to create BGP neighbors for the EBGP multihop session. Unfortunately, the static routing code sees the opportunity to use that "interface" as another next hop for the static routes, resulting in "somewhat" incorrect configuration.

### Lab topology

```
defaults.device: eos
provider: clab
module: [ bgp, routing ]
plugin: [ ebgp.multihop ]

nodes:
  r1:
    bgp.as: 65000
    routing.static:
    - node: r2
      nexthop.node: r2
  r2:
    bgp.as: 65001
    routing.static:
    - node: r1
      nexthop.node: r1

links: [ r1, r1-r2, r1-r2, r2 ]

bgp.multihop.sessions: [ r1-r2 ]
```

### Resulting static route configuration

```
ip route 10.0.0.2/32 Ethernet2 10.1.0.2
ip route 10.0.0.2/32 Ethernet3 10.1.0.6
ip route 10.0.0.2/32 _ebgp_multihop_5 10.0.0.2
```

### Version

I'm positive this bug is in every _netlab_ version that supports static routes

PR #2449, opened 2025-06-27
Dell OS10 ebgp.multihop

jbemmel:
* Passes first ebgp.multihop test (second one uses OSPFv3 in VRF which Dell OS10 does not support)
* Enables evpn/14-vxlan-ebgp-ebgp


ipspace:
Do you expect this to work in VRFs (even though it won't go through the integration test)? If so, you have to set `multihop.vrf: True`

jbemmel:
> Do you expect this to work in VRFs (even though it won't go through the integration test)? If so, you have to set `multihop.vrf: True`

Added, I tried modifying the test to use static routes but that runs into #2414 

Found the same flag missing for SR Linux - see #2452

ipspace:
> Added, I tried modifying the test to use static routes but that runs into #2414

Also, we might have implemented EBGP multihop (because it's trivial) but not static routes for some devices.

PR #2453, opened 2025-06-28
Enable iosvl2 under containerlab provider. 

DanPartelly:
Enable iosvl2 under containerlab provider. While at it, modify the provider by device matrix in documentation to reflect those changes. Default image is set to vrnetlab/iosvl2, tag 15.2.2020.

Hope this is not controversial ( didnt open an issue before to see if we want it,  but the deed is done)


ipspace:
> Enable iosvl2 under containerlab provider. While at it, modify the provider by device matrix in documentation to reflect those changes. Default image is set to vrnetlab/iosvl2, tag 15.2.2020.

I'm guessing you got that tag from the IOSvL2 image shipping with CML (or some similar)

> Hope this is not controversial ( didnt open an issue before to see if we want it, but the deed is done)

Of course it's not. Thank you!

ipspace (approved):
Looks good to me

PR #2455, opened 2025-06-28
Do not apply "isis network point-to-point" on virtual interfaces

### LINKED ISSUE #2448: [BUG] isis network point-to-point on tunnels device IOS family
### Describe the bug

isis network point-to-point is not available for tunnel interfaces on isov/iol (unnumbered or not)  Therefore the isis routing template will fail to be deployed correctly. 

@ipspace any ideas why this is so ?  Maybe it is point to point by default  on tunnels on Cisco I do not really know ISIS. 

### Repro

topology like in  #2402 but with ISIS module.

DanPartelly:
Do not apply "isis network point-to-point" on virtual interfaces for ios.  Fixes #2448.

### DIFF
--- a/netsim/ansible/templates/isis/ios.j2
+++ b/netsim/ansible/templates/isis/ios.j2
@@ -13,7 +13,7 @@ interface {{ l.ifname }}
 {%   if ('ipv6' in l or 'ipv6' in l.dhcp.client|default({})) and 'ipv6' in isis.af  %}
   ipv6 router isis {{ isis.instance }}
 {%   endif %}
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
VLAN is also a virtual interface. Did you check what happens there?

DanPartelly:
No, I did not, and I messed up. Should have checked that. It is available for Vlan, BDIs, and prolly others. Fixing.

ipspace (approved):
Looks good ;)

PR #2458, opened 2025-06-29
Fix: Configure EIGRP on unnumbered interfaces

### LINKED ISSUE #2440: [BUG] EIGRP module with ip unnumbered
### Describe the bug

EIGRP module generates wrong config when encountering ip unnumbered interfaces. Needs to be taught how to deal with them.

### To Reproduce

Same topology and config as in #2402 , but with EIGRP module enabled. Test device was iol this time.

ipspace:
This fix results in a valid EIGRP configuration for an IOS device that has unnumbered interfaces. However, it looks like EIGRP does not work over unnumbered Ethernets anyway (it might over tunnels).

Closes #2440

### DIFF
--- a/netsim/ansible/templates/eigrp/ios.j2
+++ b/netsim/ansible/templates/eigrp/ios.j2
@@ -8,7 +8,7 @@ router eigrp {{ eigrp.as }}
 {% if eigrp.router_id|ipv4 %}
  eigrp router-id {{ eigrp.router_id }}
 {% endif %}
-{% for l in netlab_interfaces|default([]) if 'eigrp' in l and 'ipv4' in l  %}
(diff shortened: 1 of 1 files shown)


DanPartelly (approved):
Thank you. Ill check it on tunnels today or tomorrow and report back.

DanPartelly:
@ipspace 

It works with both  iol and iosv  image. It forms an adjacency over the tunnel despite the loopback0 masks on the routers being /32 and in different nets  (same net loopbacks might have been a requirement in past ?? ) 
```
      10.0.0.0/8 is variably subnetted, 4 subnets, 2 masks
D        10.0.0.2/32 [90/27008000] via 10.0.0.2, 00:04:20, Tunnel0
      172.16.0.0/16 is variably subnetted, 3 subnets, 2 masks
D        172.16.1.0/24 [90/26905600] via 10.0.0.2, 00:04:20, Tunnel0

```

PR #2459, opened 2025-06-29
More nuanced 'ipv4.unnumbered' feature test checks interface type

### LINKED ISSUE #2402: [BUG] Unnumbered tunnels might work on IOSv (Classic)
It's probably possible to configure **ip unnumbered** on tunnel interfaces even on IOSv, but the current "can we do it" device feature check fails because it does not differentiate between interface types.

We need to modify that logic to support a list of interface types which can be unnumbered.

Based on discussion in #2399 started by @DanPartelly

ipspace:
Cisco IOSv supports unnumbered tunnel interfaces but not Ethernet interfaces. The 'unnumbered IPv4' feature check was modified to check the supported interface type(s) if the value of ipv4.unnumbered feature is a list.

Closes #2402


DanPartelly (approved):
Just tested it. Another bug down, another feature.

PR #2462, opened 2025-06-29
Fix: calculate BGP default values from valid values of BGP attributes

ipspace:
Instead of having a hard-coded list of default session types (and similar), this change copies them from the valid values of global BGP attributes, making it possible for plugins to add new session types.

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -14,6 +14,32 @@ from ..data.validate import validate_item
 from ..augment import devices
 from ..utils import log, routing as _rp_utils
 
+BGP_DEFAULT_SESSIONS: typing.Dict[str,list] = {
+}
+
+BGP_VALID_SESSION_TYPE: list = []
+
+BGP_DEFAULT_COMMUNITY_KW: typing.Final[dict] = {
+  'standard': 'standard',
+  'extended': 'extended'
+}
+
+BGP_INHERIT_COMMUNITY: Box
+
+def setup_bgp_constants(topology: Box) -> None:
+  global BGP_DEFAULT_SESSIONS, BGP_VALID_SESSION_TYPE, BGP_INHERIT_COMMUNITY
+
+  # Valid session types are reconstructed from the valid values of the bgp.sessions.ipv4 global attribute
+  # We're assuming there's no difference between IPv4 and IPv6
+  BGP_VALID_SESSION_TYPE = [ v for v in topology.defaults.bgp.attributes['global'].sessions.ipv4 ]
+
+  # We're assuming all address families are active on all session types
+  for af in log.AF_LIST:
+    BGP_DEFAULT_SESSIONS[af] = BGP_VALID_SESSION_TYPE
+
+  # Finally, copy the pointer to BGP community inheritance
+  BGP_INHERIT_COMMUNITY = topology.defaults.bgp.attributes._inherit_community
+
 def check_bgp_parameters(node: Box, topology: Box) -> None:
   if not "bgp" in node:  # pragma: no cover (should have been tested and reported by the caller)
     return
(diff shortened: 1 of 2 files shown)


ipspace:
Prerequisite for #2426

netsim/modules/bgp.py
```diff
@@ -14,6 +14,32 @@
...
+}
+
+BGP_INHERIT_COMMUNITY: Box
+
+def setup_bgp_constants(topology: Box) -> None:
+  global BGP_DEFAULT_SESSIONS, BGP_VALID_SESSION_TYPE, BGP_INHERIT_COMMUNITY
+
+  # Valid session types are reconstructed from the valid values of the bgp.sessions.ipv4 global attribute
+  # We're assuming there's no difference between IPv4 and IPv6
+  BGP_VALID_SESSION_TYPE = [ v for v in topology.defaults.bgp.attributes['global'].sessions.ipv4 ]
```
jbemmel: I think we should consider adding a method to api/__init__.py for a plugin to add a session type. It will need to be added to both ```ipv4``` and ```ipv6``` for both ```sessions``` and ```activate```

We could import a method from the bgp module to keep the logic where it belongs, but I'd rather invoke a method that lives in ```bgp.py``` (i.e. this file) rather than hard-code the structure inside the plugin

jbemmel:
```Python
"""
Plugin API function to add a new session type, in all the right places
"""
def add_bgp_session_type(topology: Box, session_type: str, inherit_from: str = None) -> None:
  base = topology.defaults.bgp.attributes['global']
  base.sessions.ipv4.append( session_type )
  base.sessions.ipv6.append( session_type )
  base.activate.ipv4.append( session_type )
  base.activate.ipv6.append( session_type )
  if inherit_from:
    topology.defaults.bgp.attributes._inherit_community[ session_type ] = inherit_from
```

Something like this, and then do ```from ..modules.bgp import add_bgp_session_type``` in ```api```

jbemmel:
I tried to work this in, but there is a problem: The BGP module creates the BGP sessions and validates them against bgp.sessions in one go (```node_post_transform```). The plugin doesn't have a way to update the session type in between the moment the BGP module creates them, and the moment they are validated against ```bgp.sessions```

So if I define my topology like this:
```
bgp.sessions:
  ipv4: [ ibgp, confed_ebgp ]  # To test that the new session type is available
```
only the ibgp sessions get included, the ```ebgp``` sessions are lost (before the plugin sees them)

jbemmel:
The template additions for confederation are really minimal:
```
!
router bgp {{ bgp.as }}
  bgp confederation identifier {{ bgp.confederation.as }}
{% if bgp.confederation.peers %}
  bgp confederation peers {{ bgp.confederation.peers|join(' ') }}
{% endif %}
```

So even though having a separate plugin might be cleaner, I think in this case integrating the feature in the bgp module itself may be better

netsim/modules/bgp.py
ipspace: Created a custom type to validate BGP sessions, so the change would have to be made in a single place.

ipspace:
> So even though having a separate plugin might be cleaner, I think in this case integrating the feature in the bgp module itself may be better

Let me mull this over. In the meantime, I'd appreciate your feedback on this one -- I have a few bugs to fix after this one gets merged.

ipspace:
> As is, a plugin could add a session type but it wouldn't be able to modify the bgp sessions to use the new type for specifying which bgp sessions to create.

Not true. Of course, the plugin could modify **bgp.sessions**.

However, the problem you have with your particular project is that you want to have a new session type **for the bgp sessions generated within the existing code**, and there's no good way to do that unless we'd have a callback mechanism where the BGP module would ask _someone_ "what do you think this session type should be". That's exactly why I rolled **localas_ibgp** support into the BGP module.

EBGP multihop plugin would have no such limitations. In principle, your plugin could also readjust the session types etc. in after the main BGP module has finished its job.

> One could argue that's not needed, in which case it should be documented as a known limitation

So far, I'm aware (based on the issues/discussions/other activity in this repo) of a single person writing plugins for rarely-used BGP edge cases, so consider these comments "documentation of a known limitation" 😜

PR #2463, opened 2025-06-29
Force isis three-way-handshake to "ietf" mode for p2p links and tunnels

### LINKED ISSUE #2457: [BUG] Cisco IOSv has to be configured to use IETF IS-IS P2P hellos
See #2448 for details

DanPartelly:
Force isis three-way-handshake to "ietf" mode for p2p links and tunnels across Cisco ios family (iosv is the only one which still defaults to "cisco" mode ).

Closes #2457

### DIFF
--- a/netsim/ansible/templates/isis/ios.j2
+++ b/netsim/ansible/templates/isis/ios.j2
@@ -16,6 +16,9 @@ interface {{ l.ifname }}
 {%   if l.isis.network_type is defined and l.type not in ['tunnel','loopback']  %}
   isis network {{ l.isis.network_type }}
 {%   endif %}
+{%   if l.type in ['tunnel','p2p']  %}
+  isis three-way-handshake ietf
+{%   endif %}
 {%   if l.isis.cost is defined or l.isis.metric is defined %}
   isis metric {{ l.isis.metric|default(l.isis.cost) }}
 {%     if ('ipv6' in l or 'ipv6' in l.dhcp.client|default({})) and 'ipv6' in isis.af  %}


ipspace:
> iosv is the only one which still defaults to "cisco" mode

What about iosvl2?

netsim/ansible/templates/isis/ios.j2
```diff
@@ -16,6 +16,9 @@ interface {{ l.ifname }}
 {%   if l.isis.network_type is defined and l.type not in ['tunnel','loopback']  %}
   isis network {{ l.isis.network_type }}
 {%   endif %}
+{%   if l.type in ['tunnel','p2p']  %}
```
ipspace: It probably works as-is, but you might want to check whether isis.network_type is "point-to-point" not if the interface type is p2p.

netlab can set IS-IS network type to "point-to-point" on LAN interfaces (for example, when connecting a VM to a container)

ipspace (changes requested):
Incorrect condition checked when configuring 3-way handshake. The important parameter to consider is the IS-IS network type.

DanPartelly:
> > iosv is the only one which still defaults to "cisco" mode
> 
> What about iosvl2?

Deafults to ietf. Ironically.

ipspace:
> Deafults to ietf. Ironically.

🤦‍♂️🤢

netsim/ansible/templates/isis/ios.j2
DanPartelly: Ok, it specifically checks now for isis point to point.

ipspace (approved):
Integration tests passed, merging

PR #2468, opened 2025-07-01
Better way to disable KARL and make more space

remilocherer:
Disable kernel relinking (KARL, Kernel Address Randomized Link) by removing the checksum file instead of the relinking script.
Since disk space is very scarce with the current available VM image delete all the object files that would be used for kernel relinking and additional kernels in the same directory.

### DIFF
--- a/netsim/install/libvirt/openbsd/user-data
+++ b/netsim/install/libvirt/openbsd/user-data
@@ -10,6 +10,6 @@ bootcmd:
 - echo "library_aslr=NO" >> /etc/rc.conf.local
 runcmd:
 - echo "library_aslr=NO" > /etc/rc.conf.local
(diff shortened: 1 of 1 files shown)


DanPartelly:
Does it really matter if you remove the relink script, or the checksum ?  In the end the result is the same.  More disk space on this image  is very useful.

remilocherer:
The risk I see is that deleting the script is carried over to other systems. Deleting the checksum is the official method. And with that you can still revert you decision.

ipspace:
> Does it really matter if you remove the relink script, or the checksum ? In the end the result is the same.

FYI: If anyone around here knows anything about OpenBSD, it's @remilocherer ;)

Issue #2427, opened 2025-06-24
[BUG] Installation scripts specified in 'netlab install' should be executed in dependency order


ipspace:
If you run 'netlab install libvirt ubuntu', the first script installs specific version of Vagrant, and the second script seems to upgrade Vagrant to the latest version.

In any case, we should execute the installation scripts in some logical order. For example, **grpc** script cannot be execute before **ansible** script.

ipspace:
Fixed in #2469

PR #2469, opened 2025-07-01
Sort 'netlab install' scripts in dependency order

ipspace:
* Add a 'script configuration' file
* Use the 'scripts' keys from the configuration file to execute scripts in correct order
* Now that we have that file, display nice usage summary

Cleanup:
* Move the inline PIP/APT flag settings to a function
* Remove the ancient comments from libvirt.sh

### DIFF
--- a/netsim/cli/install.py
+++ b/netsim/cli/install.py
@@ -6,26 +6,22 @@
 import typing
 import argparse
 import os
-import glob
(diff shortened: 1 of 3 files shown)


ipspace:
Fixes #2427

ipspace:
> looks good!

Thank you!

PR #2471, opened 2025-07-01
Improve OpenBSD interface configuration and enable handling of RAs.

remilocherer:
- Configure and start rad on a router to send out RAs
- Utilise SLAAC on IPv6 enabled interfaces on a host
- Allow setting only IPv6 LLA

### DIFF
--- a/netsim/ansible/templates/initial/openbsd.j2
+++ b/netsim/ansible/templates/initial/openbsd.j2
@@ -27,15 +27,14 @@ ifconfig {{ intf.ifname }} -inet6
 ifconfig {{ intf.ifname }} inet {{ intf.ipv4 }} {{ alias }}
 {% endif %}
 {% if intf.ipv6 is defined %}
+{%   set is_host = role == 'host' %}
 {%   if role == 'router' %}
-{%     if intf.ipv6 is string %}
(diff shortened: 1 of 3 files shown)


netsim/ansible/templates/initial/openbsd.j2
```diff
@@ -1,53 +1,65 @@
...
-ifconfig {{ intf.ifname }} inet6
+{%   if role == 'router' %}
+{%     if intf.ipv6 is string %}
+ifconfig {{ intf.ifname }} inet6 {{ intf.ipv6 }}
+{%     else %}
+ifconfig {{ intf.ifname }} inet6 eui64
+{%     endif %}
+echo interface {{ intf.ifname }} >> /etc/rad.conf
+{%   elif role == 'host' %}
+ifconfig {{ intf.ifname }} inet6 autoconf
```
ipspace: A host can have a static IPv6 address, and we often expect it to be reachable under that address, so it looks like we should have:

* router: fixed IPv6 or eui64
* host: fixed IPv6 (+ autoconf) or autoconf

If you're OK with it, I can fix/test that before merging.

remilocherer: Sure, go ahead. Thanks!

ipspace (approved):
Passing all integration tests 🎉

PR #2479, opened 2025-07-03
Automatic detection of Linux distro in 'netlab install'

ipspace:
'netlab install' uses /etc/os-release file to find the linux distro (ID) and its "parents" (ID_LIKE). It uses that information to check whether it's safe to execute an installation script and selects the best (= first mentioned in the config file) option from ID_LIKE when there's no exact match with ID.

### DIFF
--- a/netsim/cli/install.py
+++ b/netsim/cli/install.py
@@ -7,6 +7,8 @@ import typing
 import argparse
 import os
 import sysconfig
+import csv
+import yaml
 from pathlib import Path
 from box import Box
 
@@ -59,6 +61,28 @@ def install_parse(args: typing.List[str], setup: Box) -> argparse.Namespace:
     help='Run the specified installation script')
 
   return parser.parse_args(args)
+
+"""
+read_config_setup: Read the installation configuration file and add information
+from /etc/os-release file (if it exists)
+"""
+def read_config_setup() -> Box:
+  cf_name = 'package:install/install.yml'
+  setup = read.read_yaml(filename=cf_name)
+  if setup is None or not setup:
+    log.fatal(f'Cannot read the installation configuration file {cf_name}')
+  
+  os_release = Path('/etc/os-release')
+  if not os_release.exists():
+    log.info(f"The {os_release} file does not exist. I have no idea what operating system you're using")
+    setup.distro.ID = "unknown"
+  else:
+    try:
+      with open(os_release) as stream:
+        setup.distro = dict(csv.reader(stream, delimiter="="))
+    except Exception as ex:
+      error_and_exit(f'Cannot read {str(os_release)}: {str(ex)}')
+  return setup
 
 """
 check_crazy_pip3: deals with crazy pip3 that thinks installing Python packages in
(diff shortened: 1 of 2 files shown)


ipspace:
Next step in the "fixing installation scripts" saga: automatically identify the distro and refuse to start the installation if we don't know what to do.

@zalekz: This should fix the next set of your problems (and make containerlab script work)

zalekz:
> @zalekz: This should fix the next set of your problems...


@ipspace : I feel somewhat uncomfortable, as if all the troubles are only because of me.. Thanks a lot on behalf or everyone who'd want to use netlab on a less standard Linux distro! That's the amount of attention one rarely gets from a vendor's TAC, I suppose. I guess I'd wait with testing the `containerlab.sh` without the update to the `ubuntu.sh` to handle the linux-modules-extra-* package availability though

DanPartelly:
> I feel somewhat uncomfortable, as if all the troubles are only because of me.. Thanks a lot on behalf or everyone who'd want to use netlab on a less standard Linux distro! 

Heh. This work is important, but keep in mind, it's really Debian family only,  I use Netlab on some trully different linux distros for example on Nixos ... that is "less standard".  PopOs is as standard as it gets. As are enteprise RPM based distros Thank you as well for all the effort you put into this pull request.

ipspace:
> I use Netlab on some trully different linux distros for example on Nixos ... that is "less standard".

I think I already said this: if you ever feel like creating the installation scripts for these distros, your PR would be most welcome. IIRC you have "a few" issues with Vagrant (who doesn't), but even containerlab installation script that would include KVM/QEMU (to support vrnetlab containers) would be most appreciated.

ipspace:
> I feel somewhat uncomfortable, as if all the troubles are only because of me.

No worries. You opened a can of worms (or is it bugs) that deserve to be squashed ;)

> That's the amount of attention one rarely gets from a vendor's TAC, I suppose.

Thank you!

DanPartelly:
> > I use Netlab on some trully different linux distros for example on Nixos ... that is "less standard".
> 
> I think I already said this: if you ever feel like creating the installation scripts for these distros, your PR would be most welcome. IIRC you have "a few" issues with Vagrant (who doesn't), but even containerlab installation script that would include KVM/QEMU (to support vrnetlab containers) would be most appreciated.


This was maintained almost release to release : https://build.opensuse.org/package/show/home:danp/networklab 
You get proper dependencies including  containerlab . But no vagrant.  (jailing is not my idea of fun. I did it to test your first OpenBSd vagrant box,and if I keep that jail intact, vagrant will work till the end of time,  but it costed me (too) many hours )  You get a proper solution for Suse from my repository. I can provide Nixos flakes, but i wont make any attempt to upstream the package. 

I **strongly** believe that  proper packaging  per distro is the solution, not install scripts. I have reservations with install scripts. They can and will break python sooner or later if the machine is not dedicated to netlab.

In the end, I think our end goals are aligned here. make netlab known and easy to use.  But  the way I perceive it would take half the effort to maintain a repo for RedHat  eneprisefamily -- including rocky and alma and  one debian and descendants. And have scripts to udpate that

PR #2481, opened 2025-07-03
Initial implementation of iol2 STP

DanPartelly:
Based on Arista. 

It passes all netlab integration tests and was also validated on a lot of topologies from:

 `https://github.com/ipspace/netlab-examples/tree/master/STP/examples`

Doesn't handle stub port types (does any template ???)

### DIFF
--- a/netsim/ansible/templates/stp/ios.j2
+++ b/netsim/ansible/templates/stp/ios.j2
@@ -1,4 +1,4 @@
-{% if  stp.enable == False %}
+{% if not stp.enable|default(True)  %}
 no spanning-tree vlan 1-4094
 {% else %}
 
--- a/netsim/ansible/templates/stp/ios.j2
+++ b/netsim/ansible/templates/stp/ios.j2
@@ -21,13 +21,13 @@ spanning-tree mst 0 priority {{ stp.priority }}
 {% endif %}
 
 {# Check for per-VLAN enable and priority; implies Rapid-PVST #}
-{% if vlans is defined %}
+{% if vlans is defined and stp.protocol == 'pvrst' %}
 {%   for vname,vdata in vlans.items() %}
-{%     if not vdata.stp.enable|default(stp.enable)|default(True) %}
+{%     if not vdata.stp.enable|default(True) %}
 no spanning-tree vlan {{ vdata.id }}
 {%     elif vdata.stp.priority is defined %}
 spanning-tree vlan {{ vdata.id }} priority {{ vdata.stp.priority }}
-{%     elif stp.protocol == 'pvrst' and stp.priority is defined %}
+{%     elif  stp.priority is defined %}
 spanning-tree vlan {{ vdata.id }} priority {{ stp.priority }}
 {%     endif %}
 {%   endfor +%}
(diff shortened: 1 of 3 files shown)


netsim/ansible/templates/stp/ios.j2
```diff
@@ -0,0 +1,65 @@
+{% if not stp.enable|default(True) %}
+{%  set mode = "none" %}
+{% else %}
+{%  set proto_map = { 'stp': 'mst', 'rstp': 'mst', 'pvrst': 'rapid-pvst', 'mstp': 'mst' } %}
+{%  set mode = proto_map[stp.protocol] %}
+{% endif %}
+{# options are 'none', rstp, rapid-pvst or mstp (default if not set) #}
+{% if mode == 'none' %}
+no spanning-tree vlan 1
```
jbemmel: ```no spanning-tree vlan 1-4094``` (or whatever the max VLAN id is)

DanPartelly: Since default mode in Cisco is pvrst , if mode is none, the vlan loop disables the VLANs just in time.

netsim/ansible/templates/stp/ios.j2
```diff
@@ -0,0 +1,65 @@
+{% if not stp.enable|default(True) %}
+{%  set mode = "none" %}
```
jbemmel: Could simply move
```
no spanning-tree vlan 1-4094
```
here, and skip setting mode

jbemmel: The rest of the template should be skipped when STP is disabled?

jbemmel (changes requested):
Just some minor structural changes suggested:
```
{% if (STP disabled) %}
no spanning-tree vlan 1-4094
{% else %}
# Rest of the template
{% endif %}
```

.. although I suppose in theory one could disable stp at the top level, and enable it for a particular vlan or interface.

Up to you if you think that should be supported

netsim/ansible/templates/stp/ios.j2
DanPartelly: Does any of the templates we have do this ? Most just set mode to none than continue.  Im not sure this matters anyways, if the template result is correct.  this is not the critical path of anything

DanPartelly:
We have in docs [STP Docs](https://netlab.tools/module/stp/) in section Global, Node, Link, Interface, and VLAN Parameters a tip: You can set the ‌stp.enable parameter in the ‌vlans dictionary to enable per-VLAN STP.  

So ... Anyways, Ill change it to your preferred version since you wrote the STP module and you know best what where your intentions behind some choices.

netsim/ansible/templates/stp/ios.j2
DanPartelly: Ive changed it as you suggested.

netsim/ansible/templates/stp/ios.j2
```diff
@@ -0,0 +1,65 @@
+{% if  stp.enable == False %}
```
jbemmel: ‘’’stp.enable’’’ may not be defined, need to add ‘’’|default(True)’’’

DanPartelly: Done. PVRST default cost still an issue. Waiting for ipspace to weigh in. MST cost by default is long, so I guess we should normalize to long in all

DanPartelly:
Default pvrstp / pvstp  to long pathcost method,

PR #2491, opened 2025-07-06
Enable STP support on Iosvl2 devices

DanPartelly:
Enable STP support on Iosvl2 devices, and update platform support in stp module documentation.

Passes all current integration tests, as expected.


ipspace (approved):
As expected, the integration tests pass.

PR #2492, opened 2025-07-06
Fix: ensure bgp.neighbors is defined for BGP-enabled routing instances

### LINKED ISSUE #2486: [BUG] Disabling BGP in a VRF crashes ebgp.multihop configuration templates
Disabling BGP in a VRF results in a vdata.bgp data structure that has no neighbors (it only specifies the redistribution of connected routes into BGP). That mightily upsets **ebgp.multihop** configuration templates that expect to see **bgp.neighbors** whenever **bgp** is mentioned.

### Lab topology

```
module: [ bgp, vrf ]
plugin: [ ebgp.multihop ]

vrfs:
  tenant:
    bgp: False
    links: [ dut-h1 ]

bgp.multihop.sessions:
- dut:
  pe2:

nodes:
  dut:
    bgp.as: 65000
  pe2:
    bgp.as: 65101
  h1:
    device: linux

links:
- dut-pe2
```

### Proposed solution

As the final step in BGP module processing, make sure that the **bgp** data structure in every routing instance (if present) has **bgp.neighbors** list which can be empty.

ipspace:
Numerous implementations of ebgp.multihop configuration template assume bgp.neighbors is defined if bgp is defined (in node data or in a VRF). We could either meet that assumption or fix all those templates (and meeting that assumption could result in simpler templates overall).

Fixes #2486

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -728,6 +728,14 @@ def vlan_ebgp_role_set(topology: Box, EBGP_ROLE: str) -> None:
     vdata.pop('_as_set',None)                               # ... and clean up
     vdata.pop('_bgp_attr',None)
 
+"""
+Make sure BGP data always has the expected attributes (currently: neighbors, might be others in the future)
+"""
+def sanitize_bgp_data(node: Box) -> None:
+  for (b_data,_,_) in _rp_utils.rp_data(node,'bgp'):
+    if 'neighbors' not in b_data:
+      b_data.neighbors = []
+
 class BGP(_Module):
   """
   Node pre-transform: set bgp.rr node attribute to _true_ if the node name is in the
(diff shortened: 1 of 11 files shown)


jbemmel:
If I set ```bgp``` to ```False``` in the red vrf in test vrf/12-multi-vrf-bgp, I get:
```
red:
    af:
      ipv4: true
    bgp:
      import:
        connected:
          auto: true
      neighbors: []
    export:
```

Ideally setting ```bgp: False``` would remove all bgp config from the vrf (including the empty neighbors list added here)

ipspace:
> Ideally setting `bgp: False` would remove all bgp config from the vrf (including the empty neighbors list added here)

That would break many (symmetric IRB) EVPN, MPLS/VPN and VRF route leaking scenarios.

PR #2497, opened 2025-07-07
Fix: no underscores in topology.name generated from input directory

### LINKED ISSUE #2424: [DOC] Another Gotcha - lab directory names
Directory names caveat:  if the lab directory name contains a hyphen like in "l2-fabric", this, at least under provider clab results in a host name being added to /etc/hosts which contains the substring "l2_fabric" -> note the underscore.

This is an invalid character as per DNS rfc's and causes gettadrinfo() (probably, didnt traced it ) to fail, effectively disabling ssh and other tools in netlab chain. Even ping hostname fails.

Maybe this should also be documented along with other ssh woes ?

ipspace:
Underscores in topology.name are used by clab to generate hostnames, and some Linux distros don't grasp that a clumsy person might have underscores in hostnames.

This fix ensures there are no weird characters in topology.name that is generated from the input directory (and 'default' is used when nothing is left after the cleanup). We don't check what the user specifies in the 'name' argument (they should know better ;)

Fixes #2424

### DIFF
--- a/netsim/augment/topology.py
+++ b/netsim/augment/topology.py
@@ -15,12 +15,33 @@ from .. import data
 from ..data.validate import validate_attributes,get_object_attributes
 from ..data.types import must_be_list,must_be_string,must_be_dict
 
+"""
+Generate topology name from the lab topology file location:
+
+* Take the first entry in the 'input' list (the original topology name)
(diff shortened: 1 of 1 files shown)


DanPartelly (approved):
Looks good to me. Thank you.

PR #2501, opened 2025-07-08
Fix: More robust suzieq tool definition

ipspace:
This fix uses a different name for suzieq container and parquet volume, allowing 'netlab up' to check whether the container has started (otherwise 'docker inspect' returns volume info when asked about the container).

It also creates container without the 'auto remove' flag, so the user can inspect container logs in case of a fatal suzieq failure.

The 'tool.down' command was modified accordingly to kill the container and then remove it.

### DIFF
--- a/netsim/tools/suzieq.yml
+++ b/netsim/tools/suzieq.yml
@@ -4,9 +4,9 @@
 runtime: docker     # Default: start in a Docker container
 docker:
   up:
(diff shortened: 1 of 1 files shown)


ipspace:
@ddutt -- adding you as a reviewer just in case you have time to look at it, but I'm guessing you have plenty of more important things to do, in which case I'll merge it in a day or two ;)

PR #2504, opened 2025-07-08
Allow setting "no switchport" on link agregations on cisco l2 images.

DanPartelly:
Allow setting port channels interfaces as "no switchport". Keep current logic for everything else.

### DIFF
--- a/netsim/ansible/templates/initial/ios.j2
+++ b/netsim/ansible/templates/initial/ios.j2
@@ -51,7 +51,7 @@ interface {{ mgmt.ifname|default('GigabitEthernet0/0') }}
 interface {{ l.ifname }}
 {% if l.type == 'vlan_member' and l.vlan.access_id is defined %}
  encapsulation dot1Q {{ l.vlan.access_id }}
-{% elif l.virtual_interface is not defined and netlab_device_type in ['iosvl2','ioll2'] %}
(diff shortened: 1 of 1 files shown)


DanPartelly:
It will work. I do not feed untested changes to you, or if it's a device major feature arc   without integration tests run.   Problem is, I'm used with low level strong(er) typed languages with no run-time reflection, so most of the time Ill forget of ...  is defined  ...  or run time type checking (so you guys will get un-tastefull things)  complemented by zome wrong assumptions about data sometimes.   Your tips on how to check it helped, but It will take a while to become habit. Im getting better at Python and Jinja , but I cant say I like either of them, yet. But you know what the chefs say: Palatability often comes with habit. 

> Looks good to me (and the initial configuration test works on IOSv, so the Jinja2 syntax is OK); the real test will come when you add the LAG code to IOS.

ipspace:
> It will work. I do not feed untested changes to you, or if it's a device major feature arc without integration tests run.

👍

> Problem is, I'm used with low level strong(er) typed languages with no run-time reflection

Yeah, I feel your pain. I started with Pascal, and the "everything goes" approach of JavaScript or Python is (trying to be extra-diplomatic) suboptimal. But hey, it is what it is...

PR #2507, opened 2025-07-09
Fix: Handle the messed-up status of packages on Ubuntu derivatives

ipspace:
Pop!_OS supposedly has the 'linux-modules-extra' package, but rejects to install it with a weird error messsage that doesn't even produce a proper exit code. This patch uses a slightly different approach to figuring out if the package is there to be installed.

Tested on Pop!_OS 22.04, Ubuntu 24.04 and Debian Bookworm

### DIFF
--- a/netsim/install/ubuntu.sh
+++ b/netsim/install/ubuntu.sh
@@ -11,14 +11,6 @@ $SUDO apt-get upgrade -y $FLAG_APT
 # Install missing packages
 #
 echo "Install missing packages (also a pretty long operation)"
-# Add linux-modules-extra for e.g. vrf
-# linux-modules-extra-$(uname -r) should exist, but it won't if we're running
-# on a kernel that has been purged from the package server; in that case the
-# user will likely need to reboot.
-EXTRA=""
-if apt-cache show linux-modules-extra-$(uname -r) > /dev/null; then
-  EXTRA="linux-modules-extra-$(uname -r)"
-fi
 # Added curl, which is not installed by default on Debian - ghostinthenet - 20220417
 $SUDO apt-get -y $FLAG_APT install python3 python3-setuptools ifupdown python3-pip curl $EXTRA
 echo "Install nice-to-have packages"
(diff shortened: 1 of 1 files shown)


ipspace:
Hey @zalekz, this should fix the last remaining problem in #2441. Could you please verify it? Thank you!

ssasso (approved):

zalekz:
> Hey @zalekz, this should fix the last remaining problem in #2441. Could you please verify it? Thank you!

@ipspace : Sorry for being late, I see that the changes have already been merged, but yes, I can also confirm that everything now installs on Pop!_OS 22.04 flawlessly using the updated scripts from the current `dev` branch, those with `apt-cache search` instead of `apt-cache show`. After the installation the user is suggested to reboot the system though, so all the necessary services are started automatically.. The `clab` and `libvirt` tests are successful then after a system restart. Many thanks!

netsim/install/ubuntu.sh
```diff
@@ -11,14 +11,6 @@ $SUDO apt-get upgrade -y $FLAG_APT
...
-# Add linux-modules-extra for e.g. vrf
-# linux-modules-extra-$(uname -r) should exist, but it won't if we're running
-# on a kernel that has been purged from the package server; in that case the
-# user will likely need to reboot.
-EXTRA=""
-if apt-cache show linux-modules-extra-$(uname -r) > /dev/null; then
-  EXTRA="linux-modules-extra-$(uname -r)"
-fi
 # Added curl, which is not installed by default on Debian - ghostinthenet - 20220417
 $SUDO apt-get -y $FLAG_APT install python3 python3-setuptools ifupdown python3-pip curl $EXTRA
```
zalekz, line 15: > `$SUDO apt-get -y $FLAG_APT install python3 python3-setuptools ifupdown python3-pip curl $EXTRA`

Perhaps, there is no need in that $EXTRA variable on _this_ line anymore

ipspace: Yeah, that one should go out (the next time we're changing this script ;)

ipspace:
> I can also confirm that everything now installs on Pop!_OS 22.04 flawlessly using the updated scripts from the current `dev` branch, those with `apt-cache search` instead of `apt-cache show`.

Thanks a million for the confirmation!

PR #2512, opened 2025-07-09
VyOS: support for LAG

ssasso:


ssasso:
IMHO you can merge as is

PR #2520, opened 2025-07-09
BGP confederations on Dell OS10, EOS, FRR, and IOS

ipspace:
Based on work done by @jbemmel in #2426

### DIFF
--- a/tests/topology/expected/6pe.yml
+++ b/tests/topology/expected/6pe.yml
@@ -135,6 +135,9 @@ nodes:
       advertise_loopback: true
       as: 65001
       community:
+        confed_ebgp:
+        - standard
+        - extended
         ebgp:
         - standard
         ibgp:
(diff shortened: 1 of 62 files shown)


ipspace:
IOS configuration template is broken, fix coming tomorrow... All BGP integration test pass for EOS, FRR, OS10

tests/topology/expected/6pe.yml
```diff
@@ -135,6 +135,9 @@ nodes:
       advertise_loopback: true
       as: 65001
       community:
+        confed_ebgp:
```
jbemmel: Could limit the community types to those being used in the topology, and avoid all these changes

jbemmel (approved):
I think this makes more sense than a separate plugin

Minor tweak possible to limit changes to YAML files for stuff not being used anyway - but it will work either way

tests/topology/expected/6pe.yml
ipspace: In theory, someone might want to limit the spread of extended communities to member autonomous systems. Also, the changes to test results don't hurt if we know why they happen.

ipspace: ... however, it's trivial to enable "confed_ebgp" value only when the topology uses confederations, so why not 😎

Issue #2398, opened 2025-06-17
OSPF: Disable per AF within VRF


jbemmel:
https://github.com/ipspace/netlab/blob/f7a4fb9ea0755381a20f561e5333a565977835c2/netsim/modules/ospf.yml#L56

I can disable OSPF for a node per AF, but not within a VRF

Use case is the bgp.multihop VRF test case which requires OSPFv3 within a VRF, which Dell OS10 does not support

Suggestion:
```
vrf:
    af:
      _list_to_dict: True
      _alt_types: [ NoneType ]
      ipv4: bool
      ipv6: bool
```

ipspace:
> Use case is the bgp.multihop VRF test case which requires OSPFv3 within a VRF, which Dell OS10 does not support

I'm sorry that you have to deal with a suboptimal operating system, but I see no reason to add yet another nerd knob just so an operating system with a marginal market share passes another test.

You could work on modifying the `test.vrf_check` plugin to consider the routing protocols, but (A) I'm not sure it's worth the effort and (B) there are other half-baked/half-abandoned things that might be worth solving before this one, like #1802 or #1909

ipspace:
The original use case is still bogus, but one could have a scenario where one would run OSPFv2 for IPv4 but then add IS-IS for IPv6 (preparing for a migration to a new routing protocol), so I decided to get it done ;)

PR #2524, opened 2025-07-10
Fix: Activate VRF IGP interface only when there are viable neighbors

### LINKED ISSUE #2275: [BUG] OSPF is activated in VXLAN-backed VRF with no OSPF neighbors
When a device has no OSPF-capable neighbors in a VRF, it should not run OSPF in it. It looks like the test is incomplete -- in a VXLAN-backed VRF, the neighbor seems to be OSPF-capable even though it has no IP address on the VRF interface.

### Lab topology

It seems you can't make it much smaller -- we need VXLAN to keep OSPF running

```
---
defaults.device: frr
provider: clab

groups:
  _auto_create: True
  hosts:
    members: [ h1, h2 ]
    device: linux

vlans:
  red:
    vrf: tenant
    links: [ s1-h1 ]
  blue:
    vrf: tenant
    links: [ s1-h2 ]

vrfs:
  tenant:
    loopback: True

nodes:
  s1:
    module: [ vlan, vxlan, ospf ]
    vlan.mode: bridge
  ros:
    module: [ vlan, vxlan, ospf, vrf ]
    vlans:                                      # Bring VLANs to ROS (until we agree on a better solution)
      red:
      blue:

links:
- s1:
  ros:
  mtu: 1600
```

ipspace:
The original VRF IGP code checked whether the IGP should be active in the VRF based on whether adjacent nodes were running the same IGP, but not whether they could form an IGP adjacency with the node.

This fix is not perfect, but better than the original -- when checking whether a neighbor is a viable IGP neighbor, it checks not just the IGP module (for example, are both running OSPF), but also whether the two devices have any AF in common on the shared link. That results in IGP not being run between a router connected to a switch with L2-only VLAN (but might still result in IGP being active if the switch has an IP address but does not run IGP in that VRF).

The relevant transformation test was modified to:

* Run OSPF in the underlay
* Put VLANs into a VRF on the router-on-a-stick (to trigger the VRF interface migration code)
* Have a single OSPF router in the overlay (the original test had two routers, so we'd always get OSPF adjacencies).

Fixes #2275 (mostly, without going into too many edge cases)

### DIFF
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -321,7 +321,9 @@ def build_vrf_interface_list(
     for neighbor in l.neighbors:                                      # ... iterate over the list of neighbors
       n_data = topology.nodes[neighbor.node]
       if proto in n_data.get('module',[]):                            # ... and check if at least one of them uses the IGP
-        node.vrfs[l.vrf][proto].active = True
+        for af in ['ipv4','ipv6']:                                    # ... and has at least one AF in common
+          if af in l and af in neighbor:                              # ... with our interface
+            node.vrfs[l.vrf][proto].active = True                     # Found one? Let's keep the routing protocol active
 
   # Time to cleanup IGP data
   for vname,vdata in node.get('vrfs',{}).items():                     # ... iterate over the list of VRFs
--- a/tests/topology/expected/vxlan-router-stick.yml
+++ b/tests/topology/expected/vxlan-router-stick.yml
@@ -2,10 +2,11 @@ groups:
   router:
     members:
     - r1
-    - r2
     module:
     - vlan
     - vxlan
+    - ospf
+    - vrf
     node_data:
       vlans:
         blue: {}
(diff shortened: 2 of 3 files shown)


netsim/modules/_routing.py
```diff
@@ -321,7 +321,9 @@ def build_vrf_interface_list(
     for neighbor in l.neighbors:                                      # ... iterate over the list of neighbors
       n_data = topology.nodes[neighbor.node]
       if proto in n_data.get('module',[]):                            # ... and check if at least one of them uses the IGP
-        node.vrfs[l.vrf][proto].active = True
+        for af in ['ipv4','ipv6']:                                    # ... and has at least one AF in common
+          if af in l and af in neighbor:                              # ... with our interface
+            node.vrfs[l.vrf][proto].active = True                     # Found one? Let's keep the routing protocol active
```
jbemmel, line 326: Should we track this by address family? OSPFv2 and OSPFv3 are not the same thing...

jbemmel: Can break the outer for loop once the flag has been set

ipspace: > Should we track this by address family? OSPFv2 and OSPFv3 are not the same thing...

The address families are tracked elsewhere (based on what is configured on global/VRF interfaces). Here, we're just checking whether there's an AF match between current device and its neighbors.

However, the "for" loop could be rewritten as "for af in ...[proto].af" if the **af** dictionary is already set at that moment. Thanks for the suggestion: have to check...

ipspace: Update: the VRF **proto.af** data is not set, so it cannot be used here. We might fix this bit if we ever implement #2398.

PR #2525, opened 2025-07-10
Iosl2 lag implementation.

DanPartelly:
Enabled on ioll2 and iosvl2.

Passes integration tests lag  [01-03-*]. Test 04-lag-vlan-routed-trunk.yml crashes ioll2 image, and !!resets!! the running config of iosvl2.

### DIFF
--- a/netsim/devices/iosvl2.py
+++ b/netsim/devices/iosvl2.py
@@ -41,13 +41,7 @@ IOSv (classic) layer-2 image treats Port-Channel interfaces as physical interfac
 def lag_remove_virtual(node: Box, topology: Box) -> None:
   for intf in node.interfaces:
     if intf.get('type') == 'lag' and 'virtual_interface' in intf:
-        del intf['virtual_interface']
-        report_quirk(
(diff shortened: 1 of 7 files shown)


netsim/devices/iosvl2.py
```diff
@@ -35,12 +35,27 @@ def vlan_1_tagged(node: Box, topology: Box) -> None:
...
         category=Warning)
 
+'''
+IOSv (classic) layer-2 image treats Port-Channel interfaces as physical interfaces
+'''
+def lag_remove_virtual(node: Box, topology: Box) -> None:
+  for intf in node.interfaces:
+    if intf.get('type') == 'lag' and 'virtual_interface' in intf:
+        del intf['virtual_interface']
+        report_quirk(
```
ipspace: No need to report this -- it's an internal flag, and the user is not impacted.

ipspace (approved):
Integration tests pass (the L3 LAG just needed a bit more time to get ping to work). Thank you!

PR #2526, opened 2025-07-10
Aruba CX: Fix 2523 (soft out not working)

### LINKED ISSUE #2523: [BUG] Using hard-reset of BGP sessions in ArubaCX bgp.policy template?
For whatever stupid reason (blog post coming in 3...2...1...😜) ArubaCX takes 12 seconds to realize an outgoing route map should change MED (`bgp.policy/31-med.yml`).

Even worse, **clear bgp \* soft out** does absolutely nothing. You turn on BGP update debugging on the other end, execute the command on ArubaCX, nada. Executing "clear bgp \* soft in" is equally "fruitful".

Should we replace that with **clear bgp \*** to hard-reset the BGP sessions? /cc @ssasso

ssasso:
Fix #2523

(fall back to full session reset)

### DIFF
--- a/netsim/extra/bgp.policy/arubacx.j2
+++ b/netsim/extra/bgp.policy/arubacx.j2
@@ -47,9 +47,7 @@ router bgp {{ bgp.as }}
 {% endfor %}
 {% endif %}
 !
-do clear bgp all * soft out
-do clear bgp all * soft in
+do clear bgp *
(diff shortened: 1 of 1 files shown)


ipspace (approved):
The MED integration test passes with flying colors now; I'm pretty positive all others will too. Thank you!

PR #2527, opened 2025-07-11
Documentation: spell-checking and a note of using long method in STP calculations

DanPartelly:
- spellchecking errors
- mention that we use long stp cost method for ioll2 / iosvl2


ipspace:
FWIW:

* It's nice if the commit header mentions usable stuff like the platform and technology it refers to. Makes writing release notes a bit easer
* Also, the commit header should not be longer than 72 characters (or is it 75?) or Github splits it into two lines when doing merge.

See also: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664

DanPartelly:
@ipspace  Ok, Ill keep in mind for the future

Issue #2528, opened 2025-07-11
Lags and bridges


DanPartelly:
Should the topo below transform into into a L2 lagg or an L3 lagg and a single switched link  ?

```
module: [lag]

nodes:
  s1:
    role: bridge
  s2:
    role: bridge  

links:
  - lag.members:
    - s1:
      s2:
    - s1:
      s2:
  - s1:
    s2:    
```

jbemmel:
L2 lag and a parallel L2 link

jbemmel:
```
af:
  ipv4: true
box: ceos:4.33.0F
clab:
  env:
    CLAB_MGMT_VRF: management
    INTFTYPE: et
  kind: ceos
hostname: clab-tmp-s1
interfaces:
- ifindex: 30000
  ifname: port-channel1
  ipv4: 10.1.0.1/30
  lag:
    ifindex: 1
    lacp: fast
    lacp_mode: active
  name: s1 -> s2
  neighbors:
  - ifname: port-channel1
    ipv4: 10.1.0.2/30
    node: s2
  type: lag
  virtual_interface: true
```

It allocates an IP address - that looks like a bug

DanPartelly:
I too get a L3 lag on IOS, and a parallel L2 link. Well, we should fix them after this release. I guess more systems are bugged.

jbemmel:
> I too get a L3 lag on IOS, and a parallel L2 link. Well, we should fix them after this release. I guess more systems are bugged.

It's not specific to any device, it's in the data model (lag module most likely)

ipspace:
As the **bridge** role was created to implement simple stuff, like replacing a Linux bridge with a device the user might be more familiar with, I'd say that LAG falls way outside of the "simple stuff" box. It's not like there would be non alternative; people who want to use LAG should use it with the **vlan** module.

I would suggest the **lag** module refuses to work with the devices with **role** set to **bridge**.

PR #2530, opened 2025-07-11
bridge role: Postpone processing until lag module has created its extra links

jbemmel:
Fixes: #2528

```pre_transform``` is too early, the bridge role must wait until after ```module_pre_transform``` in the lag module has created its virtual links (if any)


jbemmel:
@ipspace I'll leave this in draft for now - it needs more work on some corner cases, I know patching things all over the place like I did is not the correct way, but I get lost in the subtle interactions between transformation hooks

Postponing the bridge logic by using a different callback works, but then some assumptions change about what's already been done, and it becomes messy

jbemmel:
Replaced by #2531

PR #2531, opened 2025-07-12
Lag: Refuse to work with bridge devices, things like IP allocation don't behave as expected

jbemmel:
Addresses #2528


ipspace (approved):
👍

PR #2536, opened 2025-07-12
Use shared warnings and wait times in ripv2 and routing tests

ipspace:


tests/integration/wait_times.yml
```diff
@@ -4,6 +4,7 @@ const.validate:
   ping: 5
   ping_long: 15
   vxlan_ping: 20
+  ping_xx_long: 40        # Needed in some VRF scenarios due to crazily slow control plane
```
jbemmel, line 7: Not referenced in this PR?

jbemmel (approved):
Assuming the extra long ping will be used elsewhere

tests/integration/wait_times.yml
ipspace: > Not referenced in this PR?

Nope, needed for the next PR (but I was doing multiple changes in parallel with running integration tests)

Issue #2534, opened 2025-07-12
Create a "links" graph that strictly reflects the topology links


ipspace:
The current "topology" graph does not exactly reflect the physical links in VLAN environments. It might be useful to have another graph that closely mirrors links (it's produced from the "links" list similarly to how we build the containerlab file)

See https://github.com/ipspace/netlab-examples/discussions/28#discussioncomment-13736038 for details

ipspace:
The problem was the incorrect usage of "lan" network type instead of checking the number of devices attached to a segment. Implemented in #2547 and #2557

Issue #2535, opened 2025-07-12
Follow the wisdom of RFC 1925 rule 6/6a when creating graphs


ipspace:
Current implementations of GraphViz and D2 graphs were brute-hacked together to have something that can produce graphs. However, as all graphs are just vertices, edges, and labels, it might make sense to have a two-step process:

* Generate graph data structure (vertices/edges/labels) from topology data (shared between the two output modules)
* Generate graph description (GraphViz or D2) from graph data structure (output-format-specific)

PR #2547, opened 2025-07-22
Refactor D2 graph generation

ipspace:
Implementing ideas from #2535, add code that creates generic graph data structure for topology and BGP graphs, and use that data structure to generate D2 graphs with generic primitives (d2_cluster, d2_node, and d2_edge)

### DIFF
--- a/netsim/outputs/_graph.py
+++ b/netsim/outputs/_graph.py
@@ -57,7 +57,6 @@ def add_groups(maps: Box, graph_groups: list, topology: Box) -> None:
 
 def graph_clusters(graph: Box, topology: Box, settings: Box) -> None:
   if 'groups' in settings:
-    print(settings.groups)
     must_be_list(
       parent=settings,
       key='groups',path='defaults.outputs.graph',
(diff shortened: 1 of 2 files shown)


ipspace:
This needs tons of testing (I probably broke a few things that worked in the past).

It will close #2534 and #2535 for D2.

DanPartelly:
@ipspace Tested it. the graph topology looks cleaner, but now we are missing IP prefixes under node names, and device type in r1[device] .

Maybe sometimes in the future we could use dashed lines -- or a similar stye for tunnel links to make them apart ? Up to you, ofc.

ipspace:
> @ipspace Tested it. the graph topology looks cleaner, but now we are missing IP prefixes under node names, and device type in r1[device] .

Thank you, will fix.

> Maybe sometimes in the future we could use dashed lines -- or a similar stye for tunnel links to make them apart ? Up to you, ofc.

Sure. I just have to apply the BGP session type logic to link types.

ipspace:
This should be equivalent to the old functionality, but refactored to use the shared "create graph" functions. Once this is merged, I'll redo the GraphViz graphs, and then start adding new functionality (like the tunnel links)

ipspace:
> This looks very nice, and it's getting to a stage where you can drop it in a lab as topology illustration and be done with it.

BTW, how do you get nice lines in your diagrams? My D2 diagrams have weird splines line this:

image

ipspace:
> An observation: if you add a stub LAN to one of the devices in AS 6500, it will be rendered outside the AS. Maybe we should find another way to represent router stubs which have no device attached. See png below.

To make this work, we'd have to trace the real extent of an AS (explore the connectivity graph until you hit links with **role** set to **external**... assuming the user hasn't changed that role). Not sure I want to go there (yet)

ipspace:
> BTW, how do you get nice lines in your diagrams? My D2 diagrams have weird splines line this:

Found it. Have to use elk layout

PR #2550, opened 2025-07-25
Add 'debug at startup' capability for IOS and FRR

ipspace:
Adds device-specific attributes (ios.debug, frr.debug) that enable debugging at the beginning of initial device configuration.

The debugging attribute can be a list of debugging options or a multi-line string.

Replaces #2544

### DIFF
--- a/netsim/devices/frr.yml
+++ b/netsim/devices/frr.yml
@@ -152,3 +152,8 @@ features:
     vtep6: true
 
 graphite.icon: router
+
+attributes:
+  node:
+    frr:
+      debug: { type: list, split_lines: True }
--- a/netsim/devices/frr.yml
+++ b/netsim/devices/frr.yml
@@ -154,6 +154,9 @@ features:
 graphite.icon: router
 
 attributes:
+  global:
+    frr:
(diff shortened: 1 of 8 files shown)


netsim/devices/frr.yml
```diff
@@ -152,3 +152,8 @@ features:
     vtep6: true
 
 graphite.icon: router
+
+attributes:
+  node:
+    frr:
+      debug: { type: list, split_lines: True }
```
jbemmel, line 162: Could consider to make this a global attribute as well, copied under node

That way, users could define debug flags once and they would apply to all IOS nodes in the lab

jbemmel (approved):
One suggestion for enhancing the attribute to allow global definition

netsim/devices/frr.yml
ipspace: Nice one. Thank you, will add.

DanPartelly:
@ipspace  Managed to test it, right after you merged it. I love it. Works great, and works with groups of devices. This will see some use from me. Thank you.

PR #2553, opened 2025-07-26
Add baremetal emulation support for libvirt devices

exitnodeio:
I've been working on adding this feature for my own use in labbing out more complex bare metal topologies.

I've built out enough of this feature that it's useful to me in accomplishing what I set out to do.  I don't think the implementation is perfect yet, but if there is interest in merging this feature upstream I'd be happy to implement any feedback, improve test coverage, and add documentation.

Some things to note that I wasn't entirely happy with:

- Including the libvirt `domain` in the device topology
  - clab devices have a similar value, so this feels like it might be the right place to put it but maybe a more descriptive name is in order
- Extending validation for integration testing
  - The vbmc validation plugin probably could have been organized differently, I initially thought the vbmc.py file in `/netsim/validate` would have matched the search path for validation plugins.  I just included the file in linux.py to work around that for the time being.

Ultimately I think this would be an interesting extension of netlab, there are emulation tools/methods specific to testing various bare metal orchestration platforms but they are all baked into the testing process of their respective projects.  VirtualBMC, used in this PR, is maintained by the openstack project but is agnostic in it's implementation.  It provides an IPMI interface which extends power control and boot device selection from the libvirt domain.


ipspace:
Thank you! Really interesting ;)

Initial suggestions:

* This is a rarely-used functionality, so it should be a plugin, not a module. There's not much (functional) difference between the two, and as you're only using module_post_transform hook, it should be trivial to migrate the module to a plugin.
* I would use 'libvirt.domain' node attribute. I'm not saying we're always succeeding, but we try to put new attributes into the relevant sub-dictionary, so we know why they're there
* Modifying all test results to add an attribute that is rarely used is a bit of a bad form ;) I would set 'libvirt.domain' parameter in the vbmc plugin for nodes that are vBMC clients.

Finally, the functionality has to be documented ;)

exitnodeio:
Thank you for the input, a plugin was actually my initial intention but I think I was just a little unfamiliar with the project when I started working on this.  Taking a look now it's a little more clear to me how that would work, and I agree that this is a better fit overall as a plugin.  I'll work on making those changes and revise this PR.

PR #2555, opened 2025-07-27
IS-IS: Enable setting circuit type at interface level

DanPartelly:
IS-IS Enable setting circuit type at interface level.
Implemented for ios, eos, frr.

### DIFF
--- a/docs/module/isis.md
+++ b/docs/module/isis.md
@@ -34,12 +34,12 @@ The following table describes per-platform support of individual IS-IS features:
 | Cisco ASAv         | ✅  |  ❌  | ✅  | ✅  |
 | Cisco IOS/XE[^18v] | ✅  | ✅  | ✅  | ✅  |
 | Cisco IOS XRv      | ✅  |  ❌  | ✅  | ✅  |
-| Cisco Nexus OS     | ✅  |  ❌  | ✅  | ✅  |
+| Cisco Nexus OS     | ✅  | ✅  | ✅  | ✅  |
 | FRR                | ✅  | ✅  | ✅  | ✅  |
-| Junos[^Junos]      | ✅  |  ❌  | ✅  | ✅  |
-| Nokia SR Linux     | ✅  |  ❌  | ✅  | ✅  |
-| Nokia SR OS        | ✅  |  ❌  | ✅  | ✅  |
-| VyOS               | ✅  |  ❌  | ✅  | ✅  |
+| Junos[^Junos]      | ✅  | ✅  | ✅  | ✅  |
+| Nokia SR Linux     | ✅  | ✅  | ✅  | ✅  |
+| Nokia SR OS        | ✅  | ✅  | ✅  | ✅  |
+| VyOS               | ✅  | ✅  | ✅  | ✅  |
 
 These platforms support additional IS-IS features:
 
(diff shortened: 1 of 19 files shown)


ipspace:
Thank you. Next steps:

* Fix the "platform support" table in the documentation (it needs to be split into two tables, and that's always a mess)
* We'd also need an integration test (with a "warning" result if the functionality does not work)
* It might be nice to have a feature check saying "hey, this is not implemented on this device". It could be either a warning or a hard error.

I can do all of the above once I get other things I'm working on cleaned up.

DanPartelly:
Hopefully, I can do all that before leaving for summer holidays in a week. 
 
>     * Fix the "platform support" table in the documentation (it needs to be split into two tables, and that's always a mess)

  [You mean the table here ?](https://netlab.tools/module/isis/) . Its already split, if that is the case. How about we just move the last column "VRF instances" in the second table before BFD, and in first table we add "Circuit type" right after "IS Type" ?  


>     * We'd also need an integration test (with a "warning" result if the functionality does not work)

```
           | L1     | L2     | L1-L2  |
---------|--------|--------|--------|
L1       | L1 ADJ | NO ADJ | L1 ADJ |
L2       | NO ADJ | L2 ADJ | L2 ADJ |
L1-L2    | L1 ADJ | L2 ADJ | BOTH   |

```

Need input. Think  we could play some VRF tricks co we can test all combinations at once with a single lab fired up ? (Using an arista to check against, and check adjacencies  formation in json output of show isis neighbors ?) ?

>     * It might be nice to have a feature check saying "hey, this is not implemented on this device". It could be either a warning or a hard error.

I prefer hard errors when functionality is not implemented.  However, if the bulk of netlab warns and merrily goes around, then a warning is mandate for consistency. 

Do you want a feature flag for this ? Something like:
```

features:
  circuit_type: 
```

ipspace:
This should be good-to-go, but it would be nice of you could check my changes.

ipspace:
> [You mean the table here ?](https://netlab.tools/module/isis/) . Its already split, if that is the case.

Not really, because the second table is specifically about BFD support

> Need input. Think we could play some VRF tricks co we can test all combinations at once with a single lab fired up?

I have tests where I conditionally disable VRFs (based on what the devices support), for IS-IS, I decided to add an extra test:

https://github.com/ipspace/netlab/tree/dev/tests/integration/isis

Anyway, it was pretty easy to add extra checks to the dual-stack test (now that all devices we use support multi-topology IS-IS, it does not matter where the extra check is).

> I prefer hard errors when functionality is not implemented.

Implemented a hard error that can be disabled with a defaults flag.

> However, if the bulk of netlab warns and merrily goes around, then a warning is mandate for consistency.

Nope, we use warnings sparingly ;)

jbemmel:
See https://github.com/ipspace/netlab/blob/e710d27926486d1ab52e24ef0c3abb804b599062/netsim/ansible/templates/isis/srlinux.macro.j2#L10

SR Linux template does support circuit type too

DanPartelly:
> This should be good-to-go, but it would be nice of you could check my changes.

This looks great.  Of course I check all those changes, they are in my own benefit also, always learning something!  Thank you.

ipspace (approved):
The integration tests pass. We're good to merge ;)

ipspace:
@jbemmel: 

> See https://github.com/ipspace/netlab/blob/e710d27926486d1ab52e24ef0c3abb804b599062/netsim/ansible/templates/isis/srlinux.macro.j2#L10

That sets the IS-type

> SR Linux template does support circuit type too

No, it did not (now it does ;), but admittedly it's not your fault that the SR Linux developers decided to call the "network type" parameter "circuit-type" whereas everyone else in the industry uses the same term for something else. You could have checked which variable is used to change it though ;)

DanPartelly:
Thanks for all the work you did in this one.

PR #2557, opened 2025-07-27
Refactor Graphviz graph generation to use shared graph functions

ipspace:
* Implements ideas from #2535 and #2547 for Graphviz graphs
* Adds generic 'object style' mechanism that replaces old 'colors' and 'margins' settings
* Corrects #2534 for Graphviz


DanPartelly (approved):
This looks good to me. Tried it with several topologies, and the results are identical to d2 graphs (save for final layout, but that's dependent on  layout  engines).  Thank  you.

PR #2560, opened 2025-07-29
SRLinux: Take global isis.type into account

jbemmel:
In ```integration/isis/12-passive.yml``` the global isis.level is defined as "level-1"

Without this fix, the SR Linux template enables level-2 on all interfaces


ipspace:
> Without this fix, the SR Linux template enables level-2 on all interfaces

Which is the default, but doesn't matter because IS type is set to level-1. A level-1 IS will never send level-2 ISH no matter what's configured on the interface(s)

PR #2563, opened 2025-07-31
New device: Nokia SR-SIM

ipspace:


docs/caveats.md
```diff
@@ -445,27 +445,52 @@ See [](generic-linux-devices)
...
-python3 -m pip install grpcio protobuf==3.20.1
-```
+Other caveats (shared with Nokia SR-SIM):
+
+* We use the `nokia.grpc` Ansible Galaxy collection to configure SR-OS. The latest version of that collection has to be installed from the git repo:
+
+  ```
+  ansible-galaxy collection install git+https://github.com/nokia/ansible-networking-collections.git#/grpc/
+  python3 -m pip install grpcio protobuf==3.20.1
+  ```
```
hellt, line 459: was it ever considered to switch from gnmi to netconf for SR OS? I thought it was once discussed?

jbemmel: See e.g. https://github.com/napalm-automation-community/napalm-sros

It would be possible, but with a significant effort for no functional gain. I initially made the choice to use gnmi, both to have consistency between SR OS and SR Linux and also because I knew that Netconf was the preferred, presumably better tested mechanism, so I figured focusing on the other one could have benefits in terms of potentially finding issues that might otherwise not be found.

People are welcome to contribute

ipspace: > was it ever considered to switch from gnmi to netconf for SR OS? I thought it was once discussed?

You once mentioned that gNMI was a bad choice, and I fully agree. "Potentially finding third-party issues that might otherwise not be found" is not necessarily a good investment of someone's time that could be better spent making the tool better, but obviously different people have different priorities 🤷‍♂️

However, unless someone figures out how to use the existing data structures with NETCONF, I don't think the migration is going to happen (anytime soon). I don't see anyone with the willpower to rewrite all SR-OS provisioning templates.

docs/module/vrf.md
sacckth, line 1: Route leaking SHOULD be supported on SR-OS for a long time, maybe this is just a missing netlab feature?

jbemmel: Correct, "supported" here means "supported by Netlab", not "supported by the device"

ipspace: @sacckth You're absolutely correct. This is a missing netlab feature; I added a footnote explaining what the problem is. It's also documented in https://github.com/ipspace/netlab/blob/d3a274f7be56be77d3be19b2a2d76c83f57a8169/docs/caveats.md?plain=1#L468

On a tangential thread, if you could point me to a document describing how to do local inter-VRF route leaking without MPLS/VPN or EVPN, I will add the functionality to _netlab_ (having a recipe to follow would be much faster than throwing random ideas at the device to see what sticks).

PR #2570, opened 2025-08-01
Fix: Make the ifname of host-side part of vEth pair unique

### LINKED ISSUE #2564: [BUG] Issue with long node names in hybrid provider scenarios
Lab:
```
---
groups:
  _auto_create: True
  nodes:
    members: [ verylongname1, verylongname2 ]
    device: frr
    provider: clab
  tor:
    members: [ r ]
    device: dellos10

links:
- r-verylongname1
- r-verylongname2
```

Result:
```
   ERROR 

  Root network namespace endpoint "verylongna_eth1" defined by multiple nodes [virbr1, virbr2].

Error executing sudo -E containerlab deploy --reconfigure -t clab-augment.yml:
```

The node name prefix needs to be unique in the first 10 characters. Possible solution: use a hex hash value of the node name, or the first 4 and last 4 characters

ipspace:
We have to connect the host-side part of a vEth pair to a Linux bridge in a multi-access or multi-provider environment. To do that, we have to have a unique host-side interface name, and the current mechanisms failed for people who couldn't possibly make their node names unique in the first 10 characters.

This fix replaces the 'use the node name prefix' trick that helped keep the device name under 16 characters (until containerlab started using long interface names for SR-SIM) with a totally cryptic interface name that shouldn't exceed 16 characters unless you're doing something truly crazy (for example, 8-digit multilab ID).

As we're not using the bridge-side interface names anywhere else in the code ('netlab capture' command is started within the container NS), this SHOULD NOT (BUT WHO KNOWS) affect anything else.

Fixes #2564

### DIFF
--- a/netsim/templates/provider/clab/clab.j2
+++ b/netsim/templates/provider/clab/clab.j2
@@ -114,12 +114,25 @@ topology:
 {%       set clab = nl.clab|default({}) %}
   - endpoints:
     - "{{ n.name }}:{{ clab.name|default(nl.ifname) }}"
-{# Use node_intf name for bridge veth interface in standard setup and more cryptic (but unique) name in multilab deployment #}
-{%       if defaults.multilab.id|default(False) %}
-    - "{{ l.bridge }}:{{ name[:6] }}_{{ n.id }}_{{ nl.ifindex }}"
-{%       else %}
-    - "{{ l.bridge }}:{{ n.name[:10] }}_{{ clab.name|default(nl.ifname) }}"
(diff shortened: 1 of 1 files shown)


jbemmel (approved):
Shortening the node names is the first thing I did, which "fixed" the issue for me locally. However, out of consideration for the fact that unsuspecting other Netlab users might encounter the same scenario, I decided to submit an issue to document the potential problem and alert the community.

I appreciate the author spending the time to fix this edge case

PR #2571, opened 2025-08-02
Add '--suffix' parameter to 'netlab collect'

ipspace:
The 'netlab collect' command usually saves configuration files with the '.cfg' suffix (which can then be used in 'netlab up --reload' process). However, when building a multi-platform solution, it might be helpful to save the configuration files with a platform-specific suffix, for example 'eos.j2'. The '--suffix' parameter allows you to replace the standard 'cfg' suffix with something else

As a few 'fetch-config' task lists used hardcoded '.cfg' suffix, most of them were either fixed to use the 'cfg_suffix' variable or (even better) set 'ansible_net_config' fact instead of creating the output file directly.

### DIFF
--- a/docs/netlab/collect.md
+++ b/docs/netlab/collect.md
@@ -1,21 +1,22 @@
 (netlab-collect)=
 # Collecting Device Configurations
 
-**netlab collect** command uses Ansible *device facts* modules (or an equivalent list of Ansible tasks) to collect device configurations and store them in specified output directory.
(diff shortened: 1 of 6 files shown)


ipspace:
I was sick-and-tired of renaming configuration files to platform-specific templates while writing the IS-IS labs, so I added another nerd knob to "netlab collect" and fixed a bit of technical debt while doing that. The FRR and SR Linux changes were tested; I hope I'm awake enough to do a find-and-replace on the CL tasklist ;)

Adding reviewers just in case someone wants to look at this or comment.

PR #2572, opened 2025-08-03
VyOS: Implement configuration templates using only 'set' commands

ipspace:
VyOS configuration templates are vbash scripts that include a lot of
setup and cleanup code. The same code is replicated in every template.
That is not good by itself, but it doesn't allow the users to use
simple custom config templates with just the 'set' commands.

This change implements configuration deployment similar to what
we're doing for FRR: if the configuration code includes 'vbash',
it's executed as-is, otherwise it's wrapped in a standard vbash
script.

Also changed:

* The 'are we ready' check has been moved from the 'deploy config'
  task list to 'readyness check' task list
* The 'fetch config' task list fetches device configuration as
  'set' commands, allowing it to be used directly as a custom config
  template. That should allow 'netlab up --reload' to work with
  VyOS.

### DIFF
--- a/netsim/ansible/tasks/deploy-config/vyos-script.j2
+++ b/netsim/ansible/tasks/deploy-config/vyos-script.j2
@@ -1,0 +1,19 @@
+{% if 'vbash' in script_content %}
+{{   script_content }}
+{% else %}
+#!/bin/vbash
+source /opt/vyatta/etc/functions/script-template
+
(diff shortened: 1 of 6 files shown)


ipspace:
@ssasso This is a follow-up work from #2571 that will be rebased before merging. The relevant changes are in the last commit.

ssasso (approved):
great work thanks!!!!

I will slowly adapt all the other templates

Issue #1817, opened 2025-01-15
Self-contained lab topologies


jbemmel:
https://github.com/ipspace/netlab/blob/2c10748b6b73519ca8fd2b77a90b15e61d1f4e3a/tests/integration/gateway/02-vrrp.yml#L46

It would be useful to have lab topologies that are fully self-contained in a single YAML file. Instead of forking off to a template with a couple of lines of config script, these snippets could be embedded within the single lab file

The above could become:
```
validate:
  r2_eth1_down:
    description: Remove R2 from the VRRP LAN
    nodes: [ r2 ]
    config:
      template:
        variable.ifstate: 'down'
        eos: |
          interface {{ interfaces[0].ifname }}
          {{ 'no ' if ifstate|default('down') == 'up' else '' }}shutdown
        frr: |
         #!/bin/bash
         ip link set {{ interfaces[0].ifname }} {{ ifstate|default('down') }}
    pass: R2 has been disconnected from the VRRP LAN
    stop_on_error: True
```

ipspace:
Good idea. Will try to figure out how to make it work both in validation and in generic custom configuration without too much hassle.

ipspace:
Implemented in #2573 and #2580

PR #2573, opened 2025-08-03
The 'files' plugin: embed files or custom config templates in topology

ipspace:
The 'files' plugin can be used to embed any file (in particular custom configuration templates) into the lab topology.

The following changes were made to other netlab components to support this functionality:

* 'netlab create' calls 'output' plugin hooks for plugins that register themselves as having on in the defaults.netlab.create.output list
* 'netlab down --cleanup' removes files specified in the _cleanup dictionary
* The 'load plugin' and 'execute plugin hook' code was moved to separate functions so it can be called directly from 'netlab create' (the Plugin list is gone by the time 'netlab create' creates output files)
* The 'fs_cleanup' function tries to remove directories when the files it removes contain a directory component.

Finally, the 'bgp/05-community' integration test uses this functionality as a proof-of-concept

Implements most of the functionality needed to support #1817

### DIFF
--- a/docs/plugins/files.md
+++ b/docs/plugins/files.md
@@ -1,0 +1,112 @@
+(plugin-files)=
+# Embed Files and Configuration Templates in Lab Topology
+
+The **files** plugin allows you to include small files and custom configuration templates directly in the lab topology, enabling the deployment of a complete solution as a single lab topology file.
+
+The plugin defines two new topology attributes:
+
+* **files** -- a dictionary or list of extra files to create in the lab directory
+* **configlets** -- a dictionary of custom configuration templates
+
+(plugin-files-create)=
+## Creating Extra Files
+
+The **files** attribute can be a list of file paths and their contents. Every element of the list has to have two attributes:
+
+* **path**: The file path relative to the lab directory
+* **content**: File content
+
+Example:
+
+```
+plugin: [ files ]
+
+files:
+- path: README.md
+  content: |
+    The README.md file
+```
+
+```{tip}
+Always use the YAML *literal ‌block scalar header* (`|`) for the file content; otherwise, the whole content will be folded into a single line.
+```
+
+The **files** attribute can also be specified as a dictionary of  file paths and content, for example:
+
+```
+plugin: [ files ]
+
+files:
+  hey/README.md: |
+    The README.md file
+```
+
+```{tip}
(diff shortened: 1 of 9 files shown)


docs/plugins/files.md
```diff
@@ -0,0 +1,112 @@
...
+Finally, you might want to create a variant of a configuration template for a specific provider, and a more generic one for all other providers. That cannot be easily expressed as a dictionary structure, so the **files** plugin treats **content** key as a special case, meaning *do not append anything to the file name*. The following example will thus create **ifup/eos.j2**  and **ifup/eos-clab.j2** files:
+
+```
+configlets:
+  mgmt:
+    eos:
+      clab: |
+        interface Management0
+          description Management interface
+      content: |
```
jbemmel: I asked Claude for alternative suggestions:
```
My top recommendations would be:

default - Most intuitive and widely understood
fallback - Very clear about the conditional behavior
base - Clean and suggests foundational template

The current content name is quite generic and doesn't clearly convey that it's special behavior. Any of these alternatives would make the configuration more self-documenting and easier for users to understand the template hierarchy.
```

docs/plugins/files.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+## Creating Custom Configuration Templates
+
+The **configlets** dictionary can be used to create custom configuration templates. Its contents are automatically converted to [elements of the **files** list](plugin-files-create), but provide a more intuitive way of specifying the templates.
+
+For example, the following **configlets** dictionary creates **ifup.j2** and **ifdown.j2** custom configuration templates in the lab directory:
+
+```
+configlets:
```
jbemmel, line 62: Could consider to call this ```files.configlets``` to make it more clear that this uses the 'files' plugin

The example at the end is not entirely intuitive, and it suggests that this 'configlets' case may be more commonly used than the plain 'files' case.

Alternatively, we could define 2 plugins: the 'files' plugin and the 'configlets' plugin (internally reusing much of the same code)

ipspace: That would clash with the 'files as a list' idea anyway, and **configlet** plugin just for winning the beauty points doesn't make too much sense ;)

docs/plugins/files.md
ipspace: 'base' is not too bad. Let's go with that. Thank you @jbemmel and Claude ;)

PR #2575, opened 2025-08-06
Set 'rr_client' and 'next_hop_self' attributes on BGP neighbor data

ipspace:
The BGP neighbor 'rr_client' and 'next_hop_self' attributes will make the configuration templates much cleaner as we'll avoid all the crazy logic that is repeated in every template.

Note that the existing attributes are not changed, so there's no rush to migrate the templates to the new attributes. That can be done at any time we feel like cleaning them up.

Also: refactor and cleanup the IBGP session generation logic to avoid duplicate code

### DIFF
--- a/netsim/modules/bgp.py
+++ b/netsim/modules/bgp.py
@@ -178,38 +178,32 @@ build_ibgp_sessions: create IBGP session data structure
 * Other nodes need IBGP sessions with all RRs in the same AS
 """
 def build_ibgp_sessions(node: Box, sessions: Box, topology: Box) -> None:
-  rrlist = find_bgp_rr(node.bgp.get("as"),topology)
-  has_ibgp     = False                            # Assume we have no IBGP sessions (yet)
+  node_as  = node.bgp['as']                       # Set up variables that will come handy, starting with node AS
+  is_rr    = node.bgp.get("rr",None)              # Is this node an RR?
+  bgp_nhs  = node.bgp.get("next_hop_self",None)   # Do we have to set next hop on IBGP sessions?
+  rrlist = [] if is_rr else find_bgp_rr(node_as,topology)
 
-  # If we don't have route reflectors, or if the current node is a route
-  # reflector, we need BGP sessions to all other nodes in the same AS
-  if not(rrlist) or node.bgp.get("rr",None):
-    for name,n in topology.nodes.items():
-      if "bgp" in n:
-        if n.bgp.get("as") == node.bgp.get("as") and n.name != node.name:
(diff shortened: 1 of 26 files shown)


netsim/modules/bgp.py
```diff
@@ -178,38 +178,32 @@ def get_remote_ibgp_endpoint(n: Box) -> Box:
...
+  node_as  = node.bgp['as']                       # Set up variables that will come handy, starting with node AS
+  is_rr    = node.bgp.get("rr",None)              # Is this node an RR?
+  bgp_nhs  = node.bgp.get("next_hop_self",None)   # Do we have to set next hop on IBGP sessions?
+  rrlist = [] if is_rr else find_bgp_rr(node_as,topology)
+
+  if is_rr or not rrlist:                         # If the current node is RR or we have a full mesh
+    ibgp_ngb_list = [                             # ... we need IBGP sessions to all nodes in the AS
+      ngb for ngb in topology.nodes.values()      # ... but while building the node list
+        if ngb.name != node.name and              # ... skip current node
+           ngb.get('bgp.as',None) == node_as]     # ... and everyone not in the current AS (or not running BGP)
```
jbemmel, line 190: Reading the code here it's not obvious how ```localas_ibgp``` neighbors are handled, but I'm assuming that happens elsewhere?

ipspace: localas_ibgp is really an EBGP session (the **bgp.as** values are different) that pretends to be IBGP. It's handled here: https://github.com/ipspace/netlab/blob/d7115a046f3d491290391c0662b5f4f0302e9f57/netsim/modules/bgp.py#L384

netsim/modules/bgp.py
```diff
@@ -178,38 +178,32 @@ def get_remote_ibgp_endpoint(n: Box) -> Box:
...
+    if not neighbor_data is None:
+      if 'loopback' in node:
+        neighbor_data._source_intf = node.loopback
+      if bgp_nhs:
+        neighbor_data.next_hop_self = True
+      if is_rr and not 'rr' in neighbor_data:
+        neighbor_data.rr_client = True
+      node.bgp.neighbors.append(neighbor_data)
+
+  if not ibgp_ngb_list:
```
jbemmel, line 206: This change leads to a warning about missing IGP even though the topology specifies ```bgp.sessions.ipv4: [ ebgp ]``` to exclude any IBGP sessions

The difference seems to be that the ```has_ibgp``` flag wouldn't get set in that case, but the new ```ibgp_ngb_list``` contains nodes even though the bgp.sessions parameter says those sessions shouldn't get built

PR #2576, opened 2025-08-06
vJunos-switch: support for explicit vlan-aware bundle

ssasso:
(using a quirk for unique RD per routing instance)


ssasso:
@ipspace 

While using "core" functions of the vrf module, I got mypy validation to fail with:

```
netsim/devices/vjunos-switch.py:42: error: Argument 1 to "get_next_vrf_id" has incompatible type "Optional[Any]"; expected "str"  [arg-type]
```

Which for now I solved forcing `str(xxx)`.
Please let me know if there's a cleaner approach.

ipspace:
> While using "core" functions of the vrf module, I got mypy validation to fail with:

mypy is still OK with using typing.Any as str, but is allergic to None values of typing.Optional. Rewrote the code to get the ASN first (because we need it for all bundles anyway), then check if we got something (we should, but mypy wants to be sure), then use it.

PR #2580, opened 2025-08-08
Use 'files' plugin to implement inline configuration templates

ipspace:
This change adds support for inline custom configuration templates in node-, group- and validation test definitions. It allows the user to specify small additions to device configurations directly in the node- or group definition, or to define small changes to device configurations directly in the validation tests.

### DIFF
--- a/docs/plugins/files.md
+++ b/docs/plugins/files.md
@@ -8,6 +8,15 @@ The plugin defines two new topology attributes:
 * **files** -- a dictionary or list of extra files to create in the lab directory
 * **configlets** -- a dictionary of custom configuration templates
 
+It also extends the meaning of the node/group **config** attribute and adds the **config.inline** attribute to validation tests.
+
+```eval_rst
+.. contents:: Table of Contents
+   :depth: 2
+   :local:
+   :backlinks: none
+```
+
 (plugin-files-create)=
 ## Creating Extra Files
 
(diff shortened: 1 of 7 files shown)


ipspace:
The final step of the configlets journey: specifying custom configurations directly in node/group/test definitions.

I will update the node/group/validate documentation once we're OK with attribute naming ;)

jbemmel (approved):
I still find it less than intuitive that, as a user, I need to include the 'files' plugin to get access to a 'config.inline' sub-element under validation

Perhaps loading the 'files' plugin by default, or otherwise making it part of standard functionality would be better.

The code is ok and the feature is great, it's just about helping users figure things out easily

jbemmel:
> I still find it less than intuitive that, as a user, I need to include the 'files' plugin to get access to a 'config.inline' sub-element under validation
> 
> Perhaps loading the 'files' plugin by default, or otherwise making it part of standard functionality would be better.
> 
> The code is ok and the feature is great, it's just about helping users figure things out easily

I just saw https://github.com/ipspace/netlab/pull/2579 so I know you were aware of this - my point is about finding a way to avoid having to go through this error in the first place

ipspace:
> The code is ok and the feature is great, it's just about helping users figure things out easily

I don't have a rock-solid reason, but as the "files and friends" functionality vastly expands the capabilities of overwriting something, I have a funny feeling that making this part of standard functionality would be giving out too much rope.

PR #2587, opened 2025-08-09
Fix: Recognize absolute/relative paths in 'netlab config'

### LINKED ISSUE #2582: [BUG] netlab config --reload command hates absolute paths
Execute **netlab config --reload /some/absolute/path** and watch the script complaining about a missing directory (even though it exists).

We might have the same problem in **netlab config** because it was designed to accept configuration template _names_ not generic directory paths, and who knows what happens in the Ansible playbook.

ipspace:
Fixes #2582


ipspace:
@PolinaOrehova -- would love to hear whether this works for you.

PR #2588, opened 2025-08-09
Use BGP neighbor attributes from #2575 in EOS, FRR, and IOS templates

ipspace:


ipspace:
@jbemmel @ssasso FYI in case you want to fix other templates (and yes, I will write the BGP developer documentation one of these days ;)

Issue #1996, opened 2025-03-04
Add nexthop-self test for localas_ibgp sessions


jbemmel:
https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp/08-ibgp-localas.yml could be enhanced to test for proper handling of nexthop-self (similar to https://github.com/ipspace/netlab/blob/2ab743438ded21a70856b7023fa440efab15f327/tests/integration/bgp/02-ibgp-ebgp-session.yml#L55)

ipspace:
Yeah, we can do that

ipspace:
You have no idea what a ginormous can of worms you opened :) Let's see if any of the devices passes the extended test...

jbemmel:
EOS needs a fix, let me know if you prefer a separate PR or adding it to this one:
```
{%       if 'ibgp' in n.type and bgp.next_hop_self|default(False) %}
  neighbor {{ peer }} route-map next-hop-self-{{ af }} out
{%       endif %}
```

None of FRR, Cumulus NVUE or EOS pass the 'IBGP routes being passed to IBGP local-as' test - perhaps that's how it should be?

PR #2591, opened 2025-08-09
A router with localas_ibgp session MUST change NH on all IBGP routes

ipspace:
The routes received over localas_ibgp session have to be reflected to other IBGP neighbors. The next hop on those reflected routes MUST be changed because we're not advertising the inter-AS subnet into IGP.

That makes the router with localas_ibgp session unsuitable to be a RR.

This change adds an extra step after the IBGP and EBGP sessions have been built. It checks for the presence of localas_ibgp session and sets neighbor.next_hop_self to 'all' on all IBGP sessions if needed.


ipspace:
The (hopefully) final step of a very long journey started by #1996

PR #2601, opened 2025-08-16
New report: IS-IS nodes, areas, system IDs and IS-types

ipspace:


ipspace:
@DanPartelly Just in case you want to try it out ;)

DanPartelly (approved):
Thank you for this new report. Tested it for all output formats on the topologies used on the latest IS-IS exercises, and everything is correct.

PR #2604, opened 2025-08-18
Fix: make maximum node_id configurable (default: 250)

### LINKED ISSUE #2603: max node_id
I have a lab with 3k lines.
Does this error come from netlab directly or might it come from a bad configured loopback adress range?

[ERRORS]  Errors found in lab.yml
[FATAL]   dataplane: Ran out of node_id values, next value would be greater than 250

ipspace:
Fixes #2603

### DIFF
--- a/docs/nodes.md
+++ b/docs/nodes.md
@@ -183,7 +183,7 @@ After the initial cleanup, the *netlab* [topology transformation code](dev/trans
 * Device interfaces created as needed during the link transformation phase and collected in the **interfaces** list.
 * [](modules.md) document describes further processing done on configuration module parameters.
 
-[^id]: Node **id** must be an integer between 1 and 250. When using the standard management interface IP addressing (where management IPv4 addresses start with .100), the node **id** should not exceed 150.
+[^id]: Node **id** must be an integer between 1 and MAX_NODE_ID (default: 250). You can change the maximum node ID with the **const.MAX_NODE_ID** [system default](topo-defaults). When using the standard management interface IP addressing (where management IPv4 addresses start with .100), the node **id** should not exceed 150.
 
 [^HOST]: Identified by **role: host** attribute
 
--- a/netsim/augment/nodes.py
+++ b/netsim/augment/nodes.py
@@ -20,7 +20,7 @@ from ..data.types import must_be_int,must_be_string,must_be_id,must_be_device
 from ..data import global_vars,is_true_int
 from ..modules._dataplane import extend_id_set,is_id_used,set_id_counter,get_next_id
 
-MAX_NODE_ID: typing.Final[int] = 250
+MAX_NODE_ID: int
 
 """
 Reserve a node ID, for example for gateway ID, return True if successful, False if duplicate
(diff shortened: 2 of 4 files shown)


jbemmel:
There's an issue here: https://github.com/ipspace/netlab/blob/0a30a7661c6717a915ee51b63a70153f80b12695/netsim/augment/nodes.py#L131

ipspace:
> There's an issue here:
> 
> https://github.com/ipspace/netlab/blob/0a30a7661c6717a915ee51b63a70153f80b12695/netsim/augment/nodes.py#L131

Thank you. So he'd have to define the management MAC addresses manually, or should we also modify the 3rd octet if needed? Also, we're currently squatting on OUI assigned to Cisco, so maybe we should change that to locally-administered range as well?

jbemmel:
> 
> Thank you. So he'd have to define the management MAC addresses manually, or should we also modify the 3rd octet if needed? Also, we're currently squatting on OUI assigned to Cisco, so maybe we should change that to locally-administered range as well?

Not completely solving the issue, but
```
addrs.mac_eui[3] = node.id // 256
addrs.mac_eui[4] = node.id % 256
```
would allow for 64K nodes.

We could adjust the OUI prefix as well (though orthogonal to this issue)

ipspace:
> ```
> addrs.mac_eui[3] = node.id // 256
> addrs.mac_eui[4] = node.id % 256
> ```

That would change the OUI, and I know you want to have plenty of MAC address space for individual nodes. Going to locally administered MAC addresses seems to be the only sane long-term solution, but after all the crap we've seen while working on this project, I'm afraid of what might break.

Should we, for the moment, just add a slight addition to that footnote saying "if you increase MAX_NODE_ID beyond 255, you'll have to deal with management MAC addresses as well"?

ssasso:
Side note: instead of "complex" math, you could use netaddr to handle mac addresses as well:

```
>>> base_mac = netaddr.EUI('02-00-00-00-00-00')
>>> id=1234
>>> node_mac = netaddr.EUI(int(base_mac)+id)
>>> node_mac
EUI('02-00-00-00-04-D2')
```

ipspace:
> Side note: instead of "complex" math, you could use netaddr to handle mac addresses as well:

That's what we had in the past. I'll leave it to @jbemmel to explain the sad history of this change 😒

jbemmel:
> > ```
> > addrs.mac_eui[3] = node.id // 256
> > addrs.mac_eui[4] = node.id % 256
> > ```
> 
> That would change the OUI, and I know you want to have plenty of MAC address space for individual nodes.
> 

Unless I’m mistaken, the OUI is in 0..2 (24 bits). Index 3 and 4 are available to change 

We could also change 0..2 - but that is a different issue 

A footnote would not help users deal with the macs - that’s a pain

ipspace:
> Unless I’m mistaken, the OUI is in 0..2 (24 bits). Index 3 and 4 are available to change

We were modifying byte #6. Then you wanted to modify byte #4. Now you claim #4 and #5 are OK. Could you please figure out what you really need (and why), otherwise we can go back to the old solution (along the lines of what @ssasso recommended) and call it a day.

Oh, and thanks for discussing things in a barely-related merged PR. That will make it so much easier to find them later.

PR #2610, opened 2025-08-19
Change the management MAC calculation, fix other large-node-id quirks

### LINKED ISSUE #2606: Deal with management MAC addresses for nodes with ID > 255
So here's what we're going to do with https://github.com/ipspace/netlab/issues/2604#issuecomment-3197093153 (unless someone presents a better solution supported with strong and reasonable arguments ;)

* We'll keep the existing logic for node ID < 255 to keep @jbemmel happy
* We'll add node ID to the base MAC address for nodes with ID > 255

There is guaranteed to be no overlap between "current" and "future" MAC addresses unless someone goes beyond 65536 nodes (good luck with that).

ipspace:
* The management MAC address is calculated using the old algorithm for node IDs below 255 (to keep Cumulus NVUE happy) and the "base + id" approach for node IDs above 255.
* The _dataplane.get_next_id function accepts a hint that is used by node ID allocation code to point the user to the system constant that needs to be changed.
* The change in const.MAX_NODE_ID requires a corresponding change in maximum value of node.id attribute. Solved by moving the node const initialization code into a separate function that is called very early in the augmentation process.

Finally, added a transformation test with a large node ID. We will add more features to it as we discover what else can go wrong.

Fixes #2606


Muddyblack:
Tested it with my big lab and worked nicely 👌
Also when it took now 35 Minutes to configure :D

netsim/augment/nodes.py
```diff
@@ -32,15 +32,22 @@ def reserve_id(n_id: int) -> bool:
...
 
+"""
+Fetch and adjust node constants
+"""
+def adjust_node_constants(topology: Box) -> None:
+  global MAX_NODE_ID                                        # Max node ID has to be initialized early in the transformation
+  MAX_NODE_ID = global_vars.get_const('MAX_NODE_ID',150)
+
+  if MAX_NODE_ID > 150:
+    topology.defaults.attributes.node.id.max_value = MAX_NODE_ID
```
jbemmel: In other cases we did it like this: https://github.com/ipspace/netlab/blob/0a30a7661c6717a915ee51b63a70153f80b12695/netsim/defaults/attributes.yml#L140

jbemmel (changes requested):
I would suggest to put ```MAX_NODE_ID``` as ```max_value``` in the attribute, rather than adjusting it if it's larger than 150.

That way, the YAML becomes readable independently, instead of getting tweaked behind the scenes

netsim/augment/nodes.py
ipspace: Thank you, will fix.

ipspace:
@jbemmel Thanks for the poke ;) In the end, it resulted in significant simplification and cleaner code ;))

PR #2613, opened 2025-08-20
JunOS routing: Fix #2612

### LINKED ISSUE #2612: [BUG] JunOS inconsistency in routing policy and prefix lists
JunOS *routing* templates is generating inconsistent prefix list names wrt usage in policy.

I think this has never been raised as error in integration tests because there the `af` top level host var always contains both ipv4 and ipv6.

Prefix lists are generated for all the topology address families (`af` var), 

https://github.com/ipspace/netlab/blob/9aac9ca4c43ac5c6fa59bf1724e6be0e9053f8d9/netsim/ansible/templates/routing/_prefix_list_junos.j2#L6

while *policy-statement* blindly tries to always use both 

https://github.com/ipspace/netlab/blob/9aac9ca4c43ac5c6fa59bf1724e6be0e9053f8d9/netsim/ansible/templates/routing/_route_policy_junos.j2#L96 
 
and this results in a configuration error in case only a single AF is enabled.

i.e., 
```
module_stderr: 'b''error: Policy error: ag_aggregate-ipv6 route-filter-list referenced (in term seq10_ipv6) but not defined\nerror: configuration check-out failed'''
 ```

ssasso:
Fix #2612

### DIFF
--- a/netsim/ansible/templates/routing/_route_policy_junos.j2
+++ b/netsim/ansible/templates/routing/_route_policy_junos.j2
@@ -93,7 +93,7 @@ policy-options {
 {#  if entry contains a prefix, we must add different (term) entries for ipv4 and ipv6 #}
 {#   as per juniper documentation: Prefixes specified in a from statement must be either all IPv4 addresses or all IPv6 addresses. #}
 {%     if 'prefix' in p_entry.match|default({}) %}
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Created a single-AF version of `10-match-prefix.yml` and it works ;) Thank you!

Issue #1387, opened 2024-10-17
netlab wrapper for tc


ipspace:
It would be nice to have a wrapper for some reasonably-simple tc CLI (**comcast** comes to mind) to introduce latency/jitter/delay/drops into topology links.

The wrapper should be similar to **netlab capture** (it can probably reuse 90% of the code) -- find the interface, execute external command to configure tc parameters

Any ideas for the CLI command we can use instead of **tc** @jbemmel @ssasso @ddutt @hellt

jbemmel:
netem - see https://containerlab.dev/manual/impairments/

I once proposed to support declarative impairments (like a WAN link with 100ms latency) but this never made it into the tool afaik

ipspace:
We could also set the link impairment parameters in the topology file ;)

ipspace:
Implemented in #2626 and #2652

PR #2626, opened 2025-08-25
Static (topology) 'tc' functionality for clab and libvirt

ipspace:
Implements #1387

### DIFF
--- a/docs/links.md
+++ b/docs/links.md
@@ -442,6 +442,23 @@ A lab device could be a networking device or a host[^HOST]. Links with attached 
 
 [^NOPASS]: To turn a link with hosts attached into a transit link, set link **role** to **lan** (or any other role).
 
+(links-netem)=
+## Link Impairment
+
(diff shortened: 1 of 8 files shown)


ipspace:
@DanPartelly @jbemmel @ssasso -- in case anyone wants to look at this (FWIW, it works beautifully)

ipspace:
> Possible further improvement: a netlab command line toggle to enable/disable the static impairment on demand, making it dynamic.

Coming in the next PR ;)

jbemmel:
See also https://github.com/srl-labs/containerlab/issues/1398

DanPartelly:
> > Possible further improvement: a netlab command line toggle to enable/disable the static impairment on demand, making it dynamic.
> 
> Coming in the next PR ;)

Then, may I suggest a topology default that defines whether TC is enabled at startup or not?

ipspace:
> See also [srl-labs/containerlab#1398](https://github.com/srl-labs/containerlab/issues/1398)

Thanks for the pointer, but I don't think I want to go down that path. I could add the "--raw" parameter to the "netlab tc set" command (once that gets implemented) to allow you to specify netem parameters directly.

Alternativey, I could make the netlab-to-tc translation map part of system defaults, and then you could do whatever you wish after adding new attributes.

PR #2627, opened 2025-08-25
Ruff

Muddyblack:
here a small ruff example

1. This one checks for unused variables ( I hope I did not break anything with that :D)
Find it important there were some sorting functions called which were never used i could imagine that can save small amount of time when doing netlab up

2. checks for unused imports 

3. maybe not needed could remove it again (sorting imports)

Added pre commit which will auto run easy fixes and a workflow to check  although i didn't test the workflow yet 😅

I only changed the files which had no auto fixes possible.

So in most cases these 3 checks have always auto fixes available.

## Usage

`ruff check` and if it should auto fix use `ruff check --fix` ( I didn't run it yet so its possible to verify that the non auto issues won't break anything.) Autofix would touch over 100 files currently

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -33,6 +33,6 @@ jobs:
       run: |
         ruff check --output-format=github .
     
-    - name: Run ruff formatter (check only)
-      run: |
-        ruff format --check --diff .
+    #- name: Run ruff formatter (check only)
+    #  run: |
+    #    ruff format --check --diff .
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,38 @@
+name: Ruff Python Linting and Formatting
+on:
+  push:
+    branches-ignore: [ devc ]
+    paths:
+    - '**.py'
+    - 'ruff.toml'
+    - '.github/workflows/ruff.yml'
+
+  pull_request:
+    paths:
+    - '**.py'
+    - 'ruff.toml'
+    - '.github/workflows/ruff.yml'
+
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.x'
+    
+    - name: Install ruff
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install ruff
+    
+    - name: Run ruff linter
+      run: |
+        ruff check --output-format=github .
+    
+    - name: Run ruff formatter (check only)
+      run: |
+        ruff format --check --diff .
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,15 @@
 ---
 repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.8.4
+  hooks:
+    # Run the linter.
+    - id: ruff
+      args: [ --fix ]
+    # Run the formatter.
+    # - id: ruff-format
+
 - repo: local
   hooks:
   - id: yamllint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,7 @@ repos:
     name: ruff check
     description: This runs ruff check. If errors occur use ruff check --fix.
     entry: ruff
+    args: [ check ]
     language: python
     types: [python]
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,11 +4,7 @@ repos:
   # Ruff version.
   rev: v0.8.4
   hooks:
-    # Run the linter.
-    - id: ruff
-      args: [ --fix ]
-    # Run the formatter.
-    # - id: ruff-format
+  - id: ruff
 
 - repo: local
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,14 @@
 ---
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  # Ruff version.
   rev: v0.8.4
   hooks:
   - id: ruff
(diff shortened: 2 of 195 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,15 @@
 ---
 repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.8.4
+  hooks:
+    # Run the linter.
+    - id: ruff
+      args: [ --fix ]
```
ipspace: I wonder how well this will work with Git, as you're about to modify files that have already been staged during the commit process. I tried to do something similar once and it was a mess.

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,38 @@
+name: Ruff Python Linting and Formatting
```
ipspace, line 1: This thing definitely works ;) See the "failing checks" section

netsim/cli/collect.py
```diff
@@ -70,7 +68,7 @@ def run(cli_args: typing.List[str]) -> None:
   (args,rest) = collect_parse(cli_args)
   log.set_logging_flags(args)
 
-  topology = load_snapshot(args)
+  load_snapshot(args)
```
ipspace, line 71: It looks like this is completely unnecessary. Have to open a separate issue for this one.

netsim/cli/restart.py
```diff
@@ -37,7 +35,7 @@ def restart_parse_args() -> argparse.ArgumentParser:
 
 def run(cli_args: typing.List[str]) -> None:
   parser = restart_parse_args()
-  args = parser.parse_args(cli_args)
+  parser.parse_args(cli_args)
```
ipspace, line 38: This seems completely unnecessary. Opening a new issue...

netsim/modules/_dataplane.py
```diff
@@ -2,11 +2,12 @@
 # Generic data plane routines
 #
 import typing
+
 from box import Box
 
-from ..utils import log
 from ..data import global_vars
-from ..data.validate import must_be_list
+from ..data.types import must_be_list
```
ipspace, line 9: Nice catch ;)

netsim/modules/vrf.py
```diff
@@ -205,7 +206,6 @@ def set_vrf_ids(obj: Box, topology: Box) -> None:
 
   asn = None
   is_global = obj is topology
-  obj_name = 'global VRFs' if is_global else obj.name
```
ipspace, line 208: This is used later in line 221 (which should have an f-string but doesn't)

.pre-commit-config.yaml
Muddyblack: Yes you are right I think the workflow is enough you dont have to check it ALL the time I just added it after a comment of yours.

I think the github workflow is sufficient maybe do checks also for the workflow not for every push only when merging sth or when pushing to the dev branch?

ipspace (changes requested):
So far, this looks really good, apart from the removed variable in vrf.py caused by lack of f-string in the subsequent error message. It would be great if you could fix that bit before proceeding, or I could do another fix after merging this.

Next step: fix everything else so the GitHub action ruff check doesn't fail, and then I think we're ready to cross the fingers and merge this.

Muddyblack:
Before one question: I did put it in the ignore files but is the "common.py" still needed?

netsim/modules/vrf.py
Muddyblack: I will fix it 👍

.pre-commit-config.yaml
ipspace: Depending on how long the check takes, I prefer having checks done before pushing (that's why I have the transformation tests in pre-push hook) so I don't have to deal with "gee, I forgot something" five minutes later.

As for github workflow, I learned the hard way to be pretty strict there. A missed check can cause failures further down the line. Yet again, it depends on how long the check takes, but it looks like ruff took only 7 seconds, so who cares 🤷‍♂️

ipspace:
> before one question i did put it in the ignore files but is the "common.py" still needed?

In theory, no. In practice, nobody ever took the time to do the cleanup, but if ruff can do it automatically, it would be nice to get it done.

However, let's do that as a separate PR, and limit this one to simple import cleanup/sorting

netsim/modules/vrf.py
ipspace: > I will fix it 👍

Thank you!

.pre-commit-config.yaml
Muddyblack: ruff is great in speed (its rust) on the current project size it takes 1 second.
Should we also keep the auto fix for the pre commit? (also 1 second just dont know the after effects in github no experience on that :D)

If I shall change it I would do it with my next commit

ipspace: > ruff is great in speed (its rust) on the current project size it takes 1 second.

👍

> Should we also keep the auto fix for the pre commit? (also 1 second just dont know the after effects in github no experience on that :D)

As I said, I'm worried about changing files in the pre-commit scripts. It might be better to fail and tell the user to run `ruff --fix`

Muddyblack:
There were more missing f strings in the file

netsim/validate/cumulus.py
```diff
@@ -7,7 +7,5 @@
 * Linux ping/route checks
 """
 
-from box import Box
 from netsim.validate.bgp.frr import *
 from netsim.validate.linux import *
-from netsim.validate.ospf.frr import show_ospf_neighbor,valid_ospf_neighbor,show_ospf_prefix,valid_ospf_prefix
```
ipspace, line 13: This should be "import *"

ipspace:
I cleaned up after ruff. Looks like it's not perfect after all 😜, which might make the "getting rid of common.py" idea harder to implement.

Anyway, the ruff GH action is still failing. That needs to be fixed before we can merge this.

Muddyblack:
i removed the format check as this would annoy to much

ipspace:
@Muddyblack Thank you! Great job ;)

Issue #2621, opened 2025-08-24
Mapping speed


Muddyblack:
Currently when i launch a yaml file with over 400 devices it gets pretty long to wait until all configs got mapped.

`[MAPPED] clab_files/xyz to xyz:/etc .... (from ....j2)`

Why is that it does 2 devices per second?

Also when i do `netlab connect xyz` it is pretty slow about 3-5 seconds.

Would it be possible to use multiprocessing depending of enduser specs and also adapt ansible.cfg forks dependent on users pc specs. Or is 10 forks where it doesn't change speed no more?

I would try to implement it if you think this is doable

Muddyblack:
sry not 5 its 14 seconds! Which is just too long for sth that should just do `docker exec labprefix+netlabprefix+xyz sh`

ipspace:
> Currently when i launch a yaml file with over 400 devices it gets pretty long to wait until all configs got mapped.
> 
> `[MAPPED] clab_files/xyz to xyz:/etc .... (from ....j2)`
> 
> Why is that it does 2 devices per second?

Because you have 400 devices? I told you you're in the "there be dragons" territory.

On a more serious note, it's pretty simple function that takes node data and node/pool addressing and passes that through a JInja2 template.

https://github.com/ipspace/netlab/blob/a4d57464e3fcca5892b6332066203442a4a1e07f/netsim/providers/__init__.py#L149

I have no idea (without some profiling info) which part of that code path becomes slower as the number of nodes increases, but I'm pretty sure there must be some O(n^2) code in there.

> Also when i do `netlab connect xyz` it is pretty slow about 3-5 seconds.

**netlab connect** does nothing else but reads the YAML file (**netlab.snapshot.yml**), extracts the device data, and executes **ssh** or **docker exec**. Either the YAML parser is slow, or you have some other issue (DNS?)

> Would it be possible to use multiprocessing depending of enduser specs and also adapt ansible.cfg forks dependent on users pc specs. Or is 10 forks where it doesn't change speed no more?

Never tested that, but you can create user templates for almost anything (you just have to figure out where to put them ;).

Anyway, if you want to test Ansible behavior, do **netlab up --no-config**, then change **ansible.cfg**, then execute **netlab initial**.

ipspace:
[...]

PR #2628, opened 2025-08-25
init multiprocessing provider

Muddyblack:
#2621

### DIFF
--- a/docs/netsim/multiprocessing.md
+++ b/docs/netsim/multiprocessing.md
@@ -1,0 +1,137 @@
+# Multiprocessing Configuration
+
+netlab automatically uses multiprocessing for file generation when processing large topologies. You can configure the exact number of workers used to optimize performance for your specific hardware.
+
+## Default Behavior
+
+netlab tries to determines the optimal number of workers based on your system:
+
+- **Small systems (≤4 cores)**: Uses all available CPU cores for maximum performance
+- **Large systems (>4 cores)**: Uses 75% of available CPU cores to avoid overwhelming the system
+
+This provides a good balance between performance and system stability across all system sizes.
+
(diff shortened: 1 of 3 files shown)


Muddyblack:
or should we keep progress as default true? 

i did a bit change it so its not 1/2 slower but still 1/4 slower at least in my tests

Muddyblack:
I should start testing locally 😂
Was sure this will happen as I touched core functions with this PR

ipspace:
I have a vague feeling that this is an overkill. You correctly identified one O(n²) bit, the other one is building the /etc/hosts file which is completely identical for all hosts.

I would pre-render the "hosts-common.j2" content (doing it once should be fast, but if you like premature optimization, check if any device needs that) and then use that in "hosts.j2".

Also, the shared values obviously should be precomputed only once (very early in the process). I would store them in topology.defaults._cache or something similar so they would be easy to fetch later on, and remove that data after the file mapping is done.

As the next optimization step, the /etc/hosts file is identical for all hosts, so it only needs to be rendered once, but let's wait with that one. Also, while the file content is the same, we should NOT use the same file -- editing /etc/hosts on one device would impact all other devices.

Last but definitely not least, DNS ;)

Muddyblack:
Glad I can learn from you :P 
For now this helps me save time. I will try it better with your suggestions for the repo :D

Muddyblack:
This brings it down to 22 seconds 😯

Muddyblack:
Yeah it is overkill 40 workers were slower than 8 workers on a over 60 core device. 

Weird I though it would use then all cpus but it does not 👀
The swap is not used neither the ram. Guess I do not fully understand how the multiprocessing package works :D

Issue #2619, opened 2025-08-22
[BUG] SR Linux fails to configure BGP default route on VRF EBGP sessions


ipspace:
04-default-originate:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp#session/04-default-originate.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/bgp#session/04-default-originate.yml-config.log)

PR #2630, opened 2025-08-25
make "bgp._session_clear" vrf aware

michelredondo:
Update bgp._session_clear to be VRF-aware to resolve [issue #2619](https://github.com/ipspace/netlab/issues/2619).
The problem surfaced after extending the bgp.session/04-default-originate.yml test to support VRFs.
It looks like ArubaCX could benefit from the same fix as well.

### DIFF
--- a/netsim/extra/bgp.session/arubacx.j2
+++ b/netsim/extra/bgp.session/arubacx.j2
@@ -71,6 +71,7 @@ router bgp {{ bgp.as }}
 {% endfor %}
 {% endif %}
 !
-{% for ngb in bgp._session_clear|default([]) %}
-do clear bgp {{ ngb }}
+do clear bgp *
+{% for vname,vdata in (vrfs|default({})).items() if vdata.bgp is defined %}
+do clear bgp vrf {{vname}} *
 {% endfor %}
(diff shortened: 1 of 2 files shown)


ipspace:
Thanks a million for fixing this one!

I will fix the transformation test results, copy the ArubaCX code from `bgp.policy` into `bgp.session`, run integration tests on both platforms, and merge this.

ipspace (approved):
Looks good to me, and the integration tests pass

netsim/extra/bgp.policy/plugin.py
```diff
@@ -157,8 +157,12 @@ def apply_config(node: Box, ngb: Box) -> None:
   global _config_name
   api.node_config(node,_config_name)                    # Remember that we have to do extra configuration
   for af in ('ipv4','ipv6'):                            # ... and add sessions that have to be cleared
-    if af in ngb:
+    if af in ngb and '_src_vrf' not in ngb:
```
jbemmel, line 160:
```
if af in ngb:
  if '_src_vrf' in ngb:
    if ngb._src_vrf not in node.vrfs:
      continue
    bgp = node.vrfs[ngb._src_vrf].bgp
  else:
    bgp = node.bgp
  data.append_to_list(bgp,'_session_clear',ngb[af])
```

ipspace: You're right. Will fix.

jbemmel: Similar in netsim/extra/bgp.session/plugin.py, perhaps a function could be defined that handles both cases

PR #2640, opened 2025-08-26
Faster extrafile mapping

Muddyblack:
https://github.com/ipspace/netlab/pull/2628#issuecomment-3222823562
REF: #2621 

So this made it from 10 minutes for 610 devices down to 1m22s

What did you mean with DNS I didnt get it fully 😅

Now the question for frr: We don't really need the ansible overhead right?
You mentioned normir? I will have to read into that ... 
But in the netlab project how would be the correct approach to add such thing as option?

I really wanna have it fast at least for frr for now :D

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -160,14 +160,6 @@ class _Provider(Callback):
     sys_folder = str(_files.get_moddir())+"/"
     out_folder = f"{self.provider}_files/{node.name}"
 
-    # Initialize cache if needed
-    if not hasattr(topology.defaults, '_cache'):
-      topology.defaults._cache = Box()
-    
-    # Cache for template content checks
-    if '_template_checks' not in topology.defaults._cache:
-      topology.defaults._cache['_template_checks'] = {}
-    
     # Pre-compute hosts entries if needed
     cache_key = '_hosts_entries_cache'
     
@@ -207,7 +199,7 @@ class _Provider(Callback):
           # For other templates, include full data
           node_data = {
             **node.to_dict(),
-            'hostvars': topology.nodes.to_dict(), 
+            'hostvars': topology.nodes.to_dict(),
             'hosts': get_host_addresses(topology).to_dict(),
             'addressing': topology.addressing.to_dict()
           }
@@ -225,8 +217,8 @@ class _Provider(Callback):
             text=f"Error rendering {template_name} into {file_name}\n{strings.extra_data_printout(str(ex))}",
             module=self.provider)
 
-        # Only print mapping info in verbose mode to reduce I/O overhead
-        log.print_verbose(f"[MAPPED] {out_folder}/{file_name} to {node.name}:{mapping} (from {template_name.replace(sys_folder,'')})")
+        strings.print_colored_text('[MAPPED]  ','bright_cyan','Mapped ')
+        print(f"{out_folder}/{file_name} to {node.name}:{mapping} (from {template_name.replace(sys_folder,'')})")
       else:
         log.error(f"Cannot find template for {file_name} on node {node.name}",log.MissingValue,'provider')
 
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -29,7 +29,11 @@ def get_cpu_model() -> str:
   elif platform.system() == "Darwin":
     processor_name = "arm64"          # Assume Apple silicon for MacOS
   elif platform.system() == "Linux":
-    processor_name = pathlib.Path("/proc/cpuinfo").read_text().splitlines()[1].split()[2]
+    # This might fail on some Linux distributions, but it's a reasonable default
+    try:
+      processor_name = pathlib.Path("/proc/cpuinfo").read_text().splitlines()[1].split()[2]
+    except:
+      processor_name = platform.processor()
   return processor_name.lower()
 
 """
@@ -128,18 +132,22 @@ class _Provider(Callback):
     map_dict = filemaps.mapping_to_dict(mappings)
     cur_binds = node.get(f'{self.provider}.{outkey}',[])
     bind_dict = filemaps.mapping_to_dict(cur_binds)
+    base_path = pathlib.Path(f"{self.provider}_files")
+
     for file,mapping in map_dict.items():
       file = file.replace('@','.')
       
-      # Check if mapping ends with :shared - if so, make it a shared file
+      # Check if mapping ends with :shared - if so, put it in the root provider folder
       if mapping.endswith(':shared'):
         mapping = mapping.rsplit(':shared', 1)[0] + ':ro'
-        out_folder = f"{self.provider}_files/shared"
+        out_path = base_path / file
       else:
-        out_folder = f"{self.provider}_files/{node.name}"
-      
-      if file in bind_dict:
+        out_path = base_path / node.name / file
+
+      out_key = out_path.as_posix()
+      if out_key in bind_dict:
         continue
+
       if not self.find_extra_template(node,file,topology):
         log.error(
           f"Cannot find template {file}.j2 for extra file {self.provider}.{inkey}.{file} on node {node.name}",
@@ -147,8 +155,7 @@ class _Provider(Callback):
           module=self.provider)
         continue
 
-      bind_dict[f"{out_folder}/{file}"] = mapping
-
+      bind_dict[out_key] = mapping
 
     node[self.provider][outkey] = filemaps.dict_to_mapping(bind_dict)
 
@@ -171,26 +178,21 @@ class _Provider(Callback):
         'hostvars': topology.nodes.to_dict(),
         'hosts': get_host_addresses(topology),
         'addressing': topology.addressing.to_dict()
-    }  
+    }
     
-    prefix = f"{self.provider}_files/"
-    prefix_len = len(prefix)
+    base_path = pathlib.Path(f"{self.provider}_files")
 
-    for file,mapping in bind_dict.items():
-      # Only handle files placed under provider_files/
-      if not file.startswith(prefix):
+    for file_str,mapping in bind_dict.items():
+      full_out_path = pathlib.Path(file_str)
+      
+      # Only handle files placed under the provider's file directory
+      try:
+        full_out_path.relative_to(base_path)
+      except ValueError:
         continue
 
-      # Derive the out_folder and the relative filename
-      rel = file[prefix_len:]
-      if rel.startswith('shared/'):
-        file_rel = rel[7:]  # Remove 'shared/' prefix
-        is_shared = True
-      else:
-        parts = rel.split('/',1)
-        file_rel = parts[1] if len(parts) > 1 else ''
-        is_shared = False
-
+      # The template name is the file name part of the path
+      file_rel = full_out_path.name
       if not file_rel:
         # nothing to render
         continue
@@ -199,12 +201,6 @@ class _Provider(Callback):
       if not template_name:
         log.error(f"Cannot find template for {file_rel} on node {node.name}",log.MissingValue,'provider')
         continue
-
-
-      if is_shared:
-        full_out_path = pathlib.Path(f"{prefix}shared") / file_rel
-      else:
-        full_out_path = pathlib.Path(file)
 
       # Create parent dirs if needed
       full_out_path.parent.mkdir(parents=True,exist_ok=True)
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -39,6 +39,7 @@ class _Provider(Callback):
     self.provider = provider
     if 'template' in data:
       self._default_template_name = data.template
+    self._template_cache = {}
 
   @classmethod
   def load(self, provider: str, data: Box) -> '_Provider':
@@ -53,10 +54,23 @@ class _Provider(Callback):
     return 'templates/provider/' + self.provider
 
   def get_full_template_path(self) -> str:
-    return str(_files.get_moddir()) + '/' + self.get_template_path()
+    return os.path.join(str(_files.get_moddir()), self.get_template_path())
 
   def find_extra_template(self, node: Box, fname: str, topology: Box) -> typing.Optional[str]:
-    if fname in node.get('config',[]):                    # Are we dealing with extra-config template?
+    """Find template with caching to avoid repeated filesystem searches"""
+    # Create a hashable key from mutable objects
+    cache_key = (node.name, node.device, fname, node.get('_daemon', False))
+    
+    if cache_key in self._template_cache:
+      return self._template_cache[cache_key]
+    
+    result = self._find_extra_template_uncached(node, fname, topology)
+    self._template_cache[cache_key] = result
+    return result
+
+  def _find_extra_template_uncached(self, node: Box, fname: str, topology: Box) -> typing.Optional[str]:
+    """Uncached version of find_extra_template"""
+    if fname in node.get('config',[]):
       path_prefix = topology.defaults.paths.custom.dirs
       path_suffix = [ fname ]
       fname = node.device
@@ -69,11 +83,11 @@ class _Provider(Callback):
           path_suffix.append(node._daemon_parent)
         path_prefix.append(str(_files.get_moddir() / 'daemons'))
 
-    path = [ pf + "/" + sf for pf in path_prefix for sf in path_suffix ]
+    path = [ os.path.join(pf, sf) for pf in path_prefix for sf in path_suffix ]
     if log.debug_active('clab'):
       print(f'Searching for {fname}.j2 in {path}')
 
-    found_file = _files.find_file(fname+'.j2',path)
+    found_file = _files.find_file(f'{fname}.j2', path)
     if log.debug_active('clab'):
       print(f'Found file: {found_file}')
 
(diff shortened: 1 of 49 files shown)


netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
     sys_folder = str(_files.get_moddir())+"/"
     out_folder = f"{self.provider}_files/{node.name}"
 
+    # Initialize cache if needed
+    if not hasattr(topology.defaults, '_cache'):
```
ipspace: No needed, we configure Box in a way where it automatically adds (sub)boxes on first reference

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
+      topology.defaults._cache = Box()
+    
+    # Cache for template content checks
+    if '_template_checks' not in topology.defaults._cache:
```
ipspace: Same. The first time you'll reference topology.defaults._cache._template_checks, the box will be created (recursively if needed)

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
             out_folder=out_folder, filename=file_name)
         except Exception as ex:
           log.fatal(
             text=f"Error rendering {template_name} into {file_name}\n{strings.extra_data_printout(str(ex))}",
             module=self.provider)
 
-        strings.print_colored_text('[MAPPED]  ','bright_cyan','Mapped ')
-        print(f"{out_folder}/{file_name} to {node.name}:{mapping} (from {template_name.replace(sys_folder,'')})")
+        # Only print mapping info in verbose mode to reduce I/O overhead
+        log.print_verbose(f"[MAPPED] {out_folder}/{file_name} to {node.name}:{mapping} (from {template_name.replace(sys_folder,'')})")
```
ipspace: Nope, I want to see stuff ;) Verbose mode turns on way too many other things.

I have no problem if you check for the "quiet" flag if you think 600 lines printed to a terminal will slow you down

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
       else:
         log.error(f"Cannot find template for {file_name} on node {node.name}",log.MissingValue,'provider')
 
+  def _compute_hosts_entries(self, topology: Box) -> str:
```
ipspace: You're reinventing way too many wheels here. Just send the whole thing through a Jinja2 template and save the string. It's only done once.

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
       if not out_folder in file:                  # Skip files that are not mapped into the temporary provider folder
         continue
       file_name = file.replace(out_folder+"/","")
       template_name = self.find_extra_template(node,file_name,topology)
       if template_name:
-        node_data = node + { 'hostvars': topology.nodes, 
-                             'hosts': get_host_addresses(topology),
-                             'addressing': topology.addressing }  # Needed for subnet prefix
+        # Check if this is a hosts file (with caching)
+        is_hosts_file = False
```
ipspace: Overcomplicating. Just go through all devices, check if any of them has anything like "hosts" in the binds, and if they do, render the "hosts-common.j2" template and save the results.

netsim/providers/__init__.py
ipspace: On an unrelated note, saving cached stuff in topology is not a good idea as we have to convert it to dict later on. It would be better to just add an optional parameter (cache: dict)

netsim/providers/__init__.py
ipspace: Also, calculate the "addressing" and "hosts" cache entries in the calling function before starting the per-node loop, and pass the results as "cache" parameter

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
+        
+        if is_hosts_file and cache_key not in topology.defaults._cache:
+          # Pre-compute hosts entries once for all nodes
+          log.print_verbose("Pre-computing hosts entries for all nodes")
+          hosts_entries = self._compute_hosts_entries(topology)
+          topology.defaults._cache[cache_key] = hosts_entries
+          log.print_verbose(f"Pre-computed {len(hosts_entries)} hosts entries (cached for all nodes)")
+        
+        # Prepare minimal node data - only include what's actually needed
+        if is_hosts_file and cache_key in topology.defaults._cache:
```
ipspace: You're assuming you know what will be needed in the future. You might be wrong, and I don't like that ;)

netsim/providers/__init__.py
```diff
@@ -161,34 +160,128 @@ def create_extra_files(
...
+            '_hosts_entries': topology.defaults._cache[cache_key],
+            'addressing': topology.addressing.to_dict()
+          }
+          log.print_verbose(f"Using cached hosts entries for {node.name}")
+        else:
+          # For other templates, include full data
+          node_data = {
+            **node.to_dict(),
+            'hostvars': topology.nodes.to_dict(), 
+            'hosts': get_host_addresses(topology).to_dict(),
```
ipspace: You just need one way to set up the node data, and if you really want to optimize stuff, you should store all "to_dict" results in that cache, either on first use, or in the calling function.

Even with FRR, when you create the "frr-daemons" file, you'll get the performance hit of three to_dict calls plus in O(n2) function (get_host_addresses)

netsim/providers/clab.py
```diff
@@ -182,9 +182,63 @@ def post_configuration_create(self, topology: Box) -> None:
...
+      return
+    
+    log.print_verbose(f"Processing {len(nodes_with_binds)} nodes for extra files")
+    
+    # Check if we should use parallel processing
+    use_parallel = len(nodes_with_binds) > 10  # Only use parallel for more than 10 nodes
+    
+    if use_parallel:
+      try:
+        import multiprocessing
```
ipspace: No. You're still using a shotgun to kill a mosquito. I understand that you love your multiprocessing code, but I don't think it's needed if we clean up the other stuff.

netsim/templates/provider/clab/linux/hosts.j2
```diff
@@ -8,4 +8,8 @@ ff00::0	ip6-mcastprefix
 ff02::1	ip6-allnodes
 ff02::2	ip6-allrouters
 #
+{% if _hosts_entries is defined %}
```
ipspace, line 11: Perfect!

ipspace:
> What did you mean with DNS I didnt get it fully 😅

Deploy a DNS server somewhere (someone would have to add DNS server support to netlab, but hey...)

> Now the question for frr: We don't really need the ansible overhead right? You mentioned normir? I will have to read into that ... But in the netlab project how would be the correct approach to add such thing as option?

I would suggest to have something outside of netlab (for the moment). You'd then start the lab with "netlab up --no-config" and then run your script to configure stuff.

jbemmel:
Since a large part of the optimization seems to be in the ```hosts``` file generation, why not simply generate one copy of this file (instead of per node) and map that same file read-only into all containers?

No need for all the multithreading and caching, just adding the concept of shared (per device type) mapping files

Muddyblack:
> Since a large part of the optimization seems to be in the `hosts` file generation, why not simply generate one copy of this file (instead of per node) and map that same file read-only into all containers?
> 
> No need for all the multithreading and caching, just adding the concept of shared (per device type) mapping files

Yes that was what I was thinking on currently 😅 Just did too much threading stuff in the last days 😂

ipspace:
> Since a large part of the optimization seems to be in the `hosts` file generation, why not simply generate one copy of this file (instead of per node) and map that same file read-only into all containers?

I love this idea, but didn't want to go down this path because someone could modify /etc/hosts and that would then be applied to all devices. However, the "read-only" part makes it viable.

What if we'd add ":shared" to the target filename to indicate it's a shared file, and then change that to ":ro". The file generation code would have to keep a cache of shared files (indexed by input template because "/etc/hosts" might be different for different devices), but could then just use the already-existing file and move on.

That plus moving the `get_host_addresses(topology)` and as many `.to_dict()` as possible out of the inner loop should make the whole thing MUCH faster.

Muddyblack:
okay this brought it down to 20-30 seconds I will look a bit over before pusing but I guess you will still find sth to do better 😁

Well 2 minutes gets only faster when ignoring some node items 😅

Muddyblack:
or should we maybe not differ between the different device types should still be same host file for all ?

ipspace:
> or should we maybe not differ between the different device types should still be same host file for all ?

"Premature optimization is the root of all evil" (supposedly Donald Knuth)

There are just a few device types that use the mapped /etc/hosts files. In most cases, you'd have FRR and Linux. One or two iterations of the same algorithm do not matter in the big scheme of things (as compared to 600). There might be other optimization opportunities.

netsim/providers/__init__.py
```diff
@@ -55,10 +54,23 @@ def get_template_path(self) -> str:
     return 'templates/provider/' + self.provider
 
   def get_full_template_path(self) -> str:
-    return str(_files.get_moddir()) + '/' + self.get_template_path()
+    return os.path.join(str(_files.get_moddir()), self.get_template_path())
```
ipspace: Nitpicking (since you decided to get into this): IIRC, the get_moddir() returns a Path object, so you could use Path operators and then convert the whole thing to string

netsim/providers/__init__.py
```diff
@@ -55,10 +54,23 @@ def get_template_path(self) -> str:
...
 
   def find_extra_template(self, node: Box, fname: str, topology: Box) -> typing.Optional[str]:
-    if fname in node.get('config',[]):                    # Are we dealing with extra-config template?
+    """Find template with caching to avoid repeated filesystem searches"""
```
ipspace: Nice one ;)

However, unfortunately I don't think it's useful. You're correct in assuming the template MIGHT depend on node name and other things, so the cache key MUST include the node name. That makes cache hits highly unlikely -- we'd have to render the same file twice for the same node

netsim/providers/__init__.py
```diff
@@ -133,6 +143,15 @@ def create_extra_files_mappings(
     bind_dict = filemaps.mapping_to_dict(cur_binds)
     for file,mapping in map_dict.items():
       file = file.replace('@','.')
+      
+      # If hosts template -> create shared per-device file (shared across nodes of same device)
+      if file == 'hosts':
```
ipspace: No. Any file that has ":shared" in target name should be shared (and ":shared" changed to ":ro"). That gives us a generic sharing mechanism, not just a solution for your current problem.

netsim/providers/__init__.py
```diff
@@ -142,52 +161,93 @@ def create_extra_files_mappings(
...
     sys_folder = str(_files.get_moddir())+"/"
-    out_folder = f"{self.provider}_files/{node.name}"
 
     bind_dict = filemaps.mapping_to_dict(binds)
+    # Process other files normally
+    node_data = {
+      **node.to_dict(),
+      'addressing': topology.addressing.to_dict(),
+      'hostvars': topology.nodes.to_dict(),
+      'hosts': topology.nodes.to_dict()
```
ipspace: Wrong

netsim/providers/__init__.py
```diff
@@ -142,52 +161,93 @@ def create_extra_files_mappings(
...
-            data=node_data.to_dict(),
-            out_folder=out_folder, filename=file_name)
-        except Exception as ex:
-          log.fatal(
-            text=f"Error rendering {template_name} into {file_name}\n{strings.extra_data_printout(str(ex))}",
-            module=self.provider)
 
-        strings.print_colored_text('[MAPPED]  ','bright_cyan','Mapped ')
-        print(f"{out_folder}/{file_name} to {node.name}:{mapping} (from {template_name.replace(sys_folder,'')})")
+      # Derive the out_folder (either node-specific or shared per-device) and the relative filename
```
ipspace: Instead of most of this complex stuff, you could keep a cache of shared file paths and use the template name as the key to that cache. Then you just check whether the template is in the cache, and return shared filename if it is.

If the same template path that is marked as :shared renders into different content we have a bigger problem.

Also, while it's nice to have shared files in "shared/device/" folder, that's not absolutely necessary.

netsim/providers/__init__.py
```diff
@@ -142,52 +161,93 @@ def create_extra_files_mappings(
...
+      rel = file[prefix_len:]  # e.g. 'node1/etc/hosts' or 'shared/<device>/etc/hosts'
+      if rel.startswith('shared/'):
+        # expected rel: 'shared/<device>/path/to/file'
+        parts = rel.split('/',2)
```
ipspace: On a more generic note, if I have to read the code multiple times and still remain unsure of what it does, there's something missing on the right-hand side of the screen (the story explaining what's going on)

ipspace (changes requested):
I believe you that this stuff works (and the speedup is amazing), but it's still overly complex. We need code that not only works but that we'll understand a year from now (trust me, I learned that the hard way)

Muddyblack:
> I believe you that this stuff works (and the speedup is amazing), but it's still overly complex. We need code that not only works but that we'll understand a year from now (trust me, I learned that the hard way)

I trust you :D And I am glad you show me how to do it correctly. I will continue on this when I have a little time besides :D

ipspace:
@Muddyblack -- you might want to cherry-pick #2645 as well. It reduced integration report generation time from 80 seconds to 72 seconds and 56fb6e071a7d78bf121847c2b3200c4c4ecc8bdc brought it down to less than 3 seconds.

It looks like Jinja2 and Box don't work too well together (but then we were already passing dicts to Jinja2 in the code that hurts you most)

Muddyblack:
Aye i will need more time currently dealing with writing analysing scripts.
But from 80 seconds to 3 is pretty nice :D

From the branch (it has a lot of html code from report) https://github.com/ipspace/netlab/commit/56fb6e071a7d78bf121847c2b3200c4c4ecc8bdc i get :
```sh
Cannot import netsim Python module, netlab installation is broken
No module named 'netsim'
```

OffTopic:
I wrote a script as you recommended which will auto gen the frr config files and update the clab binds
So when i run clab deploy they are instantly configured (Allows me to use netlab for config without the ansbible up) 
It worked as expected in my small testing env so should be also good for my big lab :D

ipspace:
> From the branch (it has a lot of html code from report) [56fb6e0](https://github.com/ipspace/netlab/commit/56fb6e071a7d78bf121847c2b3200c4c4ecc8bdc) i get :

Yeah, that's a very special branch (totally decoupled from the main one) that's used to generate results of integration testing. I run it in a different directory, so it imports netsim.* from the dev branch.

> OffTopic: I wrote a script as you recommended which will auto gen the frr config files and update the clab binds

👍

netsim/providers/__init__.py
Muddyblack: Would that mean to edit the devices yml files for linux, cumulus and frr?
netsim/devices/frr.yml: `hosts: /etc/hosts:shared`

ipspace: > Would that mean to edit the devices yml files for linux, cumulus and frr? netsim/devices/frr.yml: `hosts: /etc/hosts:shared`

Yes.

Muddyblack:
@ipspace what does `topology.defaults.processor = get_cpu_model() ` I cant find it being used anywhere else in the codebase

or is t then in the j2 files this:
`{% if "amd" in defaults.processor|lower %}`

ipspace:
> or is t then in the j2 files this: `{% if "amd" in defaults.processor|lower %}`

Yes. We need to create slightly different Vagrant files (IIRC) for some VMs running on AMD.

ipspace:
BTW, this is not going anywhere until the tests pass. When you're ready for another review, please change the PR from "draft" back to "ready for review".

Alternatively, you might decide to close this one and open a new PR because several review discussions in this PR are no longer relevant.

Muddyblack:
yes sure but before I edit the tests should be clean code xD But seems like you approve this version at least in general ^^ 
I will do the tests and do a new PR 👍

Muddyblack:
@ipspace I am a bit confused I cloned from scratch the upstream dev branch and when u run `run_tests.sh` It already throws errors ?
2 errors 

although in github workflows it works oh man ...

```
FAILED test_transformation.py::test_transformation_cases - AssertionError: assert 'module:\n- b...  as: 65000\n' == 'bgp:\n  adve...  as: 65000\n'
FAILED test_transformation.py::test_error_cases - assert ['Fatal error...ck see:', ...] == ['IncorrectTy...2, column 1 ']
```

ipspace:
Because you changed the device definitions, you'll have to recreate the test results (https://netlab.tools/dev/tests/). When you recreate all the tests, check the changes. The only change should be in the FRR mappings, otherwise we have a bigger problem.

ipspace:
> @ipspace I am a bit confused I cloned from scratch the upstream dev branch and when u run `run_tests.sh` It already throws errors ? 2 errors

Would you happen to be using Ruamel instead of pyyaml? The transformation tests are comparing YAML data structures as texts, and if a package decides to change the order of dictionary keys, we have a bit of a problem.

ipspace:
Update: I checked out your PR, recreated the transformation tests, and the only changes are in **clab.binds** and **clab.config_templates**. After that, all tests pass.

So, don't worry about the test results, I can fix them once you're done with the code. If it turns out your test problems are caused by a different YAML library, I'll rewrite the testing code to explicitly use pyyaml and force dict key sorting.

On a tangential note, you're creating the shared files directly into "clab_files" directory now. That would obviously trigger a bit of a problem if someone is creative enough to have a node named **hosts** (based on some of the issues I've seen so far, the probability of that happening is close to 1.0 😜). The safest way to deal with that would be to use a prefix like "-shared-" in the shared file name, as netlab does not allow "-" in node names.

PR #2645, opened 2025-08-30
Performance: Cache Jinja2 environments by search paths

ipspace:
The original code created a new Jinja2 environment for every rendered file because the search path might have changed. This patch uses a cache of Jinja2 environments indexed by search path(s), resulting in cached templates and thus faster rendering.

### DIFF
--- a/netsim/utils/templates.py
+++ b/netsim/utils/templates.py
@@ -1,6 +1,7 @@
 #
 # Common routines for create-topology script
 #
+import functools
 import pathlib
 import typing
 
@@ -29,6 +30,14 @@ Path parameters:
 * user_template_path -- subdirectory of user directories to search (current, home)
 """
 
+@functools.lru_cache()
+def get_jinja2_env_for_path(template_path: tuple) -> Environment:
+  ENV = Environment(loader=FileSystemLoader(template_path), \
+          trim_blocks=True,lstrip_blocks=True, \
+          undefined=make_logging_undefined(base=StrictUndefined))
+  add_ansible_filters(ENV)
+  return ENV
+
 def render_template(
       data: typing.Dict,
       j2_file: typing.Optional[str] = None,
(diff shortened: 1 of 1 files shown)


jbemmel (commented):
Could we use functools.cache instead? Python 3.9 and above, eviction functionality is not used anyway

ipspace:
> Could we use functools.cache instead? Python 3.9 and above, eviction functionality is not used anyway

We probably don't have more than a few search paths anyway, so it should probably work as well.

ipspace:
> Python 3.9 and above, eviction functionality is not used anyway

BTW, where did you get this information?

Also, cache was added in 3.9 and we still support 3.8. Maybe it's time to revisit that decision @ssasso?

jbemmel:
> > Python 3.9 and above, eviction functionality is not used anyway
> 
> BTW, where did you get this information?
> 
> Also, cache was added in 3.9 and we still support 3.8. Maybe it's time to revisit that decision @ssasso?

I started searching for functools lru_cache and the AI completed the rest

ipspace:
> I started searching for functools lru_cache and the AI completed the rest

I could find no such information in https://docs.python.org/3/library/functools.html (but of course it could be my selective blindness), so I'd really love to have the source for that claim.

jbemmel:
> > I started searching for functools lru_cache and the AI completed the rest
> 
> I could find no such information in https://docs.python.org/3/library/functools.html (but of course it could be my selective blindness), so I'd really love to have the source for that claim.

https://stackoverflow.com/questions/70071759/lru-cache-vs-cache

It suggests that cache may be just a simple wrapper around lru_cache, not sure it’s worth the fuzz.

ipspace:
> It suggests that cache may be just a simple wrapper around lru_cache, not sure it’s worth the fuzz.

Maybe it's worth reading the original documentation. It not only "suggests", it's [pretty clear](https://docs.python.org/3/library/functools.html#functools.cache) that @cache is equivalent to @lru_cache(maxsize=None).

PR #2655, opened 2025-09-04
Add GraphViz and D2 graph titles specified in graph.title global attribute

ipspace:


ipspace:
Another graph goodie: set graph title

PR #2656, opened 2025-09-04
Implement select IGP parameters on the primary loopback interface

### LINKED ISSUE #2625: Loopback validation code should accept interface- and module attributes
Guessing: set **loopback.ospf** to False? Other than that, I have no useful idea

_Originally posted by @ipspace in https://github.com/ipspace/netlab/discussions/2624#discussioncomment-14208294_

ipspace:
* The loopback data is checked as a new 'loopback' type to allow IGP modules to register a subset of link/interface attributes as begin valid on the loopback interface
* The loopback interface must be validated with the top-level 'validate_attributes' function, otherwise the module attributes are not recognized.
* Loopback data is thus checked first as a dict/bool within the node data, and then rechecked against the new 'loopback' type (assuming the 'loopback' attribute is not set to True/False).
* The 'module attribute can be False' list is extended to include the 'loopback' data type
* We have to deal with 'is the IGP set to False on loopback' in numerous places within the generic routing code until we finally remove 'igp: False' data in 'remove_unused_igp' function
* To make that work, the 'add_loopback_igp' call had to be moved, resulting in IGP information appearing on the loopback interface even when the IGP module is removed from the node.

Fixes #2625 and solves #2624

### DIFF
--- a/netsim/defaults/attributes.yml
+++ b/netsim/defaults/attributes.yml
@@ -45,7 +45,7 @@ internal:                   # Internal attributes, not validated
 
 global_extra_ns: [ providers, outputs ]
 
-can_be_false: [ link, interface ]
+can_be_false: [ link, interface, loopback ]
 
 link:                       # Global link attributes
   bandwidth: int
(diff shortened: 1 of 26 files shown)


ipspace:
@DanPartelly -- this should solve your "I only want one IGP on the loopback" challenge

@jbemmel -- instead of leaving IGP attributes on the loopback (current behavior, now with "passive" flag) when the IGP module is removed from the node due to being useless (as in: no external interfaces are using it), I could remove them. Any thoughts?

netsim/defaults/attributes.yml
```diff
@@ -122,15 +122,19 @@ node:
...
   provider: id
   cpu:
   memory: int
   unmanaged: bool
 
+loopback:
+  ipv4: { type: ipv4, use: host_prefix, _alt_types: [ bool ] }
+  ipv6: { type: ipv6, use: host_prefix, _alt_types: [ bool ] }
+  pool: addr_pool
+  _alt_types: [ bool ]
```
jbemmel: Wondering if this is needed?

jbemmel:
> 
> @jbemmel -- instead of leaving IGP attributes on the loopback (current behavior, now with "passive" flag) when the IGP module is removed from the node due to being useless (as in: no external interfaces are using it), I could remove them. Any thoughts?

I like cleaned up attributes, so yes I would prefer to remove them altogether

DanPartelly (approved):
Thank you.This is a nice surprise, didn't expected anyone to work on this untill after 25.09.

netsim/defaults/attributes.yml
ipspace: Indeed it's not. Removed. Thank you!

PR #2657, opened 2025-09-04
Enhancement/faster mapping

Muddyblack:
Re-implemented on top of latest `dev` branch

ref #2640, https://github.com/ipspace/netlab/issues/2621

### DIFF
--- a/netsim/providers/__init__.py
+++ b/netsim/providers/__init__.py
@@ -197,8 +197,16 @@ class _Provider(Callback):
       except ValueError:
         continue
 
-      # The template name is the file name part of the path
-      file_rel = full_out_path.name
+      # We have to recover the template name from the final (local) file name
+      # If the final file name is within the provider/node directory, the template
+      # name is the rest of the path (which might be a file name or a relative path)
+      # otherwise the template name is just the file name (used for shared files)
+      #
+      try:
+        file_rel = str(full_out_path.relative_to(base_path / node.name))
+      except ValueError:
+        file_rel = full_out_path.name
+
       if not file_rel:
         # nothing to render
         continue


netsim/providers/__init__.py
```diff
@@ -30,13 +29,21 @@ def get_cpu_model() -> str:
   elif platform.system() == "Darwin":
     processor_name = "arm64"          # Assume Apple silicon for MacOS
   elif platform.system() == "Linux":
-    processor_name = pathlib.Path("/proc/cpuinfo").read_text().splitlines()[1].split()[2]
+    # This might fail on some Linux distributions, but it's a reasonable default
```
ipspace, line 32: I totally understand that you can't resist fixing random stuff (I'm the same), but if we ever have to revisit the fixed code and figure out why it was changed, it would be way harder than if the fix would be submitted as a separate PR with the proper commit message (more about that after this one is merged)

Muddyblack: Yes I was thinking to not add that again from the last pr. I will do it better in the future and only edit the PR's purpose 😅

Muddyblack: Shall I revert this?

netsim/providers/__init__.py
```diff
@@ -131,63 +138,104 @@ def create_extra_files_mappings(
...
+      full_out_path = pathlib.Path(file_str)
+
+      # Only handle files placed under the provider's file directory
+      try:
+        full_out_path.relative_to(base_path)
+      except ValueError:
+        continue
+
+      # The template name is the file name part of the path
+      file_rel = full_out_path.name
```
ipspace: Showstopper: this breaks the ability to have relative paths in template names, for example:

```
nodes:
  h1:
    clab.config_templates:
      _n_r1: /etc/nr1
      a/_n_r2: /etc/nr2
```

netsim/providers/__init__.py
```diff
@@ -131,63 +138,104 @@ def create_extra_files_mappings(
...
-          log.fatal(
-            text=f"Error rendering {template_name} into {file_name}\n{strings.extra_data_printout(str(ex))}",
-            module=self.provider)
 
+      # For shared files, extract the original file name (remove 'shared-' prefix)
+      template_fname = file_rel.removeprefix(self.SHARED_PREFIX)
+
+      template_name = self.find_extra_template(node, template_fname, topology)
+      if not template_name:
+        log.error(f"Cannot find template for {file_rel} on node {node.name}",log.MissingValue,'provider')
```
ipspace, line 219: We need to provide search path in this error message or it's too confusing to troubleshoot. Will open an issue for that.

ipspace (changes requested):
Great job. Thanks a million!

There's just one tiny little problem -- because you're using `.name` instead of stripping the `out_folder` from the path, we can no longer use relative paths in `config_templates`.

The fix shouldn't be too hard, and if you don't have the time to fix it, I can do that and then merge the PR.

netsim/providers/__init__.py
ipspace: > Shall I revert this?

No, we'll manage to live with that ;)

ipspace:
The impact on 300-node topology. Before the PR:

```
real	0m47.714s
user	0m47.454s
sys	0m0.222s
```

After the PR:

```
real	0m17.743s
user	0m17.532s
sys	0m0.154s
```

We could shave off additional few seconds with custom 'output' parameter that would skip the YAML snapshot file.

Nice job ;))

ipspace:
And now for the "commit message" rant. As you might have noticed, I believe in having explanatory commit messages that might tell a future software developer trying to fix one of our blunders why we changed something and what the change is supposed to be doing.

The default commit message when merging a PR is composed from the commit messages in the PR and this is what you left for me:

> * back to lasst PR
> * tests
> * review

I hate doing copy-paste from the PR description into the commit message because GitHub doesn't care about 72-column recommendations, but a sane PR description would still be infinitely better than:

> Re-implemented on top of latest dev branch
> ref https://github.com/ipspace/netlab/pull/2640, https://github.com/ipspace/netlab/issues/2621

I could go into how useful those documents would be, but won't waste our time. I also don't have to go into full-Linus mode, because the master himself did it much better: https://github.com/torvalds/linux/pull/17#issuecomment-5659933 (found in https://news.ycombinator.com/item?id=4483566 which also contains other good info)

Anyway, in the future, please make sure that the first commit in a non-trivial PR has a succinct commit message explaining what you did (and even better, why you did it)

ipspace:
... and I broke the build when merging this 🤦‍♂️ Will fix (and make sure all GH pipelines include Python 3.8).

Before you ask why I care about an obsolete Python version: that's what Ubuntu 20.04 runs on (more @ https://github.com/ipspace/netlab/issues/1689)

Muddyblack:
I learn good git usage from you :D
I will read that and have a look at earlier commit messages from your repo

ipspace:
> I learn good git usage from you :D

Glad I could be of some use 😜

> I will read that and have a look at earlier commit messages from your repo

Thank you!

jbemmel:
> ... and I broke the build when merging this 🤦‍♂️ Will fix (and make sure all GH pipelines include Python 3.8).
> 
> Before you ask why I care about an obsolete Python version: that's what Ubuntu 20.04 runs on (more @ #1689)

FWIW, Ubuntu 20.04 LTS (Focal Fossa) reached the End of Standard Support on May 31, 2025

PR #2663, opened 2025-09-05
Fix: Implement 'str.removeprefix' in strings.utils for Python 3.8

ipspace:
As we're not yet ready to say goodbye to our good friend Python 3.8, we still can't use str.removeprefix, so I added it to the strings utility module to keep mypy (and Ubuntu 20.04 users) happy.

Also: add Python 3.8 to pull request CI/CD pipeline to ensure a similar SNAFU won't happen anytime soon.

### DIFF
--- a/.github/workflows/t-pull.yml
+++ b/.github/workflows/t-pull.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 6
       matrix:
-        python-version: [ "3.11", "3.12", "3.13" ]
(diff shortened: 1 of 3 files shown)


sdktr:
Did you intent to remove 3.11 from the PR CI/CD pipeline?

ipspace:
> Did you intent to remove 3.11 from the PR CI/CD pipeline?

Yes. If 3.8, 3.12, and 3.13 all pass, it's highly likely 3.11 would pass as well.

PR #2670, opened 2025-09-07
Add tc command description to usage.txt.

### LINKED ISSUE #2669: [BUG] The new command netlab tc is not displayed when running netlab with no CLI options.
Netlab's new command "tc" is not covered in usage.txt, hence it is not displayed when we run netlab with no command-line options.

DanPartelly:
Simply add TC command and a description to cli/usage.txt. Closes #2669.

### DIFF
--- a/netsim/cli/usage.txt
+++ b/netsim/cli/usage.txt
@@ -29,6 +29,9 @@ capture     Start packet capture on the specified node/interface
 
 exec        Executes a command on one or more network devices
 
+tc 	    Disable, enable, display, or modify link impairment parameters on 
+	    libvirt virtual machines or containerlab-created Docker containers.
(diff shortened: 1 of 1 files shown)


netsim/cli/usage.txt
```diff
@@ -29,6 +29,9 @@ capture     Start packet capture on the specified node/interface
 
 exec        Executes a command on one or more network devices
 
+tc 	    Disable, enable, display, or modify link impairment parameters on
```
ipspace: This is displayed as pure ASCII (not Markdown), so the texts should be aligned

DanPartelly: Got rid of the tabs.

PR #2676, opened 2025-09-11
Add lab host entries to SRL nodes

andywhitaker:
Adding host entries to SRL is interesting because the Clab node deployment process for SRL already adds the node name to /etc/hosts and trying to override that in the config with an inline address does not work. I went ahead and gave a go here where I'm appending .netlab to the name. I don't love it but it works. :)

### DIFF
--- a/netsim/ansible/templates/initial/srlinux.j2
+++ b/netsim/ansible/templates/initial/srlinux.j2
@@ -109,3 +109,22 @@ updates:
 
 {% endfor %}
 {% endif %}
+
+
+{# 
+Add Host Entries
+----------------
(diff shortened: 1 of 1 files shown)


ipspace:
@andywhitaker -- this gave me an interesting idea: can we just modify /etc/hosts like we do for FRR and Linux... and it works!

Could you check whether #2677 works for you?

andywhitaker:
Great idea! Yes, it's working for me too! Now I get to troubleshoot my brain to understand what form of Linux / SR Linux magic is happening to sort out the network namespaces automatically ;)

Good work!

ipspace:
Replaced by #2677

PR #2677, opened 2025-09-12
SRLinux: Use Linux-style /etc/hosts files

ipspace:
We cannot define host-to-IP mappings with the SR Linux CLI because containerlab already defined the hosts in /etc/hosts (see #2676 for details). However, we can create /etc/hosts file in advance, and then containerlab only adds its entries which due to some SRL magic work as expected across multiple network instances.


ipspace:
Tested by @andywhitaker (see #2676)

PR #2679, opened 2025-09-12
Implement per-VRF control of IGP address families

ipspace:
This change adds 'af' attribute to isis, ospf, and rip VRF parameters, allowing you to specify which address families you want to run with a VRF IGP instance.

Implements #2398

### DIFF
--- a/netsim/modules/_routing.py
+++ b/netsim/modules/_routing.py
@@ -447,9 +447,6 @@ def check_vrf_protocol_support(
         module=proto)
       return
 
-    if af and not 'af' in v_data[proto]:
-      v_data[proto].af[af] = True                 # Add the AF to VRF routing protocol data
-
 """
 check_intf_support -- check device support for optional interface parameters
 """


netsim/modules/_routing.py
```diff
@@ -431,6 +434,7 @@ def check_vrf_protocol_support(
     if not proto in v_data:                       # Are we running the target protocol in the VRF?
       continue
     if af and not af in v_data.af:                # Does the VRF use the target address family?
```
jbemmel, line 436: ```and 'af' in v_data[proto]```

Else the pop below might create an empty ```v_data[proto].af``` and line 450 would skip setting a flag

ipspace: You don't want the flag at line 450/451 set if the v_data.proto.af exists, otherwise the AFs would get reenabled regardless of what you specified (been there, experienced that ;)

docs/module/routing_protocols.md
```diff
@@ -37,10 +37,14 @@ nodes:
...
 ## Address Families
 
-Configuration modules for all IGP routing protocols that support multiple address families (IS-IS, EIGRP) or multiple protocol instances (OSPFv2, OSPFv3) support **_protocol_.af** global- or node-level module parameter. The **af** parameter can be a list- or a dictionary of address families.
+Configuration modules for all IGP routing protocols that support multiple address families (IS-IS, EIGRP) or multiple protocol instances (OSPFv2, OSPFv3) support **_protocol_.af** global-, node-, or VRF-level module parameter. The **af** parameter can be a list or a dictionary of address families.
 
 The default value of the **af** parameter is set based on address families configured on loopback- or physical interfaces -- an address family is enabled within an IGP configuration on a device if at least one interface on that device has an IP address from that address family.
 
+```{warning}
+If you decide to set a **‌_protocol_.af** parameter, you MUST specify which address families should be active. Using the **af** parameter just to set an unwanted address family to *False* without enabling any other address family will result in no address families being used.
```
jbemmel, line 45: This is a source of potential confusion and unintended errors

Basically you're saying that if vrf[proto].af is set, it should have exactly 1 af set to True. Both afs ```True``` is the default, and setting both to ```False``` is achieved by setting ```vrf[proto]: False```

The data model could check for this

netsim/modules/_routing.py
jbemmel: I'm not talking about the case where v_data.proto.af exists - it's when it doesn't exist

ipspace: Ah, but that doesn't happen anymore, because we deal with that somewhere else ;) Anyway, will check whether removing the line that bothers you (450/451) changes anything.

docs/module/routing_protocols.md
ipspace: I'm not going into the "C should check for null pointers" territory ;)) There's a simpler solution: don't use nerd knobs you're not fully familiar with ;)

netsim/modules/_routing.py
ipspace: Removing the confusing two lines changes nothing in the transformation test results, and the integration tests pass.

PR #2680, opened 2025-09-12
Add missing 'bgp.as_list.name' attribute we lost in the mists of time

### LINKED ISSUE #2675: [BUG] bgp.as_list AS definition should accept a "name" parameter
Found an ancient topology that threw an error:

https://github.com/ipspace/netlab-examples/tree/master/BGP/LocPref-Prepend/topology.yml

Looks like netlab supported AS names in the past (and would display them in BGP graphs).

### Output

```
[ERRORS]  Errors found in downloaded.yml
[ATTR]    bgp: Incorrect BGP as_list attribute 'name' in bgp.as_list.65000
[ATTR]    bgp: Incorrect BGP as_list attribute 'name' in bgp.as_list.65100
[ATTR]    bgp: Incorrect BGP as_list attribute 'name' in bgp.as_list.65101
```

### Version

25.09 (but probably everything since we introduced stricter lab topology checks)

ipspace:
In a version far far ago, the bgp.as_list entry could have a 'name' attribute that was used in topology and BGP graphs. That attribute was lost once we implemented the strict data checking (because nobody remembered it at the time), and the graph refactoring done in 25.09 borked its use in graphs.

This PR is fixing both blunders and closes #2675

### DIFF
--- a/netsim/outputs/_graph.py
+++ b/netsim/outputs/_graph.py
@@ -44,13 +44,15 @@ def build_nodes(topology: Box, g_type: str) -> Box:
     for name,n in topology.nodes.items():
       bgp_as = n.get('bgp.as',None)
       if bgp_as:
-        bgp_as = f'AS_{bgp_as}'
-        maps.bgp[bgp_as].nodes[n.name] = n
+        map_asn = f'AS_{bgp_as}'
+        maps.bgp[map_asn].nodes[n.name] = n
+        maps.bgp[map_asn].title = f'AS {bgp_as}'
 
-  if 'bgp' in topology and 'as_list' in topology.bgp:
+  if topology.get('bgp.as_list'):
     for (asn,asdata) in topology.bgp.as_list.items():
-      if 'name' in asdata and asn in maps.bgp:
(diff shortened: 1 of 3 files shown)


ipspace:
@DanPartelly -- just in case you want to take a look at yet another graph feature

netsim/outputs/_graph.py
```diff
@@ -44,13 +44,15 @@ def build_nodes(topology: Box, g_type: str) -> Box:
     for name,n in topology.nodes.items():
       bgp_as = n.get('bgp.as',None)
       if bgp_as:
-        bgp_as = f'AS_{bgp_as}'
-        maps.bgp[bgp_as].nodes[n.name] = n
+        map_asn = f'AS_{bgp_as}'
```
jbemmel, line 47: This could be a local function, the format has to match with line 53 below

ipspace: In principle, you're right. In practice, we'll probably survive ;)

DanPartelly (approved):
Thank you.

Issue #2672, opened 2025-09-10
Allow glob- and regexp matching of group members


ipspace:
Based on #2508: it would be nice to be able to select group members based on wildcards or regular expressions.

Also: it might be possible to add "select" option that would select group members based on a Python expression.

PR #2681, opened 2025-09-12
Allow wildcard- and regular expressions in group members

ipspace:
Implements #2672

### DIFF
--- a/tests/topology/input/group-data-vlan.yml
+++ b/tests/topology/input/group-data-vlan.yml
@@ -4,7 +4,7 @@ module: [vlan, ospf]
 
 groups:
   g1:
-    members: [r1, r2]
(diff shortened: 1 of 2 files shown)


docs/groups.md
```diff
@@ -38,6 +38,12 @@ groups:
     members: [ d,e,f ]
 ```
 
+The elements in the **members** list can be:
```
jbemmel, line 41: Would be good to also apply this to ```as_list.members```

ipspace: Yeah, now that we have a generic matching function that shouldn't be too hard.

Thank you, will add.

ipspace: ... opened a separate issue for that one.

PR #2691, opened 2025-09-18
Compute VRRP MAC/LLA addresses in gateway module

ipspace:
This commit moves the burden of computing VRRP MAC/LLA addresses (when needed) to the gateway module, removing all duplicated MAC/LLA address manipulation from the configuration templates.

### DIFF
--- a/netsim/ansible/templates/gateway/cumulus_nvue.j2
+++ b/netsim/ansible/templates/gateway/cumulus_nvue.j2
@@ -36,7 +36,7 @@
                 address:
 {%     if 'ipv6' in intf.gateway %}
 {# Per RFC9568 5.2.9, the link-local ipv6 address MUST be included first. Required for interop with EOS (for example) #}
-                  fe80::200:5eff:fe00:2{{ '%02x' | format(intf.gateway.vrrp.group) }}: {}
+                  {{ intf.gateway.vrrp.lla }}: {}
 - set:
     interface:
       {{ intf.ifname }}:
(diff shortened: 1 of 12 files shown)


netsim/modules/gateway.py
```diff
@@ -63,6 +63,26 @@ def cleanup_unicast_ip(node: Box) -> None:
...
+  grp = intf.get('gateway.vrrp.group',None)
+  if not grp:
+    return
+
+  gw_data = intf.gateway.vrrp
+  grp_hex = "%02x" % gw_data.group
+  for af in log.AF_LIST:
+    if af in intf.gateway:
+      gw_data.mac[af] = f'00:00:5e:00:0{ 2 if af == "ipv6" else 1 }:{ grp_hex }'
+
```
jbemmel, line 81:
```
if af=='ipv6':
  ...
```

netsim/ansible/templates/gateway/cumulus_nvue.j2
```diff
@@ -36,7 +36,7 @@
                 address:
 {%     if 'ipv6' in intf.gateway %}
 {# Per RFC9568 5.2.9, the link-local ipv6 address MUST be included first. Required for interop with EOS (for example) #}
-                  fe80::200:5eff:fe00:2{{ '%02x' | format(intf.gateway.vrrp.group) }}: {}
+                  {{ intf.gateway.vrrp.lla }}: {}
```
jbemmel, line 39: There are also valid ipv4 link local addresses, perhaps ```ipv6_lla``` would be a better name for this attribute

ipspace: While you're technically correct, the IPv4 LLAs have no special meaning in VRRP.

ipspace:
Thank you!

> Minor comments that won't likely keep anyone awake at night

... and you're right about that ;)

PR #2692, opened 2025-09-18
Generate MAC addresses for Ethernet/SVI interfaces

ipspace:
Some devices need a configured MAC address for Ethernet or VLAN interfaces. This commit moves that logic from device configuration templates into link/VLAN transformation code.

The changed MAC generation procedure applies to IOLL2, IOSvL2, EOS, FRR, and NexusOS.


ipspace:
Oops, wrong decision. I thought the "mac_address" element was applied to FRR Ethernet interfaces, but it's just EOS being used in so many transformation tests.

Issue #2615, opened 2025-08-20
[BUG] bgp.local_as not working with VLAN


ssasso:
### Describe the bug

When vlans are in use together with "vlan level" `bgp.local_as`, local config is good, but peer config still have the "previous" AS number.

### To Reproduce

Topology:

```
provider: clab
defaults.device: frr
module: [ bgp ]

nodes:
  r1:
    module: [ vlan, bgp ]
    bgp.as: 65000
    vlans:
      red:
        bgp.local_as: 65001
  r2:
    bgp.as: 65002
  r3:
    bgp.as: 65003

vlans:
  red:
  green:

links:
- r1:
    vlan.access: red
  r2:
- r1:
    vlan.access: green
  r3:
```

`netlab inspect --node r2 | yq .bgp.neighbors[0]`:
```
{
  "ifindex": 1,
  "name": "r1",
  "as": 65000,
  "type": "ebgp",
  "ipv4": "172.16.0.1",
  "activate": {
    "ipv4": true
  }
}
```

generated R2 config:

```
router bgp 65002
 neighbor 172.16.0.1 remote-as 65000
 neighbor 172.16.0.1 description r1
```

generated R1 config:

```
router bgp 65000
 neighbor 172.16.0.2 remote-as 65002
 neighbor 172.16.0.2 local-as 65001 no-prepend replace-as
 neighbor 172.16.0.2 description r2
```

### Expected behavior

R2 config:

```
router bgp 65002
 neighbor 172.16.0.1 remote-as 65001
 neighbor 172.16.0.1 description r1
```

### Version

`dev`

ipspace:
Did you really have to??? 😢🤦‍♂️ And I hoped I will never have to touch the VLAN code again 🤢

On a more serious note, thanks a million for uncovering this one. From my guesspoint perspective, it looks like the interface neighbor list is built before the (neighbor) SVI data is augmented with the node VLAN data.

No idea how long it will take me to fix it.

ssasso:
> before the SVI data is augmented with the node VLAN data

But - at least in theory - bgp transformation should happen after vlan, if I understand this correctly:

https://github.com/ipspace/netlab/blob/a8f60ea1b3f23ec4d88aa14fc00ba42767ab3279/netsim/modules/bgp.yml#L16

jbemmel:
> > before the SVI data is augmented with the node VLAN data
> 
> But - at least in theory - bgp transformation should happen after vlan, if I understand this correctly:
> 
[...]

PR #2696, opened 2025-09-20
Add module data to VLAN neighbors

ipspace:
* Copy node VLAN data to interface data during link_pre_transform to have them available when the neighbor list is built during link transformation

Also:

* Add an optional 'limit merge to these attributes' parameter to the 'data.merge_with_removed_attributes' utility function


ipspace:
This is an attempt to solve #2615. I will push several iterations because it's easier to review changes to transformation tests on GH web interface. Once the whole thing is done, I'll mark it ready for review.

netsim/modules/vlan.py
```diff
@@ -1387,6 +1387,12 @@ def link_pre_link_transform(self, link: Box, topology: Box) -> None:
       for attr in topology.defaults.vlan.attributes.copy_vlan_to_intf:
         if attr in vdata:
           intf[attr] = vdata[attr]
+      
+      # Finally, copy module attributes from node VLAN data to access interface data
+      # These attributes will be used to populate neighbor's "neighbor" list in VLANs
+      # with mode = route. We have to fix the IRB neighbors when rebuilding neighbor list.
+      #
+      data.merge_with_removed_attributes(intf,vdata,topology.nodes[intf.node].get('module',[]))
```
ipspace, line 1395: The only significant impact of this change is in the `stp.yml` results where the VLAN STP priority is applied to a VLAN access port. I would claim this is the correct behavior anyway.

In all other changes to transformation test results, the VLAN data is applied to link **interfaces** entry, which is expected, and there are no further changes.

tests/topology/expected/dhcp-vlan.yml
```diff
@@ -252,8 +266,6 @@ nodes:
         excluded:
           ipv4:
           - 172.16.0.2
-        gateway:
```
ipspace, line 255: This needs to be fixed

tests/topology/expected/group-data-vlan.yml
```diff
@@ -116,10 +116,10 @@ nodes:
       ipv4: 172.16.1.1/24
       name: VLAN blue (1000) -> [r2,r3]
       neighbors:
-      - ifname: Vlan1000
+      - ifname: Ethernet2
```
ipspace, line 119: This has no impact on the node data, but needs to be checked.

netsim/modules/vlan.py
```diff
@@ -759,21 +759,19 @@ def update_vlan_neighbor_list(vlan: str, phy_if: Box, svi_if: Box, node: Box,top
...
-        n_map[node.name].pop(af,None)
+  svi_info = node.get('module',[]) + list(log.AF_LIST)                  # Include SVI module and address information
+  svi_ifdata = get_box({ 'ifname': svi_if.ifname, 'node': node.name })  # Prepare SVI interface data for neighbor list
+  svi_ifdata += { k:v for k,v in svi_if.items() if k in svi_info }      # Copy relevant SVI info into neighbor entry
+
+  if node.name in n_map:
+    # The current node data must be based on physical interfaces
+    # Remove the current node from the neighbor list and add the SVI information
+    #
+    vlan_data.neighbors = [ n for n in vlan_data.neighbors if n.node != node.name ] + [ svi_ifdata ]
```
ipspace, line 770: Impact of this change: potentially rearranged list of VLAN neighbors, which might trigger a change in the default gateway (because a different router becomes the first neighbor) and a related change in the static routes on attached hosts.

netsim/modules/vlan.py
```diff
@@ -759,21 +759,19 @@ def update_vlan_neighbor_list(vlan: str, phy_if: Box, svi_if: Box, node: Box,top
...
-    n_data = data.get_box({
-               'ifname': svi_if.ifname,
-               'node': node.name })                                     # ... not yet, create neighbor data
-    for af in ('ipv4','ipv6'):
-      if af in svi_if:                                                  # ... copy SVI interface addresses to neighbor data
-        n_data[af] = svi_if[af]
-    vlan_data.neighbors.append(n_data)                                  # Add current node as a neighbor to VLAN neighbor list
+    # The current node is not yet in the VLAN neighbor list, add the SVI information
+    #
+    vlan_data.neighbors.append(svi_ifdata)
```
ipspace, line 774: Impact of this change: module data from SVI interface is added to the VLAN (and subsequently interface) neighbor list

netsim/modules/dhcp.py
```diff
@@ -122,11 +122,19 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
 def get_gateway_data(link: Box, topology: Box) -> Box:
-  if isinstance(link.get('gateway',None),Box):
-    return link.gateway
+  if isinstance(link.get('gateway',None),Box):    # Do we have on-link gateway data
+    return link.gateway                           # Cool, return that
   
-  ifdata = data.get_box({'neighbors': link.get('neighbors',[] )})
-  (gw_data,u_flag) = routing.create_gateway_last_resort(ifdata,data.get_box({'ipv4': True}),topology)
+  link_ngb = link.get('neighbors',[])             # Get link neighbors
+  for ngb in link_ngb:                            # Fix DHCP relays that also have DHCP client information
```
ipspace, line 129: Impact of this change: the DHCP state of link-level **neighbors** is aligned with the final interface DHCP state, removing the impact of copying SVI interface module attributes to the VLAN neighbor list.

netsim/modules/dhcp.py
```diff
@@ -122,11 +122,19 @@ def check_protocol_support(node: Box, topology: Box) -> bool:
...
+    if ngb.get('dhcp.server'):                    # ... because that attribute will get removed from the
+      ngb.dhcp.pop('client',None)                 # ... interface data anyway and just throws off the next function
+
+  ifdata = data.get_box({'neighbors': link_ngb }) # Now create a fake interface data structure
+
+  # Try to get the gateway-of-last-resort from the fake "interface" neighbors
+  #
+  (gw_data,_) = routing.create_gateway_last_resort(ifdata,data.get_box({'ipv4': True}),topology)
```
ipspace, line 137: Impact of this change: the DHCP state of link- or VLAN-level **neighbors** is aligned with the final interface DHCP state, removing the impact of copying SVI interface module attributes to the VLAN neighbor list.

ipspace:
As expected, these changes resulted in massive changes in transformation test results. I went through all of those, and they all looked legit to me. I also documented the impact of individual changes to the test results to help you understand what's going on.

The whole thing should now be working the way I expect it should work (how's that for a definitive statement 😜). It solves the BGP local_as problem and the "correct" VLAN neighbor data might help us solve a few others (for example, some reports and graphs struggled with VLANs).

PR #2697, opened 2025-09-21
OSPF support for OpenBSD devices

remilocherer:
This adds OSPFv2 and OSPFv3 support for OpenBSD.

I had to change the role from 'host' to 'router'. Overwriting it in the topology did not work (ospf templates were not evalutated).

### DIFF
--- a/netsim/ansible/templates/ospf/openbsd.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/openbsd.ospfv2.j2
@@ -1,0 +1,55 @@
+{% macro config(ospf_vrf,ospf_data,intf_data) %}
+{%- set areas = intf_data|selectattr('ipv4', 'defined')|map(attribute='ospf.area', default='')|reject('eq', '')|sort|unique %}
+#!/bin/sh
+
+cat <<OSPFD_CONF >/etc/ospfd.conf
+{% if ospf_data.router_id|ipv4 %}
+router-id {{ ospf_data.router_id }}
+{% endif %}
+spf-delay msec 100
+spf-holdtime msec 200
+{% if ospf_data.default is defined %}
+{%   set dfd = ospf_data.default %}
+redistribute default set {{ '{' }} metric {{ dfd.cost or '100' }} type {{ dfd.type.replace('e','') or '1' }} {{ '}' }}
+{% endif %}
+
+{% for a in areas %}
+area {{ a }} {{ '{' }}
+{%   for l in intf_data if l.ospf.area is defined and l.ospf.area is eq a and 'ipv4' in l %}
+{%     if l.ifname == "lo0" %}
+    interface {{ l.ifname }}:{{ l.ipv4 | ipaddr('address') }} {{ '{ '}} 
+{%     else %}
+    interface {{ l.ifname }} {{ '{ '}} 
+{%     endif %}
+{%     if l.ospf.network_type is defined and l.ospf.network_type == 'point-to-point' %}
+        type p2p
+{%     endif %}
+{%     if l.ospf.passive|default(False) %}
+        passive
+{%     endif %}
+{%     if l.ospf.cost is defined %}
+        metric {{ l.ospf.cost }}
+{%     endif %}
+{%     if l.ospf.timers.hello is defined %}
+        hello-interval {{ l.ospf.timers.hello }}
+{%     endif %}
+{%     if l.ospf.timers.dead is defined %}
+        router-dead-time {{ l.ospf.timers.dead }}
+{%     endif %}
+{%     if l.ospf.priority is defined %}
+        router-priority {{ l.ospf.priority }}
+{%     endif %}
+{%     if l.ospf.password is defined %}
+        auth-key {{ l.ospf.password }}
+        auth-type simple
+{%     endif %}
+    {{ '}' }}
+{%   endfor -%} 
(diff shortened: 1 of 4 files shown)


ipspace:
> I had to change the role from 'host' to 'router'. Overwriting it in the topology did not work (ospf templates were not evalutated).

Topology-level modules are copied into **router** nodes that do not have a **module** attribute. That's not done for **host** nodes because it's assumed hosts usually just do IP routing.

We would have to set **role: router** on **dut** device in all OSPF integration tests, but that doesn't make much sense. Once a node supports a routing protocol, it's a router ;) Setting **role** on OpenBSD device is a much better solution.

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
```diff
@@ -0,0 +1,55 @@
...
+
+cat <<OSPFD_CONF >/etc/ospfd.conf
+{% if ospf_data.router_id|ipv4 %}
+router-id {{ ospf_data.router_id }}
+{% endif %}
+spf-delay msec 100
+spf-holdtime msec 200
+{% if ospf_data.default is defined %}
+{%   set dfd = ospf_data.default %}
+redistribute default set {{ '{' }} metric {{ dfd.cost or '100' }} type {{ dfd.type.replace('e','') or '1' }} {{ '}' }}
```
ipspace: You can use { or } as simple text, no need to make an expression out of it. Jinja2 reacts just to {{ / }} and {% / %}

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
```diff
@@ -0,0 +1,55 @@
...
+{% endif %}
+
+{% for a in areas %}
+area {{ a }} {{ '{' }}
+{%   for l in intf_data if l.ospf.area is defined and l.ospf.area is eq a and 'ipv4' in l %}
+{%     if l.ifname == "lo0" %}
```
ipspace, line 19: Is this specific to "lo0" or all loopback interfaces? If it applies to all loopback interfaces, I'd go with `if l.type == "loopback"`

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
```diff
@@ -0,0 +1,55 @@
...
+    interface {{ l.ifname }}:{{ l.ipv4 | ipaddr('address') }} {{ '{ '}} 
+{%     else %}
+    interface {{ l.ifname }} {{ '{ '}} 
+{%     endif %}
+{%     if l.ospf.network_type is defined and l.ospf.network_type == 'point-to-point' %}
```
ipspace, line 24: Does OpenBSD support just "network" and "point-to-point"? We also have "point-to-multipoint" (but it's rarely implemented and thus not tested anywhere).

If OpenBSD supports "point-to-multipoint" then you might want to use a keyword map similar to what Junos does https://github.com/ipspace/netlab/blob/4e5f49f68f63d9d4bc9e50058c964a45fab0fe4b/netsim/ansible/templates/ospf/junos.macro.j2#L2

You can also define the keyword map in device **group_vars** if you hate having hard-coded stuff in templates.

netsim/ansible/templates/ospf/openbsd.ospfv3.j2
```diff
@@ -0,0 +1,47 @@
...
+
+cat <<OSPFD_CONF >/etc/ospf6d.conf
+{% if ospf_data.router_id is defined %}
+router-id {{ ospf_data.router_id }}
+{% endif %}
+spf-delay 1
+spf-holdtime 1
+{% if ospf_data.default is defined %}
+{%   set dfd = ospf_data.default %}
+redistribute default set {{ '{' }} metric {{ dfd.cost or '100' }} type {{ dfd.type.replace('e','') or '1' }} {{ '}' }}
```
ipspace: Default cost could be undefined. It would be better to use "dfd.cost|default(100)". Same for dfd.type

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
ipspace: Also, see the comment on using `|default(x)` in OSPFv3 template

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
remilocherer: This is specific to lo0. Otherwise 127.0.0.1 would be advertised ...

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
remilocherer: makes sense. 
I can adapt it towards the end of the week. Feel free to adapt if you have the time earlier.

netsim/ansible/templates/ospf/openbsd.ospfv2.j2
remilocherer: nbma is not in the OpenBSD OSPF implementation.

netsim/ansible/templates/ospf/openbsd.ospfv3.j2
remilocherer: makes sense. thanks for catching this!

PR #2703, opened 2025-09-22
FortiOS package update

a-v-popov:
This pull request introduces two changes:

## Add OpenStack cloud-init ISO
https://docs.fortinet.com/document/fortigate-private-cloud/7.6.0/kvm-administration-guide/767662/cloud-init
https://docs.fortinet.com/document/fortigate/7.6.4/administration-guide/817099/cloud-init

The user_data file is a config script, i.e. it is amending the FortiOS default factory configuration.
The exact content of this file should be discussed, e.g LLDP and other global configuration.

## Remove the `vagrant` user
From vagrant docs:
> If you are creating a base box for private use, you should try not to follow these, as they open up your base box to security risks (known users, passwords, private keys, etc.).
https://developer.hashicorp.com/vagrant/docs/boxes/base#default-user-settings

Vagrant is instead configured to use whatever user/password is configured for Ansible.

The pull request is a draft as at minimum the documentation should be updated as well, if the overall idea is acceptable.

### DIFF
--- a/netsim/templates/provider/libvirt/fortios-domain.j2
+++ b/netsim/templates/provider/libvirt/fortios-domain.j2
@@ -1,7 +1,7 @@
     {{ name }}.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
     {{ name }}.ssh.insert_key = false
-    {{ name }}.ssh.username = "{{ nodes[name].ansible_user }}"
-    {{ name }}.ssh.password = "{{ nodes[name].ansible_password }}"
+    {{ name }}.ssh.username = "{{ n.ansible_user }}"
+    {{ name }}.ssh.password = "{{ n.ansible_ssh_pass }}"
     #{{ name }}.ssh.private_key_path = "~/.ssh/id_ed25519"
 
     {{ name }}.vm.provider :libvirt do |domain|
--- a/netsim/templates/provider/libvirt/fortios-domain.j2
+++ b/netsim/templates/provider/libvirt/fortios-domain.j2
@@ -1,8 +1,13 @@
     {{ name }}.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
     {{ name }}.ssh.insert_key = false
+    {{ name }}.ssh.username = "{{ nodes[name].ansible_user }}"
+    {{ name }}.ssh.password = "{{ nodes[name].ansible_password }}"
+    #{{ name }}.ssh.private_key_path = "~/.ssh/id_ed25519"
 
     {{ name }}.vm.provider :libvirt do |domain|
-      domain.disk_bus = 'ide'
+      domain.disk_bus = 'virtio'
       domain.cpus = 1
       domain.memory = 2048
+      domain.graphics_type = "none"
+      domain.video_type = "none"
     end
--- a/docs/labs/fortios.md
+++ b/docs/labs/fortios.md
@@ -18,11 +18,13 @@ FortiOS (FortiGate) is supported by the **netlab libvirt package** command. To b
 
 ## Initial Device Configuration
 
-You'll have to copy-paste initial device configuration during the box-building process. **netlab libvirt config fortios** command displays the build recipe:
+The initial device configuration is prepared on a CD-ROM image that is read by the device during the initial boot process. All you have to do is check the applied configuration and shut down the device.
+
+**netlab libvirt config fortios** command displays the build recipe:
 
 ```{eval-rst}
-.. include:: fortios.txt
-   :literal:
+.. literalinclude:: fortios.txt
+   :language: text
 ```
 
 ```{tip}
--- a/netsim/devices/fortios.yml
+++ b/netsim/devices/fortios.yml
@@ -12,6 +12,7 @@ libvirt:
   image: fortinet/fortios
   build: https://netlab.tools/labs/fortios/
   create_template: fortios.xml.j2
+  create_iso: fortios
 clab:
   node.kind: fortinet_fortigate
   interface.name: "eth{ifindex}"
--- a/netsim/install/libvirt/fortios.txt
+++ b/netsim/install/libvirt/fortios.txt
@@ -1,17 +1,26 @@
 Creating initial configuration for FortiGate 6.x/7.0/7.4/7.6
 ============================================================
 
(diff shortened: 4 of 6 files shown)


netsim/templates/provider/libvirt/fortios-domain.j2
```diff
@@ -1,8 +1,13 @@
     {{ name }}.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
     {{ name }}.ssh.insert_key = false
+    {{ name }}.ssh.username = "{{ nodes[name].ansible_user }}"
```
ipspace: If this works, then n.ansible_user should work as well. See https://github.com/ipspace/netlab/blob/327be239e87d1b229b7fbe1d8866936429e86e3f/netsim/templates/provider/libvirt/Vagrantfile.j2#L15

a-v-popov: It works, thank you.
Also Ansible has multiple ways to provide a password and in a dictionary I see both `ansible_password` and `ansible_ssh_pass` set. Should one be preferred other the other with netlab?

I would also appreciate any thoughts on what should and what shouldn't be part of vagrant box configuration fed via bootstrap ISO.

Do you possibly have any opinion on LLDP configuration? Should it be enabled globally and disabled on port1 in vagrant box configuration?

Finally, factory config on Fortigates in 7.6 seems to include DHCP on port1 and two FortiGuard DNS servers. Should these be explicitly configured, as per current instruction in `netlab libvirt config fortios`, left alone or unset? Does anybody care about removal of `fgfm` access for FortiManager on port1?
https://community.fortinet.com/t5/FortiGate/Technical-Tip-FMG-Access-is-disabled-on-interfaces/ta-p/333746

ipspace: > It works, thank you. Also Ansible has multiple ways to provide a password and in a dictionary I see both `ansible_password` and `ansible_ssh_pass` set. Should one be preferred other the other with netlab?

Another artifact of Ansible's "we've cut it three times and it's still too short" approach to feature development. Years ago, I knew what the difference was (apart from **ansible_ssh_pass** is only used for SSH and takes precedence).

As all other devices use **ansible_ssh_pass**, I'd suggest changing that in `devices/fortios.yml` for consistency.

> I would also appreciate any thoughts on what should and what shouldn't be part of vagrant box configuration fed via bootstrap ISO.

And this is where my competence stops ;) Will wait for @ssasso or @sdargoeuves to respond.

In principle, as it doesn't cost us anything, it would be nice to most of the "static" configuration in Vagrant box configuration. However, even if that overlaps with the installation tasklists, we can't touch those because of backward compatibility (unless we tell everyone to rebuild their FortiOS boxes)

> Do you possibly have any opinion on LLDP configuration? Should it be enabled globally and disabled on port1 in vagrant box configuration?

Whenever possible, we run LLDP on all interfaces but the management interface.

sdargoeuves:
Is this change intended to simplify the process of creating the FortiGate Vagrant box by providing the configuration via cloud-init?

From my perspective, the configuration aspect isn't the most challenging part. A simple copy/paste of a few lines of configuration quickly prepares you to complete the box creation. Once the box is created, Netlab handles the more complex configurations.

In my use case, after Netlab completes its tasks (creating the VM, configuring OSPF, applying security policies), I run a couple of playbooks to create additional rules and configure management settings (SNMP, NTP, syslog). However, I'm not looking to have this as part of the box itself.

LLDP would be good for consistency, but if it's not included in the box, it could be configured during initialization.

I don't have access to a FortiManager, so I don't use ⁠`fgfm`; so if it was gone, I wouldn't cry about it!.

ipspace:
> From my perspective, the configuration aspect isn't the most challenging part. A simple copy/paste of a few lines of configuration quickly prepares you to complete the box creation.

Agreed, but it's always so nice when the Junos box auto-configures itself ;)

> LLDP would be good for consistency, but if it's not included in the box, it could be configured during initialization.

👍

a-v-popov:
> Is this change intended to simplify the process of creating the FortiGate Vagrant box by providing the configuration via cloud-init?

Yes. Plus it is removing vagrant user. And also switch FortiOS vagrantfile to virtio, because it doesn't make any sense to me to use different disk buses during vagrant box creation and in vagrantfile and PR is touching the latter anyway. I felt it would be OK to wrap all this into a single PR calling it "package update".

> From my perspective, the configuration aspect isn't the most challenging part. A simple copy/paste of a few lines of configuration quickly prepares you to complete the box creation. Once the box is created, Netlab handles the more complex configurations.

You can still copy paste whatever you have been copy pasting. As a matter of fact you can now even enjoy doing a factory reset to get a pristine box untainted by cloud-init.

> In my use case, after Netlab completes its tasks (creating the VM, configuring OSPF, applying security policies), I run a couple of playbooks to create additional rules and configure management settings (SNMP, NTP, syslog). However, I'm not looking to have this as part of the box itself.

This should be unaffected by the PR, unless you are relying on vagrant user explicitly.

> LLDP would be good for consistency, but if it's not included in the box, it could be configured during initialization.
> 
> I don't have access to a FortiManager, so I don't use ⁠`fgfm`; so if it was gone, I wouldn't cry about it!.

I guess I will only disable LLDP on port1, which on the one hand should be in line with vendor's decision, and on the other hand would leave anyone fancy the old config to do it the old way by hand.

a-v-popov:
I think I have incorporated the comments here and updated the recipe text. Please see if it suits your.

ipspace:
@ssasso @sdargoeuves -- could one of you please check this and approve/reject the PR? Thank you!

sdargoeuves:
I'll check and test this weekend, or early next week. Thanks for the
reminder, got distracted working with another fw vendor... More on this
soon!

sdargoeuves (approved):
I finally got it! I spent so much time wrapping my head around building vagrant boxes manually, I didn't realise (or maybe just forgot 😭) netlab can take that pain away by running the `netlab libvirt package forti.qcow2`
For my own understanding of vagrant, Vagrantfile, vagrant boxes... I need to spend more time on that part of the code.

I've tested the box creation, indeed it's super easy, this is great!!

I've also tested with one of my existing boxes on `7.6.3`, no impact, except the improvement of the disk_bus to `virtio` instead of `ide` 😄

ipspace:
Thank you both! Great job ;)

PR #2709, opened 2025-09-24
Fix: Warn about L3 attributes used on VLAN trunks

### LINKED ISSUE #1814: VLAN module should warn about removing L3 attributes from VLAN trunks
If you specify L3+ attributes (for example, OSPF cost) on a VLAN trunk link, they get silently removed. It would be nice to get a warning

> It could issue a warning about ignoring the ```bgp.local_as``` though

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/issues/1812#issuecomment-2591356480_

ipspace:
Closes #1814

### DIFF
--- a/tests/errors/vlan-trunk-l3-attr.log
+++ b/tests/errors/vlan-trunk-l3-attr.log
@@ -1,8 +1,8 @@
-ErrorAbort in vlan: Do not use attribute(s) ospf on VLAN trunk link links[1]
+ErrorAbort in vlan: Do not use attribute(s) ospf.* on VLAN trunk link links[1]
 ... Set defaults.vlan.warnings.trunk_attribute to False to disable this check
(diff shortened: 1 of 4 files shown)


tests/errors/vlan-trunk-l3-attr.log
```diff
@@ -0,0 +1,8 @@
+ErrorAbort in vlan: Do not use attribute(s) ospf on VLAN trunk link links[1]
```
jbemmel: 'ospf' is the module, 'ospf.cost' is the attribute. Perhaps the warning could be reworded to make this more clear?

ipspace: 'ospf' if the attribute value is non-box, 'ospf.*' if it's a Box?

PR #2711, opened 2025-09-26
Implement node rank in graph definition files

ipspace:
The 'graph.rank' node attribute allows you to group nodes into layers in GraphViz graphs, and to ensure the node positioning is more closely aligned with the network topology in D2 graphs (D2 does not have a shape rank concept)


ipspace:
@DanPartelly -- apologies for rushing ahead, but I need this functionality for tomorrow's blog post ;)

PR #2712, opened 2025-09-27
Remove intra-VRF EBGP sessions with other PE-devices

### LINKED ISSUE #2477: [BUG] In-VRF BGP sessions are established between EVPN PE-routers using EVPN-over-EBGP
In EVPN topologies using EBGP, the tenant-side BGP sessions are established between PE-routers (because they are in the same VLAN but in different BGP ASNs) unless the VLAN is in **bridge** mode or BGP is disabled in VRF.

I knew it was a bad idea supporting all the crazy vendor designs, but someone pushed really hard, so here we are 🤦‍♂️

### Lab topology

```
---
provider: clab
defaults.device: frr

module: [ vlan, vxlan, bgp, vrf, evpn ]
evpn.session: [ ebgp ]

vrfs.tenant.evpn.transit_vni: True

vlans:
  tenant:
    vrf: tenant
    links: [ r1-h1, r2-h2 ]

nodes:
  r1:
    bgp.as: 65000
  r2:
    bgp.as: 65001
  h1: { device: linux }
  h2: { device: linux }

links: [ r1-r2 ]
```

### Output

There is a BGP session between R1 and R2 in the tenant VRF:

```
$ netlab inspect --node r1 vrfs
%YAML 1.1
---
tenant:
  evpn:
    transit_vni: 200000
    evi: 1
    rd: 10.0.0.1:1
  id: 1
  rd: '65000:1'
  import:
  - '65000:1'
  export:
  - '65000:1'
  bgp:
    neighbors:
    - ifindex: 40000
      name: r2
      as: 65001
      type: ebgp
      ipv4: 172.16.0.2
      activate:
        ipv4: true
    import:
      connected:
        auto: true
  af:
    ipv4: true
  vrfidx: 100
```

That session shouldn't be there. The PE-routers should transport BGP (or any other) routes across EVPN AF.

### Workaround

Disable BGP in the tenant VRF:

```
vrfs.tenant:
  evpn.transit_vni: True
  bgp: False
```

Maybe that's good enough and just has to be documented.

ipspace:
An intra-VRF EBGP session should not be established between PE-devices running EVPN-over-EBGP when they're in the same EVPN instance.

This commit adds a 'remove VRF EBGP PE-device neighbors' function to the EVPN cleanup hook. That function performs a number of tests before deciding to remove an intra-VRF EBGP session:

* Both devices must run EBGP session in the same VRF. This preserves loopback EBGP sessions between VRFs
* Only EBGP sessions between EVPN-enabled devices are removed
* The two EVPN-enabled devices must be in the same evpn.domain -- this check enables multi-pod/multi-site designs
* The EBGP session must run over an EVPN-controlled SVI interface -- this check preserves the in-VRF EBGP sessions in Inter-AS Option A designs

Fixes #2477

### DIFF
--- a/netsim/modules/evpn.py
+++ b/netsim/modules/evpn.py
@@ -6,6 +6,7 @@ from .. import data
 from ..augment import devices
 from ..data.types import must_be_dict, must_be_int, must_be_string
 from ..utils import log
+from ..utils import routing as _rp_utils
 from . import _dataplane, _Module, _routing
 
 VALID_TRANSPORTS = [ 'vxlan', 'mpls' ] # In order of preference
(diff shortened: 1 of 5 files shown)


netsim/modules/evpn.py
```diff
@@ -447,6 +448,50 @@ def check_vlan_rt_values(node: Box, topology: Box) -> None:
...
+      continue
+
+    ngb_vlan = ngb_intf.get('vlan.name')                    # Get the VLAN associated with the SVI interface
+    if not ngb_vlan:                                        # Can't get it? Weird, but let's move on
+      continue
+
+    if ngb_vlan not in node.get('evpn.vlans',[]):           # Is this an EVPN-enabled VLAN
+      continue                                              # No? We may need the EBGP session
+
+    ngb._must_remove = True
```
jbemmel, line 488: Since this is a breaking change, we could put a warning here or at least debug tracing

ipspace:
> You often comment how annoying it is when vendors change the way their products behave without informing the users. In case there are labs that were depending on ebgp sessions that are now removed, a warning would be nice

I hope I made the "remove this session" conditions tight enough to affect only the sessions you REALLY wouldn't want to have, but I agree with you. Will add a warning.

PR #2713, opened 2025-09-30
Expose 'graph.format' and 'd2.format' node/link attributes

ipspace:
The 'graph.format' and 'd2.format' attributes were already used in the graph generation, but could not be used because the attribute definitions were missing.

Also:
* Streamline D2 attribute processing to merge object-type settings, object settings, and object format for a consistent behavior with strict format/settings/type precedence.
* Update 'platform-integration/graph/topo.yml' topology to include node and link formats for GraphViz and D2
* Rewrite/streamline large parts of graph/d2 documentation
* Add word-break CSS properties to inline code spans trying to prevent line breaks at hyphens (hint: it does not work with Chrome)


DanPartelly (approved):
Thank you.

Issue #2659, opened 2025-09-05
Implement IGP parameters on VRF loopbacks (similar to #2656)


ipspace:
Loopbacks in VRFs next?

_Originally posted by @jbemmel in https://github.com/ipspace/netlab/pull/2656#pullrequestreview-3186339733_

DanPartelly:
Maybe EBGP across loopbacks as well.

ipspace:
> Maybe EBGP across loopbacks as well.

This? https://netlab.tools/plugins/ebgp.multihop/

PR #2714, opened 2025-09-30
Implement select IGP parameters on VRF loopbacks

ipspace:
Using the infrastructure created in #2656, the VRF module validation of the loopback interface data allows module-specific loopback parameters.

The extra loopback VRF parameters are copied into the interface data when the VRF module creates VRF loopback interfaces in the node data.

Also:
* Add 'bgp.advertise' as a valid loopback parameter as we have no VRF equivalent of 'bgp.advertise_loopback' node parameter.
* Add VRF loopback OSPF area to a transformation test
* Create a new error test that checks the VRF loopback validation code

Implements #2659

### DIFF
--- a/tests/topology/expected/vrf-igp.yml
+++ b/tests/topology/expected/vrf-igp.yml
@@ -457,7 +457,7 @@ nodes:
         node: pe1
       type: p2p
     - _parent_intf: Loopback1
-      _parent_ipv4: 10.2.0.2/32
+      _parent_ipv4: 10.11.0.1/32
       _parent_vrf: blue
       ifindex: 2
       ifname: eth2
@@ -488,7 +488,8 @@ nodes:
         advertise: true
       ifindex: 10001
       ifname: Loopback1
-      ipv4: 10.2.0.2/32
+      ipv4: 10.11.0.1/32
+      ipv6: 2001:db8:cafe::1/128
       name: VRF Loopback blue
       neighbors: []
       type: loopback
@@ -557,7 +558,7 @@ nodes:
           instance: blue
           interfaces:
           - _parent_intf: Loopback1
-            _parent_ipv4: 10.2.0.2/32
+            _parent_ipv4: 10.11.0.1/32
             _parent_vrf: blue
             ifindex: 2
             ifname: eth2
@@ -594,7 +595,8 @@ nodes:
               advertise: true
             ifindex: 10001
             ifname: Loopback1
-            ipv4: 10.2.0.2/32
+            ipv4: 10.11.0.1/32
+            ipv6: 2001:db8:cafe::1/128
             isis:
               passive: false
             name: VRF Loopback blue
@@ -605,11 +607,15 @@ nodes:
           net: 49.0001.0000.0000.0002.00
           system_id: 0000.0000.0002
           type: level-2
-        loopback: true
+        loopback:
+          ipv4: 10.11.0.1/32
+          ipv6: 2001:db8:cafe::1/128
         loopback_address:
-          ipv4: 10.2.0.2/32
+          ipv4: 10.11.0.1/32
+          ipv6: 2001:db8:cafe::1/128
         networks:
-        - ipv4: 10.2.0.2/32
+        - ipv4: 10.11.0.1/32
+          ipv6: 2001:db8:cafe::1/128
         ospf:
           active: true
           af:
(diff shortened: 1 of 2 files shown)


netsim/modules/vrf.py
```diff
@@ -94,16 +94,44 @@ def get_next_vrf_id(asn: str) -> typing.Tuple[int,str]:
...
+* Global VRF loopback definition MUST NOT have ipv4/ipv6 values
+"""
+def validate_vrf_loopback(vname: str,vdata: Box,obj: Box,topology: Box) -> None:
+    lb = vdata.loopback
+    if not isinstance(lb,Box):
+      return
+    if obj is topology and ('ipv4' in lb or 'ipv6' in lb):
+      log.error(f'The "loopback" attribute in the global VRF "{vname}" cannot specify IPv4/IPv6 prefixes',
+      more_hints=[ 'Use pool attribute or specify specific IP address for a VRF loopback interface in the node VRF data'],
+      category=log.IncorrectType,
```
jbemmel: IncorrectAttr?

netsim/modules/vrf.py
```diff
@@ -343,25 +366,24 @@ def vrf_loopbacks(node : Box, topology: Box) -> None:
     lb_path = f'nodes.{node.name}.vrfs.{vrfname}.loopback'
     lb_pool: typing.Optional[str] = 'vrf_loopback'
     if isinstance(vrf_loopback,Box):
+      ifdata += { k:v for k,v in vrf_loopback.items() if k not in ['ipv4','ipv6','pool'] }
```
jbemmel: Elsewhere in the code these kinds of lists are put in YAML files. I guess you're not expecting this to change much?

netsim/modules/vrf.py
```diff
@@ -94,16 +94,44 @@ def get_next_vrf_id(asn: str) -> typing.Tuple[int,str]:
...
+* Skip the validation check if the loopback instance is not a box (must be bool)
+* Check against 'loopback' data type
+* Global VRF loopback definition MUST NOT have ipv4/ipv6 values
+"""
+def validate_vrf_loopback(vname: str,vdata: Box,obj: Box,topology: Box) -> None:
+    lb = vdata.loopback
+    if not isinstance(lb,Box):
+      return
+    if obj is topology and ('ipv4' in lb or 'ipv6' in lb):
+      log.error(f'The "loopback" attribute in the global VRF "{vname}" cannot specify IPv4/IPv6 prefixes',
```
jbemmel: IPv4/IPv6 prefix for a global VRF loopback can only be specified at the individual node level

netsim/modules/vrf.yml
```diff
@@ -28,9 +28,7 @@ _top:                               # Modification of global defaults
       id: { type: int, min_value: 1 }
       links: list
       loopback:
-        ipv4: { type: ipv4, use: host_prefix }
```
jbemmel, line 31: Where does the host_prefix check get performed now?

tests/errors/vrf-invalid-lb.yml
```diff
@@ -0,0 +1,28 @@
...
+      ospf.cost: 13
+  yellow:
+    loopback: 192.168.100.0/24
+
+nodes:
+  r1:
+    vrfs:
+      blue:
+        loopback:
+          bgp.advertise: False
```
jbemmel, line 26: - Should add a node-level VRF loopback ipv4/v6 address
- Could include a subnet prefix error case

jbemmel (changes requested):
Couple of minor items

netsim/modules/vrf.py
ipspace: I understand your reasoning, but that makes the error message and the hint a bit duplicate. Tried to find some middle ground (and I'm pretty sure I failed, but the intent should be obvious enough).

netsim/modules/vrf.py
ipspace: For the moment, that list is identical to the (global) loopback attributes, so of course I can go with that ;)

netsim/modules/vrf.yml
ipspace: https://github.com/ipspace/netlab/blob/302daf20ba8c925542a01d0ccb0b527d7bf1e2e8/netsim/modules/vrf.py#L115

tests/errors/vrf-invalid-lb.yml
ipspace: > Should add a node-level VRF loopback ipv4/v6 address

Added to the "vrf-igp" transformation test case that has it. This one is focused on "what can go wrong"

> Could include a subnet prefix error case

The "vrf-igp" transformation test case checks that we can use /32 on IPv4 loopbacks and /128 on IPv6 loopbacks. There are other tests that check the generic prefix validation logic.

ipspace:
> Couple of minor items

Thank you! Fixed.

Issue #2244, opened 2025-05-07
Using "set" action with "delete" option to delete BGP communities is silly ;)


ipspace:
Blame me -- I was blindly copying the ideas from IOS-like route maps.

Anyway, we REALLY SHOULD (RFC 6919) have a **delete** keyword that would be used to remove the communities (value- or clist-based).

Obviously, we'd have to translate the **set.community.delete** into **delete** because (unlike some vendor product managers) we care about backward compatibility and seamless migration.

Issue #2243, opened 2025-05-07
Community lists for extended and large communities


ipspace:
The **routing.community** object currently supports only standard BGP communities. It would be nice to have community lists that would match extended or large communities.

Unfortunately, that probably means we'll have to restructure the **routing.community** data format to add **type** (or something) top-level attribute that specifies which type of community list we're describing.

ipspace:
@jbemmel @ssasso @DanPartelly @Muddyblack @steffann @ddutt @nmodena -- any preference on how we'd structure the extended/large community lists?

I see three options:

### Per-entry community type

```
routing.community:
  std_1:
  - action: permit
    list: [ 65000:1 ]
  - action: deny
    regexp: _65000:1.._
  ext_1:
  - action: permit
    list: [ rt:65000:1 ]
    type: extended
  - action: deny
    list: [ rt:65000:.* ]
    type: extended
```

In this case, we could potentially split a single community list with different **type** values into multiple device ACLs (potentially confusing). Alternatively, we would report an error if **type** does not match across entries.

### Specify community list type

```
routing.community:
  std_1:          # Simplified standard community list
  - action: permit
    list: [ 65000:1 ]
  ext_1:          # Extended community list
    type: extended
    value:
    - action: permit
      list: [ rt:65000:1 ]
```

Advantage: clean data model

Drawback: more verbose for large/extended community lists

### Guesswork

If the entries are "a:b", then it's a standard community list, if they're "a:b:c" then it's a large community list, if they're "rt:..." or "soo:..." then it's an extended community list.

Drawback: doesn't work well when the user starts using regular expressons :(

jbemmel:
I would prefer the per-entry community type, and by default I would like to see Netlab "guess" the correct type ( == more basic type that can support the requested entry ). If it cannot detect automatically (e.g. in case of regular expressions), ask the user to explicitly specify what is intended (and error if it's not matching)

Then, add a global parameter to always require explicit community types (default: False), for people who are trying to teach networking and want their students to think about this in more detail

steffann:
I like the “Specify community list type” style. Guessing might be a useful fallback, but I’m generally more the in “explicit is better than implicit” camp :)

jbemmel:
> I like the “Specify community list type” style. Guessing might be a useful fallback, but I’m generally more the in “explicit is better than implicit” camp :)

We could choose to set the default to always require explicit community typing. The problem with ```#2``` is that it doesn't allow for guessing

[...]

PR #2715, opened 2025-09-30
Split 'routing' module into object-specific submodules

ipspace:
The 'routing' module became way too large and implements several almost independent objects. This commit splits it into object-specific module in 'netsim/modules/routing' directory:

* Common functionality (__init__.py)
* Shared normalize/import functions (normalize.py)
* AS path access list (aspath.py)
* Community lists (clist.py)
* Prefix lists (prefix.py)
* Routing policies (policy.py)
* Static routes (static.py)


ipspace:
This was really just a split of a single file into numerous files plus the glue needed to tie them together.

However, I didn't want to start working on #2244 or #2243 before bringing the files down to a manageable size

jbemmel (approved):
Long overdue, thanks

PR #2724, opened 2025-10-06
Change the D2 styles to match Graphviz object colors

ipspace:
Also:
* Apply default formatting to nodes and edges


ipspace:
@DanPartelly -- just in case you have time to look at it (no worries if you're busy)

PR #2725, opened 2025-10-06
Apply different styling to VLAN access/trunk links in topology graphs

ipspace:
Also:

* Add description of '-t' and '-f' values in 'netlab graph' docs


ipspace:
This PR is a follow-up to #2724. It has to be rebased after #2724 is merged. Only 2a17e4a is relevant.

PR #2734, opened 2025-10-10
Allow 'mgmt: False' to configure nodes with no management network

jbemmel:
Useful for large scale topologies

Context: #2733 

Note: Could be modified to only allow this for the 'clab' provider, need to fix test errors


netsim/augment/nodes.py
```diff
@@ -104,6 +104,8 @@ def validate(topology: Box) -> None:
 Sets missing management interface names and MAC, IPv4, and IPv6 addresses from the mgmt pool
 """
 def augment_mgmt_if(node: Box, defaults: Box, addrs: typing.Optional[Box]) -> None:
+  if not node.mgmt:
```
ipspace, line 107: Would have to check how this gets called, but "node.mgmt is False" seems safer.

jbemmel:
Closing this as there is a better solution in #2733 (provider specific attribute, hopefully at the group level)

PR #2739, opened 2025-10-11
'netlab graph' command can use Graphviz/D2 to create the final image

ipspace:
When the 'output' parameter of the 'netlab graph' command has an image file type, the code tries to execute 'dot' or 'd2' after creating the graph description file.


ipspace:
First step in streamlining **netlab graph** command. Hope you like it @DanPartelly

PR #2741, opened 2025-10-12
Avoid superfluous EBGP sessions for bgp confederation members within same subas

### LINKED ISSUE #2740: [BUG] - BGP confederation creates duplicate peerings over sub AS and main AS
### Describe the bug

I am trying to build a BGP confederation lab and I am seeing duplicate BGP sessions. Each pair of neighbors forms two sessions, one in the sub AS and one in the main AS.

### To Reproduce

You can use the topology below

### Expected behavior

**What I expected**
* r2 should have two iBGP neighbors only, r1 and r3, inside sub AS 201
* No extra sessions to AS 200

**What I see**
* r2 has four BGP neighbors in total
* Two sessions to r1, one in 201 and one in 200
* Two sessions to r3, one in 201 and one in 200
* r3 shows the same pattern with r2

```
r2#show run | sec bgp
router bgp 201
 bgp router-id 10.0.0.2
 bgp log-neighbor-changes
 bgp confederation identifier 200
 bgp confederation peers 202 
 bgp nopeerup-delay cold-boot 1
 bgp nopeerup-delay user-initiated 1
 bgp update-delay 5
 no bgp default ipv4-unicast
 neighbor 10.0.0.1 remote-as 201
 neighbor 10.0.0.1 description r1
 neighbor 10.0.0.1 update-source Loopback0
 neighbor 10.0.0.3 remote-as 201
 neighbor 10.0.0.3 description r3
 neighbor 10.0.0.3 update-source Loopback0
 neighbor 172.16.1.5 remote-as 200
 neighbor 172.16.1.5 description r1
 neighbor 172.16.1.10 remote-as 200
 neighbor 172.16.1.10 description r3
 !
 address-family ipv4
  bgp scan-time 5
  neighbor 10.0.0.1 activate
  neighbor 10.0.0.1 send-community both
  neighbor 10.0.0.1 next-hop-self
  neighbor 10.0.0.3 activate
  neighbor 10.0.0.3 send-community both
  neighbor 10.0.0.3 next-hop-self
  neighbor 172.16.1.5 activate
  neighbor 172.16.1.5 send-community
  neighbor 172.16.1.5 advertisement-interval 0
  neighbor 172.16.1.10 activate
  neighbor 172.16.1.10 send-community
  neighbor 172.16.1.10 advertisement-interval 0
 exit-address-family
ip bgp-community new-format
```

### Lab topology

```
provider: clab
defaults.device: iol
defaults.devices.iol.clab.image: iol:17.12.01

addressing:
  mgmt:
    ipv4: 192.168.200.0/24
  p2p:
    ipv4: 172.16.1.0/24
    prefix: 30

nodes:
  - name: r1
    module: [ bgp, ospf ]
  - name: r2
    module: [ bgp, ospf ]
  - name: r3
    module: [ bgp, ospf ]
  - name: r4
    module: [ bgp, ospf ]
  - name: r5
    module: [ bgp, ospf ]
  - name: p1
    module: [ bgp]
    id: 11

bgp:
  advertise_loopback: false
  confederation:
    200:
      members: [ 201, 202 ]
  as_list:
    100:
      members: [ p1]
    201:
      members: [ r1, r2, r3 ]
    202:
      members: [ r4, r5 ]

links:
  - prefix: 100.10.10.0/24
    p1:
  - p1-r1
  - r1-r2
  - r2-r3
  - r4-r5
  - r1-r4
  - r3-r5
```

### Output

```
netlab create
[CREATED] provider configuration file: clab.yml
[CREATED] transformed topology dump in YAML format in netlab.snapshot.yml
[GROUPS]  group_vars for all
[GROUPS]  group_vars for modules
[GROUPS]  group_vars for iol
[HOSTS]   host_vars for r1
[HOSTS]   host_vars for r2
[HOSTS]   host_vars for r3
[HOSTS]   host_vars for r4
[HOSTS]   host_vars for r5
[HOSTS]   host_vars for p1
[CREATED] minimized Ansible inventory hosts.yml
[CREATED] Ansible configuration file: ansible.cfg
```

### Version

netlab version 25.07

### Additional context

Image

jbemmel:
Fixes #2740

### DIFF
--- a/tests/topology/expected/bgp-confederation.yml
+++ b/tests/topology/expected/bgp-confederation.yml
@@ -15,6 +15,7 @@ bgp:
   sessions:
     ipv4:
     - confed_ebgp
+    - ibgp
     ipv6:
     - ebgp
 groups:
@@ -141,6 +142,7 @@ message: 'Use this topology to test BGP confederations on IPv4 and IPv6 EBGP ses
 
   '
 module:
+- ospf
 - bgp
 name: input
 nodes:
@@ -190,6 +192,7 @@ nodes:
       sessions:
         ipv4:
         - confed_ebgp
+        - ibgp
         ipv6:
         - ebgp
     box: none
@@ -266,6 +269,21 @@ nodes:
       ipv4: true
       ipv6: true
       neighbors:
+      - _source_intf:
+          ifindex: 0
+          ifname: Loopback0
+          ipv4: 10.0.0.2/32
+          ipv6: 2001:db8:1:2::1/64
+          neighbors: []
+          type: loopback
+          virtual_interface: true
+        activate:
+          ipv4: true
+        as: 65002
+        ipv4: 10.0.0.3
+        name: r3
+        next_hop_self: ebgp
+        type: ibgp
       - activate:
           ipv4: true
         as: 65001
@@ -279,6 +297,7 @@ nodes:
       sessions:
         ipv4:
         - confed_ebgp
+        - ibgp
         ipv6:
         - ebgp
     box: none
@@ -312,6 +331,10 @@ nodes:
         ipv4: 10.1.0.10/30
         ipv6: 2001:db8:3:2::3/64
         node: r3
+      ospf:
+        area: 0.0.0.0
+        network_type: point-to-point
+        passive: false
       type: lan
     loopback:
       ifindex: 0
@@ -319,6 +342,9 @@ nodes:
       ipv4: 10.0.0.2/32
       ipv6: 2001:db8:1:2::1/64
       neighbors: []
+      ospf:
+        area: 0.0.0.0
+        passive: false
       type: loopback
       virtual_interface: true
     mgmt:
@@ -326,8 +352,15 @@ nodes:
       ipv4: 192.168.121.102
       mac: ca:fe:00:02:00:00
     module:
+    - ospf
     - bgp
     name: r2
+    ospf:
+      af:
+        ipv4: true
+        ipv6: true
+      area: 0.0.0.0
+      router_id: 10.0.0.2
     provider: clab
   r3:
     af:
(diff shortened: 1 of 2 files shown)


ipspace:
I think the order of checks is still wrong:

* First, we have to figure out the "local" AS for both nodes
* Then, we have to figure out whether we're dealing with an IBGP session (in which case we leave), or something that looks like an EBGP session
* Only when we **know** we're dealing with an EBGP session should we start checking what the neighbor ASN should be.

FWIW, you fixed the gotcha that would better be fixed before trying to build BGP sessions: the local AS should not be in the list of confederation peers (which would fix #2740 but possibly still leave other edge cases open)

Finally, crazy stuff like an EBGP session that would turn into an IBGP session due to confederation AS should be outright rejected.

Am I missing something? Did I get it wrong?

jbemmel:
> Am I missing something? Did I get it wrong?

FWIW, I was trying to propose the minimal change that would fix the issue - having had challenges with over-zealous changes in the past.

That being said, I was wondering about the order of checks, and considered adding a check for ```bgp: False``` instead of relying on an exception for example.

If there is appetite to make more rigorous changes, we can discuss

jbemmel:
> the local AS should not be in the list of confederation peers

The problem isn't that the local AS is in the list of confederation peers, in fact adding it there could fix the issue too. The problem is that nodes with the same sub-AS are treated as being outside the confederation

ipspace:
> The problem isn't that the local AS is in the list of confederation peers, in fact adding it there could fix the issue too. The problem is that nodes with the same sub-AS are treated as being outside the confederation

Yeah, I've seen that now that I looked at the bigger picture. Too bad, that would have been an easy fix 🤷‍♂️ and apologies for wasting your time.

ipspace:
> FWIW, I was trying to propose the minimal change that would fix the issue - having had challenges with over-zealous changes in the past.

👍 Thank you!

I don't think either of us has the willpower to go through all the weird edge cases, so let's merge this and call it a day. It covers the most common use case that we got wrong (IBGP within confederation), and most other things I could think of would involve some crazy use of **local_as** within the confederation.

OK?

Issue #2736, opened 2025-10-10
The default 'all' group applies only for Ansible variables


jbemmel:
https://github.com/ipspace/netlab/blob/437fabba2d959dc15c364964d22b7c7d5a4bd4ee/docs/groups.md?plain=1#L253

Based on this, one would expect that all nodes are automatically a member of the 'all' group, and something like this would work:
```
defaults.device: frr

provider: clab

groups:
  all:
    clab.network-mode: none

nodes:
  n1:
```

Unfortunately, it doesn't

ipspace:
> The default 'all' group applies only for Ansible variables

FAD (Functions as Designed) 😜. The same is true for groups with device names.

The workaround is trivial now that you can use wildcards in group members.

jbemmel:
So you're saying all groups are groups, but some are more groups than others?

PR #2743, opened 2025-10-13
[B] Device-based groups should not have explicit members

ipspace:
Like with the 'all' group, we should not be able to specify members of groups that are based on device names as those groups are populated in Ansible output module.


ipspace:
Avoiding another potential problem before fixing #2736

PR #2745, opened 2025-10-14
Set node attributes from device/all groups

ipspace:
The device/all groups are not allowed to have static members. It was thus impossible to set node attributes in a device/all group.

This change adds dynamic calculation of group members for device/all groups. Making that work reliably required several other changes:

* The 'set node device/module' function has to apply group values in two passes. It has to apply group.device in the first pass and group.module in the second pass because changing node.device could change group membership for a device group.
* Computing members of a BGP autogroup now takes into account whether the global BGP module will be propagated into a node during the module initialization.
* The 'should we propagate modules' check was refactored into a separate function because it's used during module initialization as well as during BGP autogroup creation.
* node.device could be unspecified when BGP autogroup function calls the 'propagate_global_modules' check, but we know the only possible value that could be applied later on is the default device (because group-based devices were already set).

### DIFF
--- a/netsim/augment/groups.py
+++ b/netsim/augment/groups.py
@@ -15,8 +15,20 @@ from .. import data, modules
 from ..data import get_box
 from ..data.types import must_be_dict, must_be_id, must_be_list, transform_asdot
 from ..data.validate import get_object_attributes, validate_attributes
-from ..modules import bgp
+from ..modules import bgp, propagate_global_modules
 from ..utils import log
+
+'''
+Return computed members of special-purpose node group ('all' or device)
+'''
+def special_node_group_members(group: str, topology: Box) -> typing.Optional[list]:
+  if group == 'all':
+    return list(topology.nodes)
+
+  if group in topology.defaults.devices:
+    return [ node for node,ndata in topology.nodes.items() if ndata.get('device',None) == group ]
+
+  return None
 
 '''
 Return members of the specified group. Recurse through child groups if needed
@@ -37,10 +49,16 @@ def group_members(topology: Box, group: str, grp_type: str = 'node', count: int 
       header=True)
 
   gdata = topology.groups[group]
-  if gdata.get('type','node') != grp_type:
+  g_type = gdata.get('type','node')
+  if g_type != grp_type:
     return []
 
-  for m in gdata.members:
+  if g_type == 'node':
+    sg_members = special_node_group_members(group,topology)
+    if not sg_members is None:
+      return sg_members
+
+  for m in gdata.get('members',[]):
     m_ns = grp_type + 's'
     if m in topology[m_ns]:
       members = members + [ m ]
@@ -235,6 +253,11 @@ def check_group_data_structure(
         text=f'{grp_namespace}group "{grp}" should not have explicit members',
         more_hints='Device names or "all" cannot be used as names for user-defined groups',
         category=log.IncorrectValue,
+        module='groups')
+    if grp in topology.defaults.devices and 'device' in gdata:
+      log.error(
+        text=f'Trying to change the device to {gdata.device} in device group {grp} is ridiculous',
+        category=log.IncorrectAttr,
         module='groups')
 
     must_be_list(gdata,'module',gpath,create_empty=False,module='groups',valid_values=sorted(list_of_modules))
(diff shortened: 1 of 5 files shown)


netsim/modules/__init__.py
```diff
@@ -141,6 +141,28 @@ def remove_module(node: Box, module: str, extra: list = []) -> None:
...
+This function is called during module initialization, but also very early in the
+transformation process during BGP group creation, so we cannot assume that
+node.device is set
+"""
+def propagate_global_modules(n: Box, topology: Box) -> bool:
+  if 'device' not in n:                                     # Device might not be set yet
+    if 'device' not in topology.defaults:                   # Do we have a default device?
+      return False                                          # ... we'll have a problem anyway, so we can return whatever
+
+    n = n + {'device': topology.defaults.device}            # Assume default device
```
jbemmel, line 156: n.device = topology.defaults.device

Why the dict merge?

netsim/augment/groups.py
```diff
@@ -452,39 +475,64 @@ def reverse_topsort(topology: Box) -> list:
...
+      continue                                              # This group is not interesting, move on
 
     if log.debug_active('groups'):
-      print(f'Setting device/module for group {grp}')
+      print(f'Setting device for group {grp}: {gdata.device}')
     g_members = group_members(topology,grp)
     if not g_members and not gdata.get('_default_group',False):
       log.error(
-        f'Cannot use "module" or "device" attribute on in group {grp} that has no direct or indirect members',
+        f'Cannot use "device" attribute in group {grp} that has no direct or indirect members',
```
jbemmel, line 489: Counter example: roundabout way to set n.device
```
groups:
  g_no_members: # Serves to set the device
    members: []
    device: frr

  g_sub:
    members: [ g_no_members, n ]
    module: [ bgp ]

nodes:
  n:
```

This error isn't new and I am fine with it, just wondering what goes wrong if device is set on an empty group

netsim/modules/__init__.py
ipspace: I don't want to set n.device as a side effect in an obscure place that nobody would look at if something does wrong. I could check whether the two functions that need n.device need anything else (maybe n.provider), but that could also change, so the safest way seemed to be to recreate the final data structure.

netsim/augment/groups.py
ipspace: You know what, as contrived as the example is, I can change that into a warning ;)

ipspace: Your "counterexample" would not set the device because the attributes of a group (at least device and module) are not inherited into parent groups. The "device" is only set on members of the group, and if there are none, nothing happens.

If anything, we should also do the same check with **node_data**.

netsim/modules/__init__.py
jbemmel: Unless I’m mistaken the 2 statements are equivalent, setting n.device is easier to read / find with grep

ipspace: > Unless I’m mistaken

Spending 30 seconds in interactive Python3 gives the answer:

```
>>> from box import Box
>>> a = Box({ 'a': 1, 'b': { 'c': 2, 'd': 3 }})
>>> a
Box({'a': 1, 'b': {'c': 2, 'd': 3}})
>>> n = a.b
>>> n = n + { 'e': 4 }
>>> n
Box({'c': 2, 'd': 3, 'e': 4})
>>> a
Box({'a': 1, 'b': {'c': 2, 'd': 3}})
```

When you assign a dictionary (or Box) to a variable, you really assign a pointer. The results of an addition of two boxes (or a box and a dictionary) is a new object, so n is assigned a pointer to a different object, and the original object (a) is not modified.

However, were you to modify an inner dictionary (for example, if 'c' were a dictionary value and you'd modify n.c.x) then a.b.c.x would also be modified because the addition does not do a deep copy. I learned this particular bit the very hard way 🤷‍♂️

jbemmel: If the intent is to get a deep copy, it might be good to mention that

Didn’t we have a deep copy helper somewhere?

ipspace: > If the intent is to get a deep copy, it might be good to mention that

It is not. The intent is exactly what the code says -- merge node data with an extra attribute

PR #2746, opened 2025-10-15
[Doc] Describe how the DHCP module builds DHCP pools

ipspace:


ipspace:
@a-v-popov -- does this describe what we discussed in #2727 in enough details?

PR #2751, opened 2025-10-20
Fix: files plugin refuses to create files outside of lab directory

### LINKED ISSUE #2750: [BUG] Check the paths of files create with "files" plugin
The **path** used in a **files** element could specify a file outside of the lab directory (including absolute file paths). We should check for that and not create files outside of the lab directory.

Example:

```
plugin: [ files ]
files:
  ../README.md: |
    Creating README file in parent directory
```

ipspace:
Fixes #2750

### DIFF
--- a/netsim/extra/files/plugin.py
+++ b/netsim/extra/files/plugin.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 
-from box import Box, BoxList
+from box import Box
 
 from netsim.data import append_to_list, filemaps
 from netsim.data.types import must_be_dict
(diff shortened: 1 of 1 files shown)


netsim/extra/files/plugin.py
```diff
@@ -117,8 +117,34 @@ def process_inline_config(topology: Box, o_type: str) -> None:
...
+def check_output_paths(topology: Box) -> None:
+  if 'files' not in topology:                         # We shouldn't be called, but who knows...
+    return
+
+  lab_dir = Path('.').resolve()
+  if not isinstance(topology.files,BoxList):          # Invalid data structure? No worries, data validation
+    return                                            # ... will thrown an error
+  
+  for f_entry in topology.files:                      # Now iterate over the specified files
+    if 'path' not in f_entry:                         # No path specified? Data validation will have
```
jbemmel: Wouldn’t it make sense to perform this check as part of data validation? I.e. if path is missing that’s one type of error, if it points outside the directory that’s another

Just not sure about the partial double checks

jbemmel (approved):
Just a minor question about code structure

netsim/extra/files/plugin.py
ipspace: > Wouldn’t it make sense to perform this check as part of data validation?

That would require having a "file" data type (or some such) with additional flags like "relative". It would make sense if we were using file names in numerous places.

> Just not sure about the partial double checks

Moved the check to post_transform hook when we know the data structure has been validated.

PR #2754, opened 2025-10-21
Fix: Improve the L3VPN AF activation checks

ipspace:
Previously, the L3VPN AFs were activated on all neighbors that matched the L3VPN session type (plus the check on local L3VPN AF activation).

This fix tightens that check -- the L3VPN AF is activated only if the L3VPN session type matches, the neighbor is using MPLS module, and the local and the neighbor mpls.vpn parameters enable the AF.


ipspace:
Found this when working on an equivalent of #2752 for L3VPN and decided to submit it as a separate bug fix. The L3VPN version of #2752 will be a separate PR once this one is merged.

PR #2757, opened 2025-10-22
Fix: Set next-hop-self in Junos BGP policies only for AF IPv4/IPv6

ipspace:
The current 'next-hop-self' policy changed next hop on all external routes, effectively destroying EVPN connectivity when a Junos router propagated EVPN routes

### DIFF
--- a/netsim/ansible/templates/bgp/junos.j2
+++ b/netsim/ansible/templates/bgp/junos.j2
@@ -1,4 +1,5 @@
 {% import "junos.macro.j2" as bgpcfg with context %}
+{% set JUNOS_AF = { 'ipv4': 'inet', 'ipv6': 'inet6' } %}
 
 routing-options {
   autonomous-system {{ bgp.as }};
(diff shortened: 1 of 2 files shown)


ipspace:
@ssasso -- I know you're busy (and I will run integration tests to be on the safe(r) side), but it would be nice if you could look at these changes. Thank you!

ipspace:
FWIW, the localas-ibgp test is failing with the "old" templates as well. We never ran the updated (more thorough) test on Junos.

ssasso:
next-hop-self part looks good.

could you please open a dedicated issue for the localas-ibgp failing test? so I can have a look at it at a certain point in time :) 

thanks!

ipspace:
> next-hop-self part looks good.

Thanks a million, merging (also: I assume there's no way to match two AFs in one term, right?)

> could you please open a dedicated issue for the localas-ibgp failing test? so I can have a look at it at a certain point in time :)

Of course. The test fails because we have more thorough checks of what a node with local-as-ibgp session should do (rr-client + next-hop-self-always on local-as IBGP session).

Will try to figure out a way to add that to your Junos policy framework.

PR #2760, opened 2025-10-26
Internal ipaddr/hwaddr Jinja2 filters

ipspace:
Ansible 12.0 (ansible core 2.19) changed the way it handles dictionary keys starting with underscores. That change broke numerous netlab templates, forcing us to stay with Ansible release 11.x.

This is the first change in the get-well process -- replacing imported Ansible filters with built-in ipaddr and hwaddr filters, making netlab core functionality completely independent of Ansible.


jbemmel (approved):
_ok

PR #2762, opened 2025-10-26
OpenBSD OSPF: add missing redistribution part

remilocherer:
I missed the redistribution part previously. 
I think we do not need to filter for the protocols since the supported ones are listed in device features.

While here: move the shebang so it only gets printed once.

### DIFF
--- a/netsim/ansible/templates/ospf/openbsd.j2
+++ b/netsim/ansible/templates/ospf/openbsd.j2
@@ -1,5 +1,6 @@
 {% import "openbsd.ospfv2.j2" as ospfv2 %}
 {% import "openbsd.ospfv3.j2" as ospfv3 %}
+#!/bin/sh
 {% if ospf.af.ipv4 is defined %}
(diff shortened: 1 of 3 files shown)


ipspace:
Deleting that code from openbsd.py was a bad idea as someone could set **role: host** and **loopback: True** in a lab topology.

remilocherer:
That 2nd commit should not have landed in the PR and is not related. Need to figure out how to remove it again from the PR.

ipspace (approved):
Thank you!

FYI: this works as long as you can only redistribute connected and static into OSPF. Once you decide to implement RIP-to-OSPF redistribution, you'll have to deal with the fact that the protocol name is "ripv2" in netlab and probably "rip" or "ripng" in OpenBSD.

ipspace:
Also, it would make my life a tiny bit easier if you'd make changes in different branches (= not in "dev"). No worries though :)

PR #2763, opened 2025-10-26
Junos: use BGP neighbor next_hop_self and rr_client attributes

ipspace:
The commit optimizes Junos BGP policies and implements rr_client on local-as-IBGP peers.

The junos.policy template generates four next-hop-self policies to cover all use cases (IPv4/IPv6, ebgp/always) and uses them as needed based on neighbor af/next_hop_self attributes.

As a further optimization, the neighbor policy chain is compared to the default IBGP policy chain to apply per-neighbor policy only when absolutely needed.

For further study:

* Generate only Junos policies that are actually used
* Clean up BGP policy chain generation

### DIFF
--- a/netsim/ansible/templates/vrf/junos.bgp.j2
+++ b/netsim/ansible/templates/vrf/junos.bgp.j2
@@ -70,7 +70,7 @@ routing-instances {
           type internal;
           export [ 
 {%     if bgp.next_hop_self|default(false) %}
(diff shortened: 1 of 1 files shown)


ipspace:
Running BGP, BGP policy, and VRF integration tests on vjunos-switch to test the changes.

@ssasso -- adding you just in case you're bored ;)

PR #2764, opened 2025-10-28
Fix 'when' conditions in Ansible playbooks/task lists

ipspace:
Ansible release 12 hates having non-bool expressions in 'when:' conditions, and we sloppily used 'when: string' all over the place.

This commit ensures every expression used in a 'when' condition is a true bool and uses 'when: x is string and x != ""' expression instead of 'when: x' (adding 'x is string' because the Ansible documentation is confusing enough that I don't trust what exactly they might do in a future release)

### DIFF
--- a/netsim/ansible/collect-configs.ansible
+++ b/netsim/ansible/collect-configs.ansible
@@ -31,7 +31,7 @@
 
   - include_tasks: "{{ collect_script }}"
     name: Collect device configurations
-    when: collect_script
+    when: collect_script is string and collect_script != ''
     args:
       apply:
         vars:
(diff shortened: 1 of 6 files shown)


DanPartelly (approved):
Looks good.

ipspace:
> Correct but cringeworthy

You're describing my usual Ansible experience 🤷‍♂️🤦‍♂️

sdktr:
Wont a suffix of | bool fix this in a cleaner way?

ipspace:
> Wont a suffix of | bool fix this in a cleaner way?

That was my first though, but then I landed on the "I don't think this word means what you think it means" island: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/bool_filter.html

The test to use would be "is truthy": https://docs.ansible.com/ansible/latest/collections/ansible/builtin/truthy_test.html

sdktr:
TIL!
Why no | truthy filter then?

ipspace:
> TIL! Why no | truthy filter then?

🤷‍♂️ Personal preference (that has absolutely nothing to do with me hating a certain product that made me rewrite my code way too many times)

PR #2765, opened 2025-10-28
Warn about Ansible 12.x installation instead of giving up immediately

ipspace:
Now that we decoupled netlab core from Ansible Jinja2 filters, we can move the "Ansible 12.x is special" warning from "we're giving up" initialization code to "there be dragons" warning in CLI commands that use Ansible playbooks.

### DIFF
--- a/netsim/cli/ansible.py
+++ b/netsim/cli/ansible.py
@@ -20,7 +20,7 @@ def check_version(fatal: bool = False) -> None:
       log.warning(
         text="You're using Ansible version 12.x or higher; netlab might not work correctly",
         more_hints = [
-          "Ansible core version 2.19 introduced breaking changes in templates and playboks",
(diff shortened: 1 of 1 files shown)


netsim/cli/ansible.py
```diff
@@ -13,6 +13,24 @@
...
 
 
+def check_version(fatal: bool = False) -> None:
+  try:
+    import ansible  # type: ignore
+    if ansible.__version__ >= '2.19':           # Ansible core 2.19 contains significant templating changes
+      log.warning(
+        text="You're using Ansible version 12.x or higher; netlab might not work correctly",
+        more_hints = [
+          "Ansible core version 2.19 introduced breaking changes in templates and playboks",
```
jbemmel: playbooks

ipspace: Thank you, fixed.

PR #2767, opened 2025-10-29
Fix: ensure that the 'indent' filter is not called on an empty string

ipspace:
I have no idea why a Jinja2 macro that generates no text returns 'None' when used in Ansible 12.x, but the 'None' value crashes the standard Jinja2 indent filter (but then, nothing Ansible-related surprises me anymore).

Anyway, the sane workaround seems to be to add a condition around every '{{ some_macro()|indent }}' expressions to ensure 'some_macro' is called (and the result piped through indent) only when it returns something.


ipspace:
@jbemmel @DanPartelly -- this one seems pretty straightforward, adding you as reviewers just in case you want to look at it.

PR #2771, opened 2025-10-29
SRv6 implementation on IOS-XE (IOL, IOLL2, Cat8000v)

ipspace:
Implemented features:

* SRv6 with IS-IS
* L3VPN (IPv4 and IPv6) over SRv6

Other changes:

* srv6.allocate_loopback default value is changed to 'False' It makes no sense to use a default value that a majority of supported devices cannot work with.
* Integration tests were adjusted to use a loopback pool that is outside of SRv6 SID


DanPartelly (approved):
Thank you for this one.  Definitely needs 17.16.01 for frr interop.

PR #2772, opened 2025-10-29
Calculate missing OSPF interface timers

ipspace:
This change sets the missing OSPF interface hello/dead interval to be four times lower/higher than the interval (timer) that was specified.

It also checks that the hello interval is lower than the dead interval.

The documentation was adjusted accordingly.

### DIFF
--- a/netsim/modules/ospf.py
+++ b/netsim/modules/ospf.py
@@ -81,19 +81,22 @@ Adjust interface hello/dead timers -- if only one of them is specified, the othe
 set to be four times higher/lower.
 """
 def adjust_interface_timers(node: Box) -> None:
-  for intf in node.get('interfaces',[]):          # Scan all interfaces (function is called before moving
-    timers = intf.get('ospf.timers',None)         # ... the VRF interfaces into vrfs data)
-    if not isinstance(timers,Box):                # If the interface timers are not a dictionary
-      continue                                    # ... we have nothing to do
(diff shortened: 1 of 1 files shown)


ipspace:
Asking for a review from a few people who might be interested in OSPF timers ;)

netsim/modules/ospf.py
```diff
@@ -76,10 +76,34 @@ def propagate_node_attributes(node: Box, topology: Box) -> None:
...
+    if not isinstance(timers,Box):                # If the interface timers are not a dictionary
+      continue                                    # ... we have nothing to do
+
+    if 'hello' in timers and 'dead' not in timers:
+      timers.dead = timers.hello * 4
+    
+    if 'dead' in timers and 'hello' not in timers and timers.dead >= 3:
+      timers.hello = round(timers.dead / 4)
+
+    if timers.hello >= timers.dead:
```
jbemmel: This would seem to cause ```timers.hello``` to be set to ```{}``` when only ```timers.dead``` is set to 1 or 2

Suggestion:
```
 if 'dead' in timers and 'hello' not in timers:
      timers.hello = round(timers.dead / 4) if timers.dead >= 3 else 1
```

ipspace: Thank you, will fix. Also, the dead timer cannot be lower than three anyway (due to schema min_value), so I can remove that check, but will add max(timer,1) to the code just in case.

jbemmel (changes requested):
see comment

PR #2774, opened 2025-10-30
Simplify reporting/graphing/inspecting workflow

ipspace:
The 'graph', 'report' and 'inspect' commands were taking their data from the pickled snapshot, requiring you to start a lab or doing 'netlab create' just to generate a graph or a report.

This change makes that workflow more streamlined: the input data is loaded from the pickle file (if exists), otherwise the 'topology.yml' file is transformed and used.

The CLI flags allow you to specify a lab instance to get the pickle file from, the name of the pickle file, or the name of the lab topology

### DIFF
--- a/netsim/cli/graph.py
+++ b/netsim/cli/graph.py
@@ -13,7 +13,7 @@ from box import Box
 from ..outputs import _TopologyOutput
 from ..utils import log, strings
 from ..utils import read as _read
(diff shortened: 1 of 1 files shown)


ipspace:
> For "netlab graph", there are no -q --quiet in the arg namespace.

And that is exactly the one thing I did NOT check. Thanks a million, will fix.

PR #2775, opened 2025-10-30
Set bgp.advertise on main loopback to simplify BGP config templates

ipspace:
This change makes it simpler to write BGP configuration templates. It includes a simple fix
to the BGP module, proof-of-concept templates (EOS, FRR, IOS), and tons of changes in
transformation test results.

There's no need to adapt any other configuration template at this moment; this just makes
it easier to create new templates.

### DIFF
--- a/netsim/ansible/templates/bgp/frr.j2
+++ b/netsim/ansible/templates/bgp/frr.j2
@@ -46,16 +46,20 @@ router bgp {{ bgp.as }}
  address-family {{ af }} unicast
 !
 {{   redistribute.config(bgp,af=af) }}
-{%   if loopback[af] is defined and bgp.advertise_loopback %}
+{%   for l in netlab_interfaces if l.bgp.advertise|default("") and l[af] is defined and not 'vrf' in l %}
+{%     if loop.first %}
 !
-  network {{ loopback[af]|ipaddr(0) }}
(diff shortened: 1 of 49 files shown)


netsim/ansible/templates/bgp/frr.j2
```diff
@@ -46,16 +46,20 @@ router bgp {{ bgp.as }}
  address-family {{ af }} unicast
 !
 {{   redistribute.config(bgp,af=af) }}
-{%   if loopback[af] is defined and bgp.advertise_loopback %}
+{%   for l in netlab_interfaces if l.bgp.advertise|default("") and l[af] is defined and not 'vrf' in l %}
```
jbemmel, line 49: Can vrf be “default”?

ipspace: Not with the current data model. The "default" VRF is forbidden in this setting:

https://github.com/ipspace/netlab/blob/b6d49aa94795f320174218925941818a0489b074/netsim/modules/vrf.yml#L17

... and checked here (details in #1512)

https://github.com/ipspace/netlab/blob/b6d49aa94795f320174218925941818a0489b074/netsim/modules/vrf.py#L148

If a device expects a "default" VRF, that needs to be handled in the device template.

PR #2782, opened 2025-10-31
Generic device readyness tests (ex: ssh) specified in netlab_ready

ipspace:
This change removes the per-device symlinks that check VM-in-container SSH server readiness with a framework that provides generic readiness checks:

* The checks to perform are defined in 'netlab_ready' group variable -- a list of checks to perform
* The checks listed in 'netlab_ready' variable have to exist as task lists in 'readiness-check' directory. The only test implemented at the moment is the 'SSH ready' test (ssh.yml, a copy of vm-clab-ssh-check task list)
* 'wait-for-ready' task list first executes the tests specified in the 'netlab_ready' variable and then any device-specific task lists


ipspace:
> So far so good, worked well with everything I tried, mostly cisco devices.

Thanks a million. I think I tried it on every affected device, but it's always nice to have a second set of eyes.

PR #2785, opened 2025-11-05
Generalize vjunos/vptx device-specific readiness script

ipspace:
This change replaces a device-specific readiness script testing the presence of a specific interface on vjunos-router, vjunos-switch, and vptx, with a generic Junos script testing for the presence of first interface used in netlab topology.


ipspace:
@ssasso -- just if you have too much time ;) I tried it on all three platforms I have and it worked like a charm.

PR #2786, opened 2025-11-07
ArubaCX: Check for static route nexthop data before using it

ipspace:
The static routing configuration template on ArubaCX assumed the next-hop-interface is set for VRF routes. In some cases, that resulted in Jinja2 errors. This change checks the presence of next-hop interface before using it.


ssasso (approved):
thanks!

PR #2790, opened 2025-11-07
Move CL "is NVUE ready" check into readiness-check task list

### LINKED ISSUE #2789: [BUG] Move CL NVUE "is nvued started" check into readiness-check
The CL NVUE deploy-config script is checking for presence of NVUE daemon. That's the old approach; any "are we ready" checks should be moved into readiness-check

ipspace:
The readiness check performed within deploy-config task list has been moved to its proper place (readiness-check task list)

Closes #2789

### DIFF
--- a/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
+++ b/netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
@@ -6,23 +6,6 @@
     dest: "{{ nvue_config_file }}"
     mode: 0644
 
-- block:
-#  - name: "Start NVUE daemon"
-#    command: systemctl start nvued
-#    become: true
-
-  - name: "Wait for nvued to start"
-    service_facts:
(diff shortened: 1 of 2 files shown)


netsim/ansible/tasks/deploy-config/cumulus_nvue.yml
```diff
@@ -6,23 +6,6 @@
...
-    service_facts:
-    until: services['nvued.service'].state == 'running'
-    retries: 15
-    delay: 5
-    no_log: True
-
-  when: nvue_service_running is not defined
-
-- set_fact:
-    nvue_service_running: 1
```
jbemmel, line 24: This is missing now?

ipspace:
> Either set the flag, or remove the when clause and simply check every time

The "is NVUE ready" check has been moved into the "readiness-check" task list which is only executed once per "netlab initial"

PR #2791, opened 2025-11-07
Change default Mikrotik RouterOS7 container release to 7.18.2

ipspace:
Also: use the new vrnetlab container naming scheme


ipspace:
@ssasso -- Mikrotik BGP configuration is another dumpster fire. The last release I could get to work was 7.18.2 (see also #2794), so I changed the container tag to that release.

PR #2793, opened 2025-11-08
fix SR Linux version detection

michelredondo:
Some tests were failing for SR Linux 25.10.1 because version detection logic considered the version as "25.1" (str to float).


jbemmel:
We could use tuples e.g. ‘’’if version < (5,3) …’’’

ipspace:
@michelredondo -- Thanks a million.

@jbemmel is right -- if we split the version into a list in the Python code, we can compare lists (at least in Python). Let me try this out...

ipspace:
@michelredondo -- pushed the changes that compare lists. It works for me, could you please check if it works for you. If it does, I'll merge the PR, change the SR Linux release to 25.10 in another PR, and run the full integration tests.

Thanks again!

michelredondo:
Looks good to me. Thanks!

PR #2813, opened 2025-11-16
Fix: Enable EVPN only on IPv4 or RFC 8950 BGP sessions

### LINKED ISSUE #2810: [BUG] EVPN AF should not be enabled over IPv6 BGP sessions
Enabling EVPN AF over IPv6 BGP sessions while we have IPv4-only VTEPs is useless. We should check the VTEP AF in the EVPN template and enable EVPN AF only in the relevant session.

---

I'd like to propose an enhancement to the Arista device template to support automatic EVPN IPv6-neighbor  configuration. Currently, when using EVPN with Arista, the template does not generate ipv6 BGP neighbor statements for EVPN peering.
I add this to netsim/ansible/templates/evpn/eos.j2 and seems ok

_Originally posted by @aeangel in https://github.com/ipspace/netlab/discussions/2809_

ipspace:
netlab currently does not support EVPN with IPv6 VTEPs, so we have to limit the activation of EVPN AFs only to BGP sessions that also carry IPv4 next hops.

The EVPN module has been modified to set 'evpn' flag on a BGP neighbor only when 'ipv4' is defined on that neighbor (to be changed to a more comprehensive test when we support VXLAN-over-IPv6). The flag is thus set for real IPv4 BGP sessions, unnumbered sessions (because IPv4 is set to True) and RFC 8950 sessions over IPv6 GUA addresses (IPv4 is also set to True in that case).

Several device configuration templates have been modified as needed to select the correct transport session to use for the EVPN AF.

Finally, the 'evpn-vxlan' transformation tests adds an IPv6-only network to verify that the EBGP session running over it is not carrying EVPN AF.

In a future patch, we migth change the 'evpn' attribute to 'ipv4', 'ipv6', or 'local_if' to make it easier for templates to select the correct neighbor attribute.

Closes #2810

### DIFF
--- a/netsim/ansible/templates/bgp/sros.gnmi.macro.j2
+++ b/netsim/ansible/templates/bgp/sros.gnmi.macro.j2
@@ -76,7 +76,9 @@
       ipv4: {{ act_ipv4 }}
       ipv6: {{ act_ipv6 }}
 {% if neighbor.evpn|default(False) and not act_ipv4 and not act_ipv6 %}
-      evpn: True    # EVPN-only neighbor, have to enable it here to have at least one AF active
+      evpn: True    # EVPN-only neighbor (for example, IBGP session in IBGP-over-EBGP design)
+                    # have to enable EVPN AF here to have at least one AF active otherwise the
+                    # neighbor cannot be configured
 {% endif %}
 {% endmacro %}
 
--- a/netsim/ansible/templates/evpn/sros.j2
+++ b/netsim/ansible/templates/evpn/sros.j2
@@ -37,7 +37,10 @@ updates:
 {% endfor %}
 {#
     Now enable the EVPN AF on individual neighbors. We can't do it for the EBGP group
-    because that enables EVPN AF on IPv4 and IPv6 EBGP sessions
+    because that enables EVPN AF on IPv4 and IPv6 EBGP sessions.
+
+    Also, note that the "do we need EVPN here" check is simpler because we don't
+    implement RFC 8950 sessions on SR-OS.
 #}
 {% for n in bgp.neighbors|default([]) if n.evpn|default(False) and n.ipv4|default(False) is string %}
 {%   if loop.first %}


ipspace:
@ssasso @jbemmel -- just in case you want to look at the templates.

The major change is that the neighbor.evpn attribute is only set if the session has the capability to carry IPv4 AF (IPv4 neighbor or IPv6 GUA/LLA session with RFC 8950), then I had to fix a bunch of templates.

Next steps: EOS has VXLAN-over-IPv6, so I can get three independent implementations, and if I can get EVPN-over-IPv6 work with FRR, I should have two independent EVPN-over-IPv6 implementations ;)

netsim/ansible/templates/bgp/sros.gnmi.macro.j2
```diff
@@ -70,12 +70,14 @@
...
+{% set act_ipv6 = activate.ipv6|default(False) and ipv6 %}
     family:
-      ipv4: {{ activate.ipv4|default(False) and ipv4 }}
-      ipv6: {{ activate.ipv6|default(False) and ipv6 }}
-{%  if 'evpn' in neighbor and neighbor.evpn %}
-      evpn: True # Must have at least 1 address family enabled
-{%  endif %}
+      ipv4: {{ act_ipv4 }}
+      ipv6: {{ act_ipv6 }}
+{% if neighbor.evpn|default(False) and not act_ipv4 and not act_ipv6 %}
```
jbemmel, line 78: AI review comment:
```
**Lines 78-80:**
```jinja2
{% if neighbor.evpn|default(False) and not act_ipv4 and not act_ipv6 %}
  evpn: True    # EVPN-only neighbor, have to enable it here to have at least one AF active
{% endif %}
```

**Issues:**
- ❌ **Logic Error**: This condition can never be true if the EVPN module is working correctly. If `evpn` is set on a neighbor, `ipv4` must be present (per evpn.py line 59-60), which means `act_ipv4` should be True.
- **Recommendation**: Remove this code block or add a comment explaining it's a safety check for edge cases.
```

File: netsim/ansible/templates/evpn/sros.j2 [hunk 2]
```diff
@@ -30,13 +30,24 @@ updates:
...
 {%   if bgp.rr|default(0) or type == 'ebgp' %}
         next-hop-unchanged:
           evpn: True
 {%   endif %}
 {% endfor %}
+{#
+    Now enable the EVPN AF on individual neighbors. We can't do it for the EBGP group
+    because that enables EVPN AF on IPv4 and IPv6 EBGP sessions
+#}
+{% for n in bgp.neighbors|default([]) if n.evpn|default(False) and n.ipv4|default(False) is string %}
```
jbemmel, line 45: Suggest to add a note that this 'is string' check is because unnumbered and rfc8950 sessions are not yet supported for SROS

netsim/ansible/templates/evpn/vjunos-switch.j2
```diff
@@ -17,16 +17,15 @@ protocols {
 {% endfor %}
     group ebgp-peers {
 {% for n in bgp.neighbors if n.type == 'ebgp' and n.evpn|default(false) %}
-{%   for af in ['ipv4','ipv6'] if n[af] is defined %}
-      neighbor {{ n[af] }} {
+{%   set peer = n.ipv4 if n.ipv4 is string else n.local_if|default(n.ipv6) %}
```
jbemmel, line 20: AI comment:
```
**Issues:**
- ✅ Correctly handles string IPv4 addresses
- ✅ Correctly falls back to `local_if` for unnumbered sessions
- ⚠️ **Potential Issue**: Falls back to `n.ipv6` if `local_if` is missing. However, if EVPN is enabled, `ipv4` must be present (either string or True), so this fallback should never be needed.
- **Recommendation**: Consider removing the IPv6 fallback or adding a comment that it's defensive programming.
```

netsim/ansible/templates/bgp/sros.gnmi.macro.j2
ipspace: And you just proved why using AI for code review is suboptimal 😜

The "edge" case is EVPN-only IPv4 BGP session used for your favorite design (IBGP over EBGP). I thought we wouldn't need this, but there's no other way to configure a BGP neighbor inside the BGP module otherwise.

netsim/ansible/templates/evpn/vjunos-switch.j2
ipspace: Wrong again. What AI thinks is defensive programming is really RFC 8950 session over IPv6 GUA.

The logic is explained in the FRR template. I could add a pointer to the FRR template, but then EVPN-over-IPv6 is pretty high on my list, so this will get changed anyway.

ipspace:
> The commit added an IPv6-only network test case

Going in the right direction, but incorrect wording.

> but we should also test

That's amazingly good understanding of what's going on ;) Will add something along these lines together with the EVPN-over-IPv6 test.

ipspace:
Oh, and thanks a million for an AI review example! While it's still at the overconfident intern level, it's doing a decent job on a pretty complex code base.

PR #2817, opened 2025-11-17
[B] Cat8000v: use the new bridge-domain config model

ipspace:


ipspace:
All VLAN tests pass, so it looks like the old and new configurations are equivalent. Any reservations @DanPartelly ?

DanPartelly:
Im getting some weird errors, even at deployment time. I need to see where those are coming from, as I see you passed all tests. This for example is 23-vlan-mixed-multiple. 

```
ASK [ios_config: deploying vlan from /home/dan/projects/netlab/netsim/ansible/templates/vlan/cat8000v.j2] **********************************************************************************
  module_stderr: |-
    member GigabitEthernet4 service-instance 701
    member GigabitEthernet4 service-instance 701
     ^
    % Invalid input detected at '^' marker.
  
    s1(config-bdomain)#
  module_stdout: ''
  msg: |-
    MODULE FAILURE: No start of json char found
    See stdout/stderr for the exact error

```

DanPartelly:
@ipspace what is the exact image you used for this one ?


Cause I do not have bridge-domain members on it:

```
Cisco IOS XE Software, Version 17.16.01a
Cisco IOS Software [IOSXE], Virtual XE Software (X86_64_LINUX_IOSD-UNIVERSALK9-M), Version 17.16.1a, RELEASE SOFTWARE (fc1)

cisco C8000V (VXE) processor (revision VXE) with 1655426K/3075K bytes of memory.
Processor board ID 9HYQQSVDRNK
Router operating mode: Autonomous
4 Gigabit Ethernet interfaces
32768K bytes of non-volatile configuration memory.
3960392K bytes of physical memory.
5234688K bytes of virtual hard disk at bootflash:.

Configuration register is 0x2102

s1(config-bdomain)#mem?
% Unrecognized command
s1(config-bdomain)#mem
s1(config-bdomain)#mem?
% Unrecognized command
s1(config-bdomain)#mem

```

ipspace:
I'm using release 17.16.01a, and as I suspected, they changed the configuration syntax. Unfortunately, only the new one works with VXLAN, but is rejected by older releases.

No problem, we just need a feature flag to select the desired configuration method, and then include the correct VLAN configuration template. Once we agree on that, I'll adjust this PR.

Now for the hard part: naming (covered in #2822)

DanPartelly:
Mine is a 17.16.01a  release as well,  as you can see from the printout . Shouldn't it  work with iol-like sintax ? Im a bit confused.

ipspace:
In a bizzare twist of events: It also works on 17.13.01a running c8000v as vrnetlab container, so it's no as release-specific as I was afraid it was.

```
s1#show ver
Cisco IOS XE Software, Version 17.13.01a
Cisco IOS Software [IOSXE], Virtual XE Software (X86_64_LINUX_IOSD-UNIVERSALK9-M), Version 17.13.1a, RELEASE SOFTWARE (fc3)
...
cisco C8000V (VXE) processor (revision VXE) with 2017252K/3075K bytes of memory.
Processor board ID 9IZFBD1ZP7B
Router operating mode: Autonomous
```

Configuration:
```
redundancy
bridge-domain 700
 member GigabitEthernet3 service-instance 700
 member GigabitEthernet4 service-instance 700
!
interface GigabitEthernet1
 description Containerlab management interface
 vrf forwarding clab-mgmt
 ip address 10.0.0.15 255.255.255.0
 negotiation auto
 ipv6 address 2001:DB8::2/64
 no lldp transmit
 no lldp receive
 no mop enabled
 no mop sysid
!
interface GigabitEthernet3
 description [Access VLAN red] s1 -> h1
 no ip address
 negotiation auto
 no mop enabled
 no mop sysid
 service instance 700 ethernet
  encapsulation untagged
 !
!
interface GigabitEthernet4
 description [Access VLAN red] s1 -> h2
 no ip address
 negotiation auto
 no mop enabled
 no mop sysid
 service instance 700 ethernet
  encapsulation untagged
```

Now we have to figure out what exactly is wrong on your end. I have no idea where to start...

DanPartelly:
> Now we have to figure out what exactly is wrong on your end. I have no idea where to start...

Oh well, we will find a way. Aint' no sunshine without bridge domains :P

ipspace:
> Oh well, we will find a way. Aint' no sunshine without bridge domains :P

Throwing spaghetti at the wall (because nothing else makes sense):

* Are you using a VM or a vrnetlab container?
* Do you create the Vagrant box using the "official" recipe or do you add something more to it?

ipspace:
Found the culprit (tangentially based on suggestions by @DanPartelly) -- it looks like the newer software release wants to have `license boot level network-premier addon dna-premier` to make the "members under bridge-domain" (but also VXLAN) configuration work. We add that to the Vagrant VM, but it's not in the vrnetlab VM.

DanPartelly:
I think @hellt should revert that change. It breaks tons of functionality (now only support for new config mode) , I beleive the image built with virtio will break a lot of vlan tests even with the old config style. Hopefully he is willing to do so.

ipspace:
> I think @hellt should revert that change. It breaks tons of functionality (now only support for new config mode) , I beleive the image built with virtio will break a lot of vlan tests even with the old config style. Hopefully he is willing to do so.

A) The missing config mode seems to be caused by missing premier license
B) If I got it right, we're also using virtio for the Vagrant VM, so it doesn't matter.

DanPartelly:
> > I think @hellt should revert that change. It breaks tons of functionality (now only support for new config mode) , I beleive the image built with virtio will break a lot of vlan tests even with the old config style. Hopefully he is willing to do so.
> 
> A) The missing config mode seems to be caused by missing premier license B) If I got it right, we're also using virtio for the Vagrant VM, so it doesn't matter.

Maybe run the ROS test with old config and 17.16.01 vrnetlab see, what is happening ?  Ill doit when i get home anyways.

ipspace:
> Maybe run the ROS test with old config and 17.16.01 vrnetlab see, what is happening?

I did, and it breaks (but works with vmxnet3). So yeah, the change should be reverted (https://github.com/srl-labs/vrnetlab/issues/414).

However, I can't figure out why subinterfaces work with Vagrant VM which also uses virtio NICs.

ipspace:
Will open a new PR, this thing is too full of noise.

PR #2819, opened 2025-11-17
add tools/ to MANIFEST.in

ubaumann:
I found out the templates were missing when I installed it on a fresh server


ipspace:
Thank you!

PR #2821, opened 2025-11-18
STP: Fix node level STP protocol override setting

### LINKED ISSUE #2820: [BUG] STP Node-Level Protocol Override Not Validated
## Summary

The STP module validates the global `stp.protocol` setting against device capabilities, but does not validate node-level protocol overrides. This allows users to configure unsupported STP protocols on individual nodes without receiving an error.

## Environment

- **netlab version**: Current (as of branch `fix_minor_stp_issues`)
- **Module**: `stp`
- **Affected devices**: All devices with STP support

## Description

When a user sets `stp.protocol` at the node level (e.g., `nodes.s1.stp.protocol: mstp`), the STP module does not validate whether the device supports that protocol. The module only validates the global `stp.protocol` setting.

### Current Code Behavior

In `netsim/modules/stp.py`, lines 41-47:

```python
protocol = topology.get("stp.protocol","stp")
supported_protocols = features.get("stp.supported_protocols",[])
if protocol not in supported_protocols:
    log.error(
        f'node {node.name} (device {node.device}) does not support requested STP protocol ({protocol})',
        log.IncorrectValue,
        'stp')
```

This only checks the global protocol, not node-level overrides.

## Steps to Reproduce

1. Create a topology with a device that has limited STP protocol support (e.g., FRR which only supports `stp`)
2. Set a node-level protocol override to an unsupported protocol (e.g., `mstp`)
3. Run `netlab create` on the topology

### Minimal Test Topology

```yaml
---
defaults:
  device: frr

stp:
  protocol: stp  # Global default

addressing:
  p2p:
    ipv4: False  # STP only applies to L2 interfaces

groups:
  _auto_create: true
  switches:
    members: [s1, s2]
    module: [vlan, stp]

vlans:
  red:
    mode: bridge
    links: [s1-s2]

links:
- s1:
  s2:

nodes:
  s1:
    # Node-level protocol override - should be validated but currently isn't
    stp.protocol: mstp  # FRR doesn't support MSTP, should error
```

**File location**: `tests/ai/stp-validation-issues.yml`

## Expected Behavior

When running `netlab create` on the above topology, it should produce an error:

```
IncorrectValue in stp: node s1 (device frr) does not support requested STP protocol (mstp)
```

## Actual Behavior

The topology is created successfully without any validation errors, even though:
- FRR device only supports `stp` protocol (see `netsim/devices/frr.yml` line 138: `supported_protocols: [ stp ]`)
- Node `s1` is configured with `stp.protocol: mstp` which is not supported

## Impact

- **Severity**: Medium
- **Impact**: Users can configure invalid STP protocols on nodes, leading to:
  - Configuration generation that may not work as expected
  - Runtime errors when the lab is started
  - Confusion about why certain configurations don't work

## Suggested Fix

The validation should check node-level protocol overrides in addition to the global protocol. Suggested code change in `netsim/modules/stp.py`:

```python
# Check global protocol
protocol = topology.get("stp.protocol","stp")
supported_protocols = features.get("stp.supported_protocols",[])
if protocol not in supported_protocols:
    log.error(
        f'node {node.name} (device {node.device}) does not support requested STP protocol ({protocol})',
        log.IncorrectValue,
        'stp')

# Check node-level protocol override
node_protocol = node.get('stp.protocol', protocol)
if node_protocol not in supported_protocols:
    log.error(
        f'node {node.name} (device {node.device}) does not support requested STP protocol ({node_protocol})',
        log.IncorrectValue,
        'stp')
```

## Additional Context

- The global protocol validation works correctly
- Node-level priority validation works correctly (validates multiples of 4096)
- This is a missing validation check, not a logic error
- Related issue: The `protocol` attribute in `stp.yml` node section is missing a definition (should have `copy: global` or explicit type)

## References

- STP module: `netsim/modules/stp.py`
- STP defaults: `netsim/modules/stp.yml`
- FRR device features: `netsim/devices/frr.yml` (line 138)
- Test topology: `tests/ai/stp-validation-issues.yml`

jbemmel:
Fixes #2820

### DIFF
--- a/netsim/modules/stp.py
+++ b/netsim/modules/stp.py
@@ -92,10 +92,9 @@ class STP(_Module):
     # Check if per-VLAN priority is being used
     for vname,vdata in node.get('vlans',{}).items():
       if vdata.get('stp.priority',None):
-        stp_proto = node.stp.protocol
-        if stp_proto != 'pvrst':
+        if node.stp.protocol != 'pvrst':
           log.error(
-            f"Topology requires per-VLAN STP (pvrst) used on VLAN '{vname}' but global default is '{stp_proto}'",
+            f"Topology requires per-VLAN STP (pvrst) used on VLAN '{vname}' but global default is '{node.stp.protocol}'",
             log.IncorrectValue,
             'stp')
         elif not 'pvrst' in features.get('stp.supported_protocols',[]):


netsim/modules/stp.py
```diff
@@ -17,6 +17,13 @@ def supports_stp(intf: Box) -> bool:
     return False
   return True
 
+"""
+Get the effective STP protocol for a node, following the hierarchy:
+node-level override → topology-level → default 'stp'
+"""
+def get_effective_stp_protocol(node: Box, topology: Box) -> str:
```
ipspace: We have "modules.get_effective_module_attribute" function to get this job done (assuming you need it)

Admittedly, you'd have to add "or 'stp'" at the end as it can return None.

netsim/modules/stp.py
```diff
@@ -34,20 +41,20 @@ class STP(_Module):
 
   # Check stp.supported_protocols, stp.priority, stp.port_priority, per VLAN support and stp.enable_per_port
   def node_post_transform(self, node: Box, topology: Box) -> None:
-    if not node.get("stp.enable", True):   # if STP is disabled, don't complain about feature support
+    if not node.get('stp.enable', True):   # if STP is disabled, don't complain about feature support
       return
     features = devices.get_device_features(node,topology.defaults)
 
-    protocol = topology.get("stp.protocol","stp")
+    protocol = get_effective_stp_protocol(node,topology)
```
ipspace: Wouldn't it be easier to make "stp.protocol" inheritable? Then it would be copied into the node automatically.

ipspace (changes requested):
I think this can be made much simpler. Also, isn't **stp.protocol** already copied into node data?

netsim/modules/stp.py
jbemmel: Yes, I remembered (I asked it for similar patterns and it found the vlan module with interface - link - node - topology hierarchy)

Obviously this was just a toy experiment - I highly recommend to look into Cursor, this project is well suited to AI supported coding

netsim/modules/stp.py
jbemmel: AI found the missing copy:, will that do it?

I’ll look for the simpler way

ipspace (approved):
Looks good to me, thank you!

It does assume that the **defaults.stp.protocol** is always set, so it's propagated into topology and node data, but I'm OK with that.

PR #2826, opened 2025-11-19
Cat8000v: use the new VLAN config model with 'premier' license

ipspace:
This change introduces 'cat8000v.license' node attribute that is automatically set to 'premier' on Vagrant boxes but can be set by the user in system default or on individual nodes.

The initial impact of the 'cat8000v.license' attribute is the selection of VLAN configuration mode. With the 'premier' license, netlab configures Ethernet services as members of bridge domains. Without it, the old-style configuration (bridge-domain configured under 'service ethernet' is used).

The 'premier' license is also required for MPLS, SR-MPLS, and SRv6. These requirements are not yet checked (coming in a separate PR).

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -100,10 +100,21 @@ nodes:
 (caveats-cat8000v)=
 ## Cisco Catalyst 8000v
 
-* Apart from the VLAN configuration, Catalyst 8000v implementation uses the same configuration templates as CSR 1000v.
+Licensing:
+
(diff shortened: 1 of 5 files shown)


ipspace:
Replaces #2817

ipspace:
> I cannot test it physically

The VM tests all pass, the container tests pass (with old config syntax) apart from the subinterfaces ones. We'll deal with that later.

> because my vagrant stack is broken yet again and I have no mental energy left for dealing with vagrant.

And now you know how I feel about Ansible 🤦‍♂️🤢

ipspace:
> The code changes look OK to me.

Thank you, merging...

PR #2830, opened 2025-11-20
Cat8000v: VXLAN with static ingress replication

ipspace:


ipspace:
@DanPartelly -- wanted to ask you for a review, but this only works on Cat8000v VMs at the moment (due to licensing), and I know how much you "like" Vagrant 🤢

PR #2839, opened 2025-11-21
FRR: Enable additional daemons listed in the frr.daemons node attribute

ipspace:
Also:
* Move module-to-daemon mapping into the netlab_frr_daemons group variable
* Restructure and clean up the FRRouting caveats

### DIFF
--- a/netsim/ansible/templates/initial/frr/daemons.j2
+++ b/netsim/ansible/templates/initial/frr/daemons.j2
@@ -6,6 +6,8 @@
 echo "{{ frr_m }}=yes" >>/etc/frr/daemons
 {%   endfor %}
 {% endfor %}
-{% for frr_m in frr.daemons|default([]) %}
+{% if frr is defined %}
+{%   for frr_m in frr.daemons|default([]) %}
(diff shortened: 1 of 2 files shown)


netsim/devices/frr.yml
```diff
@@ -16,6 +16,15 @@ group_vars:
     gnupg: gpg
     curl: curl
     iproute2: bridge
+  netlab_frr_daemons:
+    bfd: [ bfdd ]
+    bgp: [ bgpd ]
+    ospf: [ ospfd, ospf6d ]
```
jbemmel, line 22: This could be split by af:
```
ospf:
   ipv4: ....
```

(or where is the logic that decides which ones to enable?)

ipspace: > This could be split by af:

Yes, but that would require some awkward checks in Jinja2 (or more Python code) for very limited gain. Also, OSPF and RIP are the only ones with this challenge.

> (or where is the logic that decides which ones to enable?)

There is no logic. You want OSPF, we enable both daemons.

PR #2842, opened 2025-11-22
Replace routing policy 'set.community.delete' with 'delete.community'

### LINKED ISSUE #2244: Using "set" action with "delete" option to delete BGP communities is silly ;)

ipspace:
The 'set.community.delete: True' is a ridiculous way to say 'delete the community'. This commit resolves #2244 and implement the more logical way of deleting communities in routing policies:

* It defines the new 'delete.community' dictionary that can contain lists of community or community list name
* It automatically converts old-style routing policies into the new format (while generating a warning)
* All device definitions and configuration templates have been updated to use the new attributes

Bonus:
* 'delete.community.list' is now implemented for Junos

### DIFF
--- a/docs/module/routing.md
+++ b/docs/module/routing.md
@@ -164,7 +164,7 @@ Apart from setting BGP communities on BGP routes, these devices can execute addi
 | VyOS                | ✅ | ❌  | ❌  |
 
 **Notes:**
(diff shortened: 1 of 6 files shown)


ipspace:
> Analysis of `delete-community` Branch Implementation

Wow. Thank you. A few suggestions definitely worth following.

PR #2844, opened 2025-11-22
feat: Prevent 'set' and 'delete' attributes with 'action: deny' in routing policies (#2841)

### LINKED ISSUE #2841: [BUG] routing.policy: do not allow 'set' or 'delete' attributes with 'action: deny'
It makes no sense to have set/delete commands in a routing policy if the policy entry denies matched
objects.

jbemmel:
Fixes #2841

### DIFF
--- a/tests/integration/routing/99-deny-action-validation.yml
+++ b/tests/integration/routing/99-deny-action-validation.yml
@@ -1,38 +1,15 @@
 ---
-message: |
-  Test that 'set' and 'delete' attributes are not allowed with 'action: deny'
+message: Test deny action validation
 
-plugin: [ bgp.policy ]
-module: [ bgp, routing ]
-
-defaults.sources.extra: [ ../wait_times.yml, ../warnings.yml ]
+module: [ routing ]
 
 nodes:
   dut:
     device: frr
-    bgp.as: 65000
-    # This should fail validation - set attribute with deny action
-    routing.policy.deny_with_set:
+    routing.policy.test_deny:
     - action: deny
       match:
         aspath: '65001'
       set:
         community:
           standard: '65000:100'
-    # This should also fail validation - delete attribute with deny action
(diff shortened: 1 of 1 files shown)


jbemmel:
Feel free to remove or move the integration test somewhere else, perhaps it would be good to start a 'regression-tests' collection with such cases. Obviously no need to run it every time for integration

ipspace:
Adding a test that is guaranteed to fail into the integration tests is a bit suboptimal ;)

I also agree that it does not make sense to run that test every time we do a push, so maybe I'll add code coverage tests (that would only be executed for earliest and latest Python version we support)

ipspace:
Thank you, but I will try to solve this (more generically) in the validation code, adding _invalid_with to be the opposite of _valid_with currently used in places like https://github.com/ipspace/netlab/blob/533f9da88fedb44e13ef1a7ad58ad608fd34d9b4/netsim/modules/routing.yml#L119

On a more serious note, I didn't notice that your PR depends on the feature branch I deleted after merging it. Apologies for the inadvertent close :(

jbemmel:
> Thank you, but I will try to solve this (more generically) in the validation code, adding _invalid_with to be the opposite of _valid_with currently used in places like
> 
> https://github.com/ipspace/netlab/blob/533f9da88fedb44e13ef1a7ad58ad608fd34d9b4/netsim/modules/routing.yml#L119
> 
> On a more serious note, I didn't notice that your PR depends on the feature branch I deleted after merging it. Apologies for the inadvertent close :(

I provided a “first order” solution. If it didn’t match your requirements, it means you under specified the issue

ipspace:
> I provided a “first order” solution.

Correct. Thanks again.

> If it didn’t match your requirements, it means you under specified the issue

It met the requirements, but when thinking more about it, I figured out a more generic solution wouldn't hurt.

PR #2871, opened 2025-11-30
New wireshark release breaks old plugins

ipspace:


ipspace:
@DanPartelly -- based on your feedback in #2869

DanPartelly (approved):
Sounds good. I will fill a bug today at Wireshark .

PR #2886, opened 2025-12-04
Fix: Copy relevant link module data to interfaces attached to a bridge

### LINKED ISSUE #2880: OSPF on netlab provided bridge node  multiaccess link
As opposed to the situation described in the previous PR, in this case netlab ignores that ospf is disabled on a multi-access link using a device with role set to "bridge" as bridge, and happily enables the protocol on all interfaces involved if the device runs the ospf module.

Comments ? We should at least be consistent in the approach.

```
groups:
    ospf_s:
      members: [hub, s1, s2]
      module: [ospf]

nodes:
    hub:
    s1:
    s2:
    br:
      device: linux
      role: bridge

links:
- hub:
  s1:
  s2:
  bridge: br
  ospf: False

```

ipspace:
The 'convert multi access link into a hub-and-spoke with a bridge' code copied the link data into bridge VLAN data. However, the VLAN code removed the module data not present on the bridge node (i.e. most of them) from the node VLAN, so they were never propagated to the interfaces attached to the bridge node.

This fix copies relevant (present on node, copying allowed) module data into interfaces attached to a bridge node when transforming the multi-access link into the hub-and-spoke links.

Resolves #2880

### DIFF
--- a/netsim/roles/bridge.py
+++ b/netsim/roles/bridge.py
@@ -9,7 +9,7 @@ Bridge-specific data transformation:
 from box import Box
 
 from ..augment import links
-from ..data import append_to_list, get_box, global_vars
+from ..data import append_to_list, get_box, global_vars, merge_with_removed_attributes
 from ..modules import _dataplane
 from ..utils import log
 from . import select_nodes_by_role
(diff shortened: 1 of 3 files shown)


DanPartelly (approved):
Works great.  Thank you.

PR #2903, opened 2025-12-07
Run host- or container configuration scripts in 'netlab initial'

ipspace:
This commit adds the ability to run container bash scripts either in the container or in the host namespace. It's currently used to configure Bird, dnsmasq, and Linux.

The core change is an extra step in the 'netlab initial' process that calls provider-specific 'deploy_node_config' hook. That clab hook executes a bash script (identified by its suffix) either on the host (in container namespace) or in the container.

Other changes:
* netlab initial configuration deployment has been refactored into a separate module (deploy.py)
* Refactor the 'do we need Ansible' code into a set of generic functions that (a) identify the configs a user wants to deploy with 'netlab initial', (b) figures out whether all configs are covered with config scripts and (c) does that for the requested nodeset
* Use the refactored functions in configuration creation ('-o') and deployment, resulting in cleaner configuration creation

Minor fixes:
* Bird implementation no longer needs extra 'config.sh' template and 'clab.exec' parameter
* Bird loopback LLA configuration does not crash on second run
* Linux routing configuration no longer prints the final routing table
* netlab initial '-v' flag is used to set log.VERBOSE variable

### DIFF
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -10,13 +10,14 @@ packages:
   bird: bird
 daemon_config:
   bird: /etc/bird/bird.conf
-  initial: /etc/config/01-initial.sh
-  lag: /etc/config/10-lag.sh
-  vlan: /etc/config/20-vlan.sh
   bgp: /etc/bird/bgp.mod.conf
   ospf: /etc/bird/ospf.mod.conf
   routing: /etc/bird/routing.mod.conf
 clab:
+  node_config:
+    initial: /etc/config/01-initial.sh:sh
+    lag: /etc/config/10-lag.sh:sh
+    vlan: /etc/config/20-vlan.sh:sh
   group_vars:
     netlab_show_command: [ birdc, 'show $@' ]
     docker_shell: bash -il


ipspace:
@Muddyblack -- Hope you'll like this one ;) Once this gets stable, I'll add FRR and VyOS implementation, and then you could (if needed) add multithreading to the "deploy config scripts" loop

ipspace:
Three more things to do:

* Device quirks and provider templates have to be in correct order (right now, initial/01-interfaces fails for cEOS containers)
* I will use the second parameter in **config_template** (currently `:ro`) to indicate the type of the script (`:sh` for container-side script, `:ns` for namespace script)
* It should be relatively easy to use `_template_cache` to do automatic refresh in **netlab initial** (fix for #2905)

netsim/ansible/templates/initial/bird-clab.j2
```diff
@@ -1,7 +1,8 @@
 set -x
 {% include 'linux-clab.j2' +%}
 #
-set -e
 {% if loopback is defined and loopback.ipv6 is defined %}
```
jbemmel, line 4: What about ```loopback.ipv6: False```?

netsim/devices/linux.yml
```diff
@@ -42,6 +42,11 @@ group_vars:
   netlab_lldp_enable: False
   netlab_net_tools: False
 clab:
+  node_config:
+    initial: /etc/config/01-initial.ns-sh
+    lag: /etc/config/10-lag.ns-sh
+    vlan: /etc/config/20-vlan.ns-sh
+    routing: /etc/config/30-routing.ns-sh
```
jbemmel: This confuses me:
* Where are these scripts located?
* Why are we encoding the dependency order manually?
* Do 30-routing.ns-sh and 02-routing.ns-sh (dnsmasq) map to the same script?

Apart from the order, is there a reason why the mapping between modules and the scripts that initialize them in Linux/container context couldn't be defined in the module?

In my mind, ideally ```netsim/devices/linux.yml``` would simply state that Linux devices support the lag module. The implication to map that to a '10-lag.ns-sh' script under /etc/config/ could be computed, rather than spelled out. What am I missing?

ipspace: > This confuses me:

Documentation coming "soon"...

> * Where are these scripts located?

This is equivalent to "clab.config_templates", but matched with config modules

> * Why are we encoding the dependency order manually?

Because this is an early implementation. Obviously, we could do optimizations once the initial implementation works well enough.

> * Do 30-routing.ns-sh and 02-routing.ns-sh (dnsmasq) map to the same script?

Think "clab.config_templates" ;)

> Apart from the order, is there a reason why the mapping between modules and the scripts that initialize them in Linux/container context couldn't be defined in the module?

Why would it be defined there?

> In my mind, ideally `netsim/devices/linux.yml` would simply state that Linux devices support the lag module. The implication to map that to a '10-lag.ns-sh' script under /etc/config/ could be computed, rather than spelled out. What am I missing?

That not all devices are configured the same way? That containers and VMs are not defined the same way?

netsim/ansible/templates/initial/bird-clab.j2
ipspace: > What about `loopback.ipv6: False`?

(A) I'm pretty sure 'ipv6: False' is removed somewhere in the transformation process.
(B) I don't think it matters too much if an LLA is defined on a loopback.
(C) This is just an adaptation of an already-working configuration template to a pure Jinja2 environment

netsim/devices/linux.yml
ipspace: > * Do 30-routing.ns-sh and 02-routing.ns-sh (dnsmasq) map to the same script?

FWIW: I cleaned this one up. The dnsmasq daemon now inherits the Linux node_config scripts

PR #2906, opened 2025-12-08
Add OSPF support for ASA OSPF

sbng:
- Enable support of OSPF in asav.yml
- Create the remplate to generate ASA OSPF configuration
- Disable OSPF point-to-point config generation in ASA. This is to workaround an issue with ASA ospf network point-to-point unable to establish OSPF neighbor when configured.

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -96,6 +96,8 @@ nodes:
 ## Cisco ASAv Caveats
 
 * Some ASAv versions use older SSH protocols. For more details, see the [Cisco IOSv SSH caveats](cisco-iosv-ssh).
+* ASAv does not have a standard implementation of OSPFv2 or IS-IS point-to-point circuits. netlab reports an error if you try to use them with ASAv nodes.
+* The ASAv OSPF and IS-IS configuration templates were not tested, as all OSPFv2/IS-IS integration tests include at least one point-to-point circuit
 
 (caveats-cat8000v)=
 ## Cisco Catalyst 8000v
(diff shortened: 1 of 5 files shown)


ipspace:
Thank you! Just a minor detail I merge it (I will also fix the documentation):

> * Disable OSPF point-to-point config generation in ASA.

So ASAv cannot do P2P OSPF? If that's the case, I'll add a device quirk to generate an error (because the adjacency wouldn't be established anyway).

sbng:
Yes. The ASA ospf implementation seems broken. This is what happen:
- IOS (point to point) <> ASA (point to point) = Broken
- IOS(broadcast) <> ASA (broadcast) = Ok
- IOS(point to point) <> ASA (broadcast) = OK

ASA refuses to send out OSPF Hello when configured as point to point. IOS on the other hand is generating OSPF Hello regardless of point to point or broadcast. So workaround is to force ASA to always use broadcast. I will try to file a Cisco TAC case and see how far this goes. Meanwhile, I think by not configuring ospf network point to point should at least get the OSPF going.

ipspace:
> IOS(point to point) <> ASA (broadcast) = OK

I doubt that works. As I [learned the hard way](https://blog.ipspace.net/2025/06/testing-ospf-configurations/), the adjacency is established, but the routing tables are not correct.

> ASA refuses to send out OSPF Hello when configured as point to point.

ASA implements "non-broadcast point-to-point mode" (the first time I've ever heard of such a beast), and you have to configure a single static neighbor to make it work.

I'll add a device quirk that prevents P2P networks with OSPF on ASA and merge this.

Thanks again, Ivan

ipspace (approved):
Nice job! Thank you, and welcome to the netlab contributors club ;) Hope it's not your last PR ;))

PR #2914, opened 2025-12-12
Optimize loading of package defaults with a pickle file

ipspace:
The change shaves ~250 msec from every 'netlab create'/'netlab up' call, resulting in enormous savings considering how fast some network devices like vJunos or IOS/XE boot.

### DIFF
--- a/netsim/utils/read.py
+++ b/netsim/utils/read.py
@@ -106,11 +106,10 @@ def read_from_pickle(path: str) -> typing.Optional[Box]:
       print(f'RFP: {str(pickle_path)} does not exist, exiting')
     return None
   try:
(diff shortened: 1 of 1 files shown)


jbemmel (approved):
I appreciate that you didn't dismiss the idea completely, even when I did a poor job at explaining where I was going with it

ipspace:
> I appreciate that you didn't dismiss the idea completely, even when I did a poor job at explaining where I was going with it

Well, the idea was on the back burner for ages, but it makes so little sense for the regular users (considering how long things boot) that it stayed there for years.

Honestly, the only thing that's faster is the CI/CD action that goes through all integration tests (because the transformation tests use cached defaults), and that's run in the background anyway and was never a show-stopper. But hey, we optimized stuff... ;)

PR #2920, opened 2025-12-13
Make netlab Jinja2 handling more Ansible-like

ipspace:
* The undefined variables are handled with a derived Undefined type that behaves like Ansible variables. For example 'x.y is defined' returns false (instead of crashing) if x is not defined.
* All Jinja2 filters are defined as having 'ansible.utils.X' synonim.

With these enhancements, we can revert the changes made to Bird and Linux templates


ipspace:
This one unfortunately escaped in the wild when I was trying to make a small fix to ipaddr templates. Closing...

PR #2927, opened 2025-12-14
Simplify STP configuration templates by normalizing data model

### LINKED ISSUE #2924: [BUG] NVUE/OS10 STP configuration templates have to be simplified
Both configuration templates use the same weird expression that looks like line noise (the author at the time of writing the code might be an exception).

We should either make the STP transformed data model easier to consume or do whatever is needed in device quirks

jbemmel:
Move complex Jinja2 filter logic from templates to STP module transformation. This makes the data model easier to consume by marking interfaces with _has_stp_config flag during transformation.

Changes:
- Add normalization in STP module to mark interfaces needing STP config
- Simplify cumulus_nvue.j2 and dellos10.j2 templates to use simple flag check
- Replaces complex filter: ifdata.vlan.trunk|default({})|dict2items|map(attribute='value')|selectattr('stp','defined')
- Applies to all devices, making the data model consistent
- Optimize trunk VLAN check using any() with generator expression
- Remove trailing whitespace

The complex Jinja2 filter expression was hard to read and maintain. By normalizing the data model during STP module transformation, we:
- Move the logic to Python where it's more readable
- Make templates simpler and consistent across devices
- Apply the normalization to all devices automatically

Fixes #2924

### DIFF
--- a/netsim/modules/stp.py
+++ b/netsim/modules/stp.py
@@ -123,10 +123,3 @@ class STP(_Module):
       trunk = intf.get('vlan.trunk',{})
       if trunk and any('stp' in vdata for vdata in trunk.values()):
         intf._has_stp_config = True
-
-  """
-  Cleanup helper attributes that shouldn't appear in final topology output
-  """
-  def node_cleanup(self, node: Box, topology: Box) -> None:
-    for intf in node.get('interfaces',[]):
-      intf.pop('_has_stp_config',None)
--- a/tests/topology/expected/lag-l2.yml
+++ b/tests/topology/expected/lag-l2.yml
@@ -107,7 +107,8 @@ nodes:
     hostname: clab-input-r1
     id: 1
     interfaces:
-    - ifindex: 30000
+    - _has_stp_config: true
+      ifindex: 30000
       ifname: port-channel1
       lag:
         ifindex: 1
@@ -253,7 +254,8 @@ nodes:
     hostname: clab-input-r2
     id: 2
     interfaces:
-    - ifindex: 30000
+    - _has_stp_config: true
+      ifindex: 30000
       ifname: port-channel1
       lag:
         ifindex: 1
--- a/tests/topology/expected/stp-port-type.yml
+++ b/tests/topology/expected/stp-port-type.yml
@@ -461,7 +461,8 @@ nodes:
     device: eos
     id: 6
     interfaces:
-    - ifindex: 1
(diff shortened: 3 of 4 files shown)


netsim/modules/stp.py
```diff
@@ -102,3 +102,24 @@ def node_post_transform(self, node: Box, topology: Box) -> None:
...
+    # Templates can now simply check: ifdata._has_stp_config|default(False)
+    # This applies to all devices, not just NVUE/OS10, making the data model consistent
+    for intf in node.get('interfaces',[]):
+      # Check for direct STP configuration
+      if 'stp' in intf:
+        intf._has_stp_config = True
+        continue
+
+      # Check for STP configuration in trunk VLANs
+      trunk = intf.get('vlan.trunk',{})
```
ipspace, line 123: Well, with "any" we could turn this into an unreadable one-liner 😜

netsim/modules/stp.py
```diff
@@ -123,3 +123,10 @@ def node_post_transform(self, node: Box, topology: Box) -> None:
       trunk = intf.get('vlan.trunk',{})
       if trunk and any('stp' in vdata for vdata in trunk.values()):
         intf._has_stp_config = True
+
+  """
+  Cleanup helper attributes that shouldn't appear in final topology output
+  """
+  def node_cleanup(self, node: Box, topology: Box) -> None:
```
ipspace: I think this is wrong as we need that attribute in the final data for the configuration templates

netsim/modules/stp.py
jbemmel: It's your choice - but be careful what you wish for ;)

netsim/modules/stp.py
jbemmel: ok, let me revert

netsim/modules/stp.py
ipspace: I didn't say you should do it ;))

netsim/modules/stp.py
ipspace: Was this an AI suggestion? I'm sure you know better...

ipspace (approved):
The Dell OS10 code works like a charm, NVUE tests fail due to #2928. Will figure out what went wrong there, fix that, rerun the NVUE STP tests (which I'm sure will succeed) and merge the PR.

Thanks again!

PR #2932, opened 2025-12-14
Clean up router_id tests in templates

### LINKED ISSUE #2929: [BUG] Clean up tests for bgp.router_id and ospf.router_id
Many templates contain this test for **bgp.router_id**

```
{% if bgp.router_id|ipv4 %}
```

That's meaningless. If the **bgp.router_id** is calculated or configured, then it will be an IPv4 address. If we want to test anything, it should be **bgp.router_id is defined**

jbemmel:
Replace meaningless router_id|ipv4 tests with router_id is defined checks. If router_id is calculated or configured, it will always be an IPv4 address, so testing for ipv4 format is redundant. The proper test is whether router_id is defined.

Fixes #2929

### DIFF
--- a/netsim/ansible/templates/bgp/arubacx.j2
+++ b/netsim/ansible/templates/bgp/arubacx.j2
@@ -4,7 +4,7 @@
 router bgp {{ bgp.as }}
   bgp log-neighbor-changes
   timers bgp connect-retry 10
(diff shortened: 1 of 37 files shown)


ipspace (approved):
Looks good to me. Thank you! Will run a few integration tests and merge it.

Issue #2933, opened 2025-12-14
Router ID Handling Inconsistencies in VRF Templates


jbemmel:
This document identifies inconsistencies in how router IDs are handled in VRF templates across different device types.

## Standard Pattern

The correct pattern for handling router IDs in VRF contexts should be:

- **BGP**: `vdata.bgp.router_id|default(bgp.router_id)`
- **OSPF**: `vdata.ospf.router_id|default(ospf.router_id)`

This pattern allows VRF-specific router_id to override the global router_id, but falls back to the global router_id if the VRF-specific one is not set.

### 1. routeros7.bgp.j2 (line 14-15)

**Current:**
```jinja2
{% if bgp.router_id is defined %}
/routing/bgp/template set vrf_{{ vname }} router-id={{ bgp.router_id }}
{% endif %}
```

**Problem:** Uses `bgp.router_id` directly without checking `vdata.bgp.router_id` first. This means VRF-specific router IDs are ignored.

**Should be:**
```jinja2
{% if vdata.bgp.router_id is defined or bgp.router_id is defined %}
/routing/bgp/template set vrf_{{ vname }} router-id={{ vdata.bgp.router_id|default(bgp.router_id) }}
{% endif %}
```

### 2. iosxr.bgp.j2 (line 18-19)

**Current:**
```jinja2
{% if vdata.bgp.router_id is defined %}
  bgp router-id {{ vdata.bgp.router_id }}
{% endif %}
```

**Problem:** Only checks for VRF-specific router_id, doesn't fall back to global `bgp.router_id`.

**Should be:**
```jinja2
{% if vdata.bgp.router_id is defined or bgp.router_id is defined %}
  bgp router-id {{ vdata.bgp.router_id|default(bgp.router_id) }}
{% endif %}
```

### 3. nxos.bgp.j2 (line 10-11)

**Current:**
```jinja2
{%   if bgp.router_id is defined %}
    router-id {{ bgp.router_id }}
{%   endif %}
```

**Problem:** Uses `bgp.router_id` directly without checking `vdata.bgp.router_id` first.

**Should be:**
```jinja2
{%   if vdata.bgp.router_id is defined or bgp.router_id is defined %}
    router-id {{ vdata.bgp.router_id|default(bgp.router_id) }}
{%   endif %}
```

### 4. junos.bgp.j2 (line 61-62)

**Current:**
```jinja2
{%   if bgp.router_id is defined %}
[...]

PR #2935, opened 2025-12-14
Add test cases for router_id inconsistencies

jbemmel:
## Maintainer summary:

**Total Issues Found:**
- BGP VRF templates: 4 real problems
- OSPF VRF templates: 2 real problems (2 templates are correct but verbose)

**Impact:**
- Templates that don't check `vdata.bgp.router_id` or `vdata.ospf.router_id` first will ignore VRF-specific router IDs
- Templates that don't fall back to global router_id may not set a router ID when only the global one is configured
- Redundant checks add unnecessary complexity

**Recommendation:**
Standardize all VRF templates to use the pattern:
```jinja2
{% if vdata.bgp.router_id is defined or bgp.router_id is defined %}
  router-id {{ vdata.bgp.router_id|default(bgp.router_id) }}
{% endif %}
```

Or, if the router-id is always required (device will use a default), simply:
```jinja2
router-id {{ vdata.bgp.router_id|default(bgp.router_id) }}
```
======
- Create test topologies for BGP and OSPF VRF router_id scenarios
- Add verification scripts (bash and Python) to check router_id handling
- Add documentation of router_id inconsistencies in test directory
- Test cases verify inconsistencies documented in router_id_inconsistencies.md
- Tests cover scenarios with VRF-specific router_ids, global fallback, and edge cases

Illustrates #2933

### DIFF
--- a/tests/router_id_inconsistencies/test-bgp-vrf-router-id.yml
+++ b/tests/router_id_inconsistencies/test-bgp-vrf-router-id.yml
@@ -55,6 +55,6 @@ vrfs:
 
 links:
 - r1-r2
-  - r2-r3
-  - r3-r4
-  - r4-r5
+- r2-r3
+- r3-r4
+- r4-r5
--- a/tests/router_id_inconsistencies/test-ospf-vrf-router-id.yml
+++ b/tests/router_id_inconsistencies/test-ospf-vrf-router-id.yml
@@ -41,4 +41,4 @@ vrfs:
 
 links:
 - r1-r2
-  - r2-r3
+- r2-r3


ipspace:
I know it's nice to have a test for every bug report, but nonetheless:

* We'll fix the old stuff (and it's easy to find with "grep")
* Someone will review the new BGP/OSPF implementations anyway, and maybe we can fine-tune the Copilot instructions to look for inconsistencies compared to other similar templates, which might turn out to be a better investment.
* Nobody will maintain this code once we fix the existing issues

I don't think it's worth spending any time on this.

ipspace:
Will be replaced by whatever we decide in #2952

Issue #2925, opened 2025-12-14
[BUG] OSPF areas should be collected in Python code


ipspace:
We already use the list of OSPF areas in the **ospf.areas** plugin. We should extend that and use the same data structure (but without the extra attributes) in the main OSPF module, simplifying the configuration templates for devices that have to use area/interface configuration structure (for example, IOS/XR)

jbemmel:
Something like this?
```Python
"""
Collect OSPF areas from interfaces and create a simplified area list.
This uses the same data structure as the ospf.areas plugin (area and _area_int)
but without the extra attributes (kind, range, etc.).
This simplifies templates for devices that use area/interface configuration structure.
"""
def collect_ospf_areas(node: Box) -> None:
  for (o_data,o_intf,vrf) in _routing.rp_data(node,'ospf'):
    if 'areas' in o_data:  # If areas already exist (from plugin), skip
      continue
    
    # Collect unique areas from interfaces
    area_set = set()
    area_map = {}  # Map area_int -> area string
    
    for intf in o_intf:
      if 'ospf' not in intf or 'area' not in intf.ospf:
        continue
      
      area = intf.ospf.area
      # Normalize area to integer (handle both int and string formats)
      if isinstance(area, int):
        area_int = area
        area_str = str(ipaddress.IPv4Address(area))
      else:
        area_str = str(area)
        area_int = int(ipaddress.IPv4Address(area_str))
      
      if area_int not in area_set:
        area_set.add(area_int)
        area_map[area_int] = area_str
    
    # Create area list with area and _area_int attributes
    if area_set:
      o_data.areas = []
      for area_int in sorted(area_set):
        o_data.areas.append(Box({
          'area': area_map[area_int],
          '_area_int': area_int
        }))
      
      # Mark as ABR if multiple areas
      if len(area_set) > 1:
        o_data._abr = True
```

ipspace:
Resolved in #2936

PR #2936, opened 2025-12-14
Collect OSPF areas in Python code (#2925)

jbemmel:
- Add collect_ospf_areas() function to main OSPF module
- Create simplified area list with area and _area_int attributes
- Use same data structure as ospf.areas plugin (without extra attributes)
- Extract normalize_ospf_area() and add_ospf_area() helper functions
- Update templates to use ospf.areas instead of collecting from interfaces:
  - iosxr.j2 (Cisco IOS XR)
  - sros.j2 (Nokia SR OS)
  - openbsd.ospfv2.j2 and openbsd.ospfv3.j2 (OpenBSD)
  - fortinet.fortios.fortios.j2 (Fortinet FortiOS)

This simplifies templates for devices that use area/interface configuration structure and makes the area list available even when ospf.areas plugin is not enabled.

### DIFF
--- a/netsim/modules/ospf.py
+++ b/netsim/modules/ospf.py
@@ -3,11 +3,11 @@
 #
 
 import ipaddress
-import typing
 
 from box import Box
 
 from ..augment import devices
+from ..data import append_to_list
 from ..utils import log
 from ..utils import routing as _ospf
 from . import _Module, _routing, bfd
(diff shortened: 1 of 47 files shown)


netsim/modules/ospf.py
```diff
@@ -76,6 +80,72 @@ def propagate_node_attributes(node: Box, topology: Box) -> None:
...
+
+"""
+Collect OSPF areas from interfaces and create a simplified area list.
+This uses the same data structure as the ospf.areas plugin (area and _area_int)
+but without the extra attributes (kind, range, etc.).
+This simplifies templates for devices that use area/interface configuration structure.
+"""
+def collect_ospf_areas(node: Box) -> None:
+  for (o_data,o_intf,vrf) in _ospf.rp_data(node,'ospf'):
+    if 'areas' in o_data:  # If areas already exist (from plugin), skip
```
ipspace: Incorrect. The user could define just a subset of areas

netsim/modules/ospf.py
```diff
@@ -76,6 +80,72 @@ def propagate_node_attributes(node: Box, topology: Box) -> None:
...
+      continue
+    
+    # Collect unique areas from interfaces
+    area_set: typing.Set[int] = set()
+    area_map: typing.Dict[int, str] = {}  # Map area_int -> area string
+    
+    # Collect areas from regular interfaces
+    for intf in o_intf:
+      if 'ospf' not in intf or 'area' not in intf.ospf:
```
ipspace: If 'ospf.area' not in intf.ospf

AI? It refuses to believe Box can use dots in key lookups

netsim/modules/ospf.py
```diff
@@ -76,6 +80,72 @@ def propagate_node_attributes(node: Box, topology: Box) -> None:
...
+    else:
+      loopback = node.get('loopback', {})
+    
+    # Check if loopback is a dictionary (not a boolean) and has OSPF area
+    if isinstance(loopback, Box) and 'ospf' in loopback and 'area' in loopback.ospf:
+      add_ospf_area(loopback.ospf.area, area_set, area_map)
+    
+    # Create area list with area and _area_int attributes
+    if area_set:
+      o_data.areas = []
```
ipspace: You have to merge collected areas with already-defined ones, then sort them by area_int. You can use "data.append_to_list" to simplify the code.

netsim/modules/ospf.py
jbemmel: It's likely a more common pattern than people using Box. In this particular case, IMHO the AI version is very usable

jbemmel:
> I cleaned up the code, the integration tests on Cisco 8000v, OpenBSD, and SR-SIM pass, so I'm merging it...
> 
> Thanks again!

If it doesn't bother you too much, I think this could work well:
* AI takes care of the boring, repetitive stuff like adapting all device templates (basically "intelligent" search/replace)
* human experts review the core code changes

ipspace:
> If it doesn't bother you too much, I think this could work well:
> 
> * AI takes care of the boring, repetitive stuff like adapting all device templates (basically "intelligent" search/replace)
> * human experts review the core code changes

Sounds like a perfect plan ;) Lets see how well it works ;))

Issue #2943, opened 2025-12-15
[BUG] SR OS OSPF Router ID Handling


jbemmel:
**Date:** 2025-12-15
**Platform:** Nokia SR OS (SROS)  
**Severity:** High  
**Status:** Open

## Executive Summary

Analysis of SR OS templates reveals inconsistencies in router ID handling across different routing protocols and contexts (global vs. VRF). The most critical issue is that OSPF router ID in VRF contexts lacks fallback to global router ID, which can cause configuration failures or unexpected behavior.

### Issue #1: OSPF Router ID Missing Fallback in VRF Context (HIGH PRIORITY)

**File:** `netsim/ansible/templates/ospf/sros.j2:54`  
**Called from:** `netsim/ansible/templates/vrf/sros.j2:25`

**Problem:**
The `ospf_config()` macro doesn't support falling back to global OSPF router ID when VRF-specific router ID is not set.

**Current Code:**
```jinja2
# In ospf_config() macro (templates/ospf/sros.j2:54)
router-id: {{ ospf.router_id }}
```

**How it's called from VRF template:**
```jinja2
# In vrf/sros.j2:25
{{ ospf_config(rtg_af,vdata.ospf,vdata.ospf.interfaces,vname) }}
```

**Root Cause:**
- The macro receives `vdata.ospf` as the `ospf` parameter
- When `vdata.ospf.router_id` is not defined, `ospf.router_id` evaluates to undefined
- The macro has no access to the global `ospf` variable to use as fallback
- Unlike BGP router ID handling in the same VRF template, which correctly implements fallback

**Impact:**
- If `vdata.ospf.router_id` is not set, the OSPF router ID may be undefined or cause template rendering errors
- OSPF may fail to start or use an unexpected router ID
- Inconsistent behavior compared to BGP, which has proper fallback handling

**Expected Behavior:**
The OSPF router ID should fall back to global `ospf.router_id` when VRF-specific router ID is not set, consistent with:
1. BGP router ID handling in the same VRF template
2. Standard pattern used by other devices
3. Data model expectations

**Recommended Fix:**
Modify the `ospf_config()` macro to accept an optional global OSPF router ID parameter, or modify the VRF template to pass the global router ID as fallback:

```jinja2
# Option 1: Modify macro signature
{% macro ospf_config(af,ospf,vrf_interfaces,vrf_name='',global_ospf=None) %}
...
[...]

PR #2944, opened 2025-12-15
Fix SR OS OSPF router ID fallback in VRF context

jbemmel:
- Modified ospf_config() macro to accept global_router_id parameter
- Updated macro to use global router ID as fallback when VRF-specific router ID is not set
- Updated VRF template to pass global OSPF router ID to the macro

This ensures OSPF router ID in VRF contexts correctly falls back to global OSPF router ID, consistent with BGP router ID handling in the same template.

Fixes: #2943

Note: We may want to consider having the Python code in the OSPF module fill out the router_id in each VRF, such that the templates can simply take it from there. That way, inconsistencies like this are less likely by structure

### DIFF
--- a/netsim/ansible/templates/ospf/sros.j2
+++ b/netsim/ansible/templates/ospf/sros.j2
@@ -32,7 +32,7 @@
 {%  endif %}
 {% endmacro %}
 
-{% macro ospf_config(af,ospf,vrf_interfaces,vrf_name='') %}
+{% macro ospf_config(af,ospf,vrf_interfaces,vrf_name='',global_router_id=None) %}
 {%   set vname = vrf_name or 'default' %}
 {%   set pid = ospf.process|default(0) %}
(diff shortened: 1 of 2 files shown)


ipspace:
> We may want to consider having the Python code in the OSPF module fill out the router_id in each VRF, such that the templates can simply take it from there. That way, inconsistencies like this are less likely by structure

Ah, you mean this: https://github.com/ipspace/netlab/blob/f47eedf4a5d65b5789a6160789e1a50195bcaf0b/netsim/modules/ospf.yml#L54

Already there and easy to verify ;)

Anyway, the templates have "a little" history and while they all work, we might go and clean them up one of these days.

ipspace:
Let me go through the tests verifying `netsim/utils/templates` generates the same configuration files as Ansible, then we'll focus on cleaning up old stuff.

PR #2950, opened 2025-12-18
Use internal templating engine for 'netlab initial -o' command

ipspace:
* The 'netlab initial -o' command generates configuration files with the internal templating engine unless the '--generate ansible' argument forces it to use an Ansible playbook.
* The configuration files can have '.sh' suffix if we know they're shell scripts
* The '--generate compare' parameter makes the config file names identical to the ones created by Ansible playbooks to make them directly comparable with diff -r
* A new testing script generates configs for all integration tests passed by a device and compares internal- and Ansible-generated configs

### DIFF
--- a/netsim/cli/initial/__init__.py
+++ b/netsim/cli/initial/__init__.py
@@ -20,7 +20,7 @@ def run_initial(cli_args: typing.List[str]) -> None:
 
   log.set_logging_flags(args)
   if args.logging or args.verbose:
-    print("Ansible playbook args: %s" % rest)
+    print(f"Unrecognized Ansible playbook args: {rest}")
 
   if args.output:
     if rest:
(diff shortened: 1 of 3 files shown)


ipspace:
@Muddyblack will love these figures (executed on large topology with 300+ nodes, mostly Linux hosts):

* **netlab create**: 27 seconds. A lot of time seems to be spent building Ansible inventory or something else after the provider files have been created; have to investigate what's going on there.
* **netlab initial -o**: 3 seconds (time needed to create all config snippets/scripts for all nodes)
* **netlab initial -o --generate ansible**: 48s, using almost 4 minutes of CPU time

We got 15x improvement in configuration generation compared to an identical Ansible playbook.

ipspace:
Adding a bunch of potential reviewers in case anyone wants to look at it (but I also know it's the year-end, so no pressure or expectations)

PR #2960, opened 2025-12-22
Fix: Light cleanup of RP router_id handling

ipspace:
* Removed the few weird calculations of BGP cluster_id
* Made SR-OS EVPN RD based on global BGP router ID (route distinguishers have global significance)
* Use VRF BGP router ID (when defined) on NX-OS
* Cleanup VyOS OSPF configs as OSPF RID is always defined

Other suboptimal practices that are not confusing (for example, checking whether OSPF or BGP router IDs exist) were not changed.

### DIFF
--- a/netsim/ansible/templates/evpn/sros.j2
+++ b/netsim/ansible/templates/evpn/sros.j2
@@ -86,7 +86,7 @@ updates:
    bgp-ipvpn:
     mpls:
      admin-state: enable
-     route-distinguisher: {{ bgp.router_id }}:{{ vdata.vrfidx }}
+     route-distinguisher: {{ vdata.evpn.id }}
      vrf-target:
       export-community: target:{{ vdata.export[0] }}
       import-community: target:{{ vdata.import[0] }}


docs/dev/config/bgp.md
```diff
@@ -56,7 +56,7 @@ _netlab_ assumes the IPv4 BGP session carries the IPv4 address family (plus othe
 * **bgp.neighbors** -- A list of IBGP and EBGP neighbors
 * **bgp.next_hop_self** -- When set to *True*, the BGP process should set itself as the next hop on EBGP routes propagated as IBGP routes. This attribute is propagated to the neighbor **next_hop_self** attribute and can be ignored if you configure **next-hop-self** on individual neighbors.
 * **bgp.originate** (optional) -- A list of additional IPv4 prefixes that should be advertised by the BGP process. You'll probably have to create static routes to support these prefixes.
-* **bgp.router_id** -- The BGP router ID
+* **bgp.router_id** -- The BGP router ID (always set). Always configure the BGP router ID to ensure your BGP configuration works in IPv6-only deployments.
```
jbemmel, line 59: Who is "always set" targeted at? Is it documenting that Netlab ensures it is always set, or should users always define it?

netsim/ansible/templates/evpn/sros.j2
```diff
@@ -86,7 +86,7 @@ updates:
    bgp-ipvpn:
     mpls:
      admin-state: enable
-     route-distinguisher: {{ vdata.bgp.router_id|default(bgp.router_id) }}:{{ vdata.vrfidx }}
+     route-distinguisher: {{ bgp.router_id }}:{{ vdata.vrfidx }}
```
jbemmel: Perhaps this should use ```vdata.evpn.rd``` instead?

ipspace:
> The evpn module sets it, we might as well use it everywhere...

Thank you. Will fix SR-OS template, and open an issue for other templates in case there's more

docs/dev/config/bgp.md
ipspace: This is developer documentation, so one would assume the developer reading it would go like "cool, I can use it"

PR #2968, opened 2025-12-24
netlab config: Add support for Ansible 12+

ipspace:
A completely rewritten 'netlab config' command combines the 'generate configlets with built-in code' approach used with 'netlab initial' with the existing Ansible playbooks to deploy rendered config files -- an approach that works well with broken networking modules "introduced" in Ansible 12.

### DIFF
--- a/docs/netlab/config.md
+++ b/docs/netlab/config.md
@@ -1,14 +1,16 @@
 (netlab-config)=
 # Deploying Custom Device Configurations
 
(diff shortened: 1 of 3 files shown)


ipspace:
As before: adding reviewers in case someone feels bored and wants to look at some code.

Will run the full integration tests before merging it.

ipspace:
Code beyond this PR has been tested, and the integration tests passed, so I'm merging this one

PR #2969, opened 2025-12-24
Add lightweight netlab API server command

captainpacket:
Adds a minimal netlab API server (stdlib HTTP) to expose CLI actions for automation. Includes new CLI command, help listing, and docs page.

### DIFF
--- a/netsim/cli/api.py
+++ b/netsim/cli/api.py
@@ -171,21 +171,21 @@ def add_opt(args: List[str], payload: Dict[str, Any], key: str, opt: str) -> Non
   args += [opt, str(value)]
 
 
-def action_runner(payload: Dict[str, Any], workdir: Path) -> Tuple[Callable[[List[str]], None], List[str]]:
+def action_runner(payload: Dict[str, Any], workdir: Path) -> Tuple[Callable[..., Any], List[str]]:
   action = (payload.get("action") or "up").strip().lower()
 
-  def run_up() -> Tuple[Callable[[List[str]], None], List[str]]:
+  def run_up() -> Tuple[Callable[..., Any], List[str]]:
     return netlab_up.run, [resolve_topology(payload, workdir)]
 
-  def run_create() -> Tuple[Callable[[List[str]], None], List[str]]:
+  def run_create() -> Tuple[Callable[..., Any], List[str]]:
     return netlab_create.run, [resolve_topology(payload, workdir)]
 
-  def run_down() -> Tuple[Callable[[List[str]], None], List[str]]:
+  def run_down() -> Tuple[Callable[..., Any], List[str]]:
     args: List[str] = []
     add_flag(args, payload, "cleanup", "--cleanup")
     return netlab_down.run, args
 
-  def run_collect() -> Tuple[Callable[[List[str]], None], List[str]]:
+  def run_collect() -> Tuple[Callable[..., Any], List[str]]:
     args: List[str] = []
     add_opt(args, payload, "instance", "--instance")
     add_opt(args, payload, "collectOutput", "--output")
@@ -193,13 +193,13 @@ def action_runner(payload: Dict[str, Any], workdir: Path) -> Tuple[Callable[[Lis
     add_flag(args, payload, "collectCleanup", "--cleanup")
     return netlab_collect.run, args
 
-  def run_status() -> Tuple[Callable[[List[str]], None], List[str]]:
+  def run_status() -> Tuple[Callable[..., Any], List[str]]:
     args: List[str] = ["--all"]
     if payload.get("instance"):
       args = ["--instance", str(payload.get("instance"))]
     return netlab_status.run, args
 
-  handlers: Dict[str, Callable[[], Tuple[Callable[[List[str]], None], List[str]]]] = {
+  handlers: Dict[str, Callable[[], Tuple[Callable[..., Any], List[str]]]] = {
     "up": run_up,
     "create": run_create,
     "down": run_down,
@@ -216,7 +216,7 @@ def action_runner(payload: Dict[str, Any], workdir: Path) -> Tuple[Callable[[Lis
 def run_netlab_action(payload: Dict[str, Any], log_fp: io.TextIOBase) -> None:
   workdir = workspace_dir(payload)
 
-  def _run_with_output(fn: Callable[[List[str]], None], args: List[str]) -> None:
+  def _run_with_output(fn: Callable[..., Any], args: List[str]) -> None:
     out = io.StringIO()
     with contextlib.redirect_stdout(out), contextlib.redirect_stderr(out):
       fn(args)
@@ -351,7 +351,7 @@ class NetlabHandler(BaseHTTPRequestHandler):
     if parts[0] == "jobs" and len(parts) >= 2:
       job_id = parts[1]
       with JOB_LOCK:
-        job = JOBS.get(job_id)
+        job: Optional[Dict[str, Any]] = JOBS.get(job_id)
       if job is None:
         return self._not_found()
       if len(parts) == 2:
@@ -386,15 +386,15 @@ class NetlabHandler(BaseHTTPRequestHandler):
     if len(parts) == 3 and parts[0] == "jobs" and parts[2] == "cancel":
       job_id = parts[1]
       with JOB_LOCK:
-        job = JOBS.get(job_id)
-        if job is None:
+        job_entry: Optional[Dict[str, Any]] = JOBS.get(job_id)
+        if job_entry is None:
           return self._not_found()
-        if job["state"] != "queued":
+        if job_entry["state"] != "queued":
           send_json(self, HTTPStatus.CONFLICT, {"error": "cannot cancel running or finished job"})
           return
-        job["state"] = "canceled"
-        job["finishedAt"] = now_iso()
-      send_json(self, HTTPStatus.OK, job_public(job))
+        job_entry["state"] = "canceled"
+        job_entry["finishedAt"] = now_iso()
+      send_json(self, HTTPStatus.OK, job_public(job_entry))
       return
 
     self._not_found()


captainpacket:
This is primarily an automation I built to hook netlab into Semaphore UI

DanPartelly:
Thank you ! I certainly see some future in this. 

Could you please write some words describing the problem you are trying to solve with this commit?  So we can start from a common understanding?

captainpacket:
Engineers here need to model and test lots of different vendors and networks.  We've had some internal tooling in the past to do this manually (some hack expect style scripts), but have switched to netlab to make the environment a bit more composable.  

We started using Semaphore UI (https://semaphoreui.com/) as a job execution engine, which works well with the built in terraform/ansible worklows in a multi-user environment.  It can also do manual bash/python scripts.  I'd started with bash scripts with sshpass, but found them kinda brittle.  I wanted to switch to an API based workflow to make it a bit more robust.

I'd initially wrote a FastAPI based wrapper that imports these functions (which I'm still using), but I had the thought that this could be easily extended in the native CLI workflow.  A typed version I think still has use and this could be extended for that.

ipspace:
@captainpacket -- Thanks a million for the PR. Definitely an interesting idea ;) and I truly appreciate you crossing all the Ts and dotting all the Is.

Give me a few days -- I'm stuck in the "we have to deal with another Ansible brokenness" bog, and have to get out of that first.

captainpacket:
Cleaned up a few semaphore specific things (project/task) that I'd left behind; added minimal authentication and TLS option

docs/netlab/cli.md
```diff
@@ -64,6 +65,7 @@ The **netlab** command is the *netlab* CLI interface. It includes data model tra
    netlab config <config.md>
    netlab connect <connect.md>
    netlab create <create.md>
+   netlab api <api.md>
```
ipspace: I think this list is sorted, so "netlab api" should be at the very top

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+from typing import Any, Callable, Dict, List, Optional
+
+from ..utils import log
+from . import collect as netlab_collect
+from . import create as netlab_create
+from . import down as netlab_down
+from . import status as netlab_status
+from . import up as netlab_up
+
+DATA_DIR = os.getenv("NETLAB_API_DATA_DIR", "/var/lib/netlab/api")
```
ipspace: I don't think this is user-writeable on most systems, so the HTTP server would have to be started as root. I usually use "/tmp/something" in these cases (or you could use "~/...")

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+  data = json.dumps(payload).encode("utf-8")
+  handler.send_response(HTTPStatus.UNAUTHORIZED)
+  handler.send_header("WWW-Authenticate", 'Basic realm="netlab api"')
+  handler.send_header("Content-Type", "application/json")
+  handler.send_header("Content-Length", str(len(data)))
+  handler.end_headers()
+  handler.wfile.write(data)
+
+
+def require_auth(handler: BaseHTTPRequestHandler) -> bool:
```
ipspace, line 102: Just wondering: isn't there a Python library to handle this?

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+    return True
+
+  send_unauthorized(handler)
+  return False
+
+
+def workspace_dir(payload: Dict[str, Any]) -> str:
+  workdir = (payload.get("workdir") or "").strip()
+  if workdir:
+    if os.path.isabs(workdir):
```
ipspace: It looks to me like you're trying to get an absolute path to the working directory. If that's the case you could probably use `payload.get("workdir",payload.get("workspaceRoot",os.getcwd()))` to get the path you want to get or loop through the keywords instead of effectively having the same code twice.

Also, I'd usually use `pathlib.Path(path).resolve()` to get an absolute path.

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+
+  root = (payload.get("workspaceRoot") or "").strip()
+  if root:
+    return root
+
+  return os.getcwd()
+
+
+def resolve_topology(payload: Dict[str, Any], workdir: str) -> str:
+  topology_url = (payload.get("topologyUrl") or "").strip()
```
ipspace: As you're using the same expression so many times, it might make sense to turn it into a function

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+    return topology_url
+  topology_path = (payload.get("topologyPath") or "").strip()
+  if not topology_path:
+    for candidate in ("netlab/topology.yml", "netlab/topology.yaml"):
+      if os.path.exists(os.path.join(workdir, candidate)):
+        topology_path = candidate
+        break
+  if not topology_path:
+    raise ValueError("topologyPath or topologyUrl required")
+  if os.path.isabs(topology_path):
```
ipspace: Yet again, `str(pathlib.Path(path).resolve())` would probably do the trick

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+    return topology_path
+  return os.path.join(workdir, topology_path)
+
+
+def list_templates(template_dir: str) -> List[Dict[str, str]]:
+  results: List[Dict[str, str]] = []
+  if not template_dir:
+    return results
+  for root, dirs, files in os.walk(template_dir):
```
ipspace: It looks like you're doing `list(glob('**/*.yml')) + list(glob('**/*yaml'))`. You could use `glob` from os.path or from pathlib

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+      topology = resolve_topology(payload, workdir)
+      _run_with_output(netlab_create.run, [topology])
+    elif action == "down":
+      args: List[str] = []
+      if payload.get("cleanup"):
+        args.append("--cleanup")
+      _run_with_output(netlab_down.run, args)
+    elif action == "collect":
+      args = []
+      if payload.get("instance"):
```
ipspace: I would probably do this with a dictionary mapping payload variable names into netlab args, and use the same approach for all commands.

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+        args = ["--instance", str(payload.get("instance"))]
+      _run_with_output(netlab_status.run, args)
+    else:
+      raise ValueError(f"unknown action {action}")
+  finally:
+    os.chdir(prev_cwd)
+
+
+def start_job(payload: Dict[str, Any]) -> Dict[str, Any]:
```
ipspace, line 238: Can you start jobs in parallel? If that's the case, you need multilab plugin, or you have to serialize the jobs.

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+  def do_GET(self) -> None:
+    if not require_auth(self):
+      return
+    parsed = urllib.parse.urlparse(self.path)
+    parts = [p for p in parsed.path.strip("/").split("/") if p]
+    if not parts:
+      send_json(self, HTTPStatus.NOT_FOUND, {"error": "not found"})
+      return
+
+    if parts == ["healthz"]:
```
ipspace: I would usually implement something like this with a dictionary mapping keywords (`parts` in you case) into functions that do the actual work.

netsim/cli/api.py
```diff
@@ -0,0 +1,384 @@
...
+        job["state"] = "canceled"
+        job["finishedAt"] = now_iso()
+      send_json(self, HTTPStatus.OK, job)
+      return
+
+    send_json(self, HTTPStatus.NOT_FOUND, {"error": "not found"})
+
+
+def run(cli_args: List[str]) -> None:
+  parser = argparse.ArgumentParser(description="netlab API server")
```
ipspace, line 404: All other CLI commands have parser in a separate function to keep the "run" function shorter (and cleaner)

ipspace (changes requested):
Mostly minor details and improvement suggestions, the only showstopper is whether jobs can be run in parallel and how to handle that.

docs/netlab/cli.md
captainpacket: Good catch — moved netlab api to the top so the list stays sorted.

netsim/cli/api.py
captainpacket: Agreed — changed the default data/log dir to a user-writable temp location; still overrideable with NETLAB_API_DATA_DIR.

netsim/cli/api.py
captainpacket: I couldn’t find a clean stdlib helper for server-side Basic Auth parsing; kept stdlib-only to avoid extra deps and refactored it into a small helper to keep the handler clean.

netsim/cli/api.py
captainpacket: Refactored workdir/workspaceRoot selection and switched to pathlib.Path(...).resolve() for absolute paths; removed duplicated branches.

netsim/cli/api.py
captainpacket: Done — factored the repeated path/payload handling into helper functions.

netsim/cli/api.py
captainpacket: Switched template discovery to pathlib globbing for *.yml/*.yaml

netsim/cli/api.py
captainpacket: Aligned with other CLI modules — moved parser construction into a separate api_parse_args() function so run() stays short.

netsim/cli/api.py
captainpacket: Good point; I’m serializing job execution (queueing requests but only running one netlab action at a time). If we later want true parallel jobs, we should detect/require multilab and unique ID per job rather than reinventing isolation.

netsim/cli/api.py
captainpacket: Refactored routing to a dict mapping endpoints/keywords to handler functions (kept the /jobs/<id>/... special-case)

netsim/cli/api.py
captainpacket: Yep — path normalization now uses str(Path(...).resolve()) via the helper.

netsim/cli/api.py
captainpacket: Agreed — refactored payload→argv construction to a consistent helper-based approach across actions.

ipspace:
@captainpacket -- Looks like you overcomplicated typing definitions and made mypy unhappy (CI check has failed). It could be a real mismatch (in which case it needs to be fixed), or just a confusing type (I would something go just with typing.List or typing.Callable without specifying the details).

captainpacket:
> @captainpacket -- Looks like you overcomplicated typing definitions and made mypy unhappy (CI check has failed). It could be a real mismatch (in which case it needs to be fixed), or just a confusing type (I would something go just with typing.List or typing.Callable without specifying the details).

Ahh yeah - I overspecified the typing a  bit.  I loosened it up and mypy passes locally.

PR #3003, opened 2026-01-06
Add Netscaler device & Dockerfile j2 template

sdargoeuves:
Following the discussion https://github.com/ipspace/netlab/discussions/2732#discussioncomment-14666829, where we mentioned adding a `netscaler.yml` file under *devices*, so we can set the user/password in the `netscaler.yml` device file when creating the docker image.

This branch contains:

1. add function to render Dockerfile.j2 template (`build.py`)

- Allows netlab clab build to detect and render .j2 Dockerfile templates
- render_j2_dockerfile() function loads system defaults and renders templates with device variables
- Includes fail() function for template validation with clear error messages

2. create netscaler Dockerfile.j2 template(`Dockerfile.j2`)

- Reads ansible_user and ansible_ssh_pass from device YAML at build time
- Validates required credentials during template rendering (fails with clear error if missing)
- Interactive SSH works (netlab connect), show commands functional (netlab connect -s)

3. Add basic support for netscaler (`netscaler.yml`)

- Minimal device support: readiness-check, deploy-config (only hostname and interface descriptions)

Here is a simple topology with a netscaler:

```yml
---
message: "Citrix ADC (VPX)"

defaults.devices.eos.clab.image: "ceos:4.29.9.1M"
provider: clab

groups:
  switches:
    _auto_create: true
    device: eos
    members: [ sw1, sw2 ]

nodes:
  lb:
    image: netlab/netscaler # 13.1 - working OK
    device: netscaler
    memory: 4096
    cpu: 2

links:
  - sw1:
    lb:
  - lb:
    sw2:
```

And it works smoothly:
```bash
╰─❯ dnetlab connect lb -s ns runnin
Connecting to container clab-ml_4-lb, executing nscli -U 127.0.0.1:clab:clab@123 "show ns runnin"
 Done
#NS13.1 Build 60.29
# Last modified Tue Jan  6 11:31:35 2026
set ns config -IPAddress 192.168.4.201 -netmask 255.255.255.0
[...]
set ns hostName lb
set interface 0/1 -haHeartbeat OFF -throughput 0 -bandwidthHigh 0 -bandwidthNormal 0 -intftype Linux -ifnum 0/1
set interface 0/2 -speed 1000 -duplex FULL -throughput 0 -bandwidthHigh 0 -bandwidthNormal 0 -intftype Linux -ifnum 0/2
set interface 0/3 -speed 1000 -duplex FULL -ifAlias "lb -> sw1" -throughput 0 -bandwidthHigh 0 -bandwidthNormal 0 -intftype Linux -ifnum 0/3 -lldpmode TRANSCEIVER
set interface 0/4 -speed 1000 -duplex FULL -ifAlias "lb -> sw2" -throughput 0 -bandwidthHigh 0 -bandwidthNormal 0 -intftype Linux -ifnum 0/4 -lldpmode TRANSCEIVER
[...]
```

Interfaces from the GUI:
image

IP from the GUI:
image

I also have an ugly playbook that builds all the Servers, Service Groups, Services and Virtual Servers, but I don't know if it has a space in `netlab` yet.

Final point, I haven't done any documentation, but I wanted to discuss first what I'm proposing before working on this.
@ipspace @jbemmel please let me know your thoughts

### DIFF
--- a/netsim/ansible/tasks/deploy-config/netscaler.yml
+++ b/netsim/ansible/tasks/deploy-config/netscaler.yml
@@ -3,16 +3,17 @@
 
 - name: "Netscaler: deploying {{ netsim_action }} from {{ config_template }}"
   ansible.builtin.raw: |
-    cat > /tmp/deploy.sh << 'NETSCALER_EOF'
-    #!/bin/bash
+    AUTHENTICATE=127.0.0.1:{{ ansible_user }}:{{ ansible_ssh_pass }}
     while IFS= read -r line; do
-      [ -n "$line" ] && nscli -U 127.0.0.1:{{ ansible_user }}:{{ ansible_ssh_pass }} "$line" || true
+      [ -n "$line" ] && nscli -U "$AUTHENTICATE" "$line" || true
     done << 'CONFIG_EOF'
     {{ deployed_config }}
     CONFIG_EOF
-    nscli -U 127.0.0.1:{{ ansible_user }}:{{ ansible_ssh_pass }} "save ns config"
-    NETSCALER_EOF
-    chmod +x /tmp/deploy.sh
-    sh /tmp/deploy.sh
+
+    if ! nscli -U "$AUTHENTICATE" "save ns config"; then
+      echo "Save failed waiting..."
+      sleep 5
+      nscli -U "$AUTHENTICATE" "save ns config"
+    fi
   when: not ansible_check_mode
   tags: [ print_action, always ]
--- a/docs/labs/clab.md
+++ b/docs/labs/clab.md
@@ -40,6 +40,7 @@ Lab topology file created by **[netlab up](../netlab/up.md)** or **[netlab creat
 | Cisco IOSv             | vrnetlab/cisco_vios:15.9.3   |
 | Cisco IOS XRd          | ios-xr/xrd-control-plane:7.11.1 |
 | Cisco Nexus OS         | vrnetlab/vr-n9kv:9.3.8       |
+| Citrix Netscaler       | netlab/netscaler:latest      |
 | Cumulus VX             | networkop/cx:4.4.0           |
 | Cumulus VX with NVUE   | networkop/cx:5.0.1           |
 | Dell OS10              | vrnetlab/vr-ftosv            |
@@ -58,8 +59,8 @@ Lab topology file created by **[netlab up](../netlab/up.md)** or **[netlab creat
 | VyOS                   | ghcr.io/sysoleg/vyos-container |
 
 * Cumulus VX, FRR, Linux, Nokia SR Linux, and VyOS images are automatically downloaded from public container registries.
-* Build the BIRD and dnsmasq images with the **netlab clab build** command.
-* Arista cEOS image has to be [downloaded and installed manually](ceos.md).
+* Build the BIRD, dnsmasq, and Netscaler images with the **netlab clab build** command.
+* The Arista cEOS image has to be [downloaded and installed manually](ceos.md).
 * Nokia SR OS and SR-SIM container images require a license; see also [vrnetlab instructions](https://containerlab.srlinux.dev/manual/vrnetlab/).
 * Follow Cisco's documentation to install the IOS XRd container, making sure the container image name matches the one _netlab_ uses (alternatively, [change the default image name](default-device-image) for the IOS XRd container).
 * Cisco 8000v containerlab image (once you manage to get it) has to be  [installed](https://containerlab.dev/manual/kinds/c8000/#getting-cisco-8000-containerlab-docker-images) with the **docker image load** command.
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -39,6 +39,7 @@
 | vJunos-router [❗](caveats-vjunos-router) | vjunos-router | full |
 | Mikrotik RouterOS 6 (CHR) [❗](caveats-routeros6) | routeros           | end of life |
 | Mikrotik RouterOS 7 (CHR) [❗](caveats-routeros7) | routeros7           | minimal |
+| Netscaler CPX | netscaler | minimal |
 | Nokia SR Linux [❗](caveats-srlinux) | srlinux | full       |
 | Nokia SR OS [❗](caveats-sros)    | sros    | best effort[^SROSBE]   |
 | Nokia SR-SIM [❗](caveats-srsim)  | srsim   | full          |
@@ -130,6 +131,7 @@ You cannot use all supported network devices with all virtualization providers. 
 | vJunos-switch       |  ❌  |  ❌  | ✅[❗](clab-vrnetlab) |
 | vJunos-router       |  ❌  |  ❌  | ✅[❗](clab-vrnetlab) | | Mikrotik RouterOS 6 | ✅  |  ❌  |  ❌  |
 | Mikrotik RouterOS 7 | [✅](build-chr7)  |  ❌  |  ❌  |
+| Netscaler CPX       |  ❌  |  ❌  | ✅  | 
 | Nokia SR Linux      |  ❌  |  ❌  | ✅  |
 | Nokia SR OS         |  ❌  |  ❌  | ✅  |
 | Nokia SR-SIM        |  ❌  |  ❌  | ✅  |
@@ -193,6 +195,7 @@ Ansible playbooks included with **netlab** can deploy and collect device configu
 | Junos[^Junos]         | ✅ | ✅ |
 | Mikrotik RouterOS 6   | ✅ | ✅ |
 | Mikrotik RouterOS 7   | ✅ | ✅ |
+| Netscaler CPX         | ✅ | ❌  |
 | Nokia SR Linux        | ✅ | ✅ |
 | Nokia SR OS[^SROS]    | ✅ | ✅ |
 | OpenBSD               | ✅ | ❌  |
@@ -229,6 +232,7 @@ The following system-wide features are configured on supported network operating
 | Junos[^Junos]            | ✅  |  ❌  | ✅  | ✅  | ✅  | 
 | Mikrotik RouterOS 6      | ✅  | ✅  | ✅[❗](caveats-routeros6) | ✅ | ✅ |
 | Mikrotik RouterOS 7      | ✅ | ✅ | ✅[❗](caveats-routeros7) | ✅ | ✅ |
+| Netscaler CPX            | ✅  |  ❌  |  ❌  |  ❌  |  ❌  |
 | Nokia SR Linux           | ✅  | ✅  | ✅  | ✅  | ✅  |
 | Nokia SR OS[^SROS]       | ✅  | ✅  | ✅  | ✅  | ✅  |
 | OpenBSD                  | ✅  | ✅  |  ❌  | ✅  | ✅  |
@@ -284,6 +288,7 @@ The following interface addresses are supported on various platforms:
 | Junos[^Junos]         | ✅  | ✅  | ✅  |  ❌  |
 | Mikrotik RouterOS 6   | ✅  | ✅  |  ❌  |  ❌  |
 | Mikrotik RouterOS 7   | ✅  | ✅  |  ❌  |  ❌  |
+| Netscaler CPX         | ✅  |  ❌  |  ❌  |  ❌  |
 | Nokia SR Linux        | ✅  | ✅  |  ❌  |  ❌  |
 | Nokia SR OS[^SROS]    | ✅  | ✅  | ✅  |  ❌  |
 | OpenBSD               | ✅  | ✅  |  ❌  |  ❌  |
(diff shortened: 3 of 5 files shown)


netsim/ansible/tasks/deploy-config/netscaler.yml
```diff
@@ -0,0 +1,9 @@
+- name: "Netscaler: deploying {{ netsim_action }} from {{ config_template }}"
+  delegate_to: localhost
+  ansible.builtin.shell: |
```
ipspace: Assuming netscaler is running in a Linux-based container, we should execute this as a bash script in the container.

netsim/ansible/tasks/readiness-check/netscaler.yml
```diff
@@ -0,0 +1,19 @@
+---
+#
+# Netscaler readiness check - wait for nscli to be available
+#
+- name: Check {{ netlab_device_type|default(inventory_hostname) }} nscli readiness
+  delegate_to: localhost
+  ansible.builtin.shell: |
```
ipspace: I think you can just use "shell" with Docker connection to get the same results. No need to delegate to localhost and then use docker exec. See what we're doing with FRR containers.

netsim/ansible/templates/initial/netscaler.j2
```diff
@@ -0,0 +1,5 @@
+set ns hostName {{ inventory_hostname.replace("_","-") }}
+
+{% for l in interfaces|default([]) %}
+set interface {{ l.ifname }} -ifAlias "{{ l.name }}{{ " ["+l.role+"]" if l.role is defined else "" }}" -ifnum {{ l.ifname }} -lldpmode TRANSCEIVER
```
ipspace: I would sometimes insert newline after {{ or {% to break a single Jinja2 line into multiple lines. Add indent to the second and subsequent lines and you get something that's somewhat readable.

ipspace: Also, where do you set IP addresses on interfaces?

netsim/cli/clab_actions/build.py
```diff
@@ -44,14 +45,26 @@ def get_dockerfiles() -> dict:
   for d_file in d_list:
     daemon = os.path.basename(os.path.dirname(d_file))
     root, ext = os.path.splitext(d_file)
-    df_dict[daemon+ext] = d_file
+    
+    # For .j2 files, use the base daemon name without .j2 extension
+    # This allows "netlab clab build netscaler" to find "netscaler/Dockerfile.j2"
+    daemon_key = daemon if not ext == '.j2' else daemon
```
ipspace: This make little sense ;)

Anyway, the idea was to allow multiple versions of the daemon to be built (see **bird**), and I think you somewhat broke that concept. You should just remove trailing ".j2" as the daemon key.

netsim/cli/clab_actions/build.py
```diff
@@ -44,14 +45,26 @@ def get_dockerfiles() -> dict:
...
+    
+    # Store with full extension for internal use, but key by daemon name
+    # If both Dockerfile and Dockerfile.j2 exist, prefer .j2
```
ipspace: If both Dockerfile and Dockerfile.j2 exist then someone messed up pretty badly. Just take the last entry and use it.

netsim/cli/clab_actions/build.py
```diff
@@ -44,14 +45,26 @@ def get_dockerfiles() -> dict:
...
+    if daemon_key not in df_dict or ext == '.j2':
+      df_dict[daemon_key] = d_file
 
   return df_dict
 
 def get_description(dfname: str) -> str:
   try:
     df_lines = pathlib.Path(dfname).read_text().split('\n')
     for line in df_lines:
+      # # Skip Jinja2 directives
```
ipspace: If you commented this out, then it's probably not needed. Should it be removed?

netsim/cli/clab_actions/build.py
```diff
@@ -65,6 +78,53 @@ def get_description(dfname: str) -> str:
...
+  
+  # Jinja2 function to fail template rendering with a custom error message
+  def fail(msg: str) -> None:
+    raise ValueError(msg)
+  
+  # Load topology defaults to get device credentials
+  try:
+    defaults = _read.system_defaults().defaults
+  except Exception as ex:
+    log.error(f'Could not load topology defaults: {ex}', log.IncorrectValue, 'build')
```
ipspace: A) I'm usually using str(ex), but maybe that's not necessary?
B) If we can't read system defaults, we're in pretty bad shape, so I'd make it log.fatal.

netsim/cli/clab_actions/build.py
```diff
@@ -65,6 +78,53 @@ def get_description(dfname: str) -> str:
...
+  with netlab device defaults before building.
+  """
+  if not df_path.endswith('.j2'):
+    return df_path  # Regular Dockerfile, use as-is
+  
+  strings.print_colored_text('[TEMPLATE] ','cyan',None)
+  print(f"Rendering Jinja2 template from {os.path.basename(df_path)}")
+  
+  # Jinja2 function to fail template rendering with a custom error message
+  def fail(msg: str) -> None:
```
ipspace: Maybe you could just make this a standard netlab Jinja2 filter (in utils.filters). I'm pretty sure it could be handy ;)

netsim/cli/clab_actions/build.py
```diff
@@ -65,6 +78,53 @@ def get_description(dfname: str) -> str:
...
+      path=os.path.dirname(df_path)
+    )
+  except Exception as ex:
+    log.fatal(
+      f'Failed to render Dockerfile template {os.path.basename(df_path)}: {str(ex)}',
+      module='build')
+  
+  # Write to temp directory
+  rendered_path = os.path.join(tmp_dir, 'Dockerfile')
+  with open(rendered_path, 'w') as f:
```
ipspace: We have utils.templates.write_template

netsim/ansible/tasks/deploy-config/netscaler.yml
sdargoeuves: I couldn't remember why I went down that path, it's because the Netscaler runs two versions of python: `2.7` and `3.5`... well, let's just say, Ansible doesn't like that very much!

To make it easier, I can just use raw, it looks simpler than what I had:

```
- ansible.builtin.set_fact:
    deployed_config: "{{ lookup('template', config_template) }}"

- name: "Netscaler: deploying {{ netsim_action }} from {{ config_template }}"
  ansible.builtin.raw: |
    cat > /tmp/deploy.sh << 'NETSCALER_EOF'
    #!/bin/bash
    {{ deployed_config }}
    NETSCALER_EOF
    chmod +x /tmp/deploy.sh
    sh /tmp/deploy.sh
  when: not ansible_check_mode
  tags: [ print_action, always ]
```

netsim/ansible/tasks/readiness-check/netscaler.yml
sdargoeuves: it goes back to the same issue as above, the archaic python version doesn't work with Ansible. I could use raw as well:

```yaml
- name: Check {{ netlab_device_type|default(inventory_hostname) }} nscli readiness
  ansible.builtin.raw: nscli -U '127.0.0.1:{{ ansible_user }}:{{ ansible_ssh_pass }}' 'show ns hostName'
  changed_when: False
  register: command_out
  until: command_out.rc == 0 and "Done" in command_out.stdout
  retries: "{{ netlab_check_retries | default(60) }}"
  delay: "{{ netlab_check_delay | default(5) }}"
  failed_when: False

- name: Confirm {{ inventory_hostname }} is ready
  ansible.builtin.debug:
    msg: "Node {{ inventory_hostname }} is ready (nscli responding)."
  when: command_out.rc == 0
```

netsim/ansible/templates/initial/netscaler.j2
sdargoeuves: at the moment you don't 😅

More seriously, after double checking the information on the [Netscaler website](https://docs.netscaler.com/en-us/citrix-adc/13-1/networking/interfaces/configuring-vlans/associate-subnet-with-interface-using-vlans.html) you have to do:

- Add a VLAN
- Bind an interface to the VLAN.
- Bind an IP address to the VLAN

Luckily, this is what we are doing with our playbook (that's a relief!)
This is the output of the `show ns run`:

```
add vlan 100 -aliasName vlan_client
add vlan 200 -aliasName vlan_server
add ns ip 172.16.30.1 255.255.255.128 -vServer DISABLED
add ns ip 172.16.30.129 255.255.255.128 -vServer DISABLED
bind vlan 100 -ifnum 0/3
bind vlan 100 -IPAddress 172.16.30.1 255.255.255.128
bind vlan 200 -ifnum 0/4
bind vlan 200 -IPAddress 172.16.30.129 255.255.255.128
```

So I could add this as part of the template, any idea what vlanID should I use? Or where to get the vlanID from?

netsim/cli/clab_actions/build.py
sdargoeuves: indeed, not sure what I was trying to do there either. I understand better now what that part was doing before. I'll update shortly

netsim/cli/clab_actions/build.py
sdargoeuves: do we want this as a filter:
`{{ None | fail(msg) }}`
or a global function:
`{{ fail(msg }}`

According to Claude:
> If you want to use it as `{{ fail('msg') }}` without the pipe, we need to make it a global function instead of a filter. Filters always receive the piped value as their first argument

In templates.py we could then add this in the add_j2_filters(), even though it's no longer technically a filter.
```python
  ENV.globals['fail'] = filters.j2_fail                    # Add fail() as a global function
```

I will go with the first option, keeping it a filter.

netsim/ansible/tasks/deploy-config/netscaler.yml
sdargoeuves: the above didn't work, I forgot to add the prefix at the beginning of each lines: `nscli -U 127.0.0.1:{{ ansible_user }}:{{ ansible_ssh_pass }}`

netsim/ansible/templates/initial/netscaler.j2
```diff
@@ -0,0 +1,15 @@
+
+set ns hostName {{ inventory_hostname.replace("_","-") }}
+
+{% for l in interfaces|default([]) %}
+set interface {{ l.ifname }} -ifAlias "{{ l.name }}{{
+  " ["+l.role+"]" if l.role is defined else "" }}" -ifnum {{
+  l.ifname }} -lldpmode TRANSCEIVER
+{% if l.ipv4 is defined %}
+{%   set vlan_id = 3990 + loop.index %}
```
sdargoeuves: this was more a test, I do not know what *good* value we should use, or how to set a variable in the device node, or interface node, to specify a value.

netsim/ansible/templates/initial/netscaler.j2
sdargoeuves: current "[solution](https://github.com/ipspace/netlab/pull/3003/changes#diff-d51e25b95d16ce27beb09487e332583c0605417d08beb9f95ec1b26e9f691668R99)" but it's not pretty, the Vlan is hardcoded...

netsim/cli/clab_actions/build.py
ipspace: > I will go with the first option, keeping it a filter.

Having a global function would make more sense (as in the templates look less convoluted). Obviously, you could register it as both and then use the filter as `msg|fail`

netsim/ansible/templates/initial/netscaler.j2
ipspace: I would go for `3990 + l.ifindex`

ipspace:
Anyway, tell me when the code is somewhat ready to be merged. I'll merge it and then maybe do a bit of polishing:

* Use the new "run scripts in containers" capability
* Assign VLANs to interfaces somewhere in Python code

netsim/ansible/templates/initial/netscaler.j2
sdargoeuves: thanks, obviously, happy to change this the day someone has a better idea on what we should use. For now it will do. I'll push the update

netsim/cli/clab_actions/build.py
sdargoeuves: I've gone with the global option, this is how you can now use it:

```
{% if not _user %}{{ fail('ansible_user must be defined in devices.netscaler.clab.group_vars') }}{% endif %}
```

sdargoeuves:
I'm quite happy with the current state, I've ran some more tests, it works great with two different versions of `13.1` but with the latest `14.1`, *netlab* completes successfully, the loadbalancer is configured as expected, I am just not able to access the GUI, and I can't figure out why. Let's just say it works fine with `13.1`, and no GUI with `14.1`

ipspace:
@sdargoeuves -- Submitted my changes as a PR in your repo. If you're OK with them and merge them, they might appear here ;)

sdargoeuves:
Excellent, thank you so much for completing, tidying and documenting everything here!
I've learned a lot from this process.

PR #3004, opened 2026-01-06
CLI `netlab status` allow `--json` for API handling

sdargoeuves:
I have no idea if this is something we want to consider. After discovering the new HTTP server, I wanted to test and thought it was a shame to not a have readable output in the browser.

If we are happy to go down this rabbit hole, I don't mind checking what other commands we could json-ify, if not feel free to close this, and leave that for a distant future!

### Overview
Added `--json` flag to `netlab status` for programmatic access, improving API usability and automation capabilities.

#### status.py
- Added `--json` argument to CLI parser
- Created data collection layer: `build_*_payload()` functions that construct structured data once
  - `build_active_labs_payload()` - all labs overview
  - `build_lab_detail_payload()` - single lab with nodes/tools
  - `build_lab_log_payload()` - lab event log
- Display functions now consume payloads instead of reimplementing logic
- Fixed duplicate snapshot file reads

#### api.py
- Using the --json when calling the `netlab status` command
- Added option to specify instance ID or ALL with `/status/<id/all>`

#### CLI

```bash
# Table (default)
netlab status

# JSON (new)
netlab status --json
netlab status --json --instance 1
netlab status --json --all
```

#### HTTP

image

### DIFF
--- a/netsim/cli/status.py
+++ b/netsim/cli/status.py
@@ -192,7 +192,7 @@ def build_lab_detail_payload(args: argparse.Namespace, lab_states: Box, topology
   iid = get_instance(args,lab_states)
   lab_state = lab_states[iid]
   lab_status = get_lab_status(lab_state)
-  payload = {
+  payload: typing.Dict[str, typing.Any] = {
     "lab": {
       "id": iid,
       "directory": lab_state.dir,
@@ -377,9 +377,10 @@ def run(cli_args: typing.List[str]) -> None:
       iid = get_instance(args,lab_states)
       wdir = lab_states[iid].dir
       snapshot = f'{wdir}/netlab.snapshot.yml'
-      topology = _read.read_yaml(filename=snapshot)
-      if topology is None:
+      topology_snapshot = _read.read_yaml(filename=snapshot)
+      if topology_snapshot is None:
         log.fatal(f'Cannot read topology snapshot file {snapshot}')
-      output_json(build_lab_detail_payload(args,lab_states,topology))
+        return
+      output_json(build_lab_detail_payload(args,lab_states,topology_snapshot))
     else:
       show_lab(args,lab_states)


netsim/cli/status.py
```diff
@@ -80,31 +86,39 @@ def display_active_labs(topology: Box,args: argparse.Namespace,lab_states: Box)
...
-  print()
-
-def load_provider_status(p_status: dict, provider: str, topology: Box) -> None:
-  p_module = providers.get_provider_module(topology,provider)
-
-  if not provider in p_status:
-    p_status[provider] = p_module.call('get_lab_status') or get_empty_box()
-
-def show_lab_nodes(topology: Box) -> None:
+def normalize_providers(providers: typing.Any) -> typing.List[str]:
```
ipspace, line 96: This should not be necessary. Hopefully we can agree (between ourselves) what data format to use ;)

netsim/cli/status.py
```diff
@@ -253,10 +361,26 @@ def run(cli_args: typing.List[str]) -> None:
     sys.exit(1)
 
   if args.all:
-    display_active_labs(topology,args,lab_states)
+    if args.json:
```
ipspace, line 364: I would love to keep this decision tree clean - one parameter, one action, and then handle the specifics in the action

netsim/cli/status.py
```diff
@@ -253,10 +361,26 @@ def run(cli_args: typing.List[str]) -> None:
...
     cleanup_lab(topology,args,lab_states)
   elif args.log:
-    show_lab_log(args,lab_states)
+    if args.json:
+      output_json(build_lab_log_payload(args,lab_states))
+    else:
+      show_lab_log(args,lab_states)
   else:
-    show_lab(args,lab_states)
+    if args.json:
```
ipspace, line 376: Every action should follow the same blueprint -- get data, create a data structure out of that, and print the data structure in table, YAML, or JSON format (see "show" commands).

Having a totally different approach to collect data for the JSON format makes little sense.

netsim/cli/status.py
```diff
@@ -80,31 +86,39 @@ def display_active_labs(topology: Box,args: argparse.Namespace,lab_states: Box)
...
+  if not providers:
+    return []
+  if isinstance(providers,(list,tuple)):
+    return list(providers)
+  # Handle comma-separated string
+  if isinstance(providers,str):
+    return [p.strip() for p in providers.split(',') if p.strip()]
+  return [str(providers)]
+
+def build_active_labs_payload(lab_states: Box) -> dict:
```
ipspace, line 107: So you're taking a data structure that we built somewhere and create a different data structure for the JSON printout? That makes no sense. We should get it right in the first place.

Also, please use Box, not dict. The code will become much easier to read (that's why I started using Box in the first place). There are two utility functions in netsim.data: get_box to convert a dict into a Box and get_empty_box that returns an empty Box. Do not create boxes by calling Box directly, the two utility functions set just the right combination of flags.

sdargoeuves:
> A few initial comments. I think it would make sense if you do another pass and simplify things, and then I'll look at them some more.
> 
> Most important: there should be a single data-gathering function, and the resulting data structure should be used directly for the JSON dump (potentially with an extra wrapper around a list), or to generate the text printout.

Thanks (and sorry) for taking the time to review this.
I got excited about the idea, and carried away trying this out with AI. While it worked, I can now see I’d taken on more than I could chew: I don’t understand the `netlab status` internals to do it cleanly, and the result is overcomplicated and too hacky.

I’ve got a few other things on the back burner right now, so I’d prefer to park/close this PR for the moment and come back in ~a month to do a clean refactor if you think the idea of adding a `--json` option is worth it.

*Lesson learned: I’ll be more careful about raising this kind of AI experiments on something that seems cool, without correctly understanding the changes.*

ipspace:
> _Lesson learned: I’ll be more careful about raising this kind of AI experiments on something that seems cool, without correctly understanding the changes._

Been there, done that ;) No worries, I love the idea. I hope you don't mind if I just go ahead and implement it?

ipspace:
One other thing (I thought I wrote that down, but evidently not) -- these changes would break the exiting API introduced in #2969. @captainpacket -- Are you OK with that, or should we return existing text in "status" field and a copy of the data in JSON-encoded format?

sdargoeuves:
> I love the idea. I hope you don't mind if I just go ahead and implement it?

I'm glad you like the idea as well, and yes, please feel free to implement it! I'll be looking at how you solve this, it should help me better understand that part of the codebase.

> these changes would break the exiting API introduced in https://github.com/ipspace/netlab/pull/2969. @captainpacket -- Are you OK with that, or should we return existing text in "status" field and a copy of the data in JSON-encoded format?

@captainpacket could you please share more information on how you are using this API?

Before seeing that PR, I was thinking it would be **nice to have** something that lets other team members view lab status and potentially take actions without needing CLI access. Our main lab is currently poorly managed by a*n unreadably long* bash script that gets the job done but isn't ideal. This HTTP server could be a great foundation for building a more user-friendly wrapper around netlab.

That said, this could grow in scope pretty quickly depending on where we want to take it. It comes down to how much we want to solve this problem — is it worth diving deeper into, or are there bigger fish to fry?

captainpacket:
@sdargoeuves I wrote a small FastAPI wrapper to solve the same problem you did - a lot of bash trickery.  The PR I submitted is a simplified version of that; it implements some extra things like multilab, graphs, etc.  I'm using a job execution engine (Semaphore UI) with a small client runner script to access the different functions, but it could easily have an actual frontend.

Feel free to use - https://github.com/forwardnetworks/netlab/blob/dev/netlab_api.py

ipspace:
Replaced by #3038

Issue #2999, opened 2026-01-04
[DOC] Document debugging flags


ipspace:
Some people (particularly those writing plugins or templates) might benefit from debugging printouts. Document the debugging options and which commands can use them.

PR #3005, opened 2026-01-07
Describe debugging flags

ipspace:
Implements #2999


ipspace:
Adding potential reviewers in case someone wants to look at this ;)

sdargoeuves:
Thank you for this, it's very useful indeed. I've only used the `-v` or with multiple `v`, but never the `debug` options.

PR #3021, opened 2026-01-14
fix netlab status to use netlab.snapshot.pickle

sdargoeuves:
I have an issue where I can't run `netlab status` on the version 26.01, as the code is looking for the netlab.snapshot.yml, but it no longer exists.

```
╰─❯ netlab status
Lab 3 in /home/sa/code/quick-netlab-lab/cisco-sdwan-lab
  status: started
  provider(s): clab

[FATAL]   Cannot read topology snapshot file /home/sa/code/quick-netlab-lab/cisco-sdwan-lab/netlab.snapshot.yml
```

This is a suggestion that works when I did my quick test.

### DIFF
--- a/netsim/cli/status.py
+++ b/netsim/cli/status.py
@@ -163,9 +163,9 @@ def show_lab(args: argparse.Namespace,lab_states: Box) -> None:
   lab_state = lab_states[iid]
   show_lab_instance(iid,lab_state)
   wdir = lab_state.dir
-  snapshot = f'{wdir}/netlab.snapshot.yml'
+  snapshot = f'{wdir}/netlab.snapshot.pickle'
 
-  topology = _read.read_yaml(filename=snapshot)
+  topology = _read.load_pickled_data(snapshot)
   if topology is None:
     log.fatal(f'Cannot read topology snapshot file {snapshot}')
 


sdargoeuves:
I was about to raise an issue, I initially thought it was again something due to my setup. During my investigation, I realised it would be easier to fix than to raise!

ipspace:
Now for the interesting question: should I push out 26.01.01 immediately or wait for a day or so? Any thoughts? /cc @sdargoeuves @ssasso @steffann @jbemmel

ipspace:
> I was about to raise an issue, I initially thought it was again something due to my setup. During my investigation, I realised it would be easier to fix than to raise!

Good call ;)) Thanks again.

DanPartelly:
> Now for the interesting question: should I push out 26.01.01 immediately or wait for a day or so? Any thoughts? /cc @sdargoeuves @ssasso @steffann @jbemmel

Yes, IMO. Important enough.

ipspace:
> Yes, IMO. Important enough.

Thank you. Working on it...

ipspace:
26.01.01 is out. Thanks again @sdargoeuves and @DanPartelly

PR #3022, opened 2026-01-15
Implement default configuration deployment mode (sh/ns) for containers

ipspace:
This change makes it the definition of Linux deployment scripts less error-prone as the 'clab' provider creates all the necessary node_config entries.


ipspace:
@jbemmel -- what you asked for a while ago

ipspace:
Will change "clab.config_mode" to "netlab_config_mode" in clab.group_vars to make the setting available to templates (and make configuration-related settings consistent based on what I learned while trying to tame FRR)

PR #3042, opened 2026-01-27
BFD config on Bird uses a daemon configuration template

ipspace:


netsim/daemons/bird/bfd.j2
```diff
@@ -0,0 +1,10 @@
+{% if bfd is defined %}
```
jbemmel, line 1: Doesn't hurt, but shouldn't this be implied by the fact that this file is getting included?

ipspace: Yeah, that happens when you say "this is easy enough for an AI agent to solve" and then glance at it and have another AI agent review it 🤷‍♂️🤦‍♂️. Proves that they definitely lack the bigger picture.

jbemmel: AI is likely to implement what it is told to do - it depends on the prompt

PR #3049, opened 2026-01-29
Integrate internal SSH readiness checks with Ansible checks

ipspace:
This commit refactors the readiness checks to integrate the internal SSH readiness checks with the Ansible checks (for example, the check for the first Junos interface):

Data structure changes:

* The devices that require readiness checks MUST have netlab_ready group variable, which should include value 'ansible' for devices with Ansible checks. This commit modifies device definitions for all devices using readiness checks
* The 'ansible' output module creates netlab_ready_ansible and netlab_ready_ssh groups

netlab initial changes:

* The 'ready' module got its own run function which is invoked with the args.ready option
* The 'ready' module first collect the nodes based on their wait-for-ready requirements, executes internal readiness checks, and starts the Ansible 'device-ready' playbook if needed
* The internal readiness checks can be disable in topology defaults (defaults.netlab.initial.ready._check_ variable)
* deploy.run function calls ready.run function as one of the first steps
* deploy.run and ready.run functions use log.section_header for improved logging functionality

Ansible-related changes:
* The 'device-ready.ansible' playbook is split into two plays (SSH readiness and Ansible checks)
* The new ansible groups are used in 'device-ready.ansible' playbook to limit the hosts involved in each play
* The 'wait-for-ready' task list is no longer included into the initial-config.ansible playbook -- the readiness check is performed solely in the device-ready.ansible playbook
* The 'wait-for-ready' task list no longer performs the generic readiness checks (ssh was the only generic check). The generic checks are performed as plays in the 'device-ready.ansible' playbook

PR #3051, opened 2026-01-30
Use Linux scripts to configure Arista cEOS containers

ipspace:
Add 'netlab_config_wrapper' and 'netlab_config_mode' to EOS clab definition to turn configuration snippets into FastCli scripts

Changes in device configs:

* Fix weird EOS OSPFv3 configuration template typo
* Fast change in interface state confuses Arista cEOS. The EOS "spooky action at a distance" has to be disabled for cEOS containers
* Add 'sleep 2' to bgp.session configuration before clearing the BGP sessions to give BGP time to start
* Wrong comment character ("#" instead of "!") in STP template
* Unsupported command (ip virtual-router mac-address mlag-peer) in LAG template

Also:

* Transformation test results have changed for tests that include cEOS containers with netlab_config_mode == 'sh'

### DIFF
--- a/netsim/ansible/templates/stp/eos.j2
+++ b/netsim/ansible/templates/stp/eos.j2
@@ -27,7 +27,7 @@ spanning-tree vlan-id {{ vdata.id }} priority {{ stp.priority }}
 {%   if ifdata.vlan.trunk_id is defined or ifdata.vlan.access_id is defined %}
 interface {{ ifdata.ifname }}
 {%    if not ifdata.stp.enable|default(True) %}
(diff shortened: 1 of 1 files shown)


ipspace:
Adding a few reviewers in case someone wants to test-drive it (you have to change a topology default to enable the functionality). It will work even better after merging #3049 (and should be merged after #3049 in any case)

ipspace:
> This will become my new default

So glad to hear you like it. Thanks a million.

> as soon as it lands in dev.

Getting there as soon as we get past #3049 (hint, hint 😎)

PR #3055, opened 2026-02-01
Fix: Use random sleep instead of PHY control in cEOS normalization

ipspace:
Older EOS versions do not have the "platform tfa phy control-frame" command. Instead of using that command to prevent a race condition in fast-configured back-to-back cEOS containers, the normalize script uses "sleep" function for a random delay.

Also:

* The EOS wrapper uses simpler 'netlab_need_shebang' logic

### DIFF
--- a/netsim/ansible/templates/normalize/eos.j2
+++ b/netsim/ansible/templates/normalize/eos.j2
@@ -6,7 +6,7 @@
 {% if netlab_config_mode == 'sh' %}
 #!/bin/bash
 set -e
-cat <<CONFIG >/mnt/flash/00-normalize.cfg
+cat <<'CONFIG' >/mnt/flash/00-normalize.cfg
 configure terminal
 {% endif %}
 {% for intf in interfaces if intf.virtual_interface is not defined %}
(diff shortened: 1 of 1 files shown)


netsim/ansible/templates/normalize/eos.j2
```diff
@@ -6,16 +6,16 @@
 {% if netlab_config_mode == 'sh' %}
 #!/bin/bash
 set -e
-cat <<CONFIG >/mnt/flash/00-normalize.cfg
+cat <<'CONFIG' >/mnt/flash/00-normalize.cfg
```
jbemmel, line 9: There are 32 other places in the code that do this:
```
jeroen@j:~/Projects/netlab$ grep "cat" netsim/ansible/* -r | grep CONFIG
netsim/ansible/tasks/deploy-config/cumulus.yml:      cat <<DEPLOYED_CONFIG >{{ cfgname }}
netsim/ansible/templates/initial/cumulus.j2:cat > /etc/network/ifupdown2/policy.d/mtu.json <<CONFIG
netsim/ansible/templates/initial/cumulus.j2:cat >/etc/network/interfaces.d/10-loopback.intf <<CONFIG
...
```

sdargoeuves:
That's great, no issues when running on this branch, thank you!

ipspace:
> That's great, no issues when running on this branch, thank you!

Thanks a million for checking!

Issue #3070, opened 2026-02-04
[DOC] MPLS support for Juniper vSRX


sdargoeuves:
### Document we *might* need to fix

https://netlab.tools/platforms/

### What's wrong

I was browsing through the docs, like anyone would do before going to bed, and realised that netlab does not support MPLS for Juniper vSRX.
What is weird, is the lab we are using, is configured with the mpls module for our Juniper vSRX, and as far I am aware, it works fine.

```yaml
groups:
  vsrxcore:
    device: vsrx
    module: [ vrf, isis, bgp, mpls, ospf ]
    members: [ c0xr01, c0xr02, c0xr03, c0xr04, c0xr05, c0xr06 ]
    bgp.as: 65056
    mpls.ldp: true
    mpls.vpn: true
    vrfs:
      customer_vpn:
      mpls_core:
```

We do run a few playbooks on top of netlab to add extra configuration, fw policies... but I did not see anything related to the MPLS part, making me think it's supported by netlab.
And if it wasn't, it might have complained that the topology contains an error.

What do you think?

sdargoeuves:
This is the vagrant box we are using with those:

```bash
juniper/vsrx3               (libvirt, 21.3R1.9)
```

ipspace:
Thank you, will fix.

Root cause: I stopped checking what vSRX does after Juniper totally botched the image download process. @ssasso added MPLS to Junos after that, and vSRX inherited feature settings from junos device.

If you notice anything else along these lines, please let me know.

PR #3083, opened 2026-02-04
Doc: Update the MPLS support tables

ipspace:


ipspace:
Closes #3070. I also updated a few other entries based on integration test results (https://tests.netlab.tools/_html/coverage.mpls)

PR #3089, opened 2026-02-07
EVPN/VXLAN support on Cat8000v, IOL, IOLL2

ipspace:
Limitations:
* No support for symmetric IRB

### DIFF
--- a/netsim/devices/cat8000v.py
+++ b/netsim/devices/cat8000v.py
@@ -1,5 +1,5 @@
 #
-# Cisco IOS-XE quirks
+# Catalyst 8000v quirks
 #
 from box import Box
 
--- a/netsim/devices/iol.py
+++ b/netsim/devices/iol.py
@@ -17,8 +17,23 @@ def check_srv6_sid(node: Box, topology: Box) -> None:
       more_hints=['Set the srv6.allocate_loopback parameter to False'],
       category=log.IncorrectValue)
 
+def check_vni(node: Box, topology: Box) -> None:
+  if 'vxlan' not in node.get('module',[]):
+    return
+  for vname,vdata in node.get('vlans',{}).items():
+    vni = vdata.get('vni',None)
+    if not vni:
+      continue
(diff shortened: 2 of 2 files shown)


ipspace:
@DanPartelly -- in case you want to try it out. Symmetric IRB coming as soon as I overcome my disgust at the fact that they expect us to configure an explicit VLAN for the transit VNI.

DanPartelly:
On IOL 17.16.01 I have been unable to get it to work. Even in a simple [scenario](https://evpn.bgplabs.net/evpn/1-bridging/).  I'll run relevant tests tomorrow and see what the heck is going on on my side, as I'm sure you have run tests already, so the problem should be here.

ipspace:
> On IOL 17.16.01 I have been unable to get it to work.

I got that lab's `solution.yml` topology working with vrnetlab/cisco_iol:17.16.01a

> as I'm sure you have run tests already

All of them, on all three platforms. Do keep in mind that VXLAN only works with EVPN

DanPartelly:
Definitely a problem on my side, then. 

Yes, I saw the notes. I've only run scenarios with EVPN as the control plane, and I haven't attempted anything with a static ingress list. Netlab will catch those scenarios anyway, for it requires EVPN for VXLAN on the platform. I have tried  solution.yml  as well (it did not work here), to make sure I didn't mess up the configuration, but the results are the same and the config is identical for all practical purposes.

Now  I have to deal with another conundrum.

DanPartelly:
Valid VNI range on Cisco IOL/IOl2  is  4096-16,777,215.  And probably on other cisco  devices as well.  

We need some quirks to account for this.

s1(config-if)#member vni 1000 in
s1(config-if)#member vni 1000 ingress-replication ?
  local-routing  Enable vxlan centralized gateway forwarding for routed traffic
  <cr>           <cr>

s1(config-if)#member vni 1000 ingress-replication 
% Valid VNI Range is 4096-16777215.

PR #3113, opened 2026-02-18
VLAN support on generic IOS/XR

ipspace:


ipspace:
It looks like the L2 data plane is not implemented in XRd. Basic 2-port bridging works, but nothing else does.

PR #3117, opened 2026-02-19
Clarify delete.community.list attributes in routing policy

snuffy22:
Minor nitpick.

Make the attribute listing similar to the match how attributes are described elsewhere in the documentation.


ipspace (approved):
Thank you!

PR #3138, opened 2026-02-22
OpenBSD: Add BGP support

remilocherer:
This is a start for supporting OpenBSD's BGP daemon with netlab.

I'm not sure I understood features.bgp.community right. By default OpenBSD sends all communities. I implemented filters that drop communities if required.

I did not yet look into policies and redistribution.

Because vrnetlab currently uses 10.0.0.2 for the management interface (hidden) I was running tests like this:
"netlab up -s addressing.loopback.ipv4=10.100.0.0/24 test.yml"
--> https://github.com/srl-labs/vrnetlab/tree/master?tab=readme-ov-file#management-interface

### BGP integration tests passed

**tests/integration/bgp:**

01-ebgp-session.yml
02-ibgp-ebgp-session.yml
03-ibgp-rr.yml
--> one fail and one warning because of hardcoded 10.0.0.10 in test
04-originate.yml
--> three failis because of hardcoded 10.0.0.2 (loopback dut2) in test
05-community.yml
07-ebgp-localas.yml
11-ipv6-ebgp.yml
12-ipv6-ibgp-ebgp.yml
13-ipv6-ibgp-rr.yml
14-ipv6-originate.yml

**tests/integration/bgp.session:**

01-allowas-in.yml
02-as-override.yml
04-default-originate.yml
--> VRF-specific tests skipped
05-gtsm.yml
--> 10.0.0.17/32 hardcoded in tests. with the options I started the topology the prefix is 10.100.0.17/32.
--> somehow it takes rather long untile the session are established.
06-passive.yml
07-password.yml
10-timers.yml
--> tests succeed but are unreliable.
--> keepalive can not be configured explicitely. openbgpd uses holdtime / 3.
11-rs-client.yml
12-rs.yml

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -111,6 +111,7 @@ These devices support BGP local-AS functionality to build EBGP or IBGP sessions:
 | JunOS                    |  ✅ |  ✅ |
 | Nokia SR Linux           |  ✅ |  ✅ |
 | Nokia SR OS[^SROS]       |  ✅ |  ✅ |
+| OpenBSD                  |  ✅ |  ❌  |
 | Sonic                    |  ✅ |  ✅ |
 | VyOS                     |  ✅ |  ❌  |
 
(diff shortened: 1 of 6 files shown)


ipspace (approved):
Great job! Everything works (I tested OpenBSD as a VM), and even the documentation is updated ;))

Thanks a million!

PR #3141, opened 2026-02-24
Juniper cRPD improvements

leec-666:
Added and fixed a few things:

* junos.mplsvpn.j2 template failed in some cases due to absent `vrfs` data.  Added check.
* Can't ping from JunOS shell so it uses Linux /etc/hosts file scripts so pinging other nodes by hostname from the Linux shell is possible
* Uses Linux ping validation module for the same reason
* Enabled EVPN but only tested on a BGP route reflector with EVPN address family
* Tested and added VRRP support
* Tested and added VRF support.  Route leaking does not work correctly.

### DIFF
--- a/netsim/devices/crpd.yml
+++ b/netsim/devices/crpd.yml
@@ -17,6 +17,9 @@ loopback_interface_name: "lo0"
 mtu: 1500
 
 features:
+  initial:
+    ipv4:
+      unnumbered: false               # The 'unnumbered-address' family inet command does not work on cRPD
   vlan: false
   vxlan: false
   evpn:


leec-666:
VRRP integration tests revealed that cRPD doesn't work with EOS or FRR due to https://github.com/ipspace/netlab/commit/cb2359c8dade06c1530016f3a9cb09ac2bbf9438.  Peering with IOL also fails: packet captures from IOL show the same checksum errors.

I cannot find the same knob in cRPD and CRPD<->CRPD works.

ipspace:
Thank you! A few minor details:

* I will have to fix the "hosts file" bit (planned to do it anyway, but never got around to doing it)
* The "license file" caveat is probably incorrect (it was copied over from vMX), I think the parameter should be clab.license
* ruff is complaining about unused imports in validate plugin 🤷‍♂️ I'll see what I can do.

netsim/ansible/templates/mpls/junos.mplsvpn.j2
```diff
@@ -1,12 +1,13 @@
+{% if vrfs is defined %}
```
ipspace, line 1: Now I know why the Junos MPLS/VPN RR tests were failing ;) Thank you!

ipspace (approved):
Did a bit of cleanup and added device quirks.

New integration tests pass, we still have to deal with unnumbered interface configuration and VRF loopbacks. Will open issues for those two.

If you're OK with my changes we're good to merge ;)

leec-666:
Looks good, thanks!

Issue #3106, opened 2026-02-16
Support for EVPN-VXLAN in vJunosEvolved (vptx)


riccardo-roveri-labs:
Right now EVPN and VXLAN configuration module is only supported on vjunos-switch. For my understanding even if junos-evo is different from junos the basic evpn-vxlan configuration seem the same to me.

I am replicating my current topology that runs on qfx5130 that run evo and for me would be nice being able to use evo. Is there some limintation of virtualized junos-evo?

Thanks,
Riccardo

ssasso:
Ciao Riccardo,
It should work, but I never found the time to verify/test it.

If you want, you can try to enable the device feature flags, copy (or symlink) the proper templates, and run integration tests.

If they work as expected, you can create a PR for it ;)

riccardo-roveri-labs:
I copy pasted from vjunos-switch and with a few tweaks it seems to work with my topology. I will try to fork it and test it properly if i find the time.

Because 80% of the stuff is the same but not 100% what do you suggest doing? Is there a smarter solution than copy paste and mainain both file? I see that for other modules thare are just shared ones named junos.
Can i create a vptx.j2 and vjunos-switch.j2 with specific config and a junos.j2 with common config?

ipspace:
> I copy pasted from vjunos-switch and with a few tweaks it seems to work with my topology. I will try to fork it and test it properly if i find the time.

Worst case, upload what you have, and I'll take it from there.

> Because 80% of the stuff is the same but not 100% what do you suggest doing? Can i create a vptx.j2 and vjunos-switch.j2 with specific config and a junos.j2 with common config?

I would do something along those lines (I did the same for IOS/XE), but as I said, just upload (or submit a PR) with your changes, and I'll do the cleanup. I have no problem doing that, but don't have time to figure out how to make things work with Junos

PS: Remember the days when Juniper claimed how they're better than the competition (= Cisco) because they have one software that runs on all platforms?

ssasso:
> Can i create a vptx.j2 and vjunos-switch.j2 with specific config and a junos.j2 with common config?

In that case you need to "include" the common file in the platform files - if I remember correctly, if there is a more specific platform file then junos.j2 is never read.

ssasso:
> they have one software that runs on all platforms?

The software runs. It's the config that won't work 🤪

rickycraft:
> PS: Remember the days when Juniper claimed how they're better than the competition (= Cisco) because they have one software that runs on all platforms?

[...]

PR #3155, opened 2026-03-01
EVPN/VXLAN on Juniper vPTX

rickycraft:
Static VXLAN is implemented as per Juniper doc but the forwarding of BUM packet is not working. Pure L2 forwarding works but arp learing is broken.

On the EVPN side the same test as vjunos-switch pass exept:
`04-vxlan-central-routing.yml` -> due to problems related to BUM forwarding
`22-ospf-ce-router.yml` -> issue on the mtu exists fails also on vjunos-switch


rickycraft:
> However, I wonder why you decided to enable the anycast gateway support.
> 
> I will see how well that works in the "traditional" (VLAN) anycast scenario, if it doesn't then I'll limit it to EVPN only (I'll have to do something similar for Cisco IOS XE and XR anyway).

I wanted to use it in my test scenario, it makes sense limiting it only to EVPN.

ipspace:
Integration tests passed, merged the PR. Thanks again!!! and welcome to netlab contributors!

What's next? vJunos-router?

rickycraft:
> What's next? vJunos-router?

I will try and let you know how it goes.

I think now we can close the issue #3106

ssasso:
> I will try and let you know how it goes.

Keep in mind that vjunos-router uses _sp-style_ configuration for VLANs, so you will need to create vlan entries inside mac-vrf and add to the vlan itself the different sub-interfaces which belongs to it - removing it from the global bridge domains.

Sth like:
```
instance-type mac-vrf;
vlans { 
  red {
    vlan-id 100;
    interface ge-0/0/0.100;
    interface ge-0/0/5.100;
}}
```

ssasso:
> and welcome to netlab contributors!

Now we have the critical mass to organize a netlab hackathon for ITNOG 2027 ;)

ipspace:
> Now we have the critical mass to organize a netlab hackathon for ITNOG 2027 ;)

🎉🎉

PR #3158, opened 2026-03-03
Cisco IOS XRd management VRF

leec-666:
Enables a management VRF on Cisco IOS XRd nodes with `clab.node.env.CLAB_MGMT_VRF` variable to be used in [ContainerLab's startup config](https://github.com/srl-labs/containerlab/blob/main/nodes/xrd/xrd.cfg)


ipspace (approved):
Perfect. Thank you!

PR #3175, opened 2026-03-05
IOS XR AS-path sets and policy matches

ipspace:
Also: split route-policy template into multiple files for increased readability

### DIFF
--- a/netsim/ansible/templates/routing/iosxr/aspath.j2
+++ b/netsim/ansible/templates/routing/iosxr/aspath.j2
@@ -2,7 +2,7 @@
 !
 as-path-set {{ asp_name }}
 {%   for p_entry in asp_data %}
-  ios-regex '{{ p_entry.path }}'
+  ios-regex '{{ p_entry.path }}'{% if not loop.last %},{% endif +%}
 {%   endfor %}
 end-set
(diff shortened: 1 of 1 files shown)


ipspace:
**Bug in `iosxr/aspath.j2`**: Missing commas between AS-path-set entries. IOS XR requires comma-separated entries in `as-path-set`, just like `prefix-set` (see how `prefix.j2` handles this with `{%  if not loop.last or not last_entry %},{% endif +%}`).

A multi-pattern set like:
```
as-path-set MY_SET
  ios-regex '_100_'
  ios-regex '_200_'
end-set
```
will be rejected by IOS XR — entries must be comma-separated:
```
as-path-set MY_SET
  ios-regex '_100_',
  ios-regex '_200_'
end-set
```

Everything else looks good — the template split is clean, the `_xr_aspath` permit/deny handling follows the established prefix pattern, and the feature flags in `xr.yml` are correct.

ipspace:
> **Bug in `iosxr/aspath.j2`**: Missing commas between AS-path-set entries.

Wasn't me but copilot CLI ;)

> will be rejected by IOS XR — entries must be comma-separated:

And it's correct. Fixing...

PR #3218, opened 2026-03-20
Use updated SuzieQ image

leec-666:
Netlab uses the [SuzieQ demo image](https://hub.docker.com/r/netenglabs/suzieq-demo) which has not been updated in 4 years.  This PR changes it to a more recent image https://hub.docker.com/r/netenglabs/suzieq


ipspace (approved):
Works for me ;) Thank you!

PR #3219, opened 2026-03-20
Enable MPLS transport for EVPN on Juniper cRPD

leec-666:
After updating to 26.03, I found that I was unable to create a cRPD EVPN route reflector. 

This PR adds `features.evpn.transport: [ mpls ]` to the Juniper cRPD device type.  

Only tested with EVPN route reflector.  Replaces `features.evpn.irb: true` as I only added that in #3141 to enable EVPN support and IRB was never tested.

### DIFF
--- a/netsim/devices/crpd.yml
+++ b/netsim/devices/crpd.yml
@@ -23,7 +23,7 @@ features:
   vlan: false
   vxlan: false
   evpn:
-    irb: true
+    transport: [ mpls ]
   gateway:
     protocol: [ vrrp ]
 


ipspace:
> This PR adds `features.evpn.transport: [ mpls ]` to the Juniper cRPD device type.

Checked the "evpn/junos/common.j2" template and it explicitly sets the encapsulation to "vxlan", which means this has no chance of working apart from your particular use case (RR).

We should relax the transport protocol check for devices that have empty evpn.vrfs and evpn.vlans lists.

ipspace:
Replaced by #3230

Issue #3221, opened 2026-03-21
[BUG] VRF configuration changed in SR Linux release 26.3


ipspace:
All VRF-related configurations fail with a message similar to this:

````
TASK [Update SRL vrf node configuration (template=/home/pipi/net101/X/node_files/rtr/vrf)] *****************************************
fatal: [rtr]: FAILED! => {"changed": false, "id": "2026-03-21 08:31:30:097347", "method": "diff", "msg": "Schema '/network-instance/inter-instance-policies/apply-policy' has no local leaf with the name 'export-policy'. Options are []: Parse error on line 1: {\"apply-policy\": {\"export-policy\": \"red_vpn_export\", \"import-policy\": \"red_vpn_import\"}}\nInput(context):\n'{\"apply-policy\": {\"export-policy\": \"red_vpn_export\", \"import-policy\": \"red_vpn_import\"}}'"}
````

If anyone feels like fixing this and implementing all the fallback mechanisms to support older versions (we did that in the past, so there's sample code available), please submit a PR.

hellt:
Hi @ipspace 

I wrote a tool that should assist in identifying the changes between any pair of releases on your actual configs https://github.com/srl-labs/srlconv

ipspace:
> I wrote a tool that should assist in identifying the changes between any pair of releases on your actual configs https://github.com/srl-labs/srlconv

Super cool. Thanks a million!

ipspace:
Closed by #3232

PR #3232, opened 2026-03-22
fix inter-instance-policies for SR Linux >= 26.3

michelredondo:
fix for https://github.com/ipspace/netlab/issues/3221


ipspace:
Wow, that was fast. Thanks a million!

Will test/merge tomorrow.

Issue #3233, opened 2026-03-23
[BUG] BGP policies do not work with SR Linux 26.3


ipspace:
Probably another small change in the data model similar to #3221

22-locpref-unnumbered:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp#policy/22-locpref-unnumbered.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/bgp#policy/22-locpref-unnumbered.yml-config.log)

21-locpref:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp#policy/21-locpref.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/bgp#policy/21-locpref.yml-config.log)

32-med-unnumbered:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp#policy/32-med-unnumbered.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/bgp#policy/32-med-unnumbered.yml-config.log)

01-locpref:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/01-locpref.yml)
* [Test results](https://tests.netlab.tools/srlinux/clab/routing/01-locpref.yml-config.log)

ipspace:
Fixed in #3235

PR #3235, opened 2026-03-23
Fix BGP policies (loc-pref/med) for SR Linux >= 26.3

michelredondo:
fix for https://github.com/ipspace/netlab/issues/3233

### DIFF
--- a/netsim/ansible/templates/routing/srlinux.j2
+++ b/netsim/ansible/templates/routing/srlinux.j2
@@ -55,10 +55,7 @@ updates:
 {%       endif %}
 {%       if 'med' in entry.set %}
           med:
-{%         if _srl_version >= [ 26, 3 ] %}
-            operation: set
-            value: {{ entry.set.med }}
-{%         elif _srl_version >= [ 25, 3 ] %}
+{%         if _srl_version >= [ 25, 3 ] %}
             operation: set
             value: {{ entry.set.med }}
 {%         else %}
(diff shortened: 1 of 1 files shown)


netsim/ansible/templates/routing/srlinux.j2
```diff
@@ -46,11 +46,19 @@ updates:
...
+            operation: set
+            value: {{ entry.set.locpref }}
+{%         else %}
             set: {{ entry.set.locpref }}
+{%         endif %}
 {%       endif %}
 {%       if 'med' in entry.set %}
           med:
-{%         if _srl_version >= [ 25, 3 ] %}
+{%         if _srl_version >= [ 26, 3 ] %}
```
ipspace: This should not be necessary as the 26.3 version does the same thing as the 25.3 version. Let me check why the MED test is failing.

ipspace:
Wow, you're fast. I couldn't use @hellt tool because I was going through another test run, and you fixed the locpref thing in the meantime. Thanks a million!

netsim/ansible/templates/routing/srlinux.j2
ipspace: ... the MED test also sets the local preference, so the root cause was the change in the local preference data model. Will remove the MED change and rerun the tests.

ipspace (approved):
Fixing Locpref was exactly what we needed. The routing policy tests work ;)

Thanks a million!

PR #3243, opened 2026-03-25
Implement bgp.advertise for openbsd.

remilocherer:


netsim/ansible/templates/bgp/openbsd.j2
```diff
@@ -9,18 +9,10 @@ log updates
...
-{%     if intf.ipv4|default(False) %}
-network {{ intf.ipv4|ansible.utils.ipaddr('0') }}
-{%     endif %}
-{%     if intf.ipv6|default(False) %}
-network {{ intf.ipv6|ansible.utils.ipaddr('0') }}
-{%     endif %}
-{%   endif %}
+{% for pfx in bgp.advertise|default([]) %}
+{%   for af in pfx %}
+network {{ pfx[af]|ansible.utils.ipaddr('0') }}
```
ipspace, line 14: This is no longer needed as the prefixes in bgp.advertise are sanitized in the BGP module, but it doesn't hurt.

ipspace (approved):
Nice job (and awesomely fast). Thank you!

PR #3254, opened 2026-03-29
Fix: Remove extra parameter in EXOS OSPFv2 timer configuration

### LINKED ISSUE #3253: [BUG] Incorrect parameters in EXOS OSPFv2 timer command
The fifth parameter in the 'configure ospf vlan timer' command should
be positive (the template value is zero) on SW release 33.1.1.31

30-timers:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv2/30-timers.yml)
* [Test results](https://tests.netlab.tools/exos/libvirt/ospf/ospfv2/30-timers.yml-config.log)

ipspace:
The fifth parameter in the 'configure ospf vlan timer' command should be positive (the template value is zero) on SW release 33.1.1.31

The parameter seems to be optional, so it might be better to remove it?

Closes #3253

### DIFF
--- a/netsim/ansible/templates/ospf/exos.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/exos.ospfv2.j2
@@ -25,7 +25,7 @@ configure ospf vlan {{ vlan_name }} cost {{ l.ospf.cost }}
 configure ospf vlan {{ vlan_name }} priority {{ l.ospf.priority }}
 {%     endif %}
 {%     if l.ospf.timers.hello is defined and l.ospf.timers.dead is defined %}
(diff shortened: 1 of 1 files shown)


sdargoeuves (commented):
I agree, I'm not sure why `0` got there initially...

Removing this value seems the best approach indeed
> wait timer interval—This interval is required by the OSPF standard to be equal to the router dead interval. Under some circumstances, setting the wait interval to smaller values can help OSPF routers on a broadcast network to synchronize more quickly at the expense of possibly electing an incorrect DR or BDR. This value should not be set to less than the hello interval. The default value is equal to the router dead interval.

([source](https://documentation.extremenetworks.com/exos_commands_32.6.1/GUID-385CD68C-199A-418A-B807-B929A6D6EC80.shtml))

PR #3255, opened 2026-03-29
Doc fix: EOS 'sh' config mode works only on recent EOS releases

### LINKED ISSUE #3246: [BUG] return of the old cEOS `platform tfa phy` issue
### Describe the bug

I'm still using the `4.29.9.1M` version of the Arista cEOS, and to be fair, if you tell me I need to update, I'd happily do it, and we can move on 😄 

When doing netlab up, and I am using the `netlab_config_mode: sh`, I get this error:

```[FATAL]   initial: normalize configuration in namespace clab-ml-1-s6xsw02 failed for node s6xsw02
[DATA]    Executed command: docker exec clab-ml-1-s6xsw02 /mnt/flash/01-normalize.sh
  >  platform tfa phy control-frame disabled

  % Invalid input at line 4

  >  platform tfa phy control-frame disabled

  % Invalid input at line 9

  >  platform tfa phy control-frame disabled

  % Invalid input at line 14

  >  platform tfa phy control-frame disabled

  % Invalid input at line 18

  >  platform tfa phy control-frame disabled

  % Invalid input at line 22

  >  platform tfa phy control-frame disabled

  % Invalid input at line 26
```
Unless I should NOT use this mode for older version of cEOS?

### Version

I'm using the `dev`, and I would expect this is due to this commit: https://github.com/ipspace/netlab/commit/69f28b315082baa2635af449dedc57091f9ecdb6

We had a similar issue with my old cEOS version here: https://github.com/ipspace/netlab/pull/3055

ipspace:
Closes #3246

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -256,7 +256,7 @@ Several other devices can use alternate (faster) configuration methods that are 
 | Dell OS10 | **startup** |
 | Junos[^Junos] | **startup** |
 
-[^EOSSH]: Arista EOS device configurations are converted into FastCli scripts and executed as Linux scripts within the cEOS container
+[^EOSSH]: Arista EOS device configurations are converted into FastCli scripts and executed as Linux scripts within the cEOS container. This method works on EOS software releases that have the **‌platform tfa phy control-frame disabled** interface configuration command (probably starting with EOS release 4.30)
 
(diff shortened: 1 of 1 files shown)


sdargoeuves (approved):
Thanks for updating this, I'm just unsure about why I had this weird issue with the latest 4.30.10M, but from what I've read, it should be fine starting from 4.30
So I'm happy with this update.

PR #3256, opened 2026-03-30
add wait timers for EXOS

sdargoeuves:
I faced some issues on a different servers than the one I've used initially for EXOS. Adding the wait time solved this.


ipspace (changes requested):
The changes in exos.yml are OK, the ones in eos.yml deployment task are probably accidental.

sdargoeuves:
definitely accidental... not sure if we need to add the same to the deploy-config for exos... But yes, it shouldn't have been eos

ipspace:
> definitely accidental... not sure if we need to add the same to the deploy-config for exos

The "wait for connection" does the same thing as netlab waiting for SSH server, which you configured in exos.yml.

ipspace (approved):
Works like a charm ;)

PR #3257, opened 2026-03-30
attempt at vlan1 quirk for EXOS

### LINKED ISSUE #3251: [BUG] VLAN 1 has to be renamed to "Default" on EXOS
The VLAN trunk test uses untagged VLAN on the trunk to verify that the untagged frames are disabled. The untagged VLAN uses VLAN ID 1 which overlaps with EXOS Default VLAN. As you cannot rename that VLAN, the only workaround seems to be a device quirk that renames the device VLAN with id=1 to "Default" and changes the name of the corresponding **vlan** attribute on all interfaces.

31-vlan-bridge-trunk:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vlan/31-vlan-bridge-trunk.yml)
* [Test results](https://tests.netlab.tools/exos/libvirt/vlan/31-vlan-bridge-trunk.yml-initial.log)

The same problem is probably causing configuration failure in another integration test:

70-vlan-1-trunk:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/vlan/70-vlan-1-trunk.yml)
* [Test results](https://tests.netlab.tools/exos/libvirt/vlan/70-vlan-1-trunk.yml-initial.log)

sdargoeuves:


ipspace (commented):
Not too bad (as in "it probably works") but way too verbose. Let me clean it up...

ipspace (approved):
The "untagged VLAN 1" test passes, the "tagged VLAN 1" test is configured but fails. Won't even try to push it any further.

sdargoeuves:
thank you so much for this! If it becomes an issue worth solving, as in, if someone complains about it, I would check further... but in the meantime, I'm happy as it is.

PR #3263, opened 2026-03-31
FRR: Add SRv6 bgp support

jbemmel:
Following the upgrade to the latest FRR 10.6 release, we can now do SRv6 with BGP as IGP

Conceptually this still somewhat mixes EBGP as an IGP underlay for SRv6 and BGP v4/v6 in the overlay, we may want to clean that up. I'm sharing an early draft with key knobs to enable this use case (and I know how much you love IBGP-over-EBGP designs, that could be changed too of course)

### DIFF
--- a/netsim/ansible/templates/srv6/frr.bgp.j2
+++ b/netsim/ansible/templates/srv6/frr.bgp.j2
@@ -1,7 +1,19 @@
+{% if 'bgp' in srv6.igp|default([]) %}
+interface ${SRv6_DEV}
+  ipv6 address {{ srv6.locator }}
+{% endif %}
+
 router bgp {{ bgp.as }}
+ ! bgp default software-version-capability latest-encoding
  segment-routing srv6
   locator {{ inventory_hostname }}
  exit
+
+ {% if 'bgp' in srv6.igp|default([]) %}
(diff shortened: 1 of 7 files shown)


ipspace:
Nice starting point, thanks a million! Unfortunately, you mixed two completely independent things in a single PR:

1. Support for SRv6 next hops for the global BGP routes (building BGP-free core with SRv6)
2. Support for SRv6 locators advertised with BGP.

At the very minimum, this should be split in two independent PRs, and I'm sure there are other devices out there supporting BGP as the routing protocol for SRv6 locators, so it would be nice to have two interoperable implementations (and yeah, I know that's a lot to wish for).

Once I got a bit of time (and some willpower), I can do the refactoring.

jbemmel:
Replaced by #3369

ipspace:
Hey, not so fast, I still have to implement your EBGP ideas ;)) Fortunately, no commit is ever truly lost.

jbemmel:
> Hey, not so fast, I still have to implement your EBGP ideas ;)) Fortunately, no commit is ever truly lost.

I didn’t think you needed more guidance on that, but appreciate the thought

Issue #3323, opened 2026-04-13
FortiOS: Inject optional firewall policy to allow transit traffic


a-v-popov:
FortiOS by default has only implicit drop rule in its firewall policy which block all transit traffic.
While it is an arguably a desirable default state for a firewall it might be an inconvenience in a lab setup.
I would suggest to add an optional firewall policy to permit traffic controlled by some config parameters.

By default no policy should be injected leaving the current behavior.
Config should allow to inject a policy and choose if it is enabled from the start or configured as disabled.
This would allow to run integration tests or run troubleshooting as need.

PR #3324, opened 2026-04-13
FortiOS: Allow any policy

a-v-popov:
This is possible variant to address https://github.com/ipspace/netlab/issues/3323

Policy is injected in initial configuration.

It is configured by device configuration:
```
$ cat ~/.netlab.yml 
---
devices.fortios:
  group_vars:
    netlab_allow:
      disable: false
```

### DIFF
--- a/netsim/ansible/templates/initial/fortios.j2
+++ b/netsim/ansible/templates/initial/fortios.j2
@@ -72,7 +72,7 @@ end
 
 {# End of `config global` #}
 
-{% if netlab_allow is defined and netlab_allow != false %}
+{% if netlab_default_policy|default(false) %}
 
 {%  if multi_vdom %}
 config vdom
@@ -81,10 +81,10 @@ config vdom
 
 config firewall policy
     edit 1000
-{%   if netlab_allow is mapping and netlab_allow.disable | default(false) %}
+{%   if not netlab_default_policy.enable|default(true) %}
         set status disable
 {%   endif %}
-        set name "netlab_allow"
+        set name "netlab_default_policy"
         set srcintf "any"
         set dstintf "any"
         set action accept
--- a/netsim/ansible/templates/initial/fortios.j2
+++ b/netsim/ansible/templates/initial/fortios.j2
@@ -69,3 +69,37 @@ end
 {% if multi_vdom %}
 end
 {% endif %}
+
+{# End of `config global` #}
+
+{% if netlab_allow is defined and netlab_allow != false %}
+
+{%  if multi_vdom %}
+config vdom
+    edit {{ vdom_traffic }}
+{%  endif %}
+
+config firewall policy
+    edit 1000
+{%   if netlab_allow is mapping and netlab_allow.disable | default(false) %}
+        set status disable
+{%   endif %}
+        set name "netlab_allow"
+        set srcintf "any"
+        set dstintf "any"
+        set action accept
+        set srcaddr "all"
+        set dstaddr "all"
+        set srcaddr6 "all"
+        set dstaddr6 "all"
+        set schedule "always"
+        set service "ALL"
+        set logtraffic disable
+    next
+end
+
+{%  if multi_vdom %}
+end
+{%  endif %}
+
+{% endif %}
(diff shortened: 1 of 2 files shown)


netsim/ansible/templates/initial/fortios.j2
```diff
@@ -69,3 +69,37 @@ end
 {% if multi_vdom %}
 end
 {% endif %}
+
+{# End of `config global` #}
+
+{% if netlab_allow is defined and netlab_allow != false %}
```
ipspace: I would call it "netlab_default_policy" or some such, and generate the firewall policy if the variable is defined.

If you want to have an override option (setting it to `false` to disable it), go with `netlab_default_policy|default(false)`-- that will be `false` if the variable is not defined or if it's set to False, and will evaluate to `false` as a boolean expression even if it has falsy value (like `None` or an empty dictionary).

netsim/ansible/templates/initial/fortios.j2
```diff
@@ -69,3 +69,37 @@ end
...
+
+{%  if multi_vdom %}
+config vdom
+    edit {{ vdom_traffic }}
+{%  endif %}
+
+config firewall policy
+    edit 1000
+{%   if netlab_allow is mapping and netlab_allow.disable | default(false) %}
```
ipspace: At this point, we know "netlab_default_policy" (or however it's called) is defined. Under our current Jinja2 setup, you could simply test `netlab_default_policy.disable|default(false)` and have the desired result even when `netlab_default_policy` is not a mapping.

Personally, I would prefer using `enable` with default `true`, "not disabling" is a double negation, and that tends to get confusing.

ipspace (changes requested):
Lovely idea. I would make a few slight changes in the variable names and streamline the tests.

ipspace:
Here's how you can test whether any particular trick works with our Jinja2 setup. As we are using the same code to generate reports and device configs, you can create a report that uses none of the topology variables and play with Jinja2 expressions. The only tricks:

* The report has to be in `reports` directory and have `.j2` extension
* There should be `topology.yml` file or a started lab in the current directory

After setting this up, just do **netlab report _reportname_** and experiment. 

Here's what I tested:

```
{% set tv = true %}
{% set fv = false %}
{% set sv = "abc" %}
{% set nv = None %}
{% set mv = {'a': true, 'b': false} %}

tv: {{ tv }} tv.a: {{ tv.a|default('DEFAULT') }}
fv: {{ fv }} fv.a: {{ fv.a|default('DEFAULT') }}
sv: {{ sv }} sv.a: {{ sv.a|default('DEFAULT') }}
nv: {{ nv }} nv.a: {{ nv.a|default('DEFAULT') }}
mv: {{ mv }} mv.a: {{ mv.a|default('DEFAULT') }} mv.x: {{ mv.x|default('DEFAULT') }}
```

And here are the test results:

```
% netlab report j2test
[INFO]    Using lab topology file topology.yml

tv: True tv.a: DEFAULT
fv: False fv.a: DEFAULT
sv: abc sv.a: DEFAULT
nv: None nv.a: DEFAULT
mv: {'a': True, 'b': False} mv.a: True mv.x: DEFAULT
```

netsim/ansible/templates/initial/fortios.j2
a-v-popov: The problem would be an empty dict, or rather an abuse of YAML like this:
```yaml
$ cat ~/.netlab.yml 
---
devices.fortios:
  group_vars:
    netlab_default_policy:
      #enable: false
```

a-v-popov:
I added a commit based on the feedback.
If anything else is required or a change in a workflow is need please let me know.

netsim/ansible/templates/initial/fortios.j2
ipspace: Even that works, see my test results for the "None" value ("YAML abuse" results in None value).

ipspace:
> I added a commit based on the feedback.

Thank you.

> If anything else is required or a change in a workflow is need please let me know.

Documentation ;) Just kidding, I added it. Merging...

ipspace (approved):
Ready to merge. Thank you!

PR #3337, opened 2026-04-19
ForitOS: Shift indexes of ospf networks by 1 to accommodate loopack

a-v-popov:
In FortiOS `edit 0` is equivalent `edit <next available>`.
As loopack has index 0 it will only work for the first time. 
The solution is to shift interface index by 1 for ospf networks list.

### DIFF
--- a/netsim/ansible/templates/ospf/fortios.j2
+++ b/netsim/ansible/templates/ospf/fortios.j2
@@ -55,7 +55,7 @@ config router ospf
 
     config network
 {%  for intf in netlab_interfaces if intf.ospf.area is defined %}
-        edit {{ intf.ifindex + 1 }}
+        edit {{ loop.index }}
             set prefix {{ intf.ipv4|ansible.utils.ipaddr('subnet') }}
             set area {{ intf.ospf.area }}
         next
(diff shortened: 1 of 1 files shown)


netsim/ansible/templates/ospf/fortios.j2
```diff
@@ -55,7 +55,7 @@ config router ospf
 
     config network
 {%  for intf in netlab_interfaces if intf.ospf.area is defined %}
-        edit {{ intf.ifindex }}
+        edit {{ intf.ifindex + 1 }}
```
sdargoeuves: @a-v-popov I've just realised that this value has nothing to do with the interface... I initially thought by doing your proposed change, it will cause an issue if you do not have a Loopback, but now I understand it has nothing to do with the `ifindex`. So any unique integer increasing for each iteration of the loop should be fine (as long as it's more than 0!)

I can't think of any reason to link it to an interface number. In which case, using `loop.index` instead of `intf.ifindex + 1` should be better.

sdargoeuves (commented):
I've added the commit with the `loop.index` instead of `intf.ifindex + 1`.

ipspace:
@a-v-popov @sdargoeuves -- either approach works for me. Please decide which one you want to use and tell me when it's OK to merge.

netsim/ansible/templates/ospf/fortios.j2
a-v-popov: Fine by me, `loop.index` looks cleaner. I guess interface index is assigned in `netlab create` and might in theory be considered more stable. In practice though I don't think there would be any difference. I do not plan to edit those topology.yml by hand anyway.

Issue #3252, opened 2026-03-29
[BUG] EXOS needs "normalize" step


ipspace:
It looks like EXOS sends one RA even when it's disabled.

```
15:08:36.868800 IP6 (class 0xc0, flowlabel 0xde3dd, hlim 255, next-header ICMPv6 (58) payload length: 64) fe80::a8c1:abff:fec9:66b0 > fe80::a8c1:abff:fec7:7cf7: [icmp6 sum ok] ICMP6, router advertisement, length 64
	hop limit 64, Flags [managed, other stateful], pref medium, router lifetime 30s, reachable time 0ms, retrans timer 0ms
	  prefix info option (3), length 32 (4): 2001:db8:1:4::/64, Flags [onlink], valid time 120s, pref. time 120s
	    0x0000:  4080 0000 0078 0000 0078 0000 0000 2001
	    0x0010:  0db8 0001 0004 0000 0000 0000 0000
	  mtu option (5), length 8 (1):  1500
	    0x0000:  0000 0000 05dc
	  source link-address option (1), length 8 (1): aa:c1:ab:c9:66:b0
	    0x0000:  aac1 abc9 66b0
```

That confuses the integration test (the tests pass after the RA information ages out). If there's no obvious solution, I could tweak the integration test to shut down and reenable the interfaces on all attached hosts to flush their caches.

08-ra:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/initial/08-ra.yml)
* [Test results](https://tests.netlab.tools/exos/libvirt/initial/08-ra.yml-validate.log)

ipspace:
Have to write a report on this

ipspace:
Took me a long while to figure this out :( Caused by EXOS leaking the RA sent by the DHCP server to all other ports.

We need the "normalize" step for EXOS to disable its interfaces before starting configuration.

ipspace:
@sdargoeuves -- I just need the "port shutdown" command that would be executed on all physical interfaces in the "normalize" script and then reverted on those interfaces that are not in **shutdown** state during initial device configuration.

sdargoeuves:
Great find! Well done and thank you for the description.
I thought the device was starting with interfaces in shutdown state, but if I remember correctly, there is that weird wizard TUI upon first boot. So it might depend on how the box/image was created. Regardless, it makes sense to force it! I'll look into this.

PR #3339, opened 2026-04-21
EXOS Normalize interface

sdargoeuves:
@ipspace I'm not sure how to run the test to see if this branch fixes the issue #3252 , do you mind having a look, or sharing how you did the test?

This PR disabled the physical ports as part of the normalization, and will re-enable those interfaces at the end of the initial task, so hopefully that should be enough to sort this issue.

I've also added a caveat, as I've realised the EXOS initial configuration is not idempotent due to the Vlan creation. I'm not sure how to workaround this at the moment, so caveat it is!

### DIFF
--- a/netsim/ansible/templates/initial/exos.j2
+++ b/netsim/ansible/templates/initial/exos.j2
@@ -50,21 +50,21 @@ disable router-discovery ipv6 vlan {{ vlan_name }}
 enable router-discovery ipv6 vlan {{ vlan_name }}
 configure vlan {{ vlan_name }} router-discovery ipv6 max-interval 5
 configure vlan {{ vlan_name }} router-discovery ipv6 min-interval 3
-{%       if ra_prefix %}
+{%         if ra_prefix %}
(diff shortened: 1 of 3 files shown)


ipspace:
> @ipspace I'm not sure how to run the test to see if this branch fixes the issue #3252 , do you mind having a look, or sharing how you did the test?

The issue pops up in tests with two IPv6 segments where a device sending RAs on one segment is configured before the tested device. Will add the test.

> This PR disabled the physical ports as part of the normalization, and will re-enable those interfaces at the end of the initial task, so hopefully that should be enough to sort this issue.

I would also remove them from VLAN 1, but I guess we already have the command to do that in one of the templates, so I can add that.

> I've also added a caveat, as I've realised the EXOS initial configuration is not idempotent due to the Vlan creation.

Thank you!

It might take a day or two before this gets merged, I'm busy with a totally unrelated project 🤷🏻‍♂️

ipspace (approved):

PR #3343, opened 2026-04-29
Juniper cSRX support

leec-666:
Juniper cSRX was added to ContainerLab 0.75.0 https://github.com/srl-labs/containerlab/releases/tag/v0.75.0.  This PR adds support for it in Netlab.

There are limited features in cSRX e.g. it does not support routing protocols or VLANs.  See https://www.juniper.net/documentation/us/en/software/csrx/csrx-consolidated-deployment-guide/topics/concept/security-csrx-docker-feature-support.html and https://www.juniper.net/documentation/us/en/software/csrx/csrx-consolidated-deployment-guide/csrx-consolidated-deployment-guide.pdf.

It might not be useful to many but I had to lab a weird NAT issue today and netlab would have sped it up so why not spend 20 times longer adding support than it took to do manually?!

### DIFF
--- a/netsim/ansible/templates/ospf/fortios.j2
+++ b/netsim/ansible/templates/ospf/fortios.j2
@@ -55,7 +55,7 @@ config router ospf
 
     config network
 {%  for intf in netlab_interfaces if intf.ospf.area is defined %}
-        edit {{ intf.ifindex }}
+        edit {{ loop.index }}
             set prefix {{ intf.ipv4|ansible.utils.ipaddr('subnet') }}
             set area {{ intf.ospf.area }}
         next


netsim/templates/provider/clab/csrx/netlab-config.j2
```diff
@@ -0,0 +1,19 @@
...
+CONFIG
+if grep -i error /tmp/cli-status >/dev/null; then
+  echo "Configuration load failed, aborting"
+  exit 1
+else
+  cat <<CONFIG | cli | tee /tmp/cli-status
+configure
+commit and-quit
+CONFIG
+  if grep -i error /tmp/cli-status | grep -vi "Device or resource busy" >/dev/null; then
```
leec-666: This was required because it always emits the error message `error: rename buf /etc/ssh/sshd_config+ to path /etc/ssh/sshd_config, errno: Device or resource busy` after `commit and-quit`

PR #3344, opened 2026-04-29
Juniper cSRX support

leec-666:
Juniper cSRX was added to ContainerLab 0.75.0 https://github.com/srl-labs/containerlab/releases/tag/v0.75.0. This PR adds support for it in Netlab.

There are limited features in cSRX e.g. it does not support routing protocols or VLANs. See https://www.juniper.net/documentation/us/en/software/csrx/csrx-consolidated-deployment-guide/topics/concept/security-csrx-docker-feature-support.html and https://www.juniper.net/documentation/us/en/software/csrx/csrx-consolidated-deployment-guide/csrx-consolidated-deployment-guide.pdf.

It might not be useful to many but I had to lab a weird NAT issue today and netlab would have sped it up so why not spend 20 times longer adding support than it took to do manually?!

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -445,7 +445,7 @@ Implementation limitations in import/export route filters (reported as errors th
 ## Juniper cSRX
 
 * cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.crpd.clab.node.license** [device default](topo-defaults).
-* If using more than 2 data interfaces, `clab.env.CSRX_PORT_NUM` must be increased to the number of interfaces + 1 (because it includes the management interface). Setting it higher will result in none of the data interfaces functioning correctly, e.g. no ARP.
+* cSRX container supports up to 17 interfaces: 1 Out-of-band management Interface (eth0) and 16 In-band interfaces (ge-0/0/0 to ge-0/0/15).
 * Since cSRX does not support any dynamic routing protocols, the use of the routing module and static routes is recommended.
 
 (caveats-crpd)=
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -441,6 +441,13 @@ Implementation limitations in import/export route filters (reported as errors th
 * When prepending an AS number different than the local one, JunOS puts the prepended AS at the beginning of the *AS_PATH* while other vendors perform the prepending and then add the device's own AS at the beginning. Most EBGP peers deny a BGP update that does not have the neighbor's AS number at the beginning of the *AS_PATH*. If needed, _netlab_ automatically adds the *local-as* to the prepend list to get it at the beginning of the *AS_PATH*.
 * *as-path* regex syntax for JunOS has different rules than other vendors. For example, a *null as-path* is represented as `()`. netlab tries some as-path conversion, but sometimes it could be wrong.
 
+(caveats-csrx)=
+## Juniper cSRX
+
+* cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.crpd.clab.node.license** [device default](topo-defaults).
+* If using more than 2 data interfaces, `clab.env.CSRX_PORT_NUM` must be increased to the number of interfaces + 1 (because it includes the management interface). Setting it higher will result in none of the data interfaces functioning correctly, e.g. no ARP.
+* Since cSRX does not support any dynamic routing protocols, the use of the routing module and static routes is recommended.
+
 (caveats-crpd)=
 ## Juniper cRPD
 
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -445,7 +445,7 @@ Implementation limitations in import/export route filters (reported as errors th
 ## Juniper cSRX
 
 * cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.crpd.clab.node.license** [device default](topo-defaults).
-* cSRX container supports up to 17 interfaces: 1 Out-of-band management Interface (eth0) and 16 In-band interfaces (ge-0/0/0 to ge-0/0/15).
+* cSRX supports up to 17 interfaces: 1 Out-of-band management Interface (eth0) and 16 In-band interfaces (ge-0/0/0 to ge-0/0/15).
 * Since cSRX does not support any dynamic routing protocols, the use of the routing module and static routes is recommended.
 
 (caveats-crpd)=
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -441,13 +441,6 @@ Implementation limitations in import/export route filters (reported as errors th
 * When prepending an AS number different than the local one, JunOS puts the prepended AS at the beginning of the *AS_PATH* while other vendors perform the prepending and then add the device's own AS at the beginning. Most EBGP peers deny a BGP update that does not have the neighbor's AS number at the beginning of the *AS_PATH*. If needed, _netlab_ automatically adds the *local-as* to the prepend list to get it at the beginning of the *AS_PATH*.
 * *as-path* regex syntax for JunOS has different rules than other vendors. For example, a *null as-path* is represented as `()`. netlab tries some as-path conversion, but sometimes it could be wrong.
 
-(caveats-csrx)=
-## Juniper cSRX
-
-* cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.crpd.clab.node.license** [device default](topo-defaults).
-* cSRX supports up to 17 interfaces: 1 Out-of-band management Interface (eth0) and 16 In-band interfaces (ge-0/0/0 to ge-0/0/15).
-* Since cSRX does not support any dynamic routing protocols, the use of the routing module and static routes is recommended.
-
 (caveats-crpd)=
 ## Juniper cRPD
 
@@ -455,6 +448,14 @@ Implementation limitations in import/export route filters (reported as errors th
 * Inter-VRF route leaking does not work correctly
 * VRRPv3 for IPv4 implementation uses the [checksum calculation that is incompatible with most other VRRP implementations](https://blog.ipspace.net/2025/01/sturgeon-law-vrrp-edition/). The `checksum-without-pseudoheader` configuration command does not seem to be available.
 * Anycast gateway is unsupported
+
+(caveats-csrx)=
+## Juniper cSRX
+
+* cSRX is a native container and should not be confused with vSRX which is a vrnetlab VM in a container. See [Juniper's cSRX documentation](https://www.juniper.net/documentation/us/en/software/csrx/csrx-consolidated-deployment-guide/topics/concept/security-csrx-docker-feature-support.html) for more details.
+* cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.csrx.clab.node.license** [device default](topo-defaults).
+* cSRX supports up to 17 interfaces: 1 out-of-band management Interface (eth0) and 16 in-band interfaces (ge-0/0/0 to ge-0/0/15).
+* Since cSRX does not support any dynamic routing protocols, use the **routing** module and static routes.
 
 (caveats-vmx)=
 ## Juniper vMX
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -220,7 +220,7 @@ Ansible playbooks included with **netlab** can deploy and collect device configu
 
 [^XE]: Includes Cisco CSR 1000v, Cisco Catalyst 8000v, Cisco IOS-on-Linux (IOL) and IOL Layer-2 image
 
-[^Junos]: Includes cRPD, vMX, vSRX, vPTX, vJunos-switch, and vJunos-router
+[^Junos]: Includes cRPD, vMX, vSRX, vPTX, vJunos-switch, and vJunos-router but not cSRX
 
 [^SROS]: Includes the Nokia SR-SIM container and the Virtualized 7750 SR and 7950 XRS Simulator (vSIM) virtual machine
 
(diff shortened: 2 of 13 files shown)


netsim/devices/csrx.yml
```diff
@@ -0,0 +1,55 @@
...
+  image: csrx:23.4R1.9
+  build: https://containerlab.dev/manual/kinds/csrx/
+  node:
+    kind: juniper_csrx
+    env:
+      # Number of interfaces to create
+      # cSRX default is 3 (management + 2 data interfaces)
+      # If more interfaces are required, this must be increased.
+      # Setting to a higher value than required will break interfaces (no arp etc)
+      CSRX_PORT_NUM: 3
```
ssasso: I think this can be dynamically generated by the quirk python module?
it can count the interfaces for the node and define the var for each node?

leec-666: Oh yeah I meant to ask about that.  So the quirks file would be responsible?  I'll see if I can figure it out.

leec-666: I think https://github.com/ipspace/netlab/pull/3344/commits/83e54f365c11f5f18c3abf5a7e1f4d2c9f26acff should do it.  

I forced an error which resulted in this output:
```
┌──────────────────────────────────────────────────────────────────────────────────┐
│ CREATING configuration files                                                     │
└──────────────────────────────────────────────────────────────────────────────────┘
[ERROR]   csrx: cSRX supports a maximum of 16 interfaces. Node fw1 has 17 interfaces.
[HINT]    Set defaults.devices.csrx.warnings.csrx_port_num to False to disable this check
```

Can the hint be disabled?  This check should not ever be disabled.

ipspace:
Starting with a naive question: is this beast different enough that we need a different device (and not just vSRX with provider=clab)?

leec-666:
Aside from feature support, there are some other annoying differences:

* Username and password are different which means vsrx.clab would need to change
* The initial config deployment fails because the junos template does not use interface units and includes LLDP.  Both cases would need to be handled.

I found it simpler to make it a new device with a couple of new templates, and not inherit from vSRX or Junos and list the features it _does_ support rather than adding overrides for those it doesn't.  This means if a user attempts to use the ospf module for example (what documentation?), lab startup will fail immediately informing them as such instead of waiting for the module to run and fail with config syntax errors.

docs/caveats.md
```diff
@@ -441,6 +441,13 @@ Implementation limitations in import/export route filters (reported as errors th
 * When prepending an AS number different than the local one, JunOS puts the prepended AS at the beginning of the *AS_PATH* while other vendors perform the prepending and then add the device's own AS at the beginning. Most EBGP peers deny a BGP update that does not have the neighbor's AS number at the beginning of the *AS_PATH*. If needed, _netlab_ automatically adds the *local-as* to the prepend list to get it at the beginning of the *AS_PATH*.
 * *as-path* regex syntax for JunOS has different rules than other vendors. For example, a *null as-path* is represented as `()`. netlab tries some as-path conversion, but sometimes it could be wrong.
 
+(caveats-csrx)=
+## Juniper cSRX
+
```
ipspace: Maybe add a line with a pointer to cSRX documentation and explaining that this is a "native" container as opposed to vSRX in vrnetlab container (which is the vsrx device)

docs/caveats.md
```diff
@@ -441,6 +441,13 @@ Implementation limitations in import/export route filters (reported as errors th
...
+* cSRX might require a license file to unlock additional features. You can specify the location of the license file with the **clab.license** node parameter or **defaults.devices.crpd.clab.node.license** [device default](topo-defaults).
+* cSRX supports up to 17 interfaces: 1 Out-of-band management Interface (eth0) and 16 In-band interfaces (ge-0/0/0 to ge-0/0/15).
+* Since cSRX does not support any dynamic routing protocols, the use of the routing module and static routes is recommended.
```
ipspace: I would be less polite and say "use **routing** module and static routes"

docs/platforms.md
```diff
@@ -391,6 +397,7 @@ Routing protocol [configuration modules](module-reference.md) are supported on t
 | Extreme Networks EXOS | ✅   |   ❌   |   ❌   |   ❌   |  ❌  |
 | Fortinet FortiOS      | ✅ [❗](caveats-fortios) |   ❌   |   ❌   |   ❌   |  ❌  |
 | FRR                   | ✅   |  ✅   |   ❌  | ✅  |  ✅  |
+| Juniper cSRX         |   ❌  |   ❌  |   ❌  |   ❌  |   ❌  |
```
ipspace: No need for a "no support" line. If needed add "but not cSRX" to the Junos footnote.

docs/platforms.md
```diff
@@ -420,6 +427,7 @@ These devices support additional control-plane protocols or BGP address families
 | Dell OS10             | ✅  | ✅  |  ❌  |  ❌  |
 | Extreme Networks EXOS |  ❌  |  ❌  |  ❌  | ✅  |
 | FRR                   | ✅  | ✅  | ✅  |  ❌  |
+| Juniper cSRX          |  ❌  |  ❌  |  ❌  |  ❌  |
```
ipspace: As above...

docs/platforms.md
```diff
@@ -469,9 +477,10 @@ The data plane [configuration modules](module-reference.md) are supported on the
 | Dell OS10             | ✅ | ✅ | ✅ |  ❌ |  ❌ |  ❌ |
 | Extreme Networks EXOS | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ |
 | FRR                   | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
+| Juniper cSRX          |  ❌ |  ❌ |   ❌ |   ❌ |   ❌ |  ❌ |
```
ipspace: No need for this line, more so as we list supported Junos platforms individually

docs/platforms.md
```diff
@@ -518,6 +527,7 @@ Core *netlab* functionality and all multi-protocol routing protocol configuratio
 | Dell OS10             | ✅ | ❌ | ❌ | ✅ | ❌ |
 | Extreme Networks EXOS | ✅ | ❌ | ❌ | ❌ | ❌ |
 | FRR                   | ✅ | ✅ | ❌ | ✅ | ❌ |
+| Juniper cSRX         | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
```
ipspace: As above; the footnote should explain the details.

netsim/ansible/templates/initial/csrx.j2
```diff
@@ -0,0 +1,65 @@
+system {
+  host-name {{ inventory_hostname }};
+  static-host-mapping {
```
ipspace: You claim cSRX does not support ipv4/ipv6 hosts in platforms.md. Guess that should be fixed.

netsim/ansible/templates/initial/csrx.j2
```diff
@@ -0,0 +1,65 @@
+system {
```
ipspace: Would it make sense to extract common parts of the Junos template to "junos/something" subdirectory and then include them for various Junos devices?

netsim/ansible/templates/routing/csrx.j2
```diff
@@ -0,0 +1,26 @@
+{% if routing.static is defined %}
```
ipspace, line 1: Is there a reason we can't just include "junos/static.j2" to generate static routes? Is the config syntax different from other Junos platforms?

netsim/devices/csrx.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ..utils import log
+from . import _Quirks, report_quirk
+
+
+def csrx_port_num(node: Box) -> None:
+  if_count = len(node.get('interfaces', []))
+  node.clab.env.CSRX_PORT_NUM = if_count + 1
+  if if_count > 16:
+    report_quirk(
```
ipspace: If you want this to be a hard error, use log.error instead of report_quirk

netsim/devices/csrx.py
```diff
@@ -0,0 +1,24 @@
...
+        f'cSRX supports a maximum of 16 interfaces. Node {node.name} has {if_count} interfaces.',
+        node=node,
+        category=log.ErrorAbort,
```
ipspace: This is a bit too harsh. `log.IncorrectValue` might be better.

netsim/devices/csrx.yml
```diff
@@ -0,0 +1,49 @@
...
+group_vars:
+  ansible_user: root
+  ansible_ssh_pass: "clab123"
+  netlab_device_type: csrx
+  netlab_check_retries: 20
+
+mgmt_if: fxp0
+ifindex_offset: 0
+interface_name: ge-0/0/{ifindex}
+loopback_interface_name: "" # Unsupported on cSRX
```
ipspace: As you're not inheriting from Junos -- if the loopback is not supported, then don't define this attribute (see asav.yml)

netsim/devices/csrx.yml
```diff
@@ -0,0 +1,49 @@
...
+    name: eth{ifindex+1}
+  features:
+    initial:
+      config_mode: [ sh ]
+  group_vars:
+    netlab_config_mode: sh
+    netlab_show_command: [ cli, -c, 'show $@' ]
+    netlab_check_command: who
+    netlab_ready: [ ssh ]
+    netlab_default_shebang: '#!/config/netlab/netlab-config.sh'
```
ipspace, line 43: Do we want to use this or should we just use NETCONF to configure it? Or is NETCONF not working?

ipspace (changes requested):
Tons of minor details, but overall a great job! Thank you!

ipspace:
> Username and password are different which means vsrx.clab would need to change

My bad. I should have checked whether we already have vSRX on clab (and we do). This is clearly not the same (or would require tons of tweaks), so a new device is absolutely warranted.

> I found it simpler to make it a new device with a couple of new templates, and not inherit from vSRX or Junos and list the features it _does_ support rather than adding overrides for those it doesn't.

Yeah, listing everything that DOES NOT work would be totally annoying.

netsim/devices/csrx.yml
ipspace: If you want to have a hard (= cannot be disabled) error, use `log.error` and set `module` to `csrx` (device type). See "requires_plugin" in "devices/_common.py" (that one uses `module=node.device` because it's a generic thing that works for any device)

netsim/devices/csrx.yml
leec-666: I based it on cRPD.  I thought the connection/deployment methods other than shell scripts relied on Ansible which I know you're not the biggest fan of.

netsim/ansible/templates/routing/csrx.j2
leec-666: I thought I was being clever by copying the template and stripping out the VRF bits but since they're conditional, I was not.

netsim/devices/csrx.py
leec-666: done :)

netsim/devices/csrx.py
leec-666: Thank you

netsim/devices/csrx.yml
ipspace: > I thought the connection/deployment methods other than shell scripts relied on Ansible

* Configuration deployment: YES
* Connecting to the device: NO

> which I know you're not the biggest fan of.

You're right ;) However, we need Ansible anyway to deploy configurations on other Junos devices, but admittedly the script-based configuration is faster. I'm OK with either one, just don't make it too complicated just to have a bash script.

ipspace:
NIce job. This looks ready-to-merge to me (any thoughts @ssasso?). Just a minor detail that I can't check myself: as you map /etc/hosts into the container, do you need to include 'junos/hosts.j2' into the initial config?

netsim/ansible/templates/initial/csrx.j2
leec-666: Good idea.  I moved the cSRX and cRPD initial interface config into its own template `junos/container_interfaces.j2`, Junos and cRPD share LLDP now, and Junos and cSRX share a static hosts template. https://github.com/ipspace/netlab/pull/3344/commits/f643b692f3020c4691a70007623a97da06e99e73

On the LLDP change: it's probably out of scope seeing that cSRX doesn't support it.  I noticed that crpd.j2 used the `interfaces` dictionary in the LLDP loop whilst junos.j2 used `netlab_interfaces`.  Both loops already exclude loopbacks so it should be safe.  I did do some limited testing on cRPD and it looks fine.

leec-666:
and I'm not sure if it's worth noting but if you ping an ipv4 address from a cSRX device, the replies are duplicated somewhere inside the host

```
root@fw1> ping r1    
PING r1 (10.0.0.3) 56(84) bytes of data.
64 bytes from r1 (10.0.0.3): icmp_seq=1 ttl=64 time=1.32 ms
64 bytes from r1 (10.0.0.3): icmp_seq=1 ttl=64 time=1.49 ms (DUP!)
64 bytes from r1 (10.0.0.3): icmp_seq=2 ttl=64 time=1.13 ms
64 bytes from r1 (10.0.0.3): icmp_seq=2 ttl=64 time=1.25 ms (DUP!)
^C[abort]


No duplicate response seen in capture:

lee@nsyt-x-test-containerlab-1:~/netlab/csrx$ netlab capture fw1 eth2 icmp
Starting packet capture on fw1/eth2: sudo ip netns exec clab-csrx-fw1 tcpdump -i eth2 --immediate-mode -l -vv icmp
tcpdump: listening on eth2, link-type EN10MB (Ethernet), snapshot length 262144 bytes
06:25:37.849146 IP (tos 0x0, ttl 64, id 54769, offset 0, flags [DF], proto ICMP (1), length 84)
    10.1.0.9 > 10.0.0.3: ICMP echo request, id 53, seq 1, length 64
06:25:37.849192 IP (tos 0x0, ttl 64, id 9087, offset 0, flags [none], proto ICMP (1), length 84)
    10.0.0.3 > 10.1.0.9: ICMP echo reply, id 53, seq 1, length 64
06:25:38.849995 IP (tos 0x0, ttl 64, id 55654, offset 0, flags [DF], proto ICMP (1), length 84)
    10.1.0.9 > 10.0.0.3: ICMP echo request, id 53, seq 2, length 64
06:25:38.850037 IP (tos 0x0, ttl 64, id 9386, offset 0, flags [none], proto ICMP (1), length 84)
    10.0.0.3 > 10.1.0.9: ICMP echo reply, id 53, seq 2, length 64
```

netsim/ansible/templates/initial/csrx.j2
ipspace: I did a few tests with cRPD and vJunos-switch, and they worked just fine.

As for "interfaces" versus "netlab_interfaces" -- both work in this scenario as the difference is only the "loopback" interface, but it might be a good idea to consistently use "netlab_interfaces" (no need to fix anything though)

ipspace:
> and I'm not sure if it's worth noting but if you ping an ipv4 address from a cSRX device, the replies are duplicated somewhere inside the host. No duplicate response seen in capture:

If I got it right: a single ICMP reply comes to the Ethernet interface of cSRX, and then cSRX magically produces two copies of it? If that's the case, leave it to Juniper to sort it out ;))

leec-666:
yeah, maybe it's too echoey inside the container because of all the missing ~~furniture~~ features

ipspace (approved):
Looks good to merge to me ;)

ssasso (approved):

PR #3353, opened 2026-05-06
Fix: Make CI/CD tests work with Ruamel

### LINKED ISSUE #3345: [BUG] tests are failing in devcontainer
Tests are failing in devcontainer
```
vscode ➜ /workspaces/netlab/tests (dev) $ PYTHONPATH="../" python3 -m pytest --tb=no -k 'xform_ or error_cases'
================================================================================================ test session starts =================================================================================================
platform linux -- Python 3.11.15, pytest-9.0.3, pluggy-1.6.0
rootdir: /workspaces/netlab
collected 5 items / 3 deselected / 2 selected                                                                                                                                                                        

test_transformation.py FF                                                                                                                                                                                      [100%]

============================================================================================== short test summary info ===============================================================================================
FAILED test_transformation.py::test_xform_cases - AssertionError: assert 'nodes:\n  c_...er: libvirt\n' == 'input:\n- to...er: libvirt\n'
FAILED test_transformation.py::test_error_cases - assert ['IncorrectVa...= 65535', ...] == ['IncorrectVa...= 65535', ...]
========================================================================================== 2 failed, 3 deselected in 17.36s ==========================================================================================
vscode ➜ /workspaces/netlab/tests (dev) $ history
    1  pip3 install -r requirements-dev.txt
    2  cd tests/
    3  PYTHONPATH="../" python3 -m pytest --tb=no -k 'xform_ or error_cases'
    4  history
vscode ➜ /workspaces/netlab/tests (dev) $
```

I also ran tests on pristine  Ubuntu 24.04.4 LTS VM in GCP and I think they were working only until I ran `netlab install ubuntu containerlab libvirt ansible`.

ipspace:
This is more of a workaround than a complete solution. The tests pass even when Ruamel is installed and you can generate new transformation test results. Error test results cannot be generated with Ruamel due to Ruamel-specific scalar types

### DIFF
--- a/docs/dev/tests.md
+++ b/docs/dev/tests.md
@@ -5,10 +5,11 @@
 * **mypy** static type checking of all Python code in `netsim/` directory
 * Transformation tests ran with **pytest** in `tests/` directory
 
-You can run the same tests with the `run-tests.sh` script in *tests* directory. It's highly recommended you run them before creating a pull request that modifies Python code, or we'll have to have a discussion before your PR is merged.
+You can run the same tests with the `run-tests.sh` script in *tests* directory. It's highly recommended that you run them before creating a pull request that modifies Python code. PRs that fail the CI/CD pipeline will not be merged.
 
 ```{tip}
-The CI/CD tests require additional Python modules. Install them with `pip3 install -r requirements-dev.txt`
+* The CI/CD tests require additional Python modules. Install them with `pip3 install -r requirements-dev.txt`
+* The CI/CD tests use PyYAML. You can run them on a system with `ruamel.yaml` installed, but they might take longer (see [bug report](https://github.com/ipspace/netlab/issues/3345) and [related PR](https://github.com/ipspace/netlab/pull/3353) for details). Uninstalling `ruamel.yaml` might not be a bad idea. 
 ```
 
 ## Automated Tests
@@ -43,7 +44,11 @@ The transformation error tests:
 
 Whenever you're creating a new error test case or modifying an existing one, you **HAVE TO** create a corresponding *expected error messages* log file.
 
-To create the *expected error messages* files run `create-error-tests.sh` script in the *tests* directory. The script assumes that your code works flawlessly and that whatever error messages are generated are the expected error messages. That might *not* be the case, so it's highly recommended that you execute `git diff errors` after running `create-errors-tests.sh` script and do a thorough check of the differences.
+To create the *expected error messages* files, run the `create-error-tests.sh` script in the *tests* directory. The script assumes that your code works flawlessly and that whatever error messages are generated are the expected error messages. That might *not* be the case, so it's highly recommended that you execute `git diff errors` after running the `create-errors-tests.sh` script and do a thorough check of the differences.
+
+```{warning}
+You cannot create a new error test on a system with `ruamel.yaml` package (details in the [bug report](https://github.com/ipspace/netlab/issues/3345) and [related PR](https://github.com/ipspace/netlab/pull/3353)). Uninstall `ruamel.yaml` before running the `create-error-tests.sh` script.
+```
 
 ## Before Submitting a PR
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,18 @@
+#
+# Pytest hooks shared across the netlab test tree
+#
+
+from utils import HAS_RUAMEL
+
+
+def pytest_configure(config):
+  if not HAS_RUAMEL:
+    return
+  config.issue_config_time_warning(
+    UserWarning(
+      "ruamel.yaml is installed; transformation tests will be slower and "
+      "`create-error-tests.sh` is unsupported "
+      "(see https://github.com/ipspace/netlab/issues/3345)."
+    ),
+    stacklevel=2,
+  )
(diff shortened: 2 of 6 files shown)


ipspace:
@a-v-popov -- this makes CI/CD tests work with ruamel, but it's an ugly workaround until we're forced to migrate to ruamel. It also makes tests run a bit slower, as the topology must be cleaned of ruamel scalar types before proceeding.

a-v-popov (commented):
A few items, mostly AI/test driven, I'd suggest addressing before merging, plus two small follow-ups, I think might be helpful:

### 1. The detection probe matches the namespace shim, not the library

`tests/utils.py` does `import ruamel`, but `ruamel` is a PEP 420 namespace package — it can survive uninstalls of `ruamel.yaml` (the `ruamel/` directory often lingers, or another package owns part of the namespace). On my system right now:

```
>>> import ruamel
>>> import ruamel.yaml
ModuleNotFoundError: No module named 'ruamel.yaml'
```

`HAS_RUAMEL` becomes `True` while `Box`/PyYAML disagree, and `clean_ruamel` runs for nothing. python-box itself probes with `from ruamel.yaml import version_info, YAML` (`box/converters.py:22`). Suggested:

```python
try:
  import ruamel.yaml  # noqa: F401
  HAS_RUAMEL = True
except ImportError:
  HAS_RUAMEL = False
```

### 2. `yaml.dump` should pin `default_flow_style=False`

`Box.to_yaml(...)` in python-box 7.x defaults to `default_flow_style=False, width=120` and calls `yaml.dump(obj, default_flow_style=default_flow_style, width=width, ...)` (`box/converters.py:195`, `box/box.py:1018`). The PR replaces it with:

```python
return yaml.dump(topology.to_dict(), width=120)
```

PyYAML's own default for `default_flow_style` is `None` ("auto: flow for inner scalar-only collections"), not `False`. Today's fixtures are all under-mappings, so the two modes coincide and the suite passes. The next fixture with a top-level scalar list, or a deeply nested scalar-only structure, will silently diverge from the format every committed `expected/*.yml` was generated in. One-line fix:

```python
return yaml.dump(topology.to_dict(), default_flow_style=False, width=120)
```

### 3. `log._ERROR_LOG = []` is redundant

The previous line was `log.err_count = 0`, which was a typo from the start — `log.err_count` doesn't exist; the public counter is `get_error_count()` at `netsim/utils/log.py:389`. The new `log._ERROR_LOG = []` does target a real attribute, but `run_test` immediately calls `log.init_log_system(header=False)`, which itself does `_ERROR_LOG = []` (`log.py:539`). The reset is redundant, and reaching into a leading-underscore symbol from outside the module is a smell. Either drop the line, or call `log.init_log_system(header=False)` if a defensive reset is desired.

### Two small follow-ups worth folding in

- **`tests/conftest.py` (new):** the standard-pytest-safe way to surface "ruamel.yaml is installed, things will be slower" is `pytest_configure` + `config.issue_config_time_warning(UserWarning(...))`. It appears in pytest's warnings summary, never fails a run, and doesn't touch collection / `-k` filtering / parallelism. Import `HAS_RUAMEL` from `tests/utils.py` so there's a single source of truth.
- **Gate the fixture generators.** The PR docs already say error-log fixtures can't be generated under ruamel, but `tests/create-error-tests.sh` doesn't enforce it — a user can still run it and commit broken `.log` files. A 4-line `python3 -c "import ruamel.yaml" && exit 2` guard at the top closes that hole. `tests/create-transformation-test-case.py` *does* work under ruamel (the `clean_ruamel` walk handles it), so a stderr advisory there is enough.

Reference implementation of all of the above on `a-v-popov/netlab@pr-3353-followup` ([compare view](https://github.com/ipspace/netlab/compare/dev...a-v-popov:netlab:pr-3353-followup)) — verified end-to-end:

- Without `ruamel.yaml`: xform + error tests pass; no warning.
- With `ruamel.yaml` 0.19.1: xform + error tests pass (+~40% wall time, matching the "slower" claim); `UserWarning` appears in pytest's warnings summary; nothing fails or skips.
- `create-error-tests.sh` with `ruamel.yaml` installed: exits 2 with an actionable message before doing anything destructive.
- `ruff check .` clean.

Happy to open the follow-up as a separate PR against `dev` after this lands, or fold the changes in here if you prefer.

ipspace:
> A few items, mostly AI/test driven, I'd suggest addressing before merging, plus two small follow-ups, I think might be helpful:

Awesome job. Thanks a million. What AI model/harness were you using?

> Happy to open the follow-up as a separate PR against `dev` after this lands, or fold the changes in here if you prefer.

It would be great if you could either add changes to this PR or submit a PR against this branch, so that we end with a single commit in the dev branch.

a-v-popov:
> What AI model/harness were you using?

That would be Claude Code running in tmux on remote VM in a cloud and using official Opus 4.7 model at xhigh effort, if I understood the question correctly.  

> It would be great if you could either add changes to this PR or submit a PR against this branch, so that we end with a single commit in the dev branch.

Submitted https://github.com/ipspace/netlab/pull/3362
Please let me know if anything is off.

ipspace:
> That would be Claude Code running in tmux on remote VM in a cloud and using official Opus 4.7 model at xhigh effort

Thank you. Exactly what I was looking for!

> Submitted #3362 Please let me know if anything is off.

The only missing detail were the type annotations in the new test file. Thanks again.

PR #3354, opened 2026-05-06
Fix netlab down within the netlab restart command

sjhloco:
When using `netlab restart` to restart a lab the `netlab down` part of it fails as the topology file name is being passed in to the command (only needed for `netlab up`) 

```bash
ste@netlab-dev:~/labs$ netlab restart simple_topo.yml
usage: netlab down [-h] [-v] [--cleanup] [--dry-run] [--force] [-i INSTANCE] [--snapshot [SNAPSHOT]]
netlab down: error: unrecognized arguments: simple_topo.yml
```

Fixed in `restart.py` by adding the topology file to the filtering for netlab down. 

```bash
ste@netlab-dev:~/labs$ netlab restart simple_topo.yml

┌──────────────────────────────────────────────────────────────────────────────────┐
│ CHECKING virtualization provider installation                                    │
└──────────────────────────────────────────────────────────────────────────────────┘
[SUCCESS] clab installed and working correctly

┌──────────────────────────────────────────────────────────────────────────────────┐
│ STOPPING clab nodes                                                              │
└──────────────────────────────────────────────────────────────────────────────────┘
provider clab: executing sudo -E containerlab destroy --cleanup -t clab.yml
warning: preserving the entire environment is not supported, `-E` is ignored
23:43:11 INFO Parsing & checking topology file=clab.yml
23:43:11 INFO Parsing & checking topology file=clab.yml
23:43:11 INFO Destroying lab name=labs
23:43:11 INFO Removed container name=ACCESS_SWI
23:43:11 INFO Removing host entries path=/etc/hosts
23:43:11 INFO Removing SSH config path=/etc/ssh/ssh_config.d/clab-labs.conf

┌──────────────────────────────────────────────────────────────────────────────────┐
│ CREATING configuration files                                                     │
└──────────────────────────────────────────────────────────────────────────────────┘
[CONFIG]  ACCESS_SWI: ifstate,normalize,initial,vlan
[CREATED] provider configuration file: clab.yml
[CREATED] pickled transformed topology data into netlab.snapshot.pickle
[GROUPS]  group_vars for all
[GROUPS]  group_vars for modules
[GROUPS]  group_vars for ioll2
[HOSTS]   host_vars for ACCESS_SWI
[CREATED] minimized Ansible inventory hosts.yml
[CREATED] Ansible configuration file: ansible.cfg

┌──────────────────────────────────────────────────────────────────────────────────┐
│ CHECKING virtualization provider installation                                    │
[...]

PR #3380, opened 2026-05-11
Rewrite the 'netlab restart' functionality

### LINKED ISSUE #3359: [BUG] netlab restart is broken
Right now, "netlab restart" works only if there's `topology.yml` in the current directory. Specifying topology file crashes the "netlab down" part, and trying to restart a lab started from any other topology file (without specifying it) crashes "netlab up" part.

According to the documentation, the command should restart the current lab from the snapshot file:

https://netlab.tools/netlab/restart/

Obviously, the code has to be modified to match the documentation :( 

### Implementation hints

The command has to recognize just a few flags (-v, -q, --no-config, --fast-config, and --snapshot). The args have to be parsed (to check them), the verbosity level has to be set, and then **down.run** is called only with the args without up-only args, and **up.run** is called with all arguments plus '--snapshot' if it's not in the original args.

_Based on observation in #3354 by @sjhloco_

ipspace:
The 'netlab restart' functionality was mostly broken and did not work with the 'more recent' additions like non-default snapshot files or explicit lab instances.

This commit completely rewrites the top-level 'netlab restart' workflow and hopefully makes the code and documentation match.

Closes #3359 and replaces #3354

### DIFF
--- a/docs/netlab/restart.md
+++ b/docs/netlab/restart.md
@@ -1,26 +1,20 @@
-# Restart Virtual Lab
+# Restart a Virtual Lab Instance
 
-**netlab restart** executes **[netlab down](down.md)** followed by **[netlab up](up.md)** to restart your lab from the transformed lab topology stored in `netlab.snapshot.pickle` snapshot file.
+Use the **netlab restart** command to shut down and restart a running lab instance. You can use this command if the lab is messed up beyond hope, or after changing the lab topology file.
 
-You can use **netlab restart** to restart the existing lab (use `--snapshot` keyword) or recreate the lab configuration files if you changed the lab topology.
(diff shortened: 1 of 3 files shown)


ipspace:
@sjhloco -- in the end, I rewrote the whole stuff. The old code was obsolete and way too trivial (I remember someone asking for it and me hacking it together in a minute or two). Would love to hear whether it works for you.

@DanPartelly -- do you want to look at this?

sjhloco:
> @sjhloco -- in the end, I rewrote the whole stuff. The old code was obsolete and way too trivial (I remember someone asking for it and me hacking it together in a minute or two). Would love to hear whether it works for you.

Yes that now works, thanks a lot for fixing 👍

ipspace:
> Yes that now works, thanks a lot for fixing 👍

Thanks for the confirmation, merging...

Issue #3397, opened 2026-05-15
vSRX vagrant box fails to boot on Ubuntu 24.04 / QEMU 8.x — kernel panic at mtx_platform_early_bootinit


sdargoeuves:
### Environment

Host OS: `Ubuntu 24.04` (**upgraded in-place from 20.04**)
QEMU: `8.2.2` (Debian 1:8.2.2+ds-0ubuntu1.16)
libvirt: `10.0.0`
Kernel: `6.8.0-111-generic`
Host CPU: `AMD EPYC Genoa`
vSRX image: `juniper/vsrx3:21.3R1.9`

### Symptom

vSRX panics immediately on every boot attempt with a kernel page fault in Juniper's early init code, despite this box working fine with Ubuntu 20.04:

```bash
mtx_platform_early_bootinit: M/T/EX/SRX Series Early Boot Initialization
Fatal trap 12: page fault while in kernel mode
fault virtual address = 0x0
instruction pointer   = 0x20:0xffffffff8057e5f0
panic: page fault
Uptime: 1s
```

The panic is deterministic — same instruction pointer every time, regardless of CPU model passed to the guest (qemu64, Haswell-noTSX, etc).

### Root cause (hypothesis)

Ubuntu 24.04 ships QEMU 8.2 which defaults to machine type `pc-i440fx-noble-v2`. The vSRX FreeBSD-based kernel appears to crash in early platform detection when running under this machine type. The same host hardware worked fine on Ubuntu 20.04 (QEMU 4.2??), which defaulted to `pc-i440fx-focal`.

### Workaround

Adding `domain.machine_type = "pc-i440fx-6.2"` to `netsim/templates/provider/libvirt/vsrx-domain.j2` resolves the issue.

### Questions for discussion

- Has anyone else hit this after upgrading to Ubuntu 24.04 / QEMU 8.x? I haven't seen any reports, so this may be specific to my setup — happy to be told I'm the only one.
- Is `machine_type` something that should be configurable via the device template `vsrx.yml` under `libvirt:` rather than hardcoded in the template? It doesn't appear to be an exposed parameter in the provider currently. I don't want to fix something for me that could break stuff for others.

### Additional note

I also updated `vsrx.xml.j2` (the box build template) with the same machine type fix, so rebuilding a box from scratch using the netlab libvirt command now works correctly too!

[...]

PR #3399, opened 2026-05-15
Change vagrant parameter for VSRX

sdargoeuves:
link to this [issue](https://github.com/ipspace/netlab/issues/3397)


sdargoeuves:
the 2nd commit makes the change simpler, as I no longer add the xml file, but edit the virt-install command to add the `--machine pc-i440fx-6.2`

ipspace:
Thanks a million. For the moment, I'd just go with your changes, no need for fancy nerd knobs.

I will try this out on my 22.04 as soon as the current batch of integration tests completes (= hours, vPTX is almost as bad as Nexus OS).

sdargoeuves:
I've now ran a quick test after the 2nd commit, to test the updated virt-install command without using the xml, the VM booted correctly.

> vPTX is almost as bad as Nexus OS

two platforms I am not yet using in the lab... but to be fair the vsrx is not fast either.

ipspace:
@DanPartelly -- You're the only one I know not using Ubuntu. Could you check whether this works for you, or should we just say "hey, we tested this on Ubuntu, if it doesn't work, use vrnetlab"?

ipspace (approved):
Looks good

Issue #3403, opened 2026-05-16
`libvirt.uplink` fails after upgrading from ubuntu 20.04 to 24.04


sdargoeuves:
## Describe the bug

When using `libvirt.uplink` in a topology to connect an ASAv firewall to a physical NIC so I can create my VPN to the cloud, vagrant fails immediately after domain creation with:

````
Error while attaching new device to domain. Call to virDomainAttachDeviceFlags failed:
unsupported configuration: setting MTU on interface type direct is not supported yet
````

## Environment

- Host OS: Ubuntu 24.04 (upgraded in-place from 20.04)
- libvirt: 10.0.0
- QEMU: 8.2.2
- vagrant-libvirt: 0.12.2
- netlab: latest `dev`

## OS/libvirt version context

This was working correctly on Ubuntu 20.04. The issue appeared after upgrading to Ubuntu 24.04 which ships libvirt 10.0.0. The situation on Ubuntu 22.04 is unknown.

## Root cause

In `netsim/templates/provider/libvirt/libvirt-bridge.j2`, `:libvirt__mtu => 9500` is set unconditionally for all interface types, including public/direct (macvtap) interfaces created by `libvirt.uplink`:

```jinja
                  :libvirt__mtu => 9500,
                  :autostart => true,
                  :auto_config => false
```

libvirt 10.x rejects the `<mtu>` element on `type='direct'` interfaces. Older libvirt silently ignored it.

Note: setting `mtu: 1500` on the affected links in the topology did not help, probably because it would be applied as configuration after the node is up during the `netlab initial` part.

## Fix applied

In `libvirt-bridge.j2`, wrap the `:libvirt__mtu` line so it is only emitted for private (bridge) networks, not public/direct ones:

```jinja
{% if not pubnet %}
                  :libvirt__mtu => 9500,
{% endif %}
```

This fix has been tested on Ubuntu 24.04 with libvirt 10.0.0 using `libvirt.uplink` and resolves the issue.

## Potential side effects

I am not certain whether skipping `:libvirt__mtu` on public interfaces has any unintended consequences, and I have not tested `libvirt.public`. 

I will create the PR for further discussion.

PR #3404, opened 2026-05-16
libvirt: do not set MTU on public/direct interfaces

sdargoeuves:
related to this [issue](https://github.com/ipspace/netlab/issues/3403)


ipspace (approved):
Thank you! The only thing I changed: I moved the MTU setting into the existing "private network" block.

PR #3405, opened 2026-05-17
CI: Capture rendered output and unified diff as pytest artifacts

a-v-popov:
Follow-up to #3391 + #3402.  After those two PRs each YAML case is its own
pytest node and per-case file writes are isolated to `tmp_path`, but when a
transformation or error-log fixture mismatches in CI the GitHub Actions log
only shows pytest's truncated assertion diff -- there is no way to retrieve
the actual rendered YAML without re-running locally.  This PR plumbs the
artifact path end-to-end and makes the in-terminal failure message useful
in its own right.

## Summary

- `_report_mismatch` helper in `tests/test_transformation.py`, used by both
  `run_transformation_test` and `run_error_case`.  On mismatch it:
  1. writes the produced content to `tmp_path/<base>.{yml,actual}`,
  2. writes a `difflib.unified_diff` (hunks + 3 lines of context) to
     `tmp_path/<base>.diff`,
  3. calls `pytest.fail(message, pytrace=...)` with a diff preview embedded
     in the failure message.
- Preview length and pytrace scale with pytest's `-v` count -- CI / default
  cap at 20 lines, `-v` doubles it, `-vv` removes the cap, `-vvv` also
  re-enables `pytrace`.  The full diff is on disk regardless.
- `run_error_case` gains a `tmp_path` parameter, threaded through
  `test_error_cases` / `test_coverage_errors` so the helper drops
  `error_log.actual` / `error_log.diff` per case.
- `tests/conftest.py`: `pytest_runtest_makereport` hookwrapper trims
  `report.longrepr.reprcrash.message` to its first line so pytest 9.x's
  short summary doesn't duplicate the diff body that the helper already
  embeds in the failure message.
- CI workflows (`t-push.yml`, `t-pull.yml`): a `Compute pytest tmp path`
  step exports `PYTEST_TMP=$RUNNER_TEMP/pytest-tmp` to `$GITHUB_ENV`,
  pytest gets `--basetemp="$PYTEST_TMP" --junit-xml="$PYTEST_TMP/junit.xml"`,
  an `actions/upload-artifact@v7` step gated on `!cancelled() &&
  steps.pytest.outcome == 'failure'` (AND'd with the existing fork-skip
  guard in `t-pull.yml`), and a small prune step between them.

### Constructing the failure message ourselves, and `pytest.fail` over `raise AssertionError`

Two related but distinct choices in `_report_mismatch`.

**Why not `assert actual == expected`:** pytest's assertion rewriter
introspects `assert a == b` and emits an `ndiff` representation.  Two
truncation layers apply:

1. At default verbosity it strips the long identical prefix/suffix and the
   output is capped at ~8 lines with `use '-v'/'-vv' to show` messages.
2. At `-v` it would print the fist ~8 lines of the file potentially containing no diff
3. At `-vv` the cap is removed -- but `ndiff` emits every line of the file,
   so a one-line fixture mismatch surfaces as thousands of matching
   lines with the single changed line lost in the middle.

`difflib.unified_diff` (hunks + 3 lines of context) is the right algorithm
for fixture comparisons -- a single-line YAML mismatch is ~7 readable lines.
Constructing the message ourselves (whether via `pytest.fail` or
`raise AssertionError`) hands the rewriter a preformatted string with no
introspectable comparison, so it lands in the test output verbatim.

**Why `pytest.fail` over `raise AssertionError`:** `pytest.fail` takes
`pytrace=False` (here `pytrace=verbose >= 3`).  A fixture mismatch needs
no helper-frame traceback at default verbosity -- the diff preview and
the on-disk pointer are the whole story.  `-vvv` re-enables `pytrace` for
harness debugging.

### Trimming `reprcrash.message`

pytest 9.x prints the full `pytest.fail(...)` message verbatim in the
`short test summary info` block whenever it detects CI (`CI` /
`BUILD_NUMBER` env vars) or runs at `-vv`+, which would duplicate the
diff body the helper deliberately embeds in the failure message.
A `pytest_runtest_makereport` hookwrapper trims
`report.longrepr.reprcrash.message` to its first line; `longrepr` is
untouched, so the `FAILURES` section still carries the diff preview and
the `Full diff: ...` pointer.  Pytest's own `ReprFileLocation.toterminal`
already applies the same first-line trim at one display site; the hook
just enforces the invariant for every consumer (short summary, JUnit XML,
custom reporters).

### `$RUNNER_TEMP` basetemp

GitHub designates `$RUNNER_TEMP` for ephemeral writes on its runners --
whatever they do with that mount today or tomorrow (tmpfs, separate
volume, faster device), the workflow inherits.  The natural shape would
be a single `jobs.tests.env` entry, but `${{ runner.* }}` is not available
outside a step context (GitHub rejects with
`Unrecognized named-value: 'runner'`).  A `Compute pytest tmp path` step
writes `PYTEST_TMP=$RUNNER_TEMP/pytest-tmp` to `$GITHUB_ENV`, after which
`$PYTEST_TMP` is in every subsequent step's shell env and
`${{ env.PYTEST_TMP }}` resolves in every subsequent step's expressions.
Single source of truth, no per-step `env:` blocks.

### `actions/upload-artifact@v7`

`v4` runs on Node 20, which GitHub forces off the runner on 2026-09-16
(and switches to Node 24 by default on 2026-06-02).  `v7` is the latest
stable, defaults to Node 24, and the inputs we use (`name`, `path`,
`retention-days`) are unchanged.

### Prune step before the artifact upload

`--basetemp` creates a directory per test invocation (~273 for
transformation + errors).  Passing cases early-return from
`_report_mismatch` and leave their case dir empty, so without pruning the
uploaded zip is dominated by them.  `find -type d -empty -delete` clears
those; `find -xtype l -delete` then catches pytest's `<test>-current`
symlinks that go dangling after the empty-dir sweep, preventing
`upload-artifact` (which follows symlinks by default) from including
zero-byte placeholders.  `junit.xml` sits at the basetemp root as a regular
file and is unaffected by either find expression.  `find -delete` returns
0 on empty match, so the step is idempotent.

### `!cancelled() && steps.pytest.outcome == 'failure'` gating

The earlier draft of this PR gated the prune + upload on `if: failure()`,
which fires on any prior step failure.  If `pip install` or `mypy` failed
before pytest ran, `$PYTEST_TMP` did not exist, the `find` in the prune
step exited nonzero, and CI showed a misleading second red step with no
diagnostic value.  `!cancelled()` overrides the implicit `success()` that
GitHub prepends to every `if:`, and `steps.pytest.outcome == 'failure'`
narrows the trigger to pytest itself -- the prune and upload steps now
run iff there is something to upload.

## What changes

- **`tests/test_transformation.py`**
  - Adds `_report_mismatch(test_case, label, actual, expected, tmp_path,
    actual_name, verbose)`.  Uses `pathlib.Path.with_suffix(".diff")` to
    derive the diff filename from the actuals filename (so transformation
    gets `result.yml` + `result.diff`, errors get `error_log.actual` +
    `error_log.diff`).
  - `run_transformation_test` and `run_error_case` become thin wrappers
    that call their respective `*_results` helper and then
    `_report_mismatch`.
  - `run_error_case` gains `tmp_path`; `test_error_cases` and
    `test_coverage_errors` pass it through.
  - All four parametrized test functions take `pytestconfig` and forward
    `max(0, pytestconfig.getoption("verbose"))` so the helper can size
    its preview.
- **`tests/conftest.py`**
  - `pytest_runtest_makereport` hookwrapper that trims
    `report.longrepr.reprcrash.message` to its first line when present.
- **`.github/workflows/t-push.yml` and `t-pull.yml`**
  - `Compute pytest tmp path` step exports
    `PYTEST_TMP=$RUNNER_TEMP/pytest-tmp` to `$GITHUB_ENV`.
  - Pytest invocation becomes
    `pytest --basetemp="$PYTEST_TMP" --junit-xml="$PYTEST_TMP/junit.xml"`
    and gains `id: pytest`.
  - `Prune empty case dirs` step runs the two `find` calls.
  - `Upload pytest artifacts` step uses `actions/upload-artifact@v7`,
    named `pytest-tmp-py${{ matrix.python-version }}` so the per-matrix-cell
    artifacts do not collide.
  - `t-pull.yml` repeats the existing fork-skip guard
    (`github.event.pull_request.head.repo.full_name != 'ipspace/netlab'`)
    on the compute, prune and upload steps.

## Anticipated questions

- **"Why no `.gitignore` entry for `pytest-tmp/`?"**  Nothing in the repo
  points pytest at an in-tree basetemp anymore: CI writes to
  `$RUNNER_TEMP/pytest-tmp` and pytest's local default is
  `/tmp/pytest-of-<user>/`.  An entry would only fire on an explicit
  `--basetemp=tests/pytest-tmp` that nothing in the repo does or
  documents -- dead weight in a config that should describe paths the
  repo actually produces.
- **"Why a `_report_mismatch` helper instead of inlining in both wrappers?"**
  The two wrappers had ~8 lines of mechanical duplication (compute diff,
  write actual, write diff, fail).  One helper, two thin call sites.
- **"Will the prune step fight with a future test that legitimately leaves
  files in `tmp_path`?"**  Only deletes *empty* directories and *broken*
  symlinks.  A test that writes anything keeps its dir; the
  inventory-output cases already do, and they survive the prune unchanged.
- **"Cosmetic: unified-diff in pytest's terminal shows `-q+Fatal...` for
  trailing lines without newlines."**  Python's `difflib.unified_diff` does
  not emit GNU diff's `\ No newline at end of file` sentinel.  The byte
  content matches what's on disk; `diff -u` on the artifact files gives
  the GNU-annotated view.

## Test plan

- [ ] `cd tests && PYTHONPATH=../ pytest -q` -- all green.
- [ ] Corrupt `topology/expected/6pe.yml` (e.g. flip
      `advertise_loopback: true` -> `false`),
      `pytest --basetemp=/tmp/pytest-tmp -k 6pe`.
      Expect: failure message contains a `--- expected` / `+++ actual`
      unified diff with hunk headers; no `Skipping N identical leading
      characters` / `use -vv to show` markers; no helper-frame traceback;
      `/tmp/pytest-tmp/<case>/result.yml` and `result.diff` written;
      short summary holds only the first line.  Revert.
- [ ] Corrupt one error-log fixture (e.g. `echo q >> errors/<case>.log`),
      rerun.  Expect: `error-log mismatch` message with diff body;
      `error_log.actual` and `error_log.diff` in the case `tmp_path`.
      Revert.
- [ ] Full run with one corruption in place: confirm
      `find <basetemp> -type d -empty -delete && find <basetemp> -xtype
      l -delete` reduces the layout to the failing case (with its
      `.diff` pair) plus the non-empty inventory-output cases plus
      `junit.xml`.
- [ ] CI on this branch: t-push and t-pull both green; failure case
      (introduced temporarily) uploads a `pytest-tmp-py3.X` artifact
      with the expected contents.

### DIFF
--- a/.github/workflows/t-pull.yml
+++ b/.github/workflows/t-pull.yml
@@ -37,11 +37,31 @@ jobs:
         for file in netsim/extra/*/plugin.py; do
           python3 -m mypy $file
         done
+    - name: Compute pytest tmp path
+      if: ${{ github.event.pull_request.head.repo.full_name != 'ipspace/netlab' }}
+      run: echo "PYTEST_TMP=$RUNNER_TEMP/pytest-tmp" >> "$GITHUB_ENV"
     - name: Run transformation tests
+      id: pytest
       if: ${{ github.event.pull_request.head.repo.full_name != 'ipspace/netlab' }}
       run: |
         cd tests
-        PYTHONPATH="../" pytest
+        PYTHONPATH="../" pytest --basetemp="$PYTEST_TMP" --junit-xml="$PYTEST_TMP/junit.xml"
+    - name: Prune empty case dirs
+      if: >-
+        ${{ !cancelled() && steps.pytest.outcome == 'failure'
+            && github.event.pull_request.head.repo.full_name != 'ipspace/netlab' }}
+      run: |
+        find "$PYTEST_TMP" -type d -empty -delete
+        find "$PYTEST_TMP" -xtype l -delete
+    - name: Upload pytest artifacts
+      if: >-
+        ${{ !cancelled() && steps.pytest.outcome == 'failure'
+            && github.event.pull_request.head.repo.full_name != 'ipspace/netlab' }}
+      uses: actions/upload-artifact@v7
+      with:
+        name: pytest-tmp-py${{ matrix.python-version }}
+        path: ${{ env.PYTEST_TMP }}
+        retention-days: 7
     - name: Check integration tests
       run: |
         cd tests
--- a/.github/workflows/t-push.yml
+++ b/.github/workflows/t-push.yml
@@ -31,7 +31,22 @@ jobs:
         for file in netsim/extra/*/plugin.py; do
           python3 -m mypy $file
         done
+    - name: Compute pytest tmp path
+      run: echo "PYTEST_TMP=$RUNNER_TEMP/pytest-tmp" >> "$GITHUB_ENV"
     - name: Run transformation tests
+      id: pytest
       run: |
         cd tests
-        PYTHONPATH="../" pytest
+        PYTHONPATH="../" pytest --basetemp="$PYTEST_TMP" --junit-xml="$PYTEST_TMP/junit.xml"
+    - name: Prune empty case dirs
+      if: ${{ !cancelled() && steps.pytest.outcome == 'failure' }}
+      run: |
+        find "$PYTEST_TMP" -type d -empty -delete
+        find "$PYTEST_TMP" -xtype l -delete
+    - name: Upload pytest artifacts
+      if: ${{ !cancelled() && steps.pytest.outcome == 'failure' }}
+      uses: actions/upload-artifact@v7
+      with:
+        name: pytest-tmp-py${{ matrix.python-version }}
+        path: ${{ env.PYTEST_TMP }}
+        retention-days: 7
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,10 +14,22 @@
 #    transformation tests will be slower and create-error-tests.sh is
 #    unsupported (see https://github.com/ipspace/netlab/issues/3345).
 #
+# 4. Normalize the failure-report "reprcrash" message to its first line so
+#    pytest's `short test summary info` does not duplicate the diff body
+#    that `_report_mismatch` puts in the `pytest.fail(...)` message. The
+#    rich multi-line message stays in `longrepr`, so the `FAILURES`
+#    section is unaffected. Without this, pytest 9.x prints the entire
+#    failure message verbatim in the short summary whenever it detects
+#    CI (`CI`/`BUILD_NUMBER` env vars) or runs at `-vv`+ -- see
+#    `_pytest/terminal.py::_get_line_with_reprcrash_message`. Pytest's
+#    own `ReprFileLocation.toterminal` already trims to the first line;
+#    we just enforce the same invariant globally.
+#
 
 import os
 import pathlib
 import sys
+import typing
 
 import pytest
 
@@ -38,3 +50,12 @@ def pytest_configure(config: pytest.Config) -> None:
       ),
       stacklevel=2,
     )
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_makereport(item: pytest.Item, call: pytest.CallInfo[None]) -> typing.Generator[None, None, None]:
+  outcome = yield
+  report = outcome.get_result()  # type: ignore[attr-defined]
+  crash = getattr(getattr(report, "longrepr", None), "reprcrash", None)
+  if crash is not None and "\n" in crash.message:
+    crash.message = crash.message.split("\n", 1)[0]
--- a/tests/test_transformation.py
+++ b/tests/test_transformation.py
@@ -4,6 +4,7 @@
 # topology file
 #
 
+import difflib
 import glob
 import os
 import pathlib
@@ -20,6 +21,8 @@ from netsim.outputs import _TopologyOutput, ansible
 from netsim.utils import log
 from netsim.utils import read as _read
 
+MAX_PREVIEW_LINES = 20
+MAX_PREVIEW_LINES_VERBOSE = 2 * MAX_PREVIEW_LINES
 
 def run_test(fname: str) -> Box:
   log.init_log_system(header = False)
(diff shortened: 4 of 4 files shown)


ipspace:
While I love the more concise error message, the whole thing is a bridge too far for me.  Storing full diffs in CI/CD artifacts seems a bit excessive, given how often we have to deal with them  (not very often), and how many people stumble upon this problem (two or three).

I will try to pull just the good parts (= concise error message) out of it.

On a tangential topic, here's what works best for me:

* Run "run-xform.sh" as part of pre-commit tests. If that fails, I haven't even triggered the CI/CD pipeline yet.
* The only scenario where the CI/CD pipeline fails are failed transformations of integration tests (which are not covered here anyway) and differences in coverage tests (which should not change that often anyway).
* When I get notified that the pipeline failed, I recreate the test results locally and then use the local "git diff" to figure out what has changed, and whether that's expected.

Should I add this to the developer documentation?

a-v-popov:
The logic was as follows:
1. `-v` could result in no useful output on a big fixture, I couldn't find a way to do fix it cleanly without manual diff
2.  if a diff is computed it could be big - it has to be truncated and it cost nothing to drop it on disk
3.  if a diff is on disk it virtually free to push it as an artifact, storing one small zip for 7 days should be OK
But I can see how this might feel an overkill. Feel free to cut it wherever you feel appropriate. 

Re documentation, I think explicit list of tests to pass before push should be sufficient, or maybe just a reference to AGENTS.md which could be considered an authoritative source.

a-v-popov:
An example of failed test run could be found here with zip at the bottom of the page.
https://github.com/a-v-popov/netlab/actions/runs/25990338590

ipspace:
Replaced by #3436

PR #3406, opened 2026-05-17
Libvirt machine type

sdargoeuves:
> Not at the moment, and until we get other similar issues, I wouldn't add one.

I know this was said in https://github.com/ipspace/netlab/issues/3397#issuecomment-4458095443 regarding configuring the machine_type for libvirt.

Still dealing with the aftermath of my upgrade from 20.04 to 24.04, this time it's my F5 load balancer vagrant box that was affected — a linux device as far as netlab is concerned. Very similar to the Juniper VSRX issue and the fix is also changing the `machine_type` for that device.

I didn't want to change it for all libvirt linux hosts, as it could have a wider blast radius, and it seems to be something quite specific for my use case. For that reason, I thought having an option to select the `machine_type` in the node attribute would work quite well.


ipspace (approved):
Perfect. Thank you! Now we have to reimplement the vSRX fix to set the `libvirt.machine_type` in device-specific provider data ;)

PR #3411, opened 2026-05-18
Use machine_type under vsrx device attribute instead of libvirt template

sdargoeuves:
https://github.com/ipspace/netlab/pull/3406#pullrequestreview-4308096656

I didn't understand why it wouldn't work, until I've realised I needed to have it under `node` inside `libvirt`

### DIFF
--- a/netsim/devices/vsrx.yml
+++ b/netsim/devices/vsrx.yml
@@ -23,6 +23,8 @@ virtualbox:
   image: juniper/vsrx3
 libvirt:
   image: juniper/vsrx3
+  node:
+    machine_type: pc-i440fx-6.2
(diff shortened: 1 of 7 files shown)


sdargoeuves:
Hopefully this doesn't fall into the new classic: "AI changed the tests to fit the (bad) code it created", because I didn't use AI to edit those tests!

More seriously, I am not sure what's the right approach when it comes to testing. There were 4 topology containing `vsrx`, I've updated 4 output files and 1 input files (to test when the value is set to false).

ipspace (approved):
Looks great to me. Thank you! Now I have to change the box-building process to use device machine_type if defined.

PR #3421, opened 2026-05-20
Feature/multiserver plugin

### LINKED ISSUE #3420: Proposal: multiserver plugin to easily split topologies across physical servers
### Idea

So I already once told about the idea of having a plugin for multiserver support.
I want to run larger or heavier topologies across multiple physical servers.

I now took my external wrapper-script and turned it into an original netlab plugin.
This outsources the logic out of my project using the direct netlab source (which just is a cleaner approach than what i had before) and maybe someone else might be interested. Who knows.

Key design aspects:
- **Zero Core Changes:** All logic is housed entirely in the plugin (`netsim/extra/multiserver/`).
- **No changes to `netlab up`:** Central workstation generates a filtered snapshot (`netlab.snapshot.pickle`) for each server. Remote servers then deploy normally using `sudo netlab up --snapshot`.
- **Interconnects:** Uses Containerlab's native VXLAN endpoints for P2P links and local host-level VXLAN tunnels for multi-access links.
- **Provider Limitation:** Currently supports the `clab` provider only.

Detailed configuration syntax, assignment modes (`explicit`/`auto`), and the workflow are documented in the accompanying PR's documentation file: `docs/plugins/multiserver.md`.

I am not sure if it really is worthy as an official netlab plugin so let me know if it is better to keep as my own.
As it does only generate a config...
Sadly this only covers the containerlab-provider.

Maybe you have ideas but this is my attempt for now.

Muddyblack:
Reference: #3420 

### Summary
This PR adds the `multiserver` plugin to distribute a single Netlab topology across multiple physical servers. 
Sadly for now containerlab-provider only.

### Key Details
- **Self-contained:** The implementation is entirely within `netsim/extra/multiserver/` and doesn't modify any core Netlab engine logic.
- **Consistent Allocations:** IP, interface, and VNI allocations are computed once on the workstation. The plugin then generates a per-server directory with a filtered `clab.yml` and `netlab.snapshot.pickle`.
- **Native Remote Deployments:** Remote servers launch using standard `sudo netlab up --snapshot -vv` without needing custom CLI options.

For the test-files I am not sure if they make any sense. But they show at least it does not interfere with the normal netlab workflow.

Explanations on how it works can be found in `docs/plugins/multiserver.md`.


docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+```eval_rst
+.. contents:: Table of Contents
+   :depth: 2
+   :local:
+   :backlinks: none
+```
+
+```{warning}
+* The *multiserver* plugin requires the **containerlab** provider on all servers.
+* Containerlab version >= `0.46` is required for native VXLAN link endpoint support.
```
ipspace: Just a thought: the plugin probably does not work with old releases, and we're enforcing 0.75.0 right now.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+| **assignment** | string | How to assign nodes to servers: `explicit` (default) or `auto` |
+| **servers** | list | List of target physical servers |
+| **vxlan** | dictionary | Global settings for VXLAN tunnels |
+| **replicate** | list | Nodes or groups that must be duplicated on all servers |
+| **output_dir** | string | Template for per-server directory names (default: `server-{server_id}`) |
+
+(multiserver-servers)=
+### Server Parameters
+
+Each entry in the **multiserver.servers** list supports these parameters:
```
ipspace: Would it make more sense to have a dictionary of servers?

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+
+| Parameter | Type | Meaning |
+|-----------|------|---------|
+| **id** | integer | Unique identifier for the server (e.g. `1`, `2`) |
```
ipspace: ID could be assigned automatically, like we do for nodes. We even have a set of functions (modules/_dataplane) to handle IDs where some objects have a static ID and others need auto-assigned ones -- used for VLANs, VRFs, and the like

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+      host: 192.168.168.129
+      groups: [ spines, leaves ]
+```
+
+### Automatic Assignment
+
+In `auto` mode, nodes that are not explicitly pinned to a server are distributed automatically using a greedy balancing algorithm:
+
+1. Nodes belonging to a *netlab* group are kept together — the entire group is placed on the server that currently has the fewest nodes. Larger groups are placed first for better balance.
+2. Remaining ungrouped nodes are assigned one at a time to the least-loaded server.
```
ipspace: Haven't looked at the code yet, but I'm guessing that "loaded" means "number of nodes", not something more complex like CPU/RAM requirements of lab devices? If that's the case, it might be worth spelling it out.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+      groups: [ core ]
+      members: [ edge-node ]
+    - id: 2
+      host: 192.168.168.129
+      groups: [ spines, leaves ]
+```
+
+### Automatic Assignment
+
+In `auto` mode, nodes that are not explicitly pinned to a server are distributed automatically using a greedy balancing algorithm:
```
ipspace: In some future version, you might want to add server capabilities for weighted distribution ;)

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+
+**Step 1: Generate configurations** on your workstation:
+
+```bash
+netlab create topology.yml
+```
+
+The plugin automatically copies all required files into each server directory — no extra bundling step is needed.
+
+**Step 2: Copy server directories to remote hosts** (e.g. via rsync):
```
ipspace: We might want to automate that in the future, but this is definitely more than good enough for version 1

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    return
+
+  # Merge plugin defaults with user config (user values take priority)
+  defaults = topology.defaults.get("multiserver", Box({}))
+  topology.multiserver = defaults + ms
+  ms = topology.multiserver
+  servers = ms.get("servers", [])
+
+  provider = topology.get("provider", "") or topology.defaults.get("provider", "")
+  if provider and provider != "clab":
```
ipspace: You could check this in a later hook when topology.provider and topology.defaults.provider are both guaranteed to be set.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+      "multiserver",
+      more_hints=["libvirt and virtualbox support may be added in a future release"],
+    )
+    return
+
+  if not servers:
+    log.error('multiserver plugin requires a "servers" list', log.MissingValue, "multiserver")
+    return
+
+  if len(servers) < 2:
```
ipspace: Once we get things more automated, it would be nice to lift this restriction to have the ability to run netlab locally against a remote server.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+# ---------------------------------------------------------------------------
+#  Hook: post_transform — resolve server assignments, classify links
+# ---------------------------------------------------------------------------
+
+
+def post_transform(topology: Box) -> None:
+  ms = topology.get("multiserver", None)
+  if not ms:
+    return
+
```
ipspace, line 67: This is where you could validate "clab" provider

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+  server_map = {s.id: s for s in ms.servers}
```
ipspace: I knew using a dictionary of servers is not a bad idea ;))

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    else:
+      _auto_distribute(unassigned, server_map, assignment, topology)
+
+  log.exit_on_error()
+
+  vni_base = ms.vxlan.get("vni_base", 10000)
+  cross_count = _classify_links(topology, assignment, replicated, vni_base)
+  log.exit_on_error()
+
+  topology._multiserver = Box({
```
ipspace: IIRC, you can assign a dict to a topology.something key and it would automatically turn into a Box.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+  assignment = ms_data.assignment
+  server_map = ms_data.server_map
+  vxlan_cfg = ms.vxlan
+  out_tpl = ms.get("output_dir", "server-{server_id}")
+
+  replicated = set(ms_data.get("replicated", []))
+  server_folders = []
+
+  for server in ms.servers:
+    sid = server.id
```
ipspace, line 133: Obviously, if we'd go for server identifiers (short names) as dictionary keys, we'd use those instead

docs/plugins/multiserver.md
Muddyblack: What do you mean? 

Well I only tried it on 0.75 true but it should work before too. It only requires a netlab version which uses the pickel system for its storage.

Netlab already enforces a higher version than 0.46 i think no?

ipspace: > Netlab already enforces a higher version than 0.46 i think no?

That was the point. You don't have to mention containerlab version.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+```bash
+rsync -avz server-1/ user@192.168.168.128:~/lab/server-1/
+rsync -avz server-2/ user@192.168.168.129:~/lab/server-2/
+```
+
+**Step 3: Deploy on each server** by running the following on each remote host:
+
+```bash
+sudo netlab up --snapshot -vv
+sudo ./vxlan-setup.sh    # only if multi-access VXLAN tunnels are present
```
ipspace: We should use CLI hooks for that https://netlab.tools/dev/cli-hooks/

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,272 @@
...
+
+Running with `topology.yml` directly on remote servers will fail because:
+1. **Consistency**: Netlab dynamically allocates IP addresses, interface IDs, and VXLAN VNIs. Independent creation runs on different hosts would result in mismatched allocations.
+2. **Recursion**: Running `netlab create` on `topology.yml` on the remote hosts would execute the `multiserver` plugin again, causing it to split the topology recursively and generate nested server subdirectories.
+```
+
+**Teardown** in reverse order:
+
+```bash
+sudo ./vxlan-teardown.sh
```
ipspace: How about "netlab down --cleanup" using CLI hooks?

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    vx_count = len(vxlan_tunnels)
+    parts = [f"{len(local_nodes)} nodes", f"{link_count} links"]
+    if vx_count:
+      parts.append(f"{vx_count} VXLAN tunnels")
+    log.info(f"Server {sid}: {out_dir}/ — {', '.join(parts)}", module="multiserver")
+
+  # Register atexit handler to copy node_files, host_vars, etc. into each server
+  # folder after netlab writes all output files.
+  if server_folders:
+    import atexit
```
ipspace: I would much prefer using yet another callback. I'll add "post-output" (or some such) callback to "netlab create" processing.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    if not sf_path.is_dir():
+      continue
+
+    # node_files: per-node dirs + shared files (names starting with -)
+    if nf_dir.is_dir():
+      dst_nf = sf_path / "node_files"
+      dst_nf.mkdir(exist_ok=True)
+      for item in nf_dir.iterdir():
+        if item.name in local_nodes or item.name.startswith("-"):
+          _copy_if_missing(item, dst_nf / item.name)
```
ipspace: This is potentially dangerous -- if you don't clean up the directories, then you could end with stale files. I would remove the target directory and then just copy stuff over.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+
+    # host_vars: per-node only
+    if hv_dir.is_dir():
+      dst_hv = sf_path / "host_vars"
+      dst_hv.mkdir(exist_ok=True)
+      for item in hv_dir.iterdir():
+        if item.name in local_nodes:
+          _copy_if_missing(item, dst_hv / item.name)
+
+    # Copy all other subdirectories (e.g. group_vars, templates, monitoring)
```
ipspace: This could copy tons of things that we might not need on the remote end. Would it be better to specify a list of paths you want to copy over?

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+
+
+def _distribute_files_atexit(lab_folder: str, server_folders: list) -> None:
+  """Distribute generated files"""
+  lab_path = Path(lab_folder)
+  nf_dir = lab_path / "node_files"
+  hv_dir = lab_path / "host_vars"
+  server_names = {Path(sf).name for sf, _ in server_folders}
+
+  for sf, local_nodes in server_folders:
```
ipspace, line 224: The whole copying process could be rewritten into "copy these things (a list) based on node names" and "always copy these other things (a list)", where both lists could potentially be adjusted with "multiserver" topology parameters.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    for item in lab_path.iterdir():
+      if not item.is_dir() or item.name in skip:
+        continue
+      # Only copy grafana directory if this server hosts the grafana node
+      if item.name == "grafana" and "grafana" not in local_nodes:
+        continue
+      _copy_if_missing(item, sf_path / item.name)
+
+    # Ansible inventory and config
+    for fname in ("ansible.cfg", "hosts.yml"):
```
ipspace: Didn't look deep enough into the code, but it would make sense to prune the Ansible files, otherwise "netlab initial" or some such on the remote end would try to reach all Ansible hosts. I'm assuming you did your tests with FRR. Add SR Linux to the mix (because it needs Ansible for deployments)

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+
+# ===========================================================================
+#  Internal helpers — post_transform
+# ===========================================================================
+
+
+def _validate_servers(servers: list) -> None:
+  seen_ids: set = set()
+  for idx, s in enumerate(servers):
+    if "id" not in s:
```
ipspace, line 342: This is what the validation routines should do automatically. It's just the question of whether you can do your magic late enough (= post-validation) or you have to be really careful not to mangle the misformed data in an early hook.

Also, see the comments about servers as a dictionary.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+      log.error(f'Server entry #{idx + 1} missing required "id" field', log.MissingValue, "multiserver")
+      continue
+    if "host" not in s:
+      log.error(f'Server {s.id} missing required "host" field', log.MissingValue, "multiserver")
+      continue
+    if s.id in seen_ids:
+      log.error(f"Duplicate server id {s.id}", log.IncorrectValue, "multiserver")
+    seen_ids.add(s.id)
```
ipspace: As already mentioned: we have routines to deal with this, and if you go for servers-as-dict, the problem solves itself.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+
+
+def _resolve_replicated(ms: Box, topology: Box) -> set:
```
ipspace, line 346: A) This needs to be done late enough to have group "members" resolved
B) It would be a good idea to add "node_or_group" data type that would automatically do what you're doing here.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+      log.error(f'multiserver.replicate: "{entry}" is not a node or group', log.IncorrectValue, "multiserver")
+  return replicated
+
+
+def _resolve_assignments(servers: list, topology: Box) -> dict:
+  assignment: dict = {}
+  for server in servers:
+    for gname in server.get("groups", []):
+      grp = topology.get("groups", {}).get(gname, None)
+      if grp is None:
```
ipspace: I think we have "group" data type for this.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    )
+
+
+# ===========================================================================
+#  Internal helpers — clab.yml generation
+# ===========================================================================
+
+
+def _to_plain(obj: object) -> object:
+  """Convert Box/BoxList to plain dict/list for clean YAML serialization."""
```
ipspace: Use "box.to_dict()". Also, if you do box.to_yaml(), you get clean YAML anyway. Finally, we have get_yaml_string helper function in utils/strings

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+  if isinstance(obj, Box):
+    return {k: _to_plain(v) for k, v in obj.items()}
+  if isinstance(obj, list):
+    return [_to_plain(v) for v in obj]
+  return obj
+
+
+def _intf_clab_name(intf: Box) -> str:
+  """Containerlab interface name for a node interface."""
+  return intf.get("clab", {}).get("name", "") or intf.get("ifname", "")
```
ipspace: `intf.get("clab.name","") or intf.get("ifname","")`

See "Using Box objects" in AGENTS.md

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+
+
+def _build_clab_node(nname: str, ndata: Box, topology: Box) -> dict:
```
ipspace: It looks like this whole part could be easier to do with a topology filter (something similar to "remove unmanaged nodes") followed by an augmented clab.yml Jinja2 template. I also wouldn't have a problem adding VXLAN support (as clab attributes) into netlab core and then using those attributes here.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+    log.info(f"Replicated on all servers: {', '.join(sorted(replicated))}", module="multiserver")
+  if cross_count:
+    log.info(
+      f"{cross_count} cross-server links (VNI {vni_base}–{vni_base + cross_count - 1})", module="multiserver"
+    )
+
+
+# ===========================================================================
+#  Internal helpers — clab.yml generation
+# ===========================================================================
```
ipspace, line 484: This whole section should be completely restructured. We're reinventing the wheel (in Python)

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+  topo_copy = Box(topology, box_dots=True)
+
+  # Filter nodes to only those on this server
+  topo_copy.nodes = Box({n: v for n, v in topo_copy.nodes.items() if n in local_nodes}, box_dots=True)
+
+  # Filter links to only those with at least one local endpoint
+  topo_copy.links = [l for l in topo_copy.links if any(i.node in local_nodes for i in l.get("interfaces", []))]
+
+  # Expand paths (add f_files / f_tasks / f_dirs computed keys).
+  make_paths_absolute(topo_copy.defaults.paths)
```
ipspace: I'm afraid there are lots of assumptions here, starting with "the directory structure MUST be the same on all the servers and the control node"

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,752 @@
...
+  # Remove prefix generators and serialize
+  cleaned = cleanup_topology(topo_copy)
+  topodict = cleaned.to_dict()
+  topodict["_netlab_version"] = __version__
+
+  with open(Path(out_dir) / "netlab.snapshot.pickle", "wb") as f:
+    pickle.dump(topodict, f)
+
+
+def _write_vxlan_scripts(out_dir: str, tunnels: list, dev: str) -> None:
```
ipspace, line 641: Any reason why you wouldn't use a Jinja2 template for this?

ipspace (changes requested):
Super-awesome-job!!! Thanks a million.

Tons of comments (as you expected ;). Some of them are just suggestions or pointers to existing helper functions, in other cases I think we can make the whole thing a lot more streamlined with significant rewrites.

Muddyblack:
I guess about 99% of them are valid 😁
I will have a look at it next week 👍

docs/plugins/multiserver.md
Muddyblack: _validate.servers now uses _dataplane.create_id_set

hence required server id is removed from defaults.yml

docs/plugins/multiserver.md
Muddyblack: adding a load definition

docs/plugins/multiserver.md
Muddyblack: added `weight: { type: int, _min: 1 }`  to the server subtype schema

using `count / weight` as laod metric

Both group placement and individual node assignment now pick the server with the lowest weighted load

docs/plugins/multiserver.md
Muddyblack: Yeah currently i mostly  do it externally automated. So an extra custom script.

netsim/extra/multiserver/plugin.py
Muddyblack: moved the check to post_transform

netsim/extra/multiserver/plugin.py
Muddyblack: alright removed the Box()

netsim/extra/multiserver/plugin.py
Muddyblack: as already said in another comment it is moved to `post_transform`

netsim/extra/multiserver/plugin.py
Muddyblack: removed it, it should already work as one server only.

As the server map will have only one entry and the links of server will not be > 1 

But I think you mean that we do here directly also rsync and netlab up on / to remote machine

netsim/extra/multiserver/plugin.py
Muddyblack: User is now able to set via the topology yml file directories he wants  to get copied on top on the default ones

I kept the possibility  to also "replace" the default ones if wanted

- `extra_copy_dirs / extra_copy_files` ->  appends to defaults 
- `copy_dirs / copy_files`  -> replaces defaults entirely (override)
- `output()` -> concatenates both lists, so they compose cleanly

netsim/extra/multiserver/plugin.py
Muddyblack: see other comment above.

netsim/extra/multiserver/plugin.py
Muddyblack: replaced `_copy_if_missing` with plain `_copy`

using `shutil.rmtree(dst_nf)` first and then doing a fresh copy

netsim/extra/multiserver/plugin.py
Muddyblack: Added SR Linux to tests and `_write_filtered_inventory` now prunes `hosts.yml` to only contain nodes assigned to the specific server.

netsim/extra/multiserver/plugin.py
Muddyblack: `output_dir` template now uses {server_name} by default,

netsim/extra/multiserver/plugin.py
Muddyblack: replaced `_to_plain` with `_plain` which uses `obj.to_dict() ` and imported `get_yaml_string` for the clab.yml output

netsim/extra/multiserver/plugin.py
Muddyblack: added setup and tear-down jinja files for vxlan

netsim/extra/multiserver/plugin.py
Muddyblack: Removed `make_paths_absolute`  and `netlab up --snapshot` loads the `pickle` anyways and the paths are computed at runtime anyways

netsim/extra/multiserver/plugin.py
Muddyblack: I could not find something in the `types.py`. The only word "group" is in `transform_to_msec()`

netsim/extra/multiserver/plugin.py
Muddyblack: servers are now a `dict` keyed by name.
`seen_ids` has been got rid of.

netsim/extra/multiserver/plugin.py
Muddyblack: I modified the clab.j2 file with ` l.clab.vxlan`

netsim/extra/multiserver/plugin.py
Muddyblack: `_validate_servers` now takes  a `box-dict` and uses `_dataplane` for the ID-assignment.

docs/plugins/multiserver.md
Muddyblack: `post_start_clab` gets registered :

`_write_hook_script("vxlan-all-setup.sh", [f"{d}/vxlan-setup.sh" for d in setup_dirs])`
`topology.defaults.netlab.up.post_start_clab = "bash vxlan-all-setup.sh"`

is that the correct approach?

docs/plugins/multiserver.md
Muddyblack: `pre_stop_clab` gets registered :

` _write_hook_script("vxlan-all-teardown.sh", [f"{d}/vxlan-teardown.sh" for d in teardown_dirs])`
` topology.defaults.netlab.down.pre_stop_clab = "bash vxlan-all-teardown.sh"`

netsim/extra/multiserver/plugin.py
Muddyblack: added `must_be_group_id` to `types.py`

netsim/extra/multiserver/plugin.py
Muddyblack: added `must_be_node_or_group` to `types.py`

docs/plugins/multiserver.md
Muddyblack: Did not seem as a correct approach.

Now the hook is set on the per-server topo_copy with a relative vxlan-setup.sh path, before the snapshot is written, so it gets baked into each server's snapshot and fires locally on the remote host.

Muddyblack: since those a are now hooks I will add a new parameter  to the defaults allowing the user to not automatically create the vxlans on netlab up. Example if you connect such with real devices you may want to first test the virtual environment and once that is okay you might want to open the vxlan connection later.

Muddyblack:
# Refactor & review-response changelog

## clab.yml generation
- Dropped ~200 lines of hand-rolled Python clab generation (`_build_server_clab` & co).
- `output()` now filters the topology and renders via the standard `clab.j2` template.
- Cross-server P2P links carry a `link.clab.vxlan` attribute; `clab.j2` got one extra
  branch under `node_count == 2`. Regular topologies are unaffected.

## Data types
- Added `must_be_group_id` and `must_be_node_or_group` (mirroring `must_be_node_id`).
- Schema now validates `groups` (group_id), `members` (node_id), `replicate` (node_or_group).
- Removed the manual existence-check loops in `_resolve_assignments` /
  `_resolve_replicated`; they only expand references now.

## Servers as a dictionary
- `multiserver.servers` is a name-keyed dict — duplicate names impossible by construction.
- Dropped the `seen_ids` loop; `_validate_servers` uses `_dataplane` IDs (like VLANs/VRFs).

## CLI hooks for VXLAN setup/teardown
- Hooks (`post_start_clab` / `pre_stop_clab`) are baked into each server's snapshot with a
  relative `vxlan-setup.sh` path, so they fire locally where `netlab up --snapshot` runs.

## Single-server `netlab up --snapshot` fixes
- `_classify_links` marked a link cross-server if endpoints spanned >1 server,
  ignoring `clab.uplink` — uplink NIC + VXLAN on the same bridge, no STP →
  broadcast storm → 100 % CPU. Fix: `clab.uplink` links are now forced `cross: False`.
- `set -e` + bare `ip link add vxlanN` aborted on "File exists" after a crash.
  Fix: `ip link del vxlanN 2>/dev/null || true` before each `add` in
  `vxlan-setup.j2` / generated script.
- New `multiserver.vxlan.auto_start` bool (default `true`). When `false`, scripts
  are generated but CLI hooks are not registered — tunnels stay down until run
  manually (staged convergence or single-server local testing).

## Remote-server snapshots *(superseded by `_resolve_snapshot_paths` above)*
- `make_paths_absolute()` no longer touches per-server snapshots.

## Bug fixes
- Fixed two `Box` crashes in `_build_server_topo` (`default_box=True` missing → `BoxKeyError`).
- Box cleanup per `AGENTS.md`: `to_dict()` and dotted-path `.get()`.
- Module config render failed in per-server dirs (`'frr.j2' not found`, OSPF/etc.):
  snapshot load skips `make_paths_absolute()`, so a bare relative `templates` dir
  reached `render_template()`, whose "not `.`/`/`" branch re-based it onto the netlab
  install dir → in-template `{% import %}` lookups missed. Fix: `_resolve_snapshot_paths`
  now anchors relative entries with `./` via `_cwd_relative()` (stays portable across hosts).

## Tests & docs
- Merged srlinux/frr coverage into `multiserver-explicit.yml`; dropped the separate test.
- `docs/plugins/multiserver.md` updated for the dict format, CLI hooks, and clab.j2 flow.

## Open / outlook
- Waiting on a core `post-output` callback (offered by maintainer) to replace the `atexit`
  file-copy handler.
- Future: optional remote orchestration — control node pushes dirs and runs `netlab`
  remotely (SSH), so the whole lab comes up from one command. Out of scope here.




--- 

I am not sure on the package:	(netlab installation path) is it baked into the snapshot file?

This would require the user to have on all their machines the same path where netlab is installed onto.
Although I would say usually you do similar setups of your machines anyways (just cloning them) .

```
if "package:" in entry:
  resolved.append(str(moddir / entry.replace("package:", "")))
```

Also unhappy with `_cwd_relative` and `_resolve_snapshot_paths`

ipspace:
A few quick thoughts -- more details after I implement the hooks this needs and we rebase it.

> Added `must_be_group_id` and `must_be_node_or_group` (mirroring `must_be_node_id`).

Will cherry-pick this change into a new PR. Don't want to have this hidden in a large blob of unrelated code.

> * Hooks (`post_start_clab` / `pre_stop_clab`) are baked into each server's snapshot with a
>   relative `vxlan-setup.sh` path, so they fire locally where `netlab up --snapshot` runs.

👍

> * New `multiserver.vxlan.auto_start` bool (default `true`). When `false`, scripts
>   are generated but CLI hooks are not registered — tunnels stay down until run
>   manually (staged convergence or single-server local testing).

👍

> * `make_paths_absolute()` no longer touches per-server snapshots.

That's a huge can of worms. We need the absolute paths in the snapshot and in the Ansible inventory. I think it would be best to have a plugin hook executed very early in "netlab up" so it can adjust the topology data and recreate the snapshot and Ansible inventory before "netlab up" does some real work.

> I am not sure on the package: (netlab installation path) is it baked into the snapshot file?

It's resolved from the current installation path in make_paths_absolute

> This would require the user to have on all their machines the same path where netlab is installed onto. Although I would say usually you do similar setups of your machines anyways (just cloning them) .

Unless you install netlab as root on lab VMs and in virtual environment on your local machine.

> Also unhappy with `_cwd_relative` and `_resolve_snapshot_paths`

Agreed. That whole thing has to be solved in a different way.

netsim/templates/provider/clab/clab.j2
```diff
@@ -98,16 +98,34 @@ topology:
 {%     endfor %}
 {%   endfor %}
 {% elif l.node_count == 2 %}
+{%   if l.clab is defined and l.clab.vxlan is defined %}
```
ipspace: Due to the way we implement undefined Jinja2 values, you only need to test `l.clab.vxlan is defined` (https://github.com/ipspace/netlab/blob/c1ccc380d6d95c02e12e7c546151ba30170bcc81/AGENTS.md?plain=1#L102)

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+(multiserver-vxlan)=
+### VXLAN Parameters
+
+Global VXLAN settings are specified in the **multiserver.vxlan** dictionary:
+
+| Parameter | Type | Meaning |
+|-----------|------|---------|
+| **vni_base** | integer | Starting VNI for cross-server links (default: `10000`) |
+| **dstport** | integer | UDP destination port for VXLAN traffic (default: `4789`) |
+| **dev** | string | Default physical interface to bind VXLAN tunnels (default: `ens33`) |
```
ipspace: I think it would make more sense to make this `eth0` (that's what you get on less-opinionated distros ;) or leave it undefined but make it a required attribute so the user is forced to define it. `ens33` is oddly specific.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+
+```yaml
+# GOOD: per-site groups — one site per server
+groups:
+  site1:
+    members: [ site1-r1, site1-r2, site1-r3, site1-r4, site1-r5 ]
+  site2:
+    members: [ site2-r1, site2-r2, site2-r3, site2-r4, site2-r5 ]
+  sites:
+    members: [ site1-r1, site1-r2, site1-r3, site1-r4, site1-r5,
```
ipspace, line 151: You might want to point out in a comment that it's even better to use `members: [ site1, site2 ]`

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+In the second example the parent `sites` group can still be used for Ansible targeting or shared configuration — it does not affect placement because the child groups (`site1`, `site2`) claim their members first during assignment.
+
+```{note}
+Groups are processed in definition order. Child groups defined **before** a parent group will claim their members first, making the parent group a no-op for assignment. Always define fine-grained groups before aggregate groups in your topology.
+```
+
+(multiserver-replicate)=
+### Replicated Nodes
+
+Nodes listed in **multiserver.replicate** are instantiated on every server. This is useful for infrastructure services that need local access on each physical host — for example, monitoring collectors, route reflectors, or DNS resolvers.
```
ipspace: I wonder how well the inevitably overlapping IP addresses work. Also, I don't think route reflectors are a good example (I can easily see how that would result in split routing).

It might be best to move this section to the end of the document and use your specific example, including an explanation of how the overlapping IP addresses are resolved as you're effectively deploying an implicit anycast service.

On a second thought, maybe that's a better way to go -- require an explicit anycast service?

I'm fine with whatever you decide is best, and this is not a showstopper. It's just that I can see too many unexpected consequences, so there should be a large enough "THERE BE DRAGONS" sign attached to this concept ;)

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+
+This places spines on `spine-host` and leaves on `leaf-host`. All four links cross servers and are provisioned as containerlab native VXLAN endpoints.
+
+## Behind the Scenes
+
+When the plugin processes the topology, it classifies links into three categories:
+
+* **Local links** connecting nodes on the same server remain as regular containerlab veth pairs or bridges.
+* **Cross-server point-to-point links** are provisioned via containerlab's native VXLAN link endpoints (`type: vxlan` in `clab.yml`).
+* **Cross-server multi-access links** use a local Linux bridge on each server, interconnected via host-level VXLAN tunnels configured by generated setup scripts.
```
ipspace: Here's a crazy idea: what if you implemented cross-server multi-access links with Linux **bridge** nodes (https://netlab.tools/node/roles/#implementing-multi-access-links-with-bridge-nodes) -- when analyzing the topology, you could add necessary bridges to **nodes** and the **bridge** attribute to links, totally removing the need for extra provisioning scripts.

OTOH, while this would make the end result simpler, you would need a very careful orchestration of steps between this plugin and the **bridge** code (https://github.com/ipspace/netlab/blob/dev/netsim/roles/bridge.py). You'd have to create the bridge nodes before the `pre_transform` hook is executed in the `bridge` role. However, looking at the code, it seems that the roles are the last plugins in the list, so just doing this in the `pre_transform` hook is probably good enough if you're OK with initializing all the required node data. However, we should not do this too early, or you'd have to crawl through VLAN and VRF links (plus there are topology components and other stuff).

Worst case, I could add another plugin hook ;)

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+* A filtered `netlab.snapshot.pickle` for use with `netlab up --snapshot`
+* A filtered `hosts.yml` containing only the nodes assigned to that server, so `netlab initial` does not attempt to configure nodes on other servers
+* Copies of `node_files/` and `host_vars/` for only the nodes on that server
+* Copies of the directories and files listed in **multiserver.copy_dirs** and **multiserver.copy_files**
+* Per-server `vxlan-setup.sh` and `vxlan-teardown.sh` scripts (when multi-access VXLAN tunnels are needed), registered in that server's snapshot as [CLI hooks](dev-cli-hooks) (`netlab.up.post_start_clab` / `netlab.down.pre_stop_clab`) so `netlab up` and `netlab down` run them automatically on the remote host
+
+(multiserver-deployment)=
+## Deployment Workflow
+
+```{note}
```
ipspace, line 237: A minor detail: You might want to use the "pre_probe" netlab up hook on the master node to abort the "netlab up" process. Set a flag in the topology defaults (or some such) after writing the worker pickles and check it in the pre_probe hook.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+```bash
+sudo netlab up --snapshot -vv
+```
+
+When multi-access VXLAN tunnels are present, `netlab up` runs `vxlan-setup.sh` automatically via a [CLI hook](dev-cli-hooks) registered by the plugin.
+
+```{important}
+**Why is `--snapshot` required on remote servers?**
+
+You must run `sudo netlab up --snapshot` on remote servers to load the topology from the pre-generated snapshot (`netlab.snapshot.pickle`) instead of the original `topology.yml`.
```
ipspace: Usually we don't need "sudo" with "netlab up", unless you need it to set up VXLAN tunnels, in which case you could add **sudo** commands to that script. However, then you'd need to check whether sudo is installed (see the installation scripts).

The "use bridge nodes and let containerlab deal with it" idea sounds more and more promising ;))

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+
+Running with `topology.yml` directly on remote servers will fail because:
+1. **Consistency**: Netlab dynamically allocates IP addresses, interface IDs, and VXLAN VNIs. Independent creation runs on different hosts would result in mismatched allocations.
+2. **Recursion**: Running `netlab create` on `topology.yml` on the remote hosts would execute the `multiserver` plugin again, causing it to split the topology recursively and generate nested server subdirectories.
+```
+
+**Teardown** on each server:
+
+```bash
+sudo netlab down
```
ipspace: Again, "sudo" is usually not needed.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+multiserver:
+  copy_dirs: [ group_vars, templates, monitoring ]
+  copy_files: [ ansible.cfg, netlab.lock ]
+```
+
+The Ansible inventory (`hosts.yml`) is always written into each server directory and is automatically filtered to contain only the nodes assigned to that server.
+
+## Limitations
+
+* Only the **containerlab** provider is supported. Libvirt and virtualbox topologies cannot be split across servers.
```
ipspace: BTW, I'm perfectly fine with this. Totally out-of-scope, but I think we should focus primarily on containerlab and pester vrnetlab maintainers where needed to make the containers start faster.

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+* Cross-server VXLAN tunnels use a flat VNI space starting at **vni_base**. The maximum VNI value is 16777215 (24-bit). Topologies with more than ~16 million cross-server links will fail validation.
```
ipspace: Wow. Made my day 😂😂

docs/plugins/multiserver.md
```diff
@@ -0,0 +1,317 @@
...
+* All physical servers must have direct IP reachability — the plugin does not support NAT traversal or relay hosts between servers.
```
ipspace: We need a better term for the workers, as they might be VMs.

netsim/data/types.py
```diff
@@ -996,6 +996,44 @@ def must_be_node_id(value: typing.Any) -> dict:
...
+@type_test()
+def must_be_node_or_group(value: typing.Any) -> dict:
+  if not isinstance(value,str):                                       # Otherwise it must be a string
+    return { '_type': 'valid node or group name (a string)' }
+
+  topology = global_vars.get_topology()               # Try to get current lab topology
+  if topology is None:                                # pragma: no-cover
+    log.fatal('Calling node_or_group validation before the topology has been initialized')
+
+  if value not in topology.nodes and value not in topology.get('groups',{}):
```
ipspace: Personal preference: when fetching a value multiple times, I would fetch it into a local variable and then use that variable to enforce consistency.

netsim/extra/multiserver/defaults.yml
```diff
@@ -0,0 +1,68 @@
...
+            type: list
+            _subtype: node_id
+          vxlan_dev: str
+          weight:
+            type: int
+            _min: 1
+      vxlan:
+        vni_base: int
+        dstport: int
+        dev: str
```
ipspace: This should be a required attribute. At the moment, it's set from the system defaults anyway, so no harm done.

netsim/extra/multiserver/defaults.yml
```diff
@@ -0,0 +1,68 @@
...
+      copy_dirs: list
+      copy_files: list
+      extra_copy_dirs: list
+      extra_copy_files: list
+
+multiserver:
+  vxlan:
+    vni_base: 10000
+    dstport: 4789
+    dev: ens33
```
ipspace: I would change this to `eth0`, or we could omit this default which would (together with `_required` on the topology attribute) force the user to specify it either in the topology or in the defaults.

netsim/extra/multiserver/defaults.yml
```diff
@@ -0,0 +1,68 @@
...
+    # When true, 'netlab up --snapshot' auto-runs vxlan-setup.sh via a CLI hook.
+    # Set false to keep cross-server tunnels inactive until you run the script
+    # manually (e.g. to stage convergence or connect servers on your own schedule).
+    auto_start: true
```
ipspace, line 64: Another minor detail that would be unnecessary with the "bridge" nodes ;))

netsim/extra/multiserver/plugin.py
ipspace: The "post-output" callback is there. Please use that.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,622 @@
...
+  _dataplane.create_id_set("multiserver_server")
+  _dataplane.extend_id_set(
+    "multiserver_server", _dataplane.build_id_set(Box({"servers": servers}), "servers", "id", "multiserver")
+  )
+  _dataplane.set_id_counter("multiserver_server", 1, max_value=65535)
+
+  for name, s in servers.items():
+    if "id" not in s:
+      s.id = _dataplane.get_next_id("multiserver_server")
+    if "host" not in s:
```
ipspace: If you set the "host" attribute to be required, it will eventually be checked, and if you don't use the "host" value very early, we should be OK.

ipspace (changes requested):
A mixed bag of nits, things that should be fixed (more info on replicated nodes, default server uplink interface name...), and Another Grand Idea (😜). I'm perfectly fine if you tell me to defer the Grand Idea to a later time and just merge this thing.

ipspace:
@Muddyblack -- I'm sorry for the long delay. I wanted to push the June release out before Autocon5 (and it turns out one has a limited daily amount of mental energy as one gets older), and then Autocon5 struck (just joking, it was a great conference).

Anyway, I think we should implement multi-access links with bridge nodes in the long run to make your life easier. I could even make that functionality part of netlab core (in the "bridge" role) to have it available regardless of whether the user uses this plugin. Hey, maybe I could even do that by default just to simplify containerlab provisioning (no need for "manual" creation of Linux bridges).

Whether we do that as part of this PR, or merge this and then work on that, is completely up to you. Just let me know ;)

ipspace:
**Update:** the "implement clab multi-access networks with bridge nodes" idea is much harder to implement than I thought if we want to support VLAN trunks or routed VLANs across multi-access networks -- we would have to split the multi-access link into P2P links very late in the transformation process.

For the moment, let's fix the other stuff and move forward with this. When I implement that other idea, the multi-access part of the code in this PR will become irrelevant, at which point we can do a cleanup.

docs/plugins/multiserver.md
jbemmel: This is starting to look more and more like a core feature, not a plugin

Perhaps we should add the notion of 'servers' to the topology, and add vxlan support for containerlab links?

Muddyblack:
No  Problem some time off the computer and charge energy is also good for me ;) 
Autocon would had been nice especially as it was in Munich this year I forgot about it after some time 😅
Maybe I find online material about it tho.

docs/plugins/multiserver.md
Muddyblack: yeah right 😂 Should add a smiley or so probably 😂

docs/plugins/multiserver.md
Muddyblack: you are right it is just a bad habit I have. Totally forgot there are machines that do not have sudo.

netsim/extra/multiserver/plugin.py
Muddyblack: The `schema's _required: True` should catch during attribute validation before the plugin code even runs?
So I just removed the if host check if thats what you meant ?

docs/plugins/multiserver.md
Muddyblack: Hm I dont really have a good alternate name for this to be honest not happy with any of them.

Chose now lab-host

ipspace: How about controller and workers?

Muddyblack: but if you have  better ideas do you also mean inside the  yml scheme or only inside the documentation and comments ?

docs/plugins/multiserver.md
Muddyblack: Hm I have this  for example for grafana and other telemetry which either use the management network or  fetch directly from  the docker. Like [frr-exporter](https://github.com/tynany/frr_exporter).
So nothing interfering in the real network only the management network thats why i originally added it. 
As having to set it up with vxlan etc is hard so decided i just inject monitoring in any server (lab-host).

Did not want it doing anything in the real network .

So basically nothing thats part of the network just outside fetching stuff? 

But I see your point I will move it to the end.

docs/plugins/multiserver.md
Muddyblack: My github did not update the page apparently but the question for the scheme part is still a valid question I have @ipspace

ipspace: Just the documentation to make things more explicit. No worries, I can do the polishing just before merging. I usually go through the documentation anyway (one of the "few" OCD kinks of mine)

Muddyblack:
Just a note:
I committed for example only one doc md file but the entire checks did run though.

Is that intended? 
`Check integration tests` or ruff linting  does not have to run if no python file was touched.
Nothing big just a remark as it might be a waste of water xD

Muddyblack:
> Just a note: I committed for example only one doc md file but the entire checks did run though.
> 
> Is that intended? `Check integration tests` or ruff linting does not have to run if no python file was touched. Nothing big just a remark as it might be a waste of water xD

Hm I see it uses the entire pr not just latest commits or so and would  require a own job or so.

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,659 @@
...
+
+
+def pre_shell_pre_probe(topology: Box) -> None:
+  """pre_probe hook fired by 'netlab up' before it probes providers.
+
+  On the controller the flag is set (output() ran during create), so we abort
+  before starting the unsplit topology. On a worker it's absent, and we
+  resolve search paths and refresh the snapshot/inventory as before.
+  """
+  abort = topology.defaults.get("multiserver", {}).get("_control_node_abort", None)
```
ipspace, line 596: Just FYI: You could do `topology.defaults.get('multiserver._control_node_abort',None)`

netsim/extra/multiserver/plugin.py
```diff
@@ -0,0 +1,659 @@
...
+    log.fatal(
+      "multiserver: per-server configurations have been generated; "
+      "'netlab up' will not start the merged topology on the controller.\n"
+      f"Generated server folders:\n{folders}\n"
+      "Copy each folder to its worker and run 'netlab up --snapshot' there "
+      "(see docs/plugins/multiserver.md, Deployment Workflow).",
+      "multiserver",
+    )
+
+  if "paths" in topology.get("defaults", {}):
```
ipspace, line 608: Just FYI: `'defaults.paths' in topology` also works

ipspace (approved):
I haven't created the infrastructure to test the plugin (yet), but I trust you thoroughly tested it. Merging...

Thanks a million. Great job!

ipspace:
> Hm I see it uses the entire pr not just latest commits or so and would require a own job or so.

Yeah, I had a problem running per-commit (or per-push) tests for PRs coming from forked repos. A working solution would be most welcome ;)

silopolis:
Getting wild! :sunglasses: :heart_eyes:

PR #3427, opened 2026-05-27
Validate clab.type value and move platform checks to quirks

### LINKED ISSUE #3426: SRLinux: Fix Containerlab type names
https://github.com/ipspace/netlab/blob/a6ef0f7dad44289b51f4b133f800af504d9d52f5/netsim/ansible/templates/initial/srlinux.j2#L44

Based on https://containerlab.dev/manual/kinds/srl/#types the types are now called ```ixr-6e``` etc. - with a dash

jbemmel:
Fixes #3426

## Summary

- Normalize `clab.type` to Containerlab-documented Nokia SR Linux hardware variants per their schema; implemented as quirk with deprecation warning for old style -less names.
- Move MTU limit and MPLS/SR-MPLS platform validation from Jinja templates into `netsim/devices/srlinux.py` device quirks, using hyphenated type names (`ixr-6e`, `ixr-d2`, etc.).
- Update MPLS platform prefix list in `srlinux.yml` and adjust tests to use valid hyphenated types.

## Test plan

- [x] Topology validation tests for clab attributes
- [ ] SR Linux integration tests with `clab.type: ixr-6e` (MPLS/SR labs)
- [x] Confirm invalid `clab.type` values are rejected at topology validation time

### DIFF
--- a/netsim/devices/srlinux.py
+++ b/netsim/devices/srlinux.py
@@ -15,21 +15,22 @@ from . import _Quirks, need_ansible_collection, report_quirk
 
 
 def check_prefix_deny(node: Box) -> None:
+  """Report a quirk when prefix filters use a deny action (unsupported on SR Linux)."""
   for pf_name,pf_list in node.get('routing.prefix',{}).items():
     for p_entry in pf_list:
       if p_entry.get('action',None) == 'deny':
         report_quirk(
-          text=f'SR Linux does not support "deny" action in prefix filters (node {node.name} prefix filter {pf_name})',
+          text=(
+            f'SR Linux does not support "deny" action in prefix filters '
+            f'(node {node.name} prefix filter {pf_name})'
+          ),
           node=node,
           quirk='prefix_deny',
           category=log.IncorrectValue)
         break
 
-'''
-Remove unused IPv4/IPv6 transport addresses -- SR Linux does not like a
-BGP neighbor with no active address families
-'''
 def cleanup_neighbor_transport(node: Box, topology: Box) -> None:
+  """Remove unused BGP transport addresses when no address family is activated."""
   for ngb in _routing.neighbors(node,vrf=True):
     if 'local_if' in ngb:               # True unnumbered, move on
       continue
@@ -59,10 +60,8 @@ def cleanup_neighbor_transport(node: Box, topology: Box) -> None:
         category=Warning)
       ngb.pop(af)
 
-"""
-Determines the SRL version based on the container image
-"""
 def set_api_version(node: Box) -> None:
+  """Set node._srl_version from the container image tag."""
   version = re.search(r'^.*/srlinux:([\d]+.[\d]+).*$', node.box)
   node._srl_version = [ 25, 3 ]         # Assume 25.3 release
   if version is not None:               # If we managed to match the SR Linux image name
@@ -72,6 +71,7 @@ def set_api_version(node: Box) -> None:
       pass                              # ... no worries, we'll use the default
 
 def check_nssa_default_cost(node: Box) -> None:
+  """Warn when OSPF NSSA areas specify a default-metric (not supported on SR Linux)."""
   for (odata,_,_) in _routing.rp_data(node,'ospf'):
     if 'areas' not in odata:
       continue
@@ -87,12 +87,8 @@ def check_nssa_default_cost(node: Box) -> None:
           category=Warning,
           quirk='ospf_nssa_default_cost')
 
-"""
-Normalize interface descriptions for SR Linux:
-- Replace '->' with '~'
-- Remove square brackets '[' and ']'
-"""
 def normalize_interface_descriptions(node: Box) -> None:
+  """Normalize interface descriptions (-> to ~, remove square brackets)."""
 
   # Normalize regular interface descriptions
   for intf in node.get('interfaces',[]):
@@ -100,6 +96,7 @@ def normalize_interface_descriptions(node: Box) -> None:
       intf.name = intf.name.replace('->','~').replace('[','').replace(']','')
 
 def license_needed(dt: str, features: Box) -> bool:
+  """Return True when the emulated platform requires an SR Linux license for MPLS/SR."""
   p_list = features.get('mpls._platforms',[])
   for p_prefix in p_list:
     if dt.startswith(p_prefix):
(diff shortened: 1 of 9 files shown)


netsim/devices/srlinux.py
```diff
@@ -143,25 +174,28 @@ def device_quirks(self, node: Box, topology: Box) -> None:
...
+                f'SR Linux on ({node.name}) before version 25.3.1 does not support '
+                'filtering out extended communities for BGP.'
+              ),
               more_data= [ f'{c}:{vals}' ],
               node=node,
               category=Warning,
               quirk='bgp_community')
 
-    if 'mpls' in mods or 'sr' in mods:
-      if not license_needed(dt,features):
```
jbemmel, line 153: We could also keep this original logic, perhaps rename ```license_needed``` to ```platform_supports_mpls```

netsim/devices/srlinux.py
```diff
@@ -15,21 +15,22 @@
 
 
 def check_prefix_deny(node: Box) -> None:
+  """Report a quirk when prefix filters use a deny action (unsupported on SR Linux)."""
   for pf_name,pf_list in node.get('routing.prefix',{}).items():
     for p_entry in pf_list:
       if p_entry.get('action',None) == 'deny':
         report_quirk(
-          text=f'SR Linux does not support "deny" action in prefix filters (node {node.name} prefix filter {pf_name})',
+          text=(
```
ipspace, line 25: No need to let AI loose on stuff that works ;) Also, the syntax (while perfectly valid) is pretty unusual for our codebase. Just a thought, though, no need to change this.

netsim/devices/srlinux.py
ipspace: I'm OK with the change in the logic. Renaming "license_needed" to "platform_supports_mpls" would be equally (potentially) confusing ;)

netsim/devices/srlinux.py
```diff
@@ -143,25 +174,28 @@ def device_quirks(self, node: Box, topology: Box) -> None:
...
-      if not license_needed(dt,features):
-        report_quirk(
-          text=f'MPLS works only on (emulated) 7250-IXR and 7730-SXR routers (node {node.name})',
-          more_hints=['Set node clab.type to a different device model (see https://containerlab.dev/manual/kinds/srl/)'],
-          node=node,
-          category=log.IncorrectValue)
+    if ('mpls' in mods or 'sr' in mods) and not is_licensed:
+      report_quirk(
+         text=f'SR-MPLS only supported on 7250 IXR and 7730 SXR routers (node {node.name})',
+         more_hints=[f'Use e.g. clab.type "ixr-6e" instead of {dt}'],
```
ipspace: I think the original text is better as it points to a resource with more detailed information.

netsim/devices/srlinux.py
```diff
@@ -143,25 +174,28 @@ def device_quirks(self, node: Box, topology: Box) -> None:
...
+         node=node,
+         quirk='sr_mpls_platform',
```
ipspace: I don't think this quirk should be silenceable

tests/topology/expected/clab-attributes.yml
```diff
@@ -35,7 +35,7 @@ nodes:
       runtime: docker
       srl-agents: []
       startup-config: start.cfg
-      type: unknown
```
ipspace, line 38: Leave this intact. Proves my point about NOT hard-coding SR Linux values into clab attributes

netsim/providers/clab.yml
```diff
@@ -34,7 +34,34 @@ attributes:
       binds:
       kind: str
       config_templates: dict
-      type: str
+      type:
+        type: str
+        valid_values:
```
ipspace: I won't start digging whether any other device uses clab.type, but hard-coding SR Linux values into provider attribute is way beyond "wrong". If we want to check the model type, we MUST do it in SR Linux quirks.

Also, as containerlab is still accepting the "old" values, we shouldn't hard-deprecate them. It would be much better to have some logic in SR Linux quirk that would transform old type into new with a warning.

tests/topology/input/clab-attributes.yml
```diff
@@ -11,7 +11,7 @@ nodes:
       binds: []
       kind: frr
       config_templates: {}
-      type: unknown
+      type: ixr-d2
```
ipspace: Leave this intact (see the other comment)

ipspace (changes requested):
Hard-coding the SR Linux model types into clab.type values is a really bad idea, all the rest are minor fixes.

netsim/devices/srlinux.py
jbemmel: The original line causes a linter warning about the line being too long, caused by a missing linter config in .pylintrc:
```yaml
# Match ruff.toml line-length
max-line-length=120
```

netsim/devices/srlinux.py
jbemmel: Reverted

netsim/devices/srlinux.py
jbemmel: agreed, hard error

netsim/providers/clab.yml
jbemmel: It is not specific to SR Linux, see https://github.com/srl-labs/containerlab/blob/05cac296b6a669cd56aab6b1950c1c54f9134742/schemas/clab.schema.json#L511

I've reverted the data model and moved the normalization logic into quirks, as suggested

tests/topology/expected/clab-attributes.yml
jbemmel: reverted

tests/topology/input/clab-attributes.yml
jbemmel: done

netsim/devices/srlinux.py
```diff
@@ -13,23 +13,81 @@
 from ..utils import routing as _routing
 from . import _Quirks, need_ansible_collection, report_quirk
 
+# containerlab nokia_srlinux type enum (schemas/clab.schema.json)
```
ipspace: This was AI, right? What happened to "insert a dash after the third character if there's no dash in the string"?

netsim/devices/srlinux.py
```diff
@@ -87,40 +144,100 @@ def check_nssa_default_cost(node: Box) -> None:
...
 
+def normalize_clab_type(node: Box, topology: Box) -> str:
+  """Validate clab.type and map legacy undashed names to hyphenated containerlab values."""
+  dt = node.get('clab.type', CLAB_DEFAULT_TYPE)
+  if a_devices.get_provider(node, topology.defaults) != 'clab':
+    return dt
+
+  canonical = _CLAB_TYPE_CANONICAL.get(dt)
+  if canonical:
+    if canonical != dt:
```
ipspace: The quirk is OK, but I would do it like "if there's no hyphen, insert it and then check it against the list of valid types"

netsim/devices/srlinux.py
```diff
@@ -13,23 +13,81 @@
...
+  'sxr-1d-32d': 'sxr-1d-32d',
+  'sxr-1-32d': 'sxr-1-32d',
+  'ixrx1b': 'ixr-x1b',
+  'ixr-x1b': 'ixr-x1b',
+  'ixrx3b': 'ixr-x3b',
+  'ixr-x3b': 'ixr-x3b',
+  'ixr-x4': 'ixr-x4',
+  'ixr-x4-d': 'ixr-x4-d',
+}
+VALID_CLAB_TYPES = tuple(sorted(set(_CLAB_TYPE_CANONICAL.values())))
```
ipspace: Also, I would have the valid types defined in srlinux.yml (it's easier to write a YAML list of strings than a Python one ,)

ipspace (changes requested):
I think the "valid types" should be moved into a list and the "old-to-new" conversion should be done algorithmically, not via a lookup table.

ipspace:
@jbemmel -- I really appreciate that you're fixing stuff, but please do me a favor and be a sanity buffer between AI and myself ;))

ipspace:
Also, I could make these changes (they are not hard or extensive), just let me know.

netsim/devices/srlinux.py
jbemmel: It failed for cases like ```ixrh432d -> ixr-h4-32d```

I can refactor into a "insert 1 or 2 dashes" based on string length, and move the list of valid types to YAML so we can validate the result against that. If it fails, we give up

ipspace: > I can refactor into a "insert 1 or 2 dashes" based on string length

Perfect.

> and move the list of valid types to YAML so we can validate the result against that. If it fails, we give up

Sounds like a perfect plan ;) Anyway, having a list or a dict in YAML makes it easier for users to add new types if needed... in which case, dict is better, because it's easier to add elements to it, but I would use None (empty) value for "no translation needed"

netsim/devices/srlinux.py
jbemmel: I've changed it to "if there's no hyphen, iterate over the known valid values and check if it matches any of those with '-' removed"

netsim/devices/srlinux.py
jbemmel: done

ipspace (approved):
Looks good to me. Thank you. Running a few integration tests.

jbemmel:
> Looks good to me. Thank you. Running a few integration tests.

We can generalize to SROS and other platforms - will create a new PR moving the core logic to clab.py

PR #3428, opened 2026-05-28
Add bgp.local_role plugin for RFC 9234 BGP Roles on FRR and BIRD

jbemmel:
Implements EBGP local role and strict-mode attributes with integration tests, topology validation. Initially supported on  FRR and Bird.


ipspace:
@jbemmel -- I love that you did this. It has been on my to-do list forever, but since only FRR was supporting it, it never got to the front of it.

However, would you agree it makes sense to include this in something like **bgp.session**, because it's really a session attribute, or maybe **bgp.policy** because it filters stuff? Also I would call the attribute **bgp.role**, not **bgp.local_role**

jbemmel:
> @jbemmel -- I love that you did this. It has been on my to-do list forever, but since only FRR was supporting it, it never got to the front of it.
> 
> However, would you agree it makes sense to include this in something like **bgp.session**, because it's really a session attribute, or maybe **bgp.policy** because it filters stuff? Also I would call the attribute **bgp.role**, not **bgp.local_role**

Sure, I merely followed - what I believed to be - your preference for an independent plugin. But it's a simple flag, we can call it what you like and merge it with whatever component you prefer

ipspace:
> Sure, I merely followed - what I believed to be - your preference for an independent plugin.

Thank you!

> But it's a simple flag

And that's exactly why I thought it would fit better into one of the existing plugins that already deal with this stuff.

> we can call it what you like

Let's go with **bgp.role**

> and merge it with whatever component you prefer

Whichever one you prefer (or is easier to fit into), with a tiny preference for **bgp.policy** (because it is a sort of route filtering policy)

jbemmel:
Replaced by #3429

PR #3429, opened 2026-05-29
Add bgp.role support to bgp.policy plugin

jbemmel:
Consolidate BGP role handling into bgp.policy as bgp.role (with optional bgp.role.strict) instead of a separate plugin, with FRR/BIRD support, tests, and documentation.

Replaces #3428

New attribute structure uses ```_value_to_dict``` enhancement to normalize and validate parameters


docs/plugins/bgp.policy.md
```diff
@@ -79,6 +87,47 @@ See [BGP Policies Test Results](https://release.netlab.tools/_html/coverage.bgp.
 **Notes:**
 
 * Arista EOS and Aruba CX do not support node-level default local preference. Node-level **bgp.locpref** attribute (if specified) is thus applied to all interfaces that do not have an explicit **bgp.locpref** attribute. That might interfere with the **bgp.policy** interface attributes.
+* FRR implements BGP Roles starting with release 8.4. BIRD implements them starting with release 2.0.11. On BIRD, BGP roles are rendered into the BGP module configuration file (`daemons/bird/bgp.j2`).
+
+(plugin-bgp-policy-role)=
```
ipspace: I would move this section to the end (just before the "Sample Topologies")

netsim/extra/bgp.policy/plugin.py
```diff
@@ -161,6 +164,44 @@ def apply_config(node: Box, ngb: Box) -> None:
   api.node_config(node,_config_name)                    # Remember that we have to do extra configuration
   _bgp.clear_bgp_session(node,ngb)
 
+def _use_role_plugin_template(node: Box) -> bool:
```
ipspace: This feels so wrong. Why would you ever hard-code a device into a plugin code?

netsim/extra/bgp.policy/plugin.py
```diff
@@ -161,6 +164,44 @@ def apply_config(node: Box, ngb: Box) -> None:
...
+  """Return False for the BIRD daemon (roles are rendered in daemons/bird/bgp.j2)."""
+  return not (node.get('_daemon') and node.device == 'bird')
+
+def apply_role_attributes(node: Box, ngb: Box, intf: Box, topology: Box) -> bool:
```
ipspace: This feels like duplicating a lot of functionality that should be already available elsewhere. If you want to ensure "role_strict" is only used when "role" is there, we could solve that in the generic validation code (if it's not there yet).

netsim/extra/bgp.policy/plugin.py
```diff
@@ -313,7 +354,7 @@ def post_transform(topology: Box) -> None:
       continue
 
     route_aggregation(ndata,topology)
-    _bgp.cleanup_neighbor_attributes(ndata,topology,_attr_list + [ 'policy' ])
+    _bgp.cleanup_neighbor_attributes(ndata,topology,_attr_list + _ROLE_ATTR_LIST + [ 'policy' ])
```
ipspace: It would be better to add role keywords into "_direct" or "_compound" attribute list? Is there a reason not to do it?

netsim/extra/bgp.policy/plugin.py
```diff
@@ -338,6 +378,7 @@ def post_transform(topology: Box) -> None:
             communities.append('extended')
       if copy_locpref and not intf.get('bgp.locpref',False):
         intf.bgp.locpref = ndata.bgp.locpref
+      apply_role_attributes(ndata,ngb,intf,topology)
```
ipspace: It looks like the only reason you're dealing with roles as an exception is because of Bird template stuff. The easiest way to solve that is to have an empty template in the plugin directory.

ipspace (changes requested):
The code needs to be cleaned up. Also, we're not running three integration tests for a simple feature. Squeeze whatever you want to have tested into one.

ipspace:
@jbemmel -- I understand you find it exciting to use AI to generate code, but I won't review the bloated results in the future (like the special "dealing with role" function). Please act as a sanity check on the code before submitting the PR.

docs/plugins/bgp.policy.md
jbemmel: Assuming you meant starting from line 92 - done

netsim/extra/bgp.policy/plugin.py
```diff
@@ -11,7 +11,8 @@
 
 _config_name = 'bgp.policy'
 
-_requires    = [ 'bgp' ]
+_requires        = [ 'bgp' ]
+_execute_after    = [ 'bgp.session' ]
```
jbemmel, line 16: AI added this following documented comments - let me know if you prefer to leave it out

netsim/extra/bgp.policy/plugin.py
jbemmel: Replaced by the dummy template file

netsim/extra/bgp.policy/plugin.py
jbemmel: Added to _direct

netsim/extra/bgp.policy/plugin.py
jbemmel: put the empty template

netsim/extra/bgp.policy/plugin.py
jbemmel: I've refactored the PR to propose the following structure:
```
bgp.role: customer
```
or
```
bgp.role:
  name: customer
  strict: true
```

The ```must_be_bgp_role``` etc validation functions could be replaced by more generic validation code (some sub cases of user defined type validation aren't implemented, see #3430

jbemmel: Now replaced

netsim/extra/bgp.policy/defaults.yml
```diff
@@ -6,29 +6,43 @@
...
-
+    global:
+      role:
+        _description: |
+          RFC 9234 strict mode for EBGP sessions. Use **defaults.bgp.role.strict** (or **bgp.role.strict**
+          in lab topology settings) to enable strict mode globally. Per-neighbor roles are set with
+          **bgp.role** on links or interfaces.
+        strict:
+          type: bool
+          _required: True   # BUG: not enforced — defaults.bgp.role is never validated; empty {} passes validate_item
```
ipspace: I think this is no longer true.

netsim/extra/bgp.policy/plugin.py
ipspace: That's OK. Thank you.

netsim/extra/bgp.policy/plugin.py
```diff
@@ -338,6 +339,10 @@ def post_transform(topology: Box) -> None:
             communities.append('extended')
       if copy_locpref and not intf.get('bgp.locpref',False):
         intf.bgp.locpref = ndata.bgp.locpref
+      role = get_effective_module_attribute(
+        'bgp.role',intf=intf,node=ndata,topology=topology,defaults=topology.defaults)
+      if role:
+        intf.bgp.role = role
```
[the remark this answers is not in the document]
ipspace: Fixed.

netsim/extra/bgp.policy/defaults.yml
```diff
@@ -6,29 +6,43 @@
...
+attributes:                   # User-defined data types (see dev/validation.md#validate-user-types)
+  bgp_role:
+    name:
+      type: str
+      valid_values: [ provider, customer, peer, rs-server, rs-client ]
+      _required: True
+    strict: bool
+    _value_to_dict:
+      name: '{value}'
+      strict: False
```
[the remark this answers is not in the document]
ipspace: Fixed.

ipspace (approved):
Cleaned up and simplified where needed.

BTW, would love to know how you managed to check the Bird configuration. Without the "bgp@policy" mapping I had to add, "netlab initial" consistently failed.

PR #3434, opened 2026-06-03
Implement _value_to_dict validation attribute

ipspace:
The _value_to_dict validation attribute transforms a non-dict value into a dictionary using the value of the _value_to_dict key as the template dictionary. You can use it whenever you have a dict attribute that can accept non-dict values.

Sample use case included in the PR: BGP communities

### DIFF
--- a/docs/dev/validation.md
+++ b/docs/dev/validation.md
@@ -501,7 +501,7 @@ ospf:
 Some _netlab_ attributes that are supposed to be dictionaries can accept non-dictionary values. You can use the **_value_to_dict** type definition key to specify a template dictionary that is used to create a replacement dictionary. The values of the replacement dictionary can be f-strings with `{value}` representing the original value.
 
 ```{warning}
-The `{value}`string could be interpreted as a YAML dictionary and MUST therefore be quoted.
+The `{value}` string could be interpreted as a YAML dictionary and MUST therefore be quoted.
 ```
 
 For example, to transform a BGP role specified as a string into a dictionary with the role name and `strict` flag, use this type definition:
--- a/docs/dev/validation.md
+++ b/docs/dev/validation.md
@@ -495,6 +495,43 @@ ospf:
         ipv6: bool
 ```
 
+(validation-value-to-dict)=
+### Dictionary Specified As a Single Value
+
+Some _netlab_ attributes that are supposed to be dictionaries can accept non-dictionary values. You can use the **_value_to_dict** type definition key to specify a template dictionary that is used to create a replacement dictionary. The values of the replacement dictionary can be f-strings with `{value}` representing the original value.
+
+```{warning}
+The `{value}`string could be interpreted as a YAML dictionary and MUST therefore be quoted.
+```
+
+For example, to transform a BGP role specified as a string into a dictionary with the role name and `strict` flag, use this type definition:
+
+```
+attributes:
+  node:
+    role:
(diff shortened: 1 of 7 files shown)


ipspace:
@jbemmel -- I think this is exactly what you need for #3429

netsim/data/validate.py
```diff
@@ -482,6 +482,23 @@ def check_valid_with(
...
+  if isinstance(mapping,list):                    # Recursively transform lists
+    return [ transform_value_to_dict(data,v) for v in mapping ]
+  if not isinstance(mapping,str) or '{' not in mapping:
+    return mapping                                # Return values that are not f-string verbatim
+
+  result = strings.eval_format(mapping,{ 'value': data })
+  try:                                            # Evalute f-string and try evaluate its results
+    return eval(result)                           # as we could use f-string to generate a list or a computed int
+  except Exception:
+    return result                                 # If the f-string result doesn't parse, return it as string
```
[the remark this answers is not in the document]
ipspace: Fixed

netsim/data/validate.py
```diff
@@ -550,6 +568,23 @@ def validate_item(
     data = parent[key]
     data_type = Box(data_type)                                        # and fix datatype definition
 
+  # Another corner case: data type is a dictionary, we have a non-dict value, and the data type definition
+  # gives us a template to transform that value into a dictionary
+  #
+  if not isinstance(data,dict) and '_value_to_dict' in data_type and parent is not None:
+    try:
+      parent[key] = transform_value_to_dict(data,data_type._value_to_dict)
+    except Exception as ex:
```
jbemmel, line 577: When does ```transform_value_to_dict``` throw?

docs/dev/validation.md
```diff
@@ -495,6 +495,43 @@ ospf:
...
+The `{value}` string could be interpreted as a YAML dictionary and MUST therefore be quoted.
+```
+
+For example, to transform a BGP role specified as a string into a dictionary with the role name and `strict` flag, use this type definition:
+
+```
+attributes:
+  node:
+    role:
+      name:
```
jbemmel, line 513: Could add ```_required``` for name

jbemmel (approved):
Elegant solution

docs/dev/validation.md
```diff
@@ -495,6 +495,43 @@ ospf:
...
+        type: str
+        valid_values: [ provider, customer, peer, rs-server, rs-client ]
+      strict: bool
+      _value_to_dict:
+        name: '{value}'
```
jbemmel, line 518: There is a bit of duplication here. Consider instead of _value_to_dict:
```
attributes:
  node:
    role:
      name: { 'type': str, 'map_single_value': True }
      strict: { 'type': bool, default: False }
```

This avoids the duplicate structure, the lack of / implict type validation and the need for eval

netsim/data/validate.py
ipspace: If the f-string evaluation fails

docs/dev/validation.md
ipspace: Up to you ;)

docs/dev/validation.md
ipspace: It also doesn't allow you to restructure the value (if needed), or create a list from a single value or...

Type validation happens after the dict is created (see test results)

Issue #3431, opened 2026-06-02
Feature request: Support additional YAML type validation cases


jbemmel:
Handwritten version of #3430 

For the ```bgp.role``` plugin I encountered some limits to the way type validation is currently implemented. There is a specific list of allowed string values for ```role```, so I tried:

```
attributes:
  bgp_role_name:
    type: str
    valid_values: [ provider, customer, peer, rs-server, rs-client ]

bgp:
  attributes:
    global:
      role:
        _description: ...
        type: bgp_role_name
        _alt_types: [ bgp_role ]
```

Unfortunately, this (i.e. using ```valid_values``` in a user defined attribute type) fails to resolve type ```bgp_role_name``` - presumably because of the ```_description``` and ```_alt_types``` members.

As a "workaround" the PR #3429 currently implements custom must_be_xxx Python validation helpers that duplicate some of the definitions (like the list of valid strings). My preference would be that this could be done purely declaratively

ipspace:
> Handwritten version of [#3430](https://github.com/ipspace/netlab/issues/3430)

Thank you!

To recap: using `role: bgp_role_name` works:

```
defaults.device: none

defaults.attributes:
  bgp_role_name:
    type: str
    valid_values: [ provider, customer, peer, rs-server, rs-client ]
  node:
    brn: bgp_role_name

nodes:
  a:
    brn: provider
  b:
    brn: customer
```

While using `role.type: bgp_role_name` does not:

```
defaults.device: none

defaults.attributes:
  bgp_role_name:
    type: str
    valid_values: [ provider, customer, peer, rs-server, rs-client ]
  node:
    brn:
      type: bgp_role_name

nodes:
  a:
    brn: provider
  b:
    brn: customer
```

Results in:

```
[INFO]    Using lab topology file topology.yml
[FATAL]   No validation function for {'type': 'bgp_role_name'}
```

I have no problem fixing this (and it probably should be fixed), HOWEVER, why do you insist on using the indirect type definition? This works as well:

```
defaults.device: none

defaults.attributes:
  bgp_role_name:
  node:
    brn:
      type: str
      valid_values: [ provider, customer, peer, rs-server, rs-client ]
      _description: none

nodes:
  a:
    brn: provider
  b:
    brn: customer
```

ipspace:
BTW, a better definition of what you're trying to achieve with **bgp.role** would be (always defining the most complex type as the primary type):

```
  node:
    brn:
      name: bgp_role_name
[...]

PR #3439, opened 2026-06-04
Fix: _alt_types list supports user-defined data types

ipspace:
Implements the related discussion in #3431

### DIFF
--- a/netsim/data/validate.py
+++ b/netsim/data/validate.py
@@ -334,24 +334,29 @@ validate_alt_type -- deal with dictionaries that could be specified as something
 """
 
 def validate_alt_type(data: typing.Any, data_type: Box) -> dict:
-  if type(data).__name__ in data_type._alt_types:       # Simple check: is type name in alt types?
+  alt_types = get_a_list(data_type._alt_types)
+  if type(data).__name__ in alt_types:                  # Simple check: is type name in alt types?
     return { '_valid': True }                           # Got it, no need for more complex validation
 
   v_alt_err: list = []
 
-  for at in data_type._alt_types:                       # Iterate over alt-types
-    validation_function = getattr(_tv,f'must_be_{at}',None)
+  for at in alt_types:                                  # Iterate over alt-types
+    at_data_type = get_box(transform_validation_shortcuts(at))
+    validation_function = getattr(_tv,f'must_be_{at_data_type.type}',None)
     if not validation_function:                         # Is alt-type a data type with a validation function?
       continue                                          # ... nope, get out of here
 
+    # Pass user-defined type constraints (for example valid_values) to the base validator.
     v_result = validation_function(                     # Try to validate
                   parent=None,                          # ... a standalone value
                   key=data,                             # ... specified in this parameter
                   path='',                              # ... no valid path, but we have to supply something 
-                  _raw_status=True)                     # ... and return raw validation status
+                  _raw_status=True,
+                  **{ k:v for k,v in at_data_type.items() if not k.startswith('_') and k != 'type' })
     if v_result.get('_valid',False):
       return v_result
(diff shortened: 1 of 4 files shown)


netsim/data/validate.py
```diff
@@ -334,24 +334,29 @@ def validate_list(
...
-                  _raw_status=True)                     # ... and return raw validation status
+                  _raw_status=True,
+                  **{ k:v for k,v in at_data_type.items() if not k.startswith('_') and k != 'type' })
     if v_result.get('_valid',False):
       return v_result
-    v_alt_err.append(v_result.get('_value','') or v_result.get('_type',''))
+    v_alt_err.append(at if v_result.get('_value',None) == 'Invalid value' else
+                     v_result.get('_value','') or v_result.get('_type',''))
 
   return { '_alt_types': v_alt_err }                    # No alt data type matched, return the collected error messages
```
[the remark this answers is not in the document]
ipspace: Valid point, was able to reproduce. Fixed.

ipspace:
@jbemmel -- This is the missing part you were asking about yesterday -- using user-defined types in _alt_types.

netsim/data/validate.py
```diff
@@ -334,26 +334,31 @@ def validate_list(
...
       continue                                          # ... nope, get out of here
 
+    # Pass user-defined type constraints (for example valid_values) to the base validator.
     v_result = validation_function(                     # Try to validate
                   parent=None,                          # ... a standalone value
                   key=data,                             # ... specified in this parameter
                   path='',                              # ... no valid path, but we have to supply something 
-                  _raw_status=True)                     # ... and return raw validation status
+                  _raw_status=True,
+                  **{ k:v for k,v in at_data_type.items() if not k.startswith('_') and k != 'type' })
```
jbemmel, line 355: Does this need to be filtered? Why not simply pass at_data_type.items()?

jbemmel (approved):
I take it that the extra context passed to the validation function is not currently used?

I'm not a big fan of the custom implicit typing approach, would suggest a more explicit definition of custom user types and passing that as user_type dict to the validation function, instead of unpacked args

netsim/data/validate.py
ipspace: > Does this need to be filtered?

Yes.

> Why not simply pass at_data_type.items()?

Check the parameters of the validation functions and then the data type parameters that can start with "_" ;)

ipspace:
> I take it that the extra context passed to the validation function is not currently used?

How do you think we pass min/max values to must_be_int?

> I'm not a big fan of the custom implicit typing approach, would suggest a more explicit definition of custom user types and passing that as user_type dict to the validation function, instead of unpacked args

Oh well, that ship has sailed years ago ;)

jbemmel:
> > I take it that the extra context passed to the validation function is not currently used?
> 
> How do you think we pass min/max values to must_be_int?
> 
> >

The reason I said this is that ```kwargs``` in ```validate_value``` is not used anywhere. Turns out the actual validation happens in validate_item

netsim/data/validate.py
jbemmel: In that case it is inconsistent with https://github.com/ipspace/netlab/blob/bf0bd0e014ad5c5dc9d5b04a6ed9830cbb98862c/netsim/data/validate.py#L616

You may want to define a helper function to filter args to be passed to custom validation functions

ipspace:
> The reason I said this is that `kwargs` in `validate_value` is not used anywhere. Turns out the actual validation happens in validate_item

Thank you, let me check...

netsim/data/validate.py
ipspace: Thank you. Done in #3447

PR #3446, opened 2026-06-04
Fix FRR bgp_prefix validation to query VRF BGP tables.

jbemmel:
The show command ignored vrf= and always used the default BGP instance. Imported L3VPN prefixes live in the VRF RIB (e.g. tenant), so checks failed even when VPN propagation worked.

Use "bgp vrf <name> <af> unicast <pfx> json" for non-default VRFs, matching EOS behavior. 

Example failure: tests/integration/mpls/15-vpnv4-rr.yml replace 'eos' with 'frr'
pfx_tenant (bgp_prefix('172.16.0.0/24', vrf='tenant') on pe2).

### DIFF
--- a/netsim/validate/bgp/frr.py
+++ b/netsim/validate/bgp/frr.py
@@ -9,6 +9,7 @@ from box import Box, BoxList
 from netsim.data import global_vars
 
 from ...utils import log
+from ...utils import routing as _rp_utils
 from .. import _common
 from . import BGP_PREFIX_NAMES, check_community_kw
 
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Thank you!

On a tangential note: the additional fix I made is a perfect example of why I'm continuously telling you to review the AI-generated code. FRR can use "vrf default" in "show" commands, making the extra code complication completely unnecessary.

jbemmel:
> Thank you!
> 
> On a tangential note: the additional fix I made is a perfect example of why I'm continuously telling you to review the AI-generated code. FRR can use "vrf default" in "show" commands, making the extra code complication completely unnecessary.

And this is a good example of how 2 people can look at something and see different things. I did review the code but didn’t realize “vrf default” would work. Good fix

ipspace:
> I did review the code but didn’t realize “vrf default” would work.

Neither did I, but somehow slipped into the "I wonder if..." mood, resulting in a really pleasant surprise ;))

Thanks again, Ivan

PR #3462, opened 2026-06-07
Correct ospf/fortios.j2 control-tag indentation post #3451

a-v-popov:
This is control-tag indentation cleanup only PR.
It can be ignored, but then we will be dragging the exception forever, or till the next change
and then that change will get the commit which after being squashed will obscure the intent. 
It should be noted that I wasn't able to find any rule for Jinja endorsed in the repo,
so the "1/3/5/7..." rule was rather inferred from the majority of other template files.
Sorry for the mess.

## What

Re-indents the Jinja control tags in `netsim/ansible/templates/ospf/fortios.j2`
to match the repository-wide convention for `.j2` templates: flush-left `{%`,
one space of inside-padding at the outermost level, and +2 per enclosing
`if`/`for` (1/3/5/7…).
## Why

The file, rewritten recently in #3451, mixed two indentation schemes:

- `{% set %}` tags used the canonical one-space inside-padding;
- the `{% if %}`/`{% for %}` ladder used a base-2, +4-per-level scheme (2/6/10).

As a result, sibling tags at the same nesting depth disagreed — for instance a
`set` at 1 space sitting directly next to an `if` at 2. This commit makes the
whole file uniform on the 1/3/5 convention used by the mature templates
(`bgp/frr.j2`, `bgp/eos.j2`, and the large majority of `.j2` files), which keeps
future diffs on this file clean.

## Render-equivalence

This is a whitespace-only change with no effect on rendered output:

- Inside-padding lives between `{%` and the keyword, so it can never reach the
  generated configuration.
- `trim_blocks` / `lstrip_blocks` (set in `netsim/utils/templates.py`) strip the
  whitespace around the tags anyway.

`git diff -w` against the prior revision is empty; the diff is purely the
re-padding of 28 control-tag lines.

## Testing

No behavioural change, so no new tests. Verified render-equivalence via
`git diff -w` (empty).

### DIFF
--- a/netsim/ansible/templates/ospf/fortios.j2
+++ b/netsim/ansible/templates/ospf/fortios.j2
@@ -10,63 +10,63 @@
 {% set passive_intfs = intf_list|selectattr('ospf.passive', 'defined')|selectattr('ospf.passive')|map(attribute='ifname')|list %}
 config router {{ proto }}
     set router-id {{ ospf.router_id }}
-{%  if ospf.reference_bandwidth is defined %}
+{% if ospf.reference_bandwidth is defined %}
     set auto-cost-ref-bandwidth {{ ospf.reference_bandwidth }}
-{%  endif %}
-{%  if passive_intfs|length > 0 %}
+{% endif %}
+{% if passive_intfs|length > 0 %}
     set passive-interface {{ passive_intfs|join(' ') }}
-{%  endif %}
+{% endif %}
 
     config area
-{%  for area_data in ospf.areas|default([]) %}
+{% for area_data in ospf.areas|default([]) %}
         edit {{ area_data.area }}
         next
-{%  endfor %}
+{% endfor %}
     end
-{%  if intf_list|length > 0 %}
+{% if intf_list|length > 0 %}
 
     config {{ intf_stanza }}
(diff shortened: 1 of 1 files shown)


a-v-popov:
An alternative could be adopting something like https://github.com/aristanetworks/j2lint
But that will require to change virtually every template as indentation parameters seems to be hardcoded.
https://github.com/aristanetworks/j2lint/blob/971cdba1c431a064b57a172e7bea117b8575e647/j2lint/linter/indenter/node.py#L19

ipspace (approved):
Thank you!

ipspace:
> It should be noted that I wasn't able to find any rule for Jinja endorsed in the repo, so the "1/3/5/7..." rule was rather inferred from the majority of other template files.

Would you like to add something along these lines to AGENTS.md (considering everyone seems to love writing code with AI these days)?

The indentation is admittedly a bit inconsistent. For short macros, the commands within a macro are indented, whereas for long sequences, they're not. Otherwise, we sometimes get too much of a mismatch between configuration commands and Jinja2 control statements.

> Sorry for the mess.

Absolutely not a problem. Thanks again, Ivan

PR #3467, opened 2026-06-10
BGP Graceful Restart implementation (EOS, FRR)

ipspace:
Implements #3456

### DIFF
--- a/docs/plugins/bgp.session.md
+++ b/docs/plugins/bgp.session.md
@@ -236,7 +236,6 @@ BGP Graceful Restart capabilities can be defined per node, per VRF, or per EBGP 
 * **bgp.gr.state** -- node Graceful Restart state (`enable`, `disable` or `helper`).
 * **bgp.gr.restart_time** -- the expected neighbor restart time (integer, up to 3600 seconds)
 * **bgp.gr.stalepath_time** -- the time after which the stale paths are purged (integer, up to 3600 seconds)
-* **bgp.gr.update_delay** -- the maximum initial BGP update delay (integer, up to 3600 seconds)
 
 ```{warning}
 * The built-in device BGP Graceful Restart default (usually: helper) is not changed unless you specify node-level **bgp.gr.state** value.
--- a/docs/plugins/bgp.session.md
+++ b/docs/plugins/bgp.session.md
@@ -18,6 +18,7 @@ The plugin adds the following BGP session attributes:
 * **bgp.as_override** is a boolean attribute that controls whether a BGP router replaces peer AS with its own AS.
 * **bgp.bfd** is a boolean attribute that enables BFD with BGP neighbors.
 * **bgp.default_originate** is a boolean attribute that controls whether a BGP router advertises a default route to its neighbor(s).
+* **bgp.gr** defines [BGP graceful restart](plugin-bgp-session-gr) capabilities
 * **bgp.gtsm** is an integer attribute that enables the Generic TTL Security Mechanism (GTSM). A *true* value sets it to 1 (the remote router can be up to one hop away).
 * **bgp.passive** is a boolean attribute that makes a node act as a passive BGP peer[^PBP] on an interface or on all applicable sessions (when specified as a node parameter).
 * **bgp.password** is a string attribute that specifies the MD5  or TCP-AO password used on EBGP sessions.
@@ -44,6 +45,7 @@ BGP session attributes can be specified at the global, node, link, or interface 
 | as_override           |    ❌   |  ❌   |  ❌   |    ✅     |
 | bfd                   |   ✅   |  ✅  |  ✅  |    ✅     |
 | default_originate     |    ❌   |  ✅  |  ❌   |    ✅     |
+| gr                    |    ❌   |  ✅  |  ✅  |    ✅     |
 | gtsm                  |   ✅   |  ✅  |  ✅  |    ✅     |
 | passive               |    ❌   |  ✅  |   ❌  |    ✅     |
 | password              |   ✅   |  ✅  |  ✅  |    ✅     |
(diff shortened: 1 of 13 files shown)


netsim/extra/bgp.session/frr.j2
```diff
@@ -43,15 +43,42 @@
...
 {#
 
+#}
+{% macro bgp_gr(bgp) %}
+{%   if bgp.gr.restart_time is defined %}
+  bgp graceful-restart restart-time {{ bgp.gr.restart_time }}
+{%   endif %}
+{%   if bgp.gr.update_delay is defined %}
+  update-delay {{ bgp.gr.update_delay }}
+{%   endif %}
```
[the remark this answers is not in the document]
ipspace: So much for FRR command consistency ;)

netsim/extra/bgp.session/defaults.yml
```diff
@@ -58,6 +59,15 @@ bgp:
       rs: bool
       tcp_ao:
         copy: global
+      gr:
+        state:
+          type: str
+          valid_values: [ enable, disable, helper ]
```
jbemmel: _required?

docs/plugins/bgp.session.md
```diff
@@ -226,6 +228,37 @@ Introducing a BGP route server to an IP subnet also prunes the full mesh of EBGP
 
 Please note that an IBGP session between route servers in the same autonomous system is retained and might generate a warning unless you also run an IGP between the route servers. To remove the IBGP session between route servers, set the `bgp.sessions.ipv4: [ ebgp ]` and `bgp.sessions.ipv6: [ ebgp ]` node attributes on route server nodes.
 
+(plugin-bgp-session-gr)=
+## BGP Graceful Restart Capabilities
+
+BGP Graceful Restart capabilities can be defined per node or per EBGP neighbor. The node-level **bgp.gr** attribute is a dictionary with these parameters (it can also be specified as the **bgp.gr.state** string value):
+
+* **bgp.gr.state** -- node Graceful Restart state (`enable`, `disable` or `helper`).
+* **bgp.gr.restart_time** -- the expected neighbor restart time (integer, up to 3600 seconds)
```
jbemmel, line 237: The implementation only allows timer tweaks at the global/node level, not per neighbor. I think that's fine, but it should be documented

netsim/extra/bgp.session/eos.j2
```diff
@@ -42,8 +42,36 @@
...
 {%- endmacro %}
 
+{% macro bgp_gr(bgp) %}
+{%   if bgp.gr.restart_time is defined %}
+  graceful-restart restart-time {{ bgp.gr.restart_time }}
+{%   endif %}
+{%   if bgp.gr.stalepath_time is defined %}
+  graceful-restart stalepath-time {{ bgp.gr.stalepath_time }}
+{%   endif %}
+{%   set gr_state = bgp.gr.state|default('helper') %}
```
jbemmel: The default may be better specified in the plugin - either using ```_required``` or by specifying this default value in YAML?

tests/integration/bgp.session/14-gr.yml
```diff
@@ -0,0 +1,80 @@
...
+    plugin: bgp_neighbor(node.bgp.neighbors,'dut',af='ipv6')
+
+  gr_enable:
+    description: Check per-neighbor GR enabled mode
+    nodes: [ x1 ]
+    show:
+      frr: "bgp neighbors json"
+    valid:
+      frr: >-
+        result["{{ bgp.neighbors[0].ipv4 }}"].gracefulRestartInfo.remoteGrMode == "Restart"
```
jbemmel, line 76:
```
plugin: |-
      bgp_neighbor_details(
        node.bgp.neighbors,'dut',
        gracefulRestartInfo={'remoteGrMode': 'Restart'})
```

using a slightly modified validation function:
```
def valid_bgp_neighbor_details(
      ngb: list,
      n_id: str,
      af: str = 'ipv4',**kwargs: typing.Any) -> str:
  _result = global_vars.get_result_dict('_result')

  n_addr = _common.get_bgp_neighbor_id(ngb,n_id,af)
  data = _result[n_addr]

  for k,v in kwargs.items():
    if not k in data:
      raise Exception(f'Neighbor data structure does not contain attribute {k}')
    if isinstance(v,dict):
      if not isinstance(data[k],(dict,Box)):
        raise Exception(f'{k} is not a dictionary')
      for sk,sv in v.items():
        if sk not in data[k]:
          raise Exception(f'Neighbor data structure does not contain attribute {k}.{sk}')
        if data[k][sk] != sv:
          raise Exception(f'{k}.{sk} expected value {sv} actual {data[k][sk]}')
    elif data[k] != v:
      raise Exception(f'{k} expected value {v} actual {data[k]}')

  return f'All specified BGP neighbor parameters have the expected values'
```

netsim/extra/bgp.session/eos.j2
```diff
@@ -42,8 +42,36 @@
...
+  no neighbor {{ peer }} graceful-restart
+  no neighbor {{ peer }} graceful-restart-helper
+{%     elif n.gr == 'helper' %}
+  neighbor {{ peer }} graceful-restart-helper
+{%     endif %}
+{%   endif %}
 {%- endmacro %}
 
+{% macro bgp_gr(bgp) %}
+{%   if bgp.gr.restart_time is defined %}
```
jbemmel, line 58: Need to configure update_delay too?

jbemmel:
EOS supports RFC9494 (Long-Lived Graceful Restart for BGP) - see https://arista.my.site.com/AristaCommunity/s/article/Understanding-BGP-Graceful-Restart

Could consider to add support for that, perhaps in a follow-up PR

docs/plugins/bgp.session.md
```diff
@@ -226,6 +228,37 @@ Introducing a BGP route server to an IP subnet also prunes the full mesh of EBGP
 
 Please note that an IBGP session between route servers in the same autonomous system is retained and might generate a warning unless you also run an IGP between the route servers. To remove the IBGP session between route servers, set the `bgp.sessions.ipv4: [ ebgp ]` and `bgp.sessions.ipv6: [ ebgp ]` node attributes on route server nodes.
 
+(plugin-bgp-session-gr)=
+## BGP Graceful Restart Capabilities
+
+BGP Graceful Restart capabilities can be defined per node or per EBGP neighbor. The node-level **bgp.gr** attribute is a dictionary with these parameters (it can also be specified as the **bgp.gr.state** string value):
```
jbemmel: I think VRF scoped configuration should be supported too. Designs where GR is a thing will likely use VRFs as well

jbemmel (changes requested):
See detailed comments, VRF scoped is SHOULD and RFC9494 long-lived is nice to have

jbemmel:
Actually, on second thought perhaps update_interval doesn’t belong in GR to begin with

netsim/extra/bgp.session/defaults.yml
ipspace: Yeah, we could go for that.

docs/plugins/bgp.session.md
ipspace: Not sure you can do tweaks per neighbor (at least in most implementations). Also, not sure how valuable these tweaks are in virtual labs anyway.

netsim/extra/bgp.session/eos.j2
ipspace: Agree. I'd go with "required"

tests/integration/bgp.session/14-gr.yml
ipspace: In theory, you're correct. In practice, I don't have that much of a problem writing an occasional expression, as it's only used on one platform in one test.

netsim/extra/bgp.session/eos.j2
ipspace: Couldn't find the relevant parameter in Arista EOS.

docs/plugins/bgp.session.md
ipspace: FRR and EOS both seem to have VRF-aware GR configuration, so sure.

ipspace:
> Actually, on second thought perhaps update_interval doesn’t belong in GR to begin with

Agree, I'll throw it out.

a-v-popov:
What configuration is required in a topology file  to simply enable GR for all neighbors?

## Node-level-only `bgp.gr` is silently ignored

`api.node_config()` is reached only through `mark_plugin_config()` inside `apply_neighbor_attributes()` (`plugin.py:57`). Because `gr` is in `no_propagate`, a node-level `bgp.gr` never produces a neighbor attribute, so a node whose *only* session attribute is node-level `bgp.gr` never gets the `bgp.session` config deployed.

Reproduction: two FRR nodes, `dut.bgp.gr.state: disable`, nothing else. After `netlab create`, the transformed topology carries `bgp.gr.state`, but `nodes.dut.config` is absent and `node_files/dut` contains no `bgp.session` file — the device keeps its default GR behaviour with no warning.

The integration test does not catch it because the DUT also carries interface-level `gr` on two links, which schedules the config. A possible fix: call `api.node_config(ndata, _config_name)` in `post_transform()` whenever `ndata.bgp.gr` is defined.

Related gap: interface-level `gr` flows through `check_device_attribute_support()` (`plugin.py:56`), but the node-level value bypasses it, so `bgp.gr` on a device without GR support produces neither configuration nor warning. The bypass also affects devices that will support only a subset of states via the established `gr.valid: [ ... ]` feature form (cf. `remove_private_as.valid`. Running the node-level `state` through the same valid-list check would close both holes.

ipspace:
> Node-level-only `bgp.gr` is silently ignored

Thank you, missed that.

tests/integration/bgp.session/14-gr.yml
```diff
@@ -0,0 +1,94 @@
...
+links:
+- dut.bgp.gr: enable
+  x1:
+- dut.bgp.gr: helper
+  x2:
+- dut:
+  x3:
+- dut.bgp.gr: enable
+  v1:
+  vrf: tenant
```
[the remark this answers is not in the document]
ipspace: Nope, this works ;)

ipspace:
@jbemmel -- Added VRF support (which required some crazy attribute gymnastics, have to document that one of these days). Would you take another look at it?

@a-v-popov -- Still have to work on that fix. Might take a bit...

jbemmel (approved):
I could see a use case for enabling BGP GR globally, but that's just my minor preference. It is documented consistently.

Approved (assuming the minor issues blocking the tests will be fixed of course)

netsim/extra/bgp.session/plugin.py
```diff
@@ -81,6 +87,34 @@ def copy_local_attributes(ndata: Box, topology: Box) -> None:
...
+def check_node_attributes(ndata: Box, topology: Box) -> None:
+  '''
+  Some bgp.session attributes (like bgp.gr) can be specified on the node level.
+  Check that the device supports them, and mark the device as needing extra config
+  if we find at least one of them.
+  '''
+
+  global _config_name
+
+  attrs = topology.defaults.bgp.attributes.session.attr
```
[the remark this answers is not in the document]
ipspace: Nice catch, but this is intentional

docs/plugins/bgp.session.md
```diff
@@ -18,6 +18,7 @@ The plugin adds the following BGP session attributes:
 * **bgp.as_override** is a boolean attribute that controls whether a BGP router replaces peer AS with its own AS.
 * **bgp.bfd** is a boolean attribute that enables BFD with BGP neighbors.
 * **bgp.default_originate** is a boolean attribute that controls whether a BGP router advertises a default route to its neighbor(s).
+* **bgp.gr** defines [BGP graceful restart](plugin-bgp-session-gr) capabilities
```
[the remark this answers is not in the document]
ipspace: Nice try, but it works

PR #3470, opened 2026-06-12
BIRD: Add multi-stage Dockerfile to build latest v2.19.1 from source

jbemmel:
Provides a slim runtime image with BuildKit apt caching and parallel compilation for faster rebuilds.

Build: ```netlab clab build bird.v2_from_src``` with optional ```--sw-version```

Use: ```netlab up -d bird -p clab 01-interfaces.yml -s nodes.dut.box=netlab/bird.v2_from_src:latest```

Builds Bird v2.19.1 released on June 9th 2026, which reportedly supports
* BGP: Automatic peering based on discovered neighbors
* BGP/MPLS Ethernet VPNs using VXLAN tunnels

Now using a .j2 Dockerfile template to render user requested version

### DIFF
--- a/netsim/daemons/bird/Dockerfile.v2_from_src
+++ b/netsim/daemons/bird/Dockerfile.v2_from_src
@@ -16,14 +16,18 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     wget \
     && rm -rf /var/lib/apt/lists/*
 
-# Set BIRD version, 2.19.1 was released on 2026-06-09
-ENV BIRD_VERSION=2.19.1
+ARG SW_VERSION
+ARG SW_DOWNLOAD_URL
 
 # Download and extract source code
-RUN wget https://bird.nic.cz/download/bird-${BIRD_VERSION}.tar.gz \
-    && tar -xzf bird-${BIRD_VERSION}.tar.gz
+RUN set -e; \
+    if ! wget -q -O /tmp/sw-src.tar.gz "$SW_DOWNLOAD_URL"; then \
+      echo "ERROR: Cannot download BIRD ${SW_VERSION} from ${SW_DOWNLOAD_URL} — version may not exist"; \
+      exit 1; \
+    fi; \
+    tar -xzf /tmp/sw-src.tar.gz
 
-WORKDIR /bird-${BIRD_VERSION}
+WORKDIR /bird-${SW_VERSION}
 
 # Configure, compile, and install
 RUN ./configure --prefix=/usr --sysconfdir=/etc/bird --localstatedir=/var \
--- a/netsim/daemons/bird/Dockerfile.v2_from_src
+++ b/netsim/daemons/bird/Dockerfile.v2_from_src
@@ -1,0 +1,59 @@
+# syntax=docker/dockerfile:1
+# --- Stage 1: Build Stage ---
+FROM debian:bookworm-slim AS builder
+
+# Install compilation dependencies
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt-get update && apt-get install -y \
+    build-essential \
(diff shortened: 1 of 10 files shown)


netsim/daemons/bird/Dockerfile.v2_from_src
```diff
@@ -0,0 +1,59 @@
...
+    bison \
+    libncurses-dev \
+    libreadline-dev \
+    libssh-dev \
+    autoconf \
+    wget \
+    && rm -rf /var/lib/apt/lists/*
+
+# Set BIRD version, 2.19.1 was released on 2026-06-09
+ENV BIRD_VERSION=2.19.1
```
ipspace: And why would it be a good idea to hardcode a software version into a build file?

netsim/daemons/bird/Dockerfile.v2_from_src.j2
```diff
@@ -0,0 +1,59 @@
...
+# Copy binaries and built directories from the builder stage
+COPY --from=builder /usr/sbin/bird /usr/sbin/bird
+COPY --from=builder /usr/sbin/birdcl /usr/sbin/birdcl
+COPY --from=builder /etc/bird /etc/bird
+
+# Create runtime directory for control sockets
+RUN mkdir -p /var/run
+
+# Expose BGP (179) and OSPF (89) routing ports
+EXPOSE 179
```
ipspace: Seriously? Tell Cursor to do better next time 😜

ipspace (changes requested):
Nice job. However, please try to figure out how to set the BIRD version via an environment variable or some such, and document how to build various versions of the BIRD container, including building from source. Best place would be `labs/bird.md`, linked from `labs/clab.md`.

Also, it would be nice to set the container label to the actual software version when building from source.

netsim/daemons/bird/Dockerfile.v2_from_src
jbemmel: It was a first draft to get the idea across. I considered making the version configurable, but chose not to go there until asked. I'll make the changes

jbemmel:
> Also, it would be nice to set the container label to the actual software version when building from source.

My thoughts exactly, but again a bigger change. I'll make the adjustments

netsim/daemons/bird/Dockerfile.v2_from_src.j2
jbemmel: no comment ;)

netsim/daemons/bird/Dockerfile.v2_from_src
jbemmel: Moved the default version to ```clab.sw_version``` (could use ```clab.build.sw_version``` if preferred)

User can override using ```--sw-version``` CLI param (but only for targets where ```clab.sw_version``` is defined)

jbemmel: > And why would it be a good idea to hardcode a software version into a build file?

See also https://github.com/ipspace/netlab/blob/4c94745df35e07cbfcbf1c356ebeefa3d2ebee51/netsim/daemons/netscaler/Dockerfile.j2#L1

ipspace: Well, now we have a mechanism to fix that ;)

ipspace: > Moved the default version to `clab.sw_version` (could use `clab.build.sw_version` if preferred)

I think we should go with `clab.build.sw_version`. No worries, I can fix that.

ipspace: > I think we should go with `clab.build.sw_version`. No worries, I can fix that.

And of course I was wrong 🤷🏻‍♂️ Doing that would most probably trigger changes in clab.py code (where it displays how to build the container). I'll leave it as-is.

ipspace (approved):
Nice job, thank you!

PR #3471, opened 2026-06-12
Fix: capture validate log in device-module-test on warnings

jbemmel:
## Summary

- Fix a crash in `device-module-test` when `netlab validate` exits with code 3 (warnings only).
- Always capture validate output to the log file; `extract_warning()` now tolerates a missing log file.

## What triggered the issue

Running `./device-module-test -d bird -p clab bgp` reached `bgp/03-ibgp-rr.yml`, where the `inact` check is marked `level: warning`. BIRD does not reflect inactive routes, so validate correctly exited with code 3.

In normal (non-`--quiet`, non-`--batch`) mode, validate output went to the terminal and no `-validate.log` file was written. The harness then called `extract_warning()`, which unconditionally read that log path and raised `FileNotFoundError`.

## Test plan

- [x] Reproduce: `./device-module-test -d bird -p clab bgp 03-ibgp-rr` — validate reports a warning and the test completes instead of crashing
- [ ] Confirm other integration tests still pass with validate warnings recorded in `results.yaml`

### DIFF
--- a/tests/integration/device-module-test
+++ b/tests/integration/device-module-test
@@ -210,7 +210,10 @@ def execute(
 
 def extract_warning(test: str, args: argparse.Namespace, step: str = 'validate') -> list:
   log_name = f"{args.logdir}/{os.path.basename(test)}-{step}.log"
-  log_lines = pathlib.Path(log_name).read_text().split('\n')
+  log_path = pathlib.Path(log_name)
+  if not log_path.exists():
+    return []
+  log_lines = log_path.read_text().split('\n')
   w_prefixes = ['[WARNING]','Warning in ']
   return [ line.replace(w_pfx,'').strip() for w_pfx in w_prefixes for line in log_lines if line.startswith(w_pfx) ]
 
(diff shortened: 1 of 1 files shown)


tests/integration/device-module-test
```diff
@@ -337,9 +340,8 @@ def run_test(test: str, args: argparse.Namespace,first: bool) -> None:
   if args.batch:
     valid_cmd += " --dump result"
 
-  valid_capture = "validate" if args.quiet or args.batch else None
   valid_cmd = test_def.get("validate",valid_cmd)
-  valid = execute(valid_cmd,args,test=test,log=valid_capture,err=valid_capture)
+  valid = execute(valid_cmd,args,test=test,log="validate",err="validate")
```
ipspace: This defies the whole idea of "show the user how the test is progressing"

tests/integration/device-module-test
```diff
@@ -210,7 +210,10 @@ def execute(
 
 def extract_warning(test: str, args: argparse.Namespace, step: str = 'validate') -> list:
   log_name = f"{args.logdir}/{os.path.basename(test)}-{step}.log"
-  log_lines = pathlib.Path(log_name).read_text().split('\n')
+  log_path = pathlib.Path(log_name)
```
ipspace, line 213: This part makes sense. Do you want to revert the other part of commit or should I copy it into a new PR?

ipspace (changes requested):
There's a reason we show the user stdout/stderr

tests/integration/device-module-test
jbemmel: agreed, reverted

tests/integration/device-module-test
jbemmel: reverted

ipspace (approved):
Lovely. Thank you!

PR #3472, opened 2026-06-12
Bird: BGP confederation support

jbemmel:
Passes ```40-confederation```, full bgp test suite after #3471

### DIFF
--- a/docs/dev/config/bgp.md
+++ b/docs/dev/config/bgp.md
@@ -82,7 +82,7 @@ A description of a BGP neighbor might have these parameters:
 * **local_as** (optional) -- Local autonomous system (when using **local-as** functionality)
 * **local_if** (optional) -- Local interface name for "unnumbered" BGP neighbors (EBGP sessions over IPv6 LLA configured as interface sessions)
 * **name** -- Neighbor name (used in descriptions)
-* **next_hop_self** (optional) -- use "next-hop-self" for this neighbor. Can contain **ebgp** (change next hop for EBGP routes) or **always** (change next hop for all routes, including reflected IBGP routes).
+* **next_hop_self** (optional) -- use "next-hop-self" for this neighbor. Can contain **ebgp** (change next hop for EBGP routes) or **all** (change next hop for all routes, including reflected IBGP routes).
 * **rr_client** (optional) -- the neighbor is a route-reflector client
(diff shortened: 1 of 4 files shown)


netsim/daemons/bird/bgp.j2
```diff
@@ -133,8 +139,8 @@ protocol bgp bgp_{{ n.name }}_{{ af }} {
     import all;
     export filter { bgp_export_{{ n.type }}( {{ 'true' if n.get('default_originate') else 'false' }},
                                              {{ 'true' if n.get('remove_private_as') else 'false' }}); };
-{%       if 'next_hop_self' in bgp and bgp.next_hop_self and 'ibgp' in n.type %}
-    next hop self {{ 'on' if n.type == 'localas_ibgp' else 'ebgp' }};
+{%       if n.next_hop_self|default(false) or ('next_hop_self' in bgp and bgp.next_hop_self and 'ibgp' in n.type) %}
```
ipspace: Since the introduction of n.next_hop_self, we no longer need to check other parameters (unless there's something really special about BIRD, but it doesn't look like that). See https://netlab.tools/dev/config/bgp/#bgp-neighbors

netsim/daemons/bird/bgp.j2
```diff
@@ -133,8 +139,8 @@ protocol bgp bgp_{{ n.name }}_{{ af }} {
...
+    next hop self {{ 'on' if n.next_hop_self|default(false) == 'all' or n.type == 'localas_ibgp' else 'ebgp' }};
```
ipspace: I don't think we need to check n.type anymore. The n.next_hop_self should be reliably set to "all" or "ebgp"

docs/dev/config/bgp.md
```diff
@@ -263,41 +263,6 @@ You also have to modify the "set peer name" logic in AF configuration:
 {%   endfor %}
 ```
 
-(dev-bgp-confederation-bird)=
```
ipspace: It looks there's a good reason I don't let AI commit changes or submit PRs ;))

ipspace (changes requested):
I think the next-hop logic needs to be cleaned up. Thank you!

netsim/daemons/bird/bgp.j2
jbemmel: The documentation link says "ebgp" or "always" - I'll double check

ipspace (approved):
Works like a charm. Thank you (and thanks for the doc fix)!

PR #3474, opened 2026-06-13
Fix anycast-check crash when device lacks gateway.protocol

jbemmel:
## Summary
- Default `features.get('gateway.protocol')` to `[]` in the EVPN anycast-check test plugin
- Prevents `TypeError` when `netlab create -d <device>` uses a default device without gateway features (for example, bird)

## Test plan
- [x] `netlab create -d bird -p clab evpn/02-vxlan-asymmetric-irb.yml` no longer crashes in `anycast-check` pre_transform (fails later on missing gateway module, as expected)


ipspace (approved):
Thank you! Looks like all other devices have some gateway functionality, but not anycast.

PR #3475, opened 2026-06-14
Bird: add bgp/evpn support with vxlan and anycast gateway

jbemmel:
This depends on #3470 for the new 2.19.1 release

* Adds support for evpn, vxlan and gateway (anycast) modules
* Reusing FRR logic/scripts for VLAN and VXLAN provisioning

Couple of highlights:
* Needed a way to sync daemon startup with Netlab running config scripts -> bird-specific config script touches a 'done' file that the container waits for
* Had to make frr scripts more idempotent and reusable in some places
* Retains bird :ns provisioning
* Putting bird error log on stderr so that docker logs shows it - including when using older version with EVPN
* bird/vlan.j2 is a bash script while bird/vxlan.j2 is config, might benefit from cleaner separation (but then lookup paths need to be changed)

device-module-test: 4 new cases OK
```
./device-module-test -d bird -p clab evpn
Pre-test cleanup: 
* 19:53:59 evpn/01-vxlan-bridging.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
* 19:54:26 evpn/02-vxlan-asymmetric-irb.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
19:54:48 evpn/03-vxlan-symmetric-irb.yml: create(FAIL) 
19:54:48 evpn/04-vxlan-central-routing.yml: create(FAIL) 
19:54:49 evpn/05-vxlan-l3only.yml: create(FAIL) 
19:54:50 evpn/06-vxlan-bridging-vlan-bundle.yml: create(FAIL) 
19:54:50 evpn/10-vxlan-rr.yml: create(ok) up(FAIL) cleanup(ok)
* 19:54:52 evpn/11-vxlan-ebgp.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
* 19:55:13 evpn/12-vxlan-ibgp-ebgp.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
19:55:22 evpn/13-vxlan-ebgp-allowas.yml: create(FAIL) 
19:55:23 evpn/14-vxlan-ebgp-ebgp.yml: create(FAIL) 
19:55:23 evpn/15-vxlan-ebgp-unnumbered.yml: create(FAIL) 
19:55:24 evpn/20-vxlan-irb-ospf.yml: create(FAIL) 
19:55:25 evpn/21-bgp-ce-router.yml: create(FAIL) 
19:55:25 evpn/22-ospf-ce-router.yml: create(FAIL) 
19:55:26 evpn/30-cs-bridging.yml: create(FAIL) 
19:55:26 evpn/41-vxlan-ipv6-bridging.yml: create(FAIL) 
19:55:27 evpn/51-mpls-bridging.yml: create(FAIL) 
19:55:28 evpn/52-mpls-asymmetric-irb.yml: create(FAIL) 
19:55:28 evpn/53-evpn-l3only.yml: create(FAIL) 
19:55:29 evpn/54-evpn-l3vpn.yml: create(FAIL) 
19:55:29 evpn/61-sr-bridging.yml: create(FAIL) 
```
Most failures are due to missing vrf support; 10 fails because I don't have cEOS locally


jbemmel:
Could add a version check that aborts when the image isn't 2.19.0 or newer

netsim/ansible/templates/gateway/frr.j2
```diff
@@ -53,4 +53,6 @@ if [ ! -e /sys/class/net/{{ v_if }} ]; then
   ip link set dev {{ v_if }} up
 fi
 {% endfor %}
+{% if interfaces | selectattr('gateway.protocol', 'defined') | selectattr('gateway.protocol', 'equalto', 'vrrp') | list %}
```
ipspace: OK, why is this needed? Is there something in the VRRP config script that breaks stuff if no interfaces use VRRP?

After reviewing the rest of the PR: if this is a hack to say "only add FRR config when running on FRR", then split the template into "frr.j2" and "frr.data-plane.j2" and use only the data-plane part with Bird.

netsim/ansible/templates/initial/bird.j2
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+#
+# Initial configuration for BIRD containers (docker exec / sh mode).
+# Do not use the -clab suffix here: that convention is for netns execution on the host.
+#
+set -x
+set -e
+{% if 'vxlan' in module|default([]) %}
```
ipspace: These scripts are supposed to be executed **well after** containerlab says "I'm done", so I don't understand why the interfaces would not be ready.

netsim/ansible/templates/vxlan/frr.j2
```diff
@@ -6,10 +6,12 @@ set -e # Exit immediately when any command fails
 {% set use_evpn = vxlan.flooding|default('') == 'evpn' %}
 
 {% macro create_vxlan_interface(vni,br_name,vrf=None,mtu=1500) %}
+if [ ! -e /sys/devices/virtual/net/vxlan{{ vni }} ]; then
```
ipspace, line 9: Nice touch, but then it would be nice to have the contents of the IF block indented.

netsim/daemons/bird/Dockerfile.v2_from_src.j2
```diff
@@ -57,8 +61,23 @@ COPY --from=builder /usr/sbin/bird /usr/sbin/bird
 COPY --from=builder /usr/sbin/birdcl /usr/sbin/birdcl
 COPY --from=builder /etc/bird /etc/bird
 
-# Create runtime directory for control sockets
-RUN mkdir -p /var/run
+# Wait for netlab initial to execute bound /etc/config scripts, then start BIRD.
```
ipspace, line 60: These changes would have to be made to all Bird Dockerfiles, otherwise we'll have a fun troubleshooting session once the "standard" release starts supporting VXLAN/EVPN.

netsim/providers/clab/configs.py
```diff
@@ -12,6 +12,24 @@
 from ...utils import log, strings
 from . import utils
 
+
```
ipspace: It would be much better to add a script in the bird config quirk to set the flag you need.

netsim/daemons/bird.yml
```diff
@@ -16,12 +19,16 @@ daemon_config:
   bgp@policy: /etc/bird/bgp.policy.conf
   ospf@areas: /etc/bird/ospf.areas.conf
   ospf: /etc/bird/ospf.mod.conf
+  vxlan: /etc/bird/vxlan.mod.conf
+  evpn: /etc/bird/evpn.mod.conf
   routing: /etc/bird/routing.mod.conf
 clab:
+  node_config: null             # Do not inherit linux :ns scripts; use netlab_config_mode: sh
```
ipspace: (A) -- this looks like a hack
(B) -- You could also override the initial value
(C) -- Would it not be better to configure the interfaces from the "outside", then you wouldn't need all the tools within the container?

ipspace (changes requested):
Nice one, but I think a bunch of things might have to be changed.

netsim/ansible/templates/gateway/frr.j2
jbemmel: done

netsim/ansible/templates/vxlan/frr.j2
jbemmel: ok, also updated indentation further down the file

netsim/ansible/templates/initial/bird.j2
jbemmel: Glitch during development, now that the dust has settled they come up

netsim/daemons/bird/Dockerfile.v2_from_src.j2
jbemmel: Added, now using a local bird config-done script to set the marker

netsim/daemons/bird.yml
jbemmel: Switched back to :ns config mode

netsim/providers/clab/configs.py
jbemmel: Done, needs docker exec

netsim/extra/config-done/bird.j2
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+#
+set -e
+docker exec {{ ansible_host }} touch /var/run/netlab-config-done
```
jbemmel, line 4: Alternative: See #3498 where this is done after initial/bird-clab.j2 without requiring this odd 'config-done' quasi-plugin

netsim/daemons/bird/bgp.j2
```diff
@@ -39,6 +45,9 @@ function bgp_prefixes( bool originate_default ) {
   if source ~ [ {{ netlab_import_map[proto] }} ]
     then accept "{{ proto }} route:", net;
 {%   endfor %}
+{% elif 'ospf' in module|default([]) %}
```
ipspace: WTAF is this supposed to fix? We implemented route redistribution for Bird, so there should be no need for hacks like this one.

netsim/daemons/bird/Dockerfile
```diff
@@ -4,8 +4,27 @@ ENV DEBIAN_FRONTEND=noninteractive
...
 LABEL description="BIRD Internet Routing Daemon (bird.network.cz)"
 
-RUN mkdir /etc/bird && mkdir /run/bird && apt-get update && apt-get install -y iputils-ping net-tools iproute2 bird2
+RUN mkdir /etc/bird && mkdir /run/bird && apt-get update && apt-get install -y bird2
 
 WORKDIR /root
 
-CMD [ "bird", "-c", "/etc/bird/bird.conf", "-d" ]
+# Wait for netlab initial to execute bound /etc/config scripts, then start BIRD.
+RUN cat >/entrypoint.sh <<'ENTRYPOINT'
```
ipspace, line 12: Embedding the same script into three Dockerfiles is a perfect recipe for reusable code :(( More later...

netsim/daemons/bird/vlan.j2
```diff
@@ -0,0 +1,4 @@
+{% include 'vlan/frr.j2' +%}
+{% if 'vxlan' in module|default([]) %}
```
ipspace, line 2: Why do you need this dirty hack???

netsim/daemons/bird/vxlan.j2
```diff
@@ -0,0 +1,5 @@
+{% if vxlan.flooding|default('') == 'evpn' %}
```
ipspace, line 1: If this thing generates content only when EVPN is enabled, don't you think it belongs to the EVPN config?

netsim/daemons/bird.yml
```diff
@@ -1,11 +1,14 @@
 ---
 description: BIRD Internet Routing Daemon
+parent: linux
```
ipspace, line 3: Why exactly do you need this?

netsim/daemons/bird.yml
```diff
@@ -1,11 +1,14 @@
...
+role: router
```
ipspace, line 4: And why have you changed the default role?

netsim/daemons/bird.yml
```diff
@@ -1,11 +1,14 @@
...
 group_vars:
   netlab_import_map:
     bgp: RTS_BGP
     ospf: RTS_OSPF
     connected: RTS_DEVICE
     static: RTS_STATIC_DEVICE,RTS_STATIC
+  netlab_initial: always
```
ipspace, line 11: Isn't this relevant only with configuration reload?

netsim/daemons/bird.yml
```diff
@@ -16,20 +19,22 @@ daemon_config:
...
   ospf: /etc/bird/ospf.mod.conf
+  vxlan: /etc/bird/vxlan.mod.conf
+  evpn: /etc/bird/evpn.mod.conf
   routing: /etc/bird/routing.mod.conf
 clab:
   group_vars:
     netlab_show_command: [ birdc, 'show $@' ]
     docker_shell: bash -il
-    netlab_config_mode: sh
+    netlab_config_mode: ns
```
ipspace, line 29: Isn't this inherited from Linux?

ipspace (changes requested):
I've seen hacks before, but this one is a winner. Including VXLAN configs into VLAN script, and then generating EVPN configs in VXLAN template -- seriously?

You've stretched the definition of what a daemon is way beyond the original intent. No problem with that. However, now we have to figure out how to do data-plane configuration before the daemon is started. Reusing the same code in three Dockerfiles, templates that configure something other than what they're supposed to do, and quick hacks like your "config-done" sort-of plugin are not exactly a stellar first implementation of that concept.

I don't have much confidence in your AI friend generating a decent solution to a problem we obviously have a hard time defining. Please leave the PR as-is; I'll try to figure out a nicer way to make things work.

netsim/daemons/bird.yml
jbemmel: See #3499 as to what triggered it - I agree it would be better to fix that specific test

It could make sense that a "Routing Daemon" would have a default "router" role - but we can leave it out

netsim/daemons/bird.yml
jbemmel: 'linux' is the implicit default so it's not strictly needed

netsim/daemons/bird.yml
ipspace: Although it's not defined (precisely) anywhere in the documentation, the daemon devices were designed to provide control-plane-only functionality (for example, BGP RR). That's also the most common BIRD use case (IXP BGP route server)

netsim/daemons/bird.yml
ipspace: > 'linux' is the implicit default so it's not strictly needed

So why did you add it?

netsim/daemons/bird.yml
jbemmel: no, but individual ```clab.node_config``` entries are

ipspace: > no, but individual `clab.node_config` entries are

So???

netsim/daemons/bird.yml
jbemmel: I initially modeled the Bird config after ```frr```, hence this got copied - but you are right that it currently isn't used

netsim/daemons/bird/Dockerfile
jbemmel: "Plain" Dockerfiles and Jinja2 templates get built differently, making it harder to implement a 'universal' wait mechanism. Including the same bash script in each was the easiest way to get the consistency that was asked for - but I agree a more general core mechanism would be nicer

netsim/daemons/bird.yml
jbemmel: > Although it's not defined (precisely) anywhere in the documentation, the daemon devices were designed to provide control-plane-only functionality (for example, BGP RR). That's also the most common BIRD use case (IXP BGP route server)

ok, so in my mind that is a "router" role more than a "host" role. Looking at the codebase, I now realize that Linux daemons are explicitly exempted from requiring a loopback interface - is that the more common way BIRD gets deployed?

netsim/daemons/bird/vxlan.j2
jbemmel: It does fit better in evpn.j2, which already has the check

netsim/daemons/bird/vlan.j2
jbemmel: Because daemons have 2 types of template files: Bash scripts and config file snippets

```vxlan.j2``` is already mapped as a config snippet, so it cannot also be a Bash script. This was a workaround to get both without core changes

If we move the logic from ```vxlan.j2``` into ```evpn.j2```, it frees up that file to include ```vxlan/frr.j2``` instead and we can remove this

netsim/daemons/bird/bgp.j2
jbemmel: It's wrong, I'll remove it

jbemmel:
Replaced by #3509 and #3548

PR #3478, opened 2026-06-14
Add Router OS 7 VXLAN / EVPN support

snuffy22:
Add the ability to use VXLAN and VXLAN + EVPN

Tested with the `netlab-examples` repository

Seems to work for the following (when as the router):
* VXLAN/vxlan-router-stick
* VXLAN/vxlan-bridging
* EVPN/vxvlan-bridging

**NOTE**: I have not done other extensive regressions for other parts of the stack

PR #3480, opened 2026-06-14
Allow Router OS 7 post 7.18

snuffy22:
I have made a bunch of minor changes to allow for 7.21.4  to work.

Was not sure how to handle the difference between versions so older stuff was just commented out.

Have yet to see if the newer style also works on the older revision.

### DIFF
--- a/netsim/ansible/templates/bgp/routeros7.j2
+++ b/netsim/ansible/templates/bgp/routeros7.j2
@@ -1,10 +1,15 @@
 {% import "routeros7.macro.j2" as bgpcfg %}
 
 /routing/bgp/template add name=main as={{ bgp.as }} output.network=bgp-networks
+{# Needed for > 7.18 #}
+/routing/bgp/instance add name=main as={{ bgp.as }}
 
 {% if bgp.router_id is defined %}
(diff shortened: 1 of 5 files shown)


ssasso:
Imho we can forget about old version and keep only the newer ones, clearly stating this in the caveats documents.

ssasso (approved):

ipspace (changes requested):
Most importantly: thanks a million!

However, this one seems to overlap with #3478. If that's the case (to keep the code history clean), pretty please:

* Submit a new PR with just the BGP and image changes (no EVPN/VXLAN/feature flags)
* Once we merge that, rebase the EVPN/VXLAN PR, and then we'll merge that one.

I can do the split/new PR if needed.

snuffy22:
I see what you mean.. 
Will remove the EVPN/VXLAN from this one (feel like it snuck in by accident)

snuffy22:
> I see what you mean.. Will remove the EVPN/VXLAN from this one (feel like it snuck in by accident)

I was using the github online UI it appears when I branched it did off the wrong one (not dev master)

snuffy22:
Updated please see newer one #3485

PR #3482, opened 2026-06-14
Add FortiOS BGP support

a-v-popov:
This PR makes FortiOS BGP-capable (it was OSPF/static only):

* **bgp module**: IPv4/IPv6 unicast EBGP/IBGP, route reflection, prefix
  origination (`config network` / `config network6`), community propagation,
  and route redistribution (`redistribute` / `redistribute6`).
* **bgp.session plugin**: password, passive, timers, and graceful restart
  (`bgp.gr`, per the model merged in #3467).

Design notes:

* FortiOS treats each numbered neighbor address as its own session; a
  dual-stack neighbor yields two `edit` blocks, and a session is emitted only
  for an activated address family (an `edit` block with the family disabled
  would still reach Established with no negotiated AF, which does not match
  netlab's separate-session-per-AF model).
* Cross-AF next hops (RFC 8950), unnumbered/LLA sessions, local-as, and
  confederations are intentionally not declared in `features.bgp`, so netlab
  rejects topologies that need them with a clear error. VRF and EVPN support
  is deferred.
* Graceful restart mapping was verified live on FortiOS 8.0.0: the
  per-neighbor `capability-graceful-restart` knob drives all protocol-visible
  behaviour, while the global `graceful-restart` switch only preserves the
  kernel FIB across a routing restart. Any GR-enabled neighbor therefore
  forces the global switch on. FortiOS has no helper-only mode: `helper` is
  rejected per neighbor via `features.bgp.gr.valid` and ignored at the node
  level (documented in `docs/caveats.md`).
* The FortiOS GR timer range is 1-3600 s, narrower than the schema's 0-3600;
  a `bgp.gr.restart_time` of 0 passes schema validation but is rejected by the
  device, so the device quirk `check_bgp_gr_restart_time` rejects it at
  transform time (covered by `tests/errors/fortios-bgp-gr-restart-time`).
  If features declaration and validation allowed specification of min/max values
  this would not be necessary, but I couldn't find any option to do so.
* All bgp.session attributes are rendered in the `bgp/fortios.j2` module
  template; an empty `extra/bgp.session/fortios.j2` satisfies the plugin's
  deployment lookup (the `openbsd.j2` precedent). This keeps each neighbor's
  configuration in a single `edit` block and works for node-level-only
  `bgp.gr` (the FGCP HA-cluster pattern), which schedules the snippet via
  `check_node_attributes()`.
* Because the single-edit-block approach makes `extra/bgp.session/fortios.j2`
  render empty, `deploy-config/fortios.yml` now renders the template once into
  a fact and gates both the config-script API push and the CLI fallback on
  non-empty content, instead of deploying an empty configuration. **This guard
  applies to every FortiOS configuration action, not only BGP.**

## Test evidence

Live-verified on real FortiOS (libvirt) on 2026-06-14, running the PR-branch
code: **all 14 transformation-eligible BGP integration topologies passed
`netlab validate`**, one lab at a time, with per-test licensing and teardown.

| Test | Checks |
| --- | --- |
| bgp/01-ebgp-session | 6 |
| bgp/02-ibgp-ebgp-session | 10 |
| bgp/03-ibgp-rr | 26 |
| bgp/04-originate | 30 |
| bgp/05-community | 14 |
| bgp/11-ipv6-ebgp | 6 |
| bgp/12-ipv6-ibgp-ebgp | 10 |
| bgp/13-ipv6-ibgp-rr | 19 |
| bgp/14-ipv6-originate | 24 |
| bgp/20-dual-stack-activate | 17 |
| bgp/30-import-ds-ospf | 17 |
| bgp.session/06-passive | 6 |
| bgp.session/07-password | 5 |
| bgp.session/10-timers | 9 |

The remaining BGP/bgp.session topologies are correctly rejected at transform
time because FortiOS does not declare the feature: unnumbered/RFC 8950
(06/16/21/22/24), local-as (07/08/09), confederation (40), route-server
(11-rs/12-rs), import-policy route-maps (31), and the unsupported bgp.session
attributes allowas-in, as-override, default-originate, gtsm, remove-private-as,
tcp-ao, bfd, rs-client.

Negative checks:

* Per-neighbor `bgp.gr: helper` on FortiOS fails with `IncorrectValue in
  bgp.session: ... does not support BGP attribute gr value helper`.
* `bgp.gr.restart_time: 0` fails at transform time
  (`tests/errors/fortios-bgp-gr-restart-time`).

Gates: `mypy -p netsim` clean (309 files); `ruff check` clean;
transformation/error suites pass (`pytest -k error_cases`: 113 passed).

Note: the canonical `tests/integration/bgp.session/14-gr.yml` cannot run
against FortiOS as written — it bundles a VRF segment (FortiOS has no VRF
support yet) and a per-neighbor `helper` link (correctly rejected by
`features.bgp.gr.valid`). The GR mapping itself was live-verified separately on
FortiOS 8.0.0. The GR column in the
platform table therefore carries a caveat marker.


netsim/ansible/templates/bgp/fortios.j2
```diff
@@ -0,0 +1,157 @@
...
+    blocks below) drives everything protocol-visible -- capability advertisement,
+    helper behaviour, the R-bit cycle -- while this global switch only preserves the
+    kernel FIB across a routing restart. A neighbor capability without the global
+    switch would advertise an F-bit the dataplane cannot honour, so any enabled
+    neighbor forces the switch on, even when the node-level state is 'disable'.
+
+    FortiOS accepts the GR timers only in the 1-3600 range; the netlab schema allows
+    0, which fails at deployment.
+#}
+{% set gr_enabled_neighbors = bgp.neighbors|default([])|selectattr('gr','defined')|selectattr('gr','eq','enable')|list %}
```
ipspace, line 35: FWIW (just a remark), you could have set this flag in the device quirk if you don't like long Jinja2 pipelines.

netsim/ansible/templates/bgp/fortios.j2
```diff
@@ -0,0 +1,157 @@
...
+{#         Activate the address family matching this session's transport, disable the other.
+           The loop above guarantees the matching family is the activated one. #}
+{%       if af == 'ipv4' %}
+            set activate enable
+            set activate6 disable
+{%       else %}
+            set activate disable
+            set activate6 enable
+{%       endif %}
+{%       if n.next_hop_self|default(False) %}
```
ipspace, line 86: Another minor remark: you could have set a variable and used it in subsequent commands.

netsim/ansible/templates/bgp/fortios.j2
```diff
@@ -0,0 +1,157 @@
...
+        edit {{ loop.index }}
+            set prefix6 {{ pfx.ipv6 }}
+        next
+{% endfor %}
+    end
+
+{#  Redistribution into BGP. FortiOS keeps separate IPv4 (redistribute) and IPv6
+    (redistribute6) tables; emit both for each imported protocol so dual-stack import
+    works. netlab proto names match FortiOS keywords; under redistribute6 "ospf" is OSPFv3. #}
+{% for proto in bgp.import|default({}) %}
```
ipspace, line 145: You might want to check the node AFs, but if it works ;)

ipspace (approved):
I can't test it 🤷🏻‍♂️😞, but it looks perfect. I only made a few tiny remarks, which are more of a personal preference than anything else.

The only change I made: as some of us run transformation/error tests on every push, I moved the FortiOS-specific quirk test into the "coverage" part of the tests.

Thanks a million!

PR #3486, opened 2026-06-15
Fix vMX SSH-readiness timeout regression (#3483)

### LINKED ISSUE #3483: [BUG] Upgrading netlab from 25.11 to 26.6 breaks vmx
### Describe the bug

After upgrading netlab from version 25.11 to 26.6, all vmx fail to complete netlab up. The topology (~20 nodes) times out under `netlab up` while waiting for SSH on the vmx devices to become ready.

### To Reproduce

```
pip3 -m pip install netlab==25.11
netlab up
# No longer works
pip3 -m pip upgrade netlab==26.6
netlab up
```

### Expected behavior

netlab up should work for vmx ... still.

### Lab topology

```
provider: clab
defaults.device: vmx
nodes:
  vr1:
  vr2:
links:
- vr1-vr2

```

### Output

```
[INFO]    Checking SSH server(s) on vr1,vr2,vr3,vr4,vr5,vr6,vr7,vr8,vr9,vrr1,vrr2,vce1,vce2,vce3,vce4,vce5,vce
          6,vce7,vce8,vce9,vce10,vce11
[SSH]     SSH server on node vr4 (device vmx) is not ready after 102.5 seconds
[SSH]     SSH server on node vr3 (device vmx) is not ready after 102.5 seconds
[SSH]     SSH server on node vr1 (device vmx) is not ready after 102.5 seconds
...
[SSH]     SSH server on node vix2 (device vmx) is not ready after 102.5 seconds
[SSH]     SSH server on node vtransit (device vmx) is not ready after 102.6 seconds
[FATAL]   initial: SSH server did not start in time on node(s) vr1,vr2,vr3,vr4,vr5,vr6,vr7,vr8,vr9,vrr1,vrr2,v
          ce1,vce2,vce3,vce4,vce5,vce6,vce7,vce8,vce9,vce10,vce11
[FATAL]   netlab up: netlab initial failed, aborting...
```

### Version

```
sudo containerlab version upgrade

~/.../inter-as/option-c$ netlab version
netlab version 26.06
  executable location: /home/foo/.local/bin/netlab
  package location: ['/home/foo/.local/lib/python3.10/site-packages/netsim']
  user defaults: ['/home/hstfadmin/.netlab.yml']

Required packages:
  python-box: 7.3.2
  Jinja2: 3.1.5
  PyYAML: 6.0.2
  netaddr: 1.3.0
  requests: 2.25.1
  filelock: 3.17.0
  rich: 13.9.4

Optional packages:
  ruamel.yaml: 0.18.10

Ansible components:
  ansible: 10.7.0
  ansible-core: 2.17.9
  ansible-pylibssh: 1.2.2
```

### Additional context

Is there a way to override the timeout value? Possible this [PR](https://github.com/ipspace/netlab/pull/2782) broke it?

a-v-popov:
This one was only verified by in memory transform without live tests,
as I don't have any Junos uploaded on netlab server.
What could possibly go wrong?

## Problem

After upgrading from 25.11 to 26.06, vMX nodes fail during `netlab up` with:

```
SSH server on node <name> (device vmx) is not ready after 102.5 seconds
```

## Root cause

Moving the SSH readiness check from the Ansible `until` loop
(`netsim/ansible/tasks/readiness-check/ssh.yml`) to internal Python code
(`netsim/cli/initial/ready.py`, introduced by #3049) changed how
`netlab_check_retries` and `netlab_check_delay` are interpreted. The per-device
values were never altered — only their meaning.

| | 25.11 (Ansible loop) | 26.06 (Python check) |
|---|---|---|
| total budget | ≈ `retries × (per-attempt + delay)` | flat cap `retries × delay` (`ready.py:61`, `ready.py:115`) |
| per-attempt timeout | fixed 10 s (`timeout -k 10s 10s`) | `delay` seconds (`ready.py:98`) |

vMX carried `netlab_check_retries: 20` with the default `delay` of 5. Under the
old semantics that allowed roughly 250–300 s of wall-clock time; under the new
flat cap it yields exactly `20 × 5 = 100 s`, which is below the boot time of a
full vMX (VCP + VFP). vMX was the only Juniper virtual device whose effective
budget fell close enough to its real boot time to start failing — its tuning
was anomalously low relative to its siblings (vsrx/vjunos-router at 60/10,
vjunos-switch/vptx at 40/10).

## Fix

- Establish a **60/10 readiness floor (600 s)** in `junos.yml` top-level
  `group_vars`, so the whole Juniper virtual family inherits a sane budget.
- Remove vMX's stale `netlab_check_retries: 20` so it picks up that floor.
- Remove the now-redundant `60/10` overrides from `vsrx` and `vjunos-router`;
  they continue to resolve to 60/10 through inheritance.
- `vjunos-switch` and `vptx` deliberately set `40/10` in their `clab` blocks,
  which still overrides the floor, so they are left unchanged.

Raising vMX's `delay` to 10 s also restores a 10 s per-attempt probe timeout,
matching the old `timeout -k 10s 10s` behaviour rather than halving it.

## Provider paths

These timers only govern the **clab** readiness path. The two providers differ:

- **clab (vrnetlab):** `netlab_ready: [ ssh, ansible ]`. The internal SSH check
  in `ready.py` runs and consumes `netlab_check_retries`/`netlab_check_delay` —
  this is the path that regressed and the path this PR fixes.
- **libvirt (native VM):** `netlab_ready: [ ansible ]`. The internal SSH check
  does not run, so `netlab_check_*` are resolved but never consulted. Readiness
  is governed solely by `tasks/readiness-check/junos.yml`, which hardcodes
  `interval: 5`, `retries: 60` (≈300 s). The libvirt path was identical in
  25.11 and 26.06 and was never affected by the regression.

Two notes follow. First, vMX is **clab-only**: neither `vmx.yml` nor the
`junos` parent defines a `libvirt` block, so vMX runs exclusively under
vrnetlab — exactly the reporter's path. Second, under libvirt `vjunos-switch`
and `vptx` resolve to the 60/10 floor rather than their `clab`-block 40/10,
but since the timers are unused on that path the difference is immaterial.

## Verification

Transformed in-memory topologies and resolved the values through the same
`get_node_group_var` path that `ready.py` uses.

**clab** (one node per device) — the path that governs the SSH check:

| node | device | retries | delay | wait | status |
|---|---|---|---|---|---|
| r_vmx | vmx | 60 | 10 | **600 s** | inherits junos floor (was 100 s) |
| r_vsrx | vsrx | 60 | 10 | 600 s | inherits junos (override removed) |
| r_vjr | vjunos-router | 60 | 10 | 600 s | inherits via vmx → junos (override removed) |
| r_vjs | vjunos-switch | 40 | 10 | 400 s | own clab override still wins |
| r_vptx | vptx | 40 | 10 | 400 s | unchanged |

**libvirt** (vMX excluded — no libvirt support) — confirms the SSH check is
not active and the timers are not consulted:

| node | device | netlab_ready | (timers unused) |
|---|---|---|---|
| r_vsrx | vsrx | `[ ansible ]` | readiness via hardcoded junos check |
| r_vjr | vjunos-router | `[ ansible ]` | readiness via hardcoded junos check |
| r_vjs | vjunos-switch | `[ ansible ]` | readiness via hardcoded junos check |
| r_vptx | vptx | `[ ansible ]` | readiness via hardcoded junos check |

`yamllint` passes on all four changed files.

Fixes #3483.

### DIFF
--- a/netsim/devices/junos.yml
+++ b/netsim/devices/junos.yml
@@ -14,6 +14,8 @@ group_vars:
   ansible_connection: netconf
   netlab_console_connection: ssh
   netlab_ready: [ ansible ]
+  netlab_check_retries: 60
+  netlab_check_delay: 10
 
 features:
   initial:
--- a/netsim/devices/vjunos-router.yml
+++ b/netsim/devices/vjunos-router.yml
@@ -32,7 +32,5 @@ clab:
   group_vars:
     ansible_user: admin
     ansible_ssh_pass: admin@123
-    netlab_check_retries: 60
-    netlab_check_delay: 10
 
 graphite.icon: router
--- a/netsim/devices/vmx.yml
+++ b/netsim/devices/vmx.yml
@@ -9,7 +9,6 @@ group_vars:
   ansible_user: admin
   ansible_ssh_pass: "admin@123"
   netlab_device_type: vmx
-  netlab_check_retries: 20
   netlab_license_url: https://www.juniper.net/content/dam/www/assets/dm/us/en/E421992502.txt
 
 features:
--- a/netsim/devices/vsrx.yml
+++ b/netsim/devices/vsrx.yml
@@ -44,7 +44,5 @@ clab:
   group_vars:
     ansible_user: admin
     ansible_ssh_pass: "admin@123"
-    netlab_check_retries: 60
-    netlab_check_delay: 10
 
 graphite.icon: firewall


ipspace:
@a-v-popov -- so you're effectively saying "we're assuming Junos is ridiculously slow unless proven otherwise" 😂 I have no problem with that.

Any thoughts @ssasso?

a-v-popov:
> so you're effectively saying "we're assuming Junos is ridiculously slow unless proven otherwise"

No, not even close. Junos provide proper dataplane and boot emulation, which actually makes testing on their virtual appliances meaningful, unlike most other platforms. The price is somewhat slower boot time, but it is only visible to people who adopted SW developers throwaway VM approach which is completely inadequate for any descent network OS, so it is OK to make them suffer.

With this PR I am merely suggesting to use single source of timers for all Junos types for clab provider, unless somebody want to go and actually test the relative boot time of different Junos types. I would say that libvirt should also use the same timers, but currently it works, so I don't want to touch it.

ipspace:
> I would say that libvirt should also use the same timers, but currently it works, so I don't want to touch it.

Vagrant checks for SSH server reachability before declaring the VM operational, so we don't need to do the SSH checks there (and that's why they're usually declared as **clab.group_vars**). Using them with Vagrant is not too harmful, though.

ipspace (changes requested):
Joking aside, I looked at the bigger picture:

* Even though the code is still there, netlab is no longer using Ansible-based SSH checks
* The readiness checks are driven by the "netlab_ready" group variable, which is set to "ansible" for Junos devices (to check interface status) and "ansible, ssh" for Junos clab devices (because containerlab does not check the SSH server)
* The "netlab_check_delay" and "netlab_check_retries" are only relevant for SSH tests on Junos devices. The Junos "are interfaces there" checks have hard-coded values.

Based on that, the "netlab_check_delay" and "netlab_check_retries" variables are only relevant for Junos clab devices, so I would suggest we set them only in **clab.group_vars**

What do you think @a-v-popov @ssasso

a-v-popov:
I am not using clab much, so I have no real stakes in this.
I think one can argue both ways.

Do the values correspond to an image, or to a provider?
In Junos case they are always VMs so they more related to platform itself than the provider.
When Vagrant is ditched and replaced with something adequate these will likely be be global vars.
 
And here is a little present from one of my AI friends, post-PR numbers.
Pre-PR it would instead be vmx in the global row, and vsrx + vjunos-router in the clab row — 11 clab vs 6 global.

| Placement                                  | Count | Devices                                                                  |
| ------------------------------------------ | ----- | ----------------------------------------------------------------------- |
| **Provider block** (`<provider>.group_vars`) | 9     | arubacx, cisco8000v, dellos10, exos, iol, ios, routeros7, vjunos-switch, vptx |
| **Global** (top-level `group_vars`)        | 6     | asav, crpd, csrx, junos, nxos, openbsd                                   |

ipspace:
> And here is a little present from one of my AI friends, post-PR numbers. Pre-PR it would instead be vmx in the global row, and vsrx + vjunos-router in the clab row — 11 clab vs 6 global.

Thank you -- that's the showstopper. Either we have _everything_ Junos-related under `clab` or we'll face nasty inheritance challenges. OK, let's go with global settings.

ipspace:
Totally off-topic:

> When Vagrant is ditched and replaced with something adequate these will likely be be global vars.

We might ditch Vagrant, but I don't think we'll find an adequate replacement. It makes no sense to reinvent that wheel, so it would have to do everything Vagrant does for us.

I think it's more realistic that we move everything to containerlab, assuming they get _vrnetlab_ to the point where all the containers it builds are equivalent to Vagrant boxes (snapshot after the initial image baking is done).

PR #3490, opened 2026-06-16
Add OSPF graceful restart config with initial support for FRR

jbemmel:
## Summary - a minimal proposal to add GR support for OSPF as part of ospf module

- Add `ospf.gr` to the OSPF module with `state` (`enable`, `disable`, `helper`) and optional `grace_period` (BGP-style string shorthand supported)
- Render FRR OSPF/VRF graceful restart configuration
- Add OSPFv2 integration test `41-gr.yml` and minimal documentation
- updated AGENTS.md based on feedback, hoping to improve future PRs

Addresses https://github.com/ipspace/netlab/discussions/3488

## Test plan

- [x] `netlab up tests/integration/ospf/ospfv2/41-gr.yml` with `NETLAB_DEVICE=frr NETLAB_PROVIDER=clab`
- [x] `netlab validate` — adjacency and helper GR checks on FRR probes

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,8 +91,6 @@ With the Box settings used in the topology data structure:
 
 * The interim dictionaries are created automatically: `a.b.c=1` works even when `a.b` does not exist yet.
 * The dotted paths can be used in indices (for example, `a['b.c'] = 1`) and in **in** tests (for example `'b.c' in a`)
-* Prefer `box.get('dotted.path', default)` over nested existence checks like `'gr' in box and box.gr.get('state', ...)`
-* When testing whether an optional feature is active, default to "off" (e.g. `'disable'`), not to an enabled state. Do not add `.get()` defaults for `_required` attributes — validation guarantees they are set when the parent object is present.
 
 IMPORTANT -- the Box objects must have the default_box and box_dots flags set to work as expected:
 
@@ -119,7 +117,6 @@ IMPORTANT -- the Box objects must have the default_box and box_dots flags set to
 - YAML files for configuration and topology definitions
 - Jinja2 templates for configuration generation
 - Global settings in `netsim/defaults/`
-- In module/plugin YAML (`netsim/modules/*.yml`, `netsim/extra/*/defaults.yml`), define user-defined attribute types only when the same structure is reused in multiple places. Inline `{ type: ..., valid_values: ... }` (and similar constraints) for single-use attributes instead of adding a named type in `_top` or `attributes`. See `docs/dev/validation.md` (User-Defined Data Types).
 
 ### Documentation
 - Docstrings follow Google style (not strictly enforced but preferred)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -91,6 +91,8 @@ With the Box settings used in the topology data structure:
 
 * The interim dictionaries are created automatically: `a.b.c=1` works even when `a.b` does not exist yet.
 * The dotted paths can be used in indices (for example, `a['b.c'] = 1`) and in **in** tests (for example `'b.c' in a`)
+* Prefer `box.get('dotted.path', default)` over nested existence checks like `'gr' in box and box.gr.get('state', ...)`
+* When testing whether an optional feature is active, default to "off" (e.g. `'disable'`), not to an enabled state. Do not add `.get()` defaults for `_required` attributes — validation guarantees they are set when the parent object is present.
 
 IMPORTANT -- the Box objects must have the default_box and box_dots flags set to work as expected:
 
@@ -117,6 +119,7 @@ IMPORTANT -- the Box objects must have the default_box and box_dots flags set to
 - YAML files for configuration and topology definitions
 - Jinja2 templates for configuration generation
 - Global settings in `netsim/defaults/`
+- In module/plugin YAML (`netsim/modules/*.yml`, `netsim/extra/*/defaults.yml`), define user-defined attribute types only when the same structure is reused in multiple places. Inline `{ type: ..., valid_values: ... }` (and similar constraints) for single-use attributes instead of adding a named type in `_top` or `attributes`. See `docs/dev/validation.md` (User-Defined Data Types).
 
 ### Documentation
 - Docstrings follow Google style (not strictly enforced but preferred)
--- a/netsim/ansible/templates/ospf/frr.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/frr.ospfv2.j2
@@ -15,7 +15,10 @@ router ospf
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
 {% endif %}
-{% include 'frr.gr.j2' %}
+{% if ospf_data.gr is defined %}
+ capability opaque
+{%   include 'frr.gr.j2' %}
+{% endif %}
  timers throttle spf 10 50 500
  timers throttle lsa all 100
  timers lsa min-arrival 100
--- a/netsim/ansible/templates/ospf/frr.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/frr.ospfv2.j2
@@ -15,8 +15,8 @@ router ospf
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
 {% endif %}
-{% if ospf_data.gr is defined and ospf_data.gr.state|default('enable') != 'disable' %}
-{%   set gr_state = ospf_data.gr.state|default('enable') %}
+{% if ospf_data.get('gr.state','disable') != 'disable' %}
+{%   set gr_state = ospf_data.gr.state %}
 {%   if gr_state == 'enable' %}
 {%     if ospf_data.gr.grace_period is defined %}
  graceful-restart grace-period {{ ospf_data.gr.grace_period }}
--- a/netsim/ansible/templates/ospf/frr.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/frr.ospfv2.j2
@@ -15,6 +15,21 @@ router ospf
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
 {% endif %}
+{% if ospf_data.gr is defined and ospf_data.gr.state|default('enable') != 'disable' %}
+{%   set gr_state = ospf_data.gr.state|default('enable') %}
+{%   if gr_state == 'enable' %}
+{%     if ospf_data.gr.grace_period is defined %}
+ graceful-restart grace-period {{ ospf_data.gr.grace_period }}
+{%     else %}
+ graceful-restart
+{%     endif %}
+{%   elif gr_state == 'helper' %}
+ graceful-restart helper enable
+{%     if ospf_data.gr.grace_period is defined %}
+ graceful-restart helper supported-grace-time {{ ospf_data.gr.grace_period }}
+{%     endif %}
+{%   endif %}
+{% endif %}
  timers throttle spf 10 50 500
  timers throttle lsa all 100
  timers lsa min-arrival 100
--- a/netsim/ansible/templates/ospf/frr.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/frr.ospfv2.j2
@@ -15,7 +15,7 @@ router ospf
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
 {% endif %}
-{% if ospf_data.get('gr.state','disable') != 'disable' %}
+{% if ospf_data.gr.state|default('disable') != 'disable' %}
 {%   set gr_state = ospf_data.gr.state %}
 {%   if gr_state == 'enable' %}
 {%     if ospf_data.gr.grace_period is defined %}
--- a/netsim/ansible/templates/ospf/frr.ospfv2.j2
+++ b/netsim/ansible/templates/ospf/frr.ospfv2.j2
@@ -16,14 +16,12 @@ router ospf
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
 {% endif %}
 {% if ospf_data.gr is defined %}
-{%   if ospf_data.gr.restart is true or ospf_data.gr.restart.grace_period is defined %}
- graceful-restart{{ " grace-period " ~ ospf_data.gr.restart.grace_period if ospf_data.gr.restart.grace_period is defined else "" }}
+{%   if ospf_data.gr.restart.grace_period is defined %}
+ graceful-restart grace-period {{ ospf_data.gr.restart.grace_period }}
 {%   endif %}
-{%   if ospf_data.gr.helper is true or ospf_data.gr.helper.grace_period is defined %}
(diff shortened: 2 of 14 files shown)


netsim/modules/ospf.yml
```diff
@@ -1,6 +1,13 @@
 # OSPFv2/OSPFv3 default settings and attributes
 #
 ---
+_top:
```
ipspace: There is no need to define a new type that's only used once.

netsim/modules/ospf.py
```diff
@@ -139,6 +139,23 @@ def adjust_interface_timers(node: Box) -> None:
         module='ospf',
         category=log.IncorrectValue)
 
+def check_gr_support(node: Box, features: Box) -> None:
+  for (o_data,_,vrf) in _ospf.rp_data(node,'ospf'):
+    if 'gr' not in o_data or o_data.gr.get('state','enable') == 'disable':
```
ipspace: You know we can do o_data.get('gr.state','disable'). Also, there's no way a _required value would need a default value, and in any case, 'enable' is a bad default value. AI? 😉

netsim/modules/ospf.py
```diff
@@ -139,6 +139,23 @@ def adjust_interface_timers(node: Box) -> None:
...
+      continue
+
+    vrf_info = f' in VRF {vrf}' if vrf else ''
+    if 'ipv4' not in o_data.af:
```
ipspace: When you decide to add a new core feature PRETTY PLEASE look beyond the device you're interested in at that moment. Other devices support OSPF GR for IPv6, so we need to do per-AF checks.

netsim/devices/frr.yml
```diff
@@ -146,6 +146,7 @@ features:
     unnumbered: true
     import: [ bgp, ripv2, isis, connected, static, vrf ]
     default: true
+    gr: true
```
ipspace: This could be a list of supported AFs

netsim/ansible/templates/ospf/frr.ospfv2.j2
```diff
@@ -14,6 +14,21 @@ router ospf
 {% endif %}
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
+{% endif %}
+{% if ospf_data.gr is defined and ospf_data.gr.state|default('enable') != 'disable' %}
```
ipspace: Too complex, you could simply check ospf_data.gr.state is defined or its default value. Also 'enable' is not a good default value, and you defined this attribute as '_required' and so on...

ipspace (changes requested):
Good idea, needs a bit of a cleanup.

netsim/modules/ospf.py
jbemmel: Even FRR does support it, I was focusing on a minimal PR to get started - thinking ipv6 support could be added later. I can include it from the start.

It's not that I'm particularly interested in FRR, it's just an easy Proof-of-Concept device

I'll change global feature flag ```ospf.gr``` to be ```bool``` or a list of supported AF

netsim/devices/frr.yml
jbemmel: ok

netsim/modules/ospf.py
jbemmel: ok

netsim/modules/ospf.yml
jbemmel: ok

netsim/ansible/templates/ospf/frr.ospfv2.j2
jbemmel: Fixed

netsim/ansible/templates/ospf/frr.ospfv2.j2
```diff
@@ -14,6 +14,21 @@ router ospf
 {% endif %}
 {% if ospf_data.reference_bandwidth is defined %}
  auto-cost reference-bandwidth {{ ospf_data.reference_bandwidth }}
+{% endif %}
+{% if ospf_data.get('gr.state','disable') != 'disable' %}
```
jbemmel: I considered moving this logic to a common macro file, but decided it wasn't worth it right now

ipspace (approved):
Code looks decent. The GR config is pretty verbose, but I couldn't find an easy way to streamline it considering FRR commands.

OSPFv2 integration test seems to check whether X2 is properly configured ;), and there's no OSPFv3 integration test. Will try to fix that and then merge.

ipspace (changes requested):
This looked good until I ran the tests and tried to figure out why they don't work :(

Starting with the obvious: when using your template, my version of Jinja2 (3.1.6) didn't render the "graceful restart" commands at all; it looks like Jinja2 wasn't happy with your AI's idea of using `.get()` in a template. Was there a change in Jinja2 behavior, or did your friend forget to look at the rendered files?

Even worse, when using your code, your integration test failed. After my fix, it worked, even though it still tested the wrong thing.

After reading RFC 3623, it seems like there's nothing one could do on adjacent routers to check the GR configuration, so the best we could do would be to render device configs (to test Jinja2 templates).

Finally:

* GR is not a functionality that you would configure. You have to trigger a planned outage from the CLI (or some such), and at that time, the router advertises opaque LSA including the configure **grace period**
* GR helper is configurable, but not advertised.

That means the data model is wrong. The "grace period" (int) and "helper" (bool) should be configurable, but that's it.

AGENTS.md
```diff
@@ -91,6 +91,8 @@ With the Box settings used in the topology data structure:
 
 * The interim dictionaries are created automatically: `a.b.c=1` works even when `a.b` does not exist yet.
 * The dotted paths can be used in indices (for example, `a['b.c'] = 1`) and in **in** tests (for example `'b.c' in a`)
+* Prefer `box.get('dotted.path', default)` over nested existence checks like `'gr' in box and box.gr.get('state', ...)`
```
a-v-popov: I strongly disagree on this one, in my opinion this is one of the worst things in netlab code.

For static analysis and AI tools, the dotted-string form reduces transparency. The field path is hidden inside an arbitrary string, so ordinary type checkers, refactoring tools, “find references”, dead-field detection, and schema-aware linters cannot reliably connect "gr.state" to the gr.state data structure unless they have project-specific knowledge of Box(box_dots=True). AI tools can infer the convention from context, but the inference is weaker than with syntactic member access or typed structures; it is easier to miss a rename, accept a misspelled path, or misunderstand whether the string is a literal key or a nested selector.

Putting this line in AGENTS.md is essentially asking AI tool to create problems for itself.

Also sneaking code styling guidelines on top of an OSPF PR is just wrong.

jbemmel: > Also sneaking code styling guidelines on top of an OSPF PR is just wrong.

This proposal was made in direct response to previous feedback on this PR, intended as a way to improve the consistency of future AI-generated contributions. I am happy to discuss whether this guideline is the right approach for the project, but I would appreciate keeping the feedback focused on the technical merits rather than the presumed intent behind the suggestion.

a-v-popov: Your PR is called "Add OSPF graceful restart support for FRR", which is already misleading as the primary change is the OSPF GR data model. But your pushed AGENT.md changes has no connection to any of these
- OSPF
- GR
- FRR

Since you claimed your PR to be intersection of those, I would argue it is wrong to push your AGENT.md changes on this PR regardless of how you think you came up with those or how good you think your intentions were.
And I don't need to assume any intent behind it, the changes are material part of your commit.

jbemmel:
Since the original PR, this updates the OSPF GR work to:

* Split GR configuration into separate ospf.gr.restart and ospf.gr.helper settings, with boolean shortcuts and separate grace-period validation.
* Generate FRR GR config for both OSPFv2 and OSPFv3 restart/helper modes.
* Replace the basic OSPFv2 helper-config check with route-retention integration tests for both OSPFv2 and OSPFv3.
* Update user and template-developer docs to match the new model.

Sample:
```
ospf.gr:
  restart:
    grace_period: 60
  helper:
    grace_period: 123
```

This allows a node to be configured for local restart only, as a helper, or both. ```ospf.gr: True``` gets normalized to ```ospf.gr.restart: True``` with the device default grace_period.

I'll leave it to you whether you want to include the ```AGENTS.md``` modifications or not, could be removed or submitted as a separate PR if preferred

ipspace:
The "dict or bool" approach results in messy code. At the very least, you should check what happens when you set "gr.restart" to "false" -- does it disappear (in which case, no need for convoluted tests, just check if gr.restart is defined) or not.

Anyway, wouldn't it be simpler to transform "gr.restart: True" (and the same for helper) into "gr.restart.grace_period: x" (either some sane value like 300, or the maximum value allowed by the RFC, 1800)?

AGENTS.md
ipspace: > I strongly disagree on this one, in my opinion this is one of the worst things in netlab code.

Blame me for that. I'm old-school and prefer concise code like "a.b.c|default(None) == 1" over bloated stuff like "a is defined and a.b is defined and a.b.c is defined and a.b.c == 1"

> For static analysis and AI tools, the dotted-string form reduces transparency.

When the AI tools become as reliable as compilers are today, I won't care how convoluted their resulting code is. I don't care about the machine code generated by the Rust or C compiler either.

For the moment, AI tools are as "reliable" as some humans, and I know who will be involved in fixing that code, so I care.

> Putting this line in AGENTS.md is essentially asking AI tool to create problems for itself.

Agreed.

> Also sneaking code styling guidelines on top of an OSPF PR is just wrong.

I couldn't agree more, but I don't know how hard I'd have to yell "DON'T PUT UNRELATED STUFF IN PRs" to have any effect.

a-v-popov: I opened a separate discussion thread https://github.com/ipspace/netlab/discussions/3513

jbemmel:
> The "dict or bool" approach results in messy code. At the very least, you should check what happens when you set "gr.restart" to "false" -- does it disappear (in which case, no need for convoluted tests, just check if gr.restart is defined) or not.

ospf.gr.restart: False remains in the data model, it does not disappear

jbemmel:
> Anyway, wouldn't it be simpler to transform "gr.restart: True" (and the same for helper) into "gr.restart.grace_period: x" (either some sane value like 300, or the maximum value allowed by the RFC, 1800)?

Implemented as suggested, PR updated

ipspace:
Looks good to me ;) apart from two details:

* I hoped we could do the True-to-dict transformation in the validation code with `true_value`. Let me try out a few things...
* Please tell your AI friend it cannot execute validation actions on the tested device 🤦‍♂️ I'm pretty sure you would never make such mistake.

However, you're on to something -- if DUT is willing to be a GR helper, then it will lie about the adjacency. If we add another probe (without GR) and kill both OSPF processes, then we can measure when the second loopback disappears on yet another probe, wait a few more seconds, and check whether the first one is still there.

ipspace:
Hey, I already started working on the changes ;) I promised I'd do them ;))

ipspace:
... the only thing left is the documentation cleanup and integration tests. Working...

jbemmel:
> ... the only thing left is the documentation cleanup and integration tests. Working...

Discovered that FRR needs ```capability opaque``` for OSPFv2:
```
{% if ospf_data.gr is defined %}
 capability opaque
{%   if ospf_data.gr.restart.grace_period is defined %}
 graceful-restart grace-period {{ ospf_data.gr.restart.grace_period }}
{%   endif %}
{%   if ospf_data.gr.helper.grace_period is defined %}
 graceful-restart helper enable
 graceful-restart helper supported-grace-time {{ ospf_data.gr.helper.grace_period }}
{%   endif %}
{% endif %}
```

Without it, the exec command fails (which is not detected by Netlab - opening a PR for that)

ipspace:
> Without it, the exec command fails (which is not detected by Netlab - opening a PR for that)

Already fixed ;) Discovered it while trying to make integration test work.

ipspace:
@jbemmel -- I think we're done. Do you want to have another look at the code before I merge it?

ipspace:
FWIW: no plan survives the first encounter with the reality...

> If we add another probe (without GR) and kill both OSPF processes, then we can measure when the second loopback disappears on yet another probe, wait a few more seconds, and check whether the first one is still there.

Killing the second OSPF process also stops GR helper due to topology change 🤷‍♂️

netsim/modules/ospf.py
```diff
@@ -128,25 +128,47 @@ def adjust_interface_timers(node: Box) -> None:
...
+    if 'gr' not in o_data:
+      continue
+
+    if len(gr_af) == 2:                               # Both IPv4 and IPv6 are supported -> ok
+      continue
+
+    vrf_info = f' in VRF {vrf}' if vrf else ''
+
+    for af in o_data.af:
+      if af not in gr_af:                             # Technically it could be gr.restart: False
```
jbemmel, line 158: This is no longer true...

ipspace: Well, it is, because "ospf.gr.restart: False" results in "_removed_attr" key in "ospf.gr" dictionary (we need that to ensure things further down the line don't overwrite it).

tests/integration/ospf/ospfv2/41-gr.yml
```diff
@@ -0,0 +1,60 @@
...
+    wait_msg: Waiting for SPF to complete
+    wait: ospfv2_spf
+    nodes: [ xp ]
+    plugin: ospf_prefix(nodes.xgr.loopback.ipv4)
+  gr_stop:
+    description: Start planned OSPF graceful restart on XGR
+    nodes: [ xgr ]
+    devices: [ frr ]
+    exec: >-
+      vtysh -c "graceful-restart prepare ip ospf" && kill -STOP $(pidof ospfd)
```
jbemmel, line 53: I encountered a minor issue where ```netlab validate``` would hang indefinitely after killing ospfd.

We may want to add this at the end:
```
gr_resume:
    description: Resume OSPF on X1
    nodes: [ x1 ]
    devices: [ frr ]
    exec: >-
      sh -c 'kill -CONT $(pidof ospfd) || true'
    pass: OSPF daemon resumed on X1
```

ipspace: > I encountered a minor issue where `netlab validate` would hang indefinitely after killing ospfd.

Where exactly did it hang? In the `kill` command? There's no reason it would hang further down the line.

> We may want to add this at the end:

We could 🤷‍♂️ Your choice.

>     exec: >-
>       sh -c 'kill -CONT $(pidof ospfd) || true'

Just wondering: why was 'kill' good enough to kill the process, but here you use 'sh -c'?

jbemmel: over-zealous defensive scripting, it's not needed because Netlab behind the scenes already combines everything into a bash -c command

jbemmel: > Where exactly did it hang? In the `kill` command? There's no reason it would hang further down the line.

It hangs the second time I run ```netlab validate```:
```
jbemmel@jvb-laptop:~/Projects/3rd-party/netlab/tests/integration$ netlab validate 
[adj]         Check OSPF adjacency with DUT [ node(s): xgr,xp ]
(... crickets ...)
```

ipspace: > It hangs the second time I run `netlab validate`:

Ah, of course. You **stopped** the OSPF process, and now vtysh is stuck... Life was so much easier when you were only able to kill (= terminate) processes.

PR #3492, opened 2026-06-16
Add vpp device support

jbemmel:
Summary
Adds VPP as a buildable containerlab device (vpp), similar to BIRD: ```netlab clab build vpp``` produces netlab/vpp:latest, then lab nodes can use it under provider: clab.

This is a rework of the earlier VPP branch. It deliberately does not use Containerlab’s fdio_vpp kind; it uses the same daemon/image pattern as BIRD while keeping comparable dataplane plumbing (VPP + Linux CP in the dataplane netns).

Highlights
* Local image build (netlab clab build vpp, optional --sw-version)
* Router and bridge roles (not host)
* Selectable control plane: bird (default) or frr via defaults.daemons.vpp.control_plane or per-node control_plane
* Initial interface, loopback, MTU, RA, and unnumbered support
* Full VLAN support using native VPP mechanisms
* Docs: docs/labs/vpp.md

initial test results:
```
./device-module-test -d vpp -p clab initial
Pre-test cleanup: 
17:15:44 initial/01-interfaces.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:15:55 initial/01a-scripts.yml: create(FAIL) 
17:15:56 initial/01b-startup-config.yml: create(FAIL) 
17:15:56 initial/02-loopback.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:16:05 initial/03-unnumbered.yml: create(ok) up(ok) config(ok) validate(SKIPPED) cleanup(ok) OK
17:16:07 initial/04-mtu.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:17:01 initial/05-host.yml: create(FAIL) 
17:17:02 initial/06-bridge.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:17:11 initial/07-router.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:17:23 initial/08-ra.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
17:17:59 initial/10-reload.yml: create(FAIL) 
```


netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,59 @@
...
+
+exec /etc/vpp/clab-interfaces.vpp
+
+{% for l in netlab_interfaces %}
+{%   if l.type|default('') == 'loopback' %}
+{%     set lb_instance = l.ifname[4:] %}
+create loopback interface instance {{ lb_instance }}
+{%     set vpp_if = l.ifname %}
+{%   else %}
+{%     set vpp_if = 'host-' ~ l.ifname %}
```
jbemmel, line 37: I tried using ```interface_name: host-eth{ifindex}``` but that breaks the bridge scenario

netsim/augment/nodes.py
```diff
@@ -552,7 +552,7 @@ def cleanup_non_ansible_config(n: Box) -> None:
       # ... also, build netlab_ansible_skip_module list because the module will be configured
       # ... in another way
       #
-      if kn == n.device or kn in n.get('module',[]) or kn in n.get('config',[]) or kn == 'initial':
+      if kn == n.device or kn in n.get('module',[]) or kn in n.get('config',[]) or kn == 'initial' or kn.endswith('.conf'):
```
jbemmel: See #3491 I needed at least 2 daemon config files

netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,62 @@
+#
+# bound to /etc/vpp/setup.conf
```
pimvanpelt: I would not call this file `*.conf` - it is idiomatic to refer to CLI-executable files as `*.vpp`, as you do below while executing clab-interfaces.vpp. I suggest calling it /etc/vpp/config/setup.vpp instead, and bind-mounting /etc/vpp/config/ rather than only setup.conf.

netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,62 @@
...
+#
+
+mpls table add 0
```
pimvanpelt: enabling this requires the VM or container to have kernel module 'mpls_router' and also sysctl `net.mpls.platform_labels` to a non-zero value (default 65535)

netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,62 @@
...
+set ip neighbor-config ip4 age 900
+set ip neighbor-config ip6 age 900
+lcp default netns dataplane
+lcp lcp-sync on
+# lcp lcp-sync-unnumbered on  # not in VPP 26.02-release
```
pimvanpelt: you're right - it was changed /just after/ the cutoff in commit:

```
commit 466fb6da7c82b298e21295d31c87ec893cee0e02
Author: Pim van Pelt <pim@ipng.nl>
Date:   Tue Feb 10 23:16:03 2026 +0100
```

You can probably remove the setting alltogether. What it does is inhibits Linux CP from copying forward the unnumbered IPv4/IPv6 address to the Linux TAP. It is possible/harmless to leave it 'off', in which case VPP will configure the unnumbered IPs on the Linux side as well. The purpose of the flag is to guard things from trying to manipulate those TAP-side IP addresses, as VPP will ignore any add/replace/delete on them if it's configured as an unnumbered interface.

netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,62 @@
...
+{%   if is_loopback %}
+{%     set lb_instance = l.ifname[4:] %}
+create loopback interface instance {{ lb_instance }}
+{%     set vpp_if = l.ifname %}
+{%   else %}
+{%     set vpp_if = 'host-' ~ l.ifname %}
+{%   endif %}
+set interface state {{ vpp_if }} {{ 'down' if l.shutdown|default(False) else 'up' }}
+{% if l.ipv4 is defined and (l.ipv4 is string or l._parent_ipv4 is defined) %}
+set interface ip address {{ vpp_if }} {{ l.ipv4 if l.ipv4 is string else l._parent_ipv4 }}
```
pimvanpelt: Routing protocols will only start to work if you additionally create a Linux Interface Pair (_LIP_) on this interface. Intended order of operations:

1.    Bind veth to dataplane (with `create host-interface`)
2.    Expose dataplane interface with Linux CP (with `lcp create foo host-if bar`)
3.    Use Linux CP netlink listener on 'bar' to sync routes/addresses/links

(1) is what is special in the case of clab or docker hosts, where we offer a veth pair to the container. In bare metal this would be a pre-existing DPDK or RDMA interface. In VMs this would be a pre-existing virtio interface (which will be bound by the DPDK virtio PMD).

Critical detail: Linux CP does not listen to or interact with the 'host-interface' in any way. It only governs interfaces that were created with `lcp create` in the network namespace, typically `dataplane` as you referenced on line 8 in this file.

netsim/daemons/vpp/setup.vpp.j2
```diff
@@ -0,0 +1,62 @@
...
+{%           endif %}
+{%         endif %}
+{%         if l.ra.dhcp|default(false) == 'all' %}
+ip6 nd {{ vpp_if }} ra-managed-config-flag
+{%         endif %}
+{%         if l.ra.dhcp|default(false) == 'other' %}
+ip6 nd {{ vpp_if }} ra-other-config-flag
+{%         endif %}
+{%       endif %}
+{%     elif role == 'host' %}
```
pimvanpelt: In the comparison table in platforms.md you noted a red-cross in the 'Host' column for VPP. I think you should make it a green-checkmark as VPP has a complete host-stack and is often used to accelerate host traffic as well (eg nginx or iperf). See also my note for the platform.md file.

If you do decide to leave out 'host' capabilities, then this branch becomes moot due to your setting in netsim/daemons/vpp.yml L35,L47

docs/platforms.md
```diff
@@ -83,6 +84,7 @@ Most devices behave as routers (or layer-3 switches); the following devices can
 |-----------------------|:--:|:--:| :--:|
 | Arista EOS            | ✅ | ✅ | ✅ |
 | Bird                  | ✅ | ✅ | ❌  |
+| VPP                   | ✅ | ❌  | ✅ |
```
pimvanpelt: Regarding host capabilities: I think you should make it a green-checkmark as VPP has a complete host-stack and is often used to accelerate host traffic as well (eg nginx or iperf). See also my note for the platform.md file.

netsim/augment/nodes.py
pimvanpelt: I don't think you do. startup.conf is a daemon config file, but your envisioned `setup.conf` is a completely different type of file, akin to the 'startup-conf' file in other NOS (the one you manipulate with `write mem` and `show runnung-config`). It's not a daemon config file, it's some form of runtime configuration. See also my critique of its pre-compiled use, which I think is a bad idea.

netsim/daemons/vpp/vpp.j2
```diff
@@ -0,0 +1,40 @@
+unix {
+  interactive
+  nosyslog
+  
+  full-coredump
+  cli-listen /run/vpp/cli.sock
+  cli-prompt {{ inventory_hostname }}-vpp$
+  cli-no-pager
+  poll-sleep-usec 100
+  startup-config /etc/vpp/setup.conf
```
pimvanpelt: In clab, I found that bind-mount'ing /etc/vpp/config/* with the dataplane config was more flexible than referencing only one file. You might want to put this file in /etc/vpp/config/setup.vpp (note the suffix, to disambiguate 'daemon config file' from 'vpp runtime configuration').

netsim/daemons/vpp.yml
```diff
@@ -0,0 +1,50 @@
...
+virtualbox:
+  image:
+features:
+  initial:
+    ipv6:
+      lla: true
+      use_ra: true
+    ra: true
+    reload: false
+    roles: [ router, bridge ]
```
pimvanpelt, line 69: and host?

netsim/daemons/vpp.yml
```diff
@@ -0,0 +1,50 @@
...
+  build: True
+  # node:
+  #  kind: fdio_vpp
+  sw_version: # Marks versioning support, latest release is used by default
+  repo: release
+  mtu: 1500
+  features:
+    initial:
+      config_mode: [ sh ]
+      roles: [ router, bridge ]
```
pimvanpelt, line 51: and host?

pimvanpelt (changes requested):
I love this! I left a bunch of nits and comments.

The one thing I think you should not do, is autogen a `setup.conf` or `setup.vpp` file. It makes it ~impossible to make any changes to the dataplane configuration while operating the netlab. What if you wanted to add a sub-interface, or remove an address, or make an L2XC between two interfaces, or bind an MPLS tunnel? Your users would end up resorting to 'vppctl' debug CLI, which is cumbersome at best, and perilous at worst, because ordering matters, and the CLI/API do very limited syntax and semantic checking. 

Also, there will be no config persistence if they restart the container. There is no `vppctl show run` or `vppctl write mem`, by design it is a configurable dataplane with an API that is meant to be configured by another system. Users will risk crashing the dataplane if they do things in the wrong order, or with the wrong arguments.

Consider using `vppcfg` and have ansible generate `/etc/vpp/vppcfg.yaml`, and explain to users how to edit that file and run `vppcfg plan` and `vppcfg apply`. When starting the container, you can generate the file you're now hand-crafting with `vppcfg plan --novpp -c vppcfg.yaml -o vppcfg.vpp` and start the dataplane with that instead. I guarantee you that is a better user experience, even if you rely on an external component (you can `pip install` it).

Good luck with the netlab integration!

netsim/daemons/vpp/setup.vpp.j2
jbemmel: Netlab currently does not have an easy way to bind-mount full directories per node, it's individual file based. That could be changed of course, but for now I can definitely call it .vpp to start with

netsim/augment/nodes.py
jbemmel: Agreed, what I meant to say is that I need 2 .j2 template files - they serve different purposes. One can be called ```vpp.j2``` but then the other got removed by Netlab internal logic. I've now found a different way to do this without modifying augment/nodes.py (which we prefer not to do).

netsim/daemons/vpp/setup.vpp.j2
jbemmel: For clab ```mpls-router``` module is automatically loaded when the mpls module is used: https://github.com/ipspace/netlab/blob/f0c554da382d106cc914fe71ba997773ccc4186d/netsim/providers/clab.yml#L26

For FRR we similarly set that sysctl: https://github.com/ipspace/netlab/blob/f0c554da382d106cc914fe71ba997773ccc4186d/netsim/ansible/templates/mpls/frr.j2#L20

jbemmel: Added sysctl to initial script, and moved mpls table command to separate mpls.vpp config script (to be extended)

netsim/daemons/vpp/setup.vpp.j2
jbemmel: ok, I'll leave it commented out for now

netsim/daemons/vpp/setup.vpp.j2
jbemmel: The initial tests only use ICMP which appears to work as expected, appreciate the tip - will revisit when adding BGP/OSPF support

netsim/daemons/vpp/setup.vpp.j2
jbemmel: I initially included 'host' role support, but that triggers a bunch of Linux host scripts to get included which will probably need tweaking (static routes and such). Leaving out 'host' for now is more of a priority call, not  an end state. You're correct that this branch currently doesn't get used - but it will be needed once 'host' role is added back.

docs/platforms.md
jbemmel: Yes, eventually I think we'd also want to enable usage of VPP as a host. The red check mark says "currently doesn't work"

netsim/daemons/vpp.yml
jbemmel: Yes, probably in the near future. Briefly tested but currently fails

netsim/daemons/vpp/vpp.j2
jbemmel: Updated the location, Netlab currently only maps individual files (one per config module)

jbemmel:
> I love this! I left a bunch of nits and comments.
> 
> The one thing I think you should not do, is autogen a `setup.conf` or `setup.vpp` file. It makes it ~impossible to make any changes to the dataplane configuration while operating the netlab. What if you wanted to add a sub-interface, or remove an address, or make an L2XC between two interfaces, or bind an MPLS tunnel? Your users would end up resorting to 'vppctl' debug CLI, which is cumbersome at best, and perilous at worst, because ordering matters, and the CLI/API do very limited syntax and semantic checking.
> 
> Also, there will be no config persistence if they restart the container. There is no `vppctl show run` or `vppctl write mem`, by design it is a configurable dataplane with an API that is meant to be configured by another system. Users will risk crashing the dataplane if they do things in the wrong order, or with the wrong arguments.
> 
> Consider using `vppcfg` and have ansible generate `/etc/vpp/vppcfg.yaml`, and explain to users how to edit that file and run `vppcfg plan` and `vppcfg apply`. When starting the container, you can generate the file you're now hand-crafting with `vppcfg plan --novpp -c vppcfg.yaml -o vppcfg.vpp` and start the dataplane with that instead. I guarantee you that is a better user experience, even if you rely on an external component (you can `pip install` it).
> 
> Good luck with the netlab integration!

I hear what you're saying, and in a production setup that makes total sense. The Netlab workflow / mental model deviates from that though, in ways that fundamentally impact these usage concerns.

Netlab provides an abstracted single topology definition file from which all configuration is derived. This "vendor neutral" intent gets mapped to platform specific syntax - soon including VPP.

That means that if a user wants to add a sub interface, they'd edit this Netlab topology file and redeploy the lab. That would work for VPP, but also Arista, SR Linux, FRR or any of the other platforms. Same for addressing, MPLS tunnels, etc.

Basically the same idea as your vppcfg - abstract YAML syntax mapped to platform specific commands in the right order and validated - but applied to all platforms. ```netlab create``` will (eventually) be a close alternative for ```vppcfg plan```.

A major benefit for you/VPP will be that once we finalize this config mapping, we can validate against the Netlab 250+ integration tests, and run virtual interop tests between any and all supported platforms.

DanPartelly:
> That means that if a user wants to add a sub interface, they'd edit this Netlab topology file and redeploy the lab. That would work for VPP, but also Arista, SR Linux, FRR or any of the other platforms. Same for addressing, MPLS tunnels, etc.

Well, there is no canonical "one true way"  to use netlab. Certainly  I would not always restart Netlab with an edited topology file just to make some changes to the running config of a device.  Sometimes it makes sense to do that, sometimes it doesn't. And if the topology device it's one of the devices with a long boot time, yeah, forget that, I wont restart. To save time Ill just duplicate the topology , modify it, and dump  the new config. Netlab create only, not a full deploy.  Reconcile manually on the long running lab, if needed,  copy / paste does the rest. 
   
Vppctl is really not the end of the world.  As of today, it's the only way to really dig and learn  VPP. Regardless you use vppcfg as planner or write the config yourself, there is no other way to learn VPP but vppctl command line. And crashing things or getting invalid configs is part of the learning experience.

I am good with it as is, without using vppcfg planner.

jbemmel:
> Well, there is no canonical "one true way" to use netlab

You’re right of course, it is not that black and white

DanPartelly:
@jbemmel What is the status of this one? Do you want to add more things, or should we focus on merging it in the near future? I have a personal interest in the platform.

jbemmel:
> @jbemmel What is the status of this one? Do you want to add more things, or should we focus on merging it in the near future? I have a personal interest in the platform.

Thanks for the trigger - I think this is ready for review. It adds vpp as a device with either BIRD or frr as control_plane; the initial integration tests pass (with some minor exceptions).

Adding other things to make it useful (ospf, bgp, etc.) should be done in separate PRs

docs/netlab/clab.md
```diff
@@ -95,6 +95,8 @@ $ netlab clab build bird --version 2.17.4 --tag netlab/bird:latest
 
 To use a custom container image, specify it with the **image** node parameter or the **defaults.daemons._daemon_.clab.image** [default setting](topo-defaults).
 
+See [Building BIRD containers](../labs/bird.md) and [Building VPP containers](../labs/vpp.md) for configurable software releases.
```
ipspace: Please use anchors

docs/caveats.md
```diff
@@ -328,6 +328,12 @@ Netlab enables VRRPv3 by default on Dell OS10, overriding any platform defaults.
 
 * You have to build the *dnsmasq* container image with the **netlab clab build dnsmasq** command.
 
+(caveats-vpp)=
+## VPP (Vector Packet Processor)
+
+* You must build the VPP container image with the **netlab clab build vpp** command. See [](build-vpp) for version options.
+* VPP is available only on the [containerlab](labs/clab.md) provider.
```
ipspace: Isn't this clear enough from the "platforms" table?

docs/platforms.md
```diff
@@ -83,6 +84,7 @@ Most devices behave as routers (or layer-3 switches); the following devices can
 |-----------------------|:--:|:--:| :--:|
 | Arista EOS            | ✅ | ✅ | ✅ |
 | BIRD                  | ✅ | ✅ | ❌  |
+| VPP                   | ✅ | ❌  | ✅ |
```
ipspace: Sorting...

docs/platforms.md
```diff
@@ -232,6 +234,7 @@ _netlab_ uses Ansible playbooks and device-specific task lists to deploy device
 |--------|----------|---------------------------------|
 | bird   | clab     | **bash** scripts or daemon configuration files[^BBS] |
 | dnsmasq | clab    | **bash** scripts or daemon configuration files[^DBS] |
+| vpp    | clab     | **bash** scripts, daemon configuration files, and VPP CLI configuration files[^VPPC] |
```
ipspace: Please use case-insensitive sort ;)

ipspace (changes requested):
Starting with a few documentation nits and a major question: why do you think it's a good idea to define VPP as a daemon (as opposed to a regular device)?

After all, you start the VPP daemon at the end of the initial configuration and delay the start of the BIRD process until the dataplane configuration is completed.

netsim/daemons/vpp/Dockerfile.j2
```diff
@@ -0,0 +1,143 @@
...
+	export VPP_VERSION="{{ _sw_version or '' }}"; \
+	export REPO="{{ _clab.get('repo','release') }}"; \
+	bash <<'GETVPP'
+#!/bin/bash
+
+[ -z "$REPO_URL" ] && REPO_URL="https://packagecloud.io/install/repositories/fdio/${REPO:=release}"
+
+function get_vpp () {
+	ls "*.deb" 2>/dev/null && { die "remove existing *.deb files"; }
+
```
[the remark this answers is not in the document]
ipspace: @jbemmel -- Copilot is probably right ;)

netsim/daemons/vpp/start.vpp.j2
```diff
@@ -0,0 +1,52 @@
...
+#
+set -e
+
+NETNS=dataplane
+VPP_CONTROL_PLANE={{ control_plane|default('bird') }}
+
+/usr/bin/vpp -c /etc/vpp/startup.conf &
+VPP_PID=$!
+until [ -S /run/vpp/cli.sock ]; do sleep 0.1; done
+vppctl show version >/dev/null
```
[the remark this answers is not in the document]
ipspace: I agree with the "timeout 30" suggestion

jbemmel:
> why do you think it's a good idea to define VPP as a daemon (as opposed to a regular device)?
> 
> After all, you start the VPP daemon at the end of the initial configuration and delay the start of the BIRD process until the dataplane configuration is completed.

I initially conceptualized this as "BIRD with a different data plane", hence I started from a daemon. It wasn't so much "a good idea" but more "a logical progression".

I guess I never really paid much attention to the difference between devices and daemons, it's not a distinction I tend to make naturally.

We could easily reconceive VPP as a device - I'll make that change

ipspace:
> We could easily reconceive VPP as a device - I'll make that change

Thank you!

ipspace:
> We could easily reconceive VPP as a device - I'll make that change

Thanks for making that change, but why do you still have files in the "daemons/vpp" directory?

jbemmel:
> Thanks for making that change, but why do you still have files in the "daemons/vpp" directory?

```clab build``` currently only looks under daemons/ so I left Dockerfile.j2, I moved the rest

```
./device-module-test -d vpp -p clab initial
NETLAB_DEVICES_VPP_CONTROL__PLANE=frr ./device-module-test -d vpp -p clab initial
```
both pass

The bridge test was failing for frr, so I also added proper vlan support. Those tests now pass too :)

ipspace (approved):
Great job, with a few leftover bits and pieces.

I cleaned up a few documentation quirks. You also forgot to add VPP to a lot of "supported platforms" tables, but if neither your AI friend nor you care about that, neither do I 🤷🏻‍♂️

I also disabled features that work only on Linux but not VPP, or on FRR but not BIRD. If you want to have them, add corresponding device quirks. Starting a device, not getting any errors, and having a broken network is not the best user experience.

MTU on VLANs is probably broken, and the loopback interface is not working with OSPFv3. Will open separate bug reports for that.

jbemmel:
> MTU on VLANs is probably broken, and the loopback interface is not working with OSPFv3. Will open separate bug reports for that.

I tried setting MTU on the VLANs but it failed with an error that the underlying device doesn't support it - so I left it out. Note that all the VLAN tests pass - perhaps MTU on VLANs could use a more specific test, so it gets flagged?

I didn't fully test OSPF yet - I figured it might be better to merge the baseline first, and then incrementally fix some remaining quirks.

jbemmel:
> I also disabled features that work only on Linux but not VPP, or on FRR but not BIRD. If you want to have them, add corresponding device quirks. Starting a device, not getting any errors, and having a broken network is not the best user experience.

Thinking more about this, it may be better to model this as 2 devices e.g. ```vpp-bird``` and ```vpp-frr``` as opposed to a 'vpp' device with a control_plane parameter. We could declare the ```vpp``` device as an abstract parent (```_meta_device: True```), and model the difference in features appropriately. No quirks necessary

ipspace:
> Thinking more about this, it may be better to model this as 2 devices e.g. `vpp-bird` and `vpp-frr` as opposed to a 'vpp' device with a control_plane parameter.

I was thinking along the exact same lines ;) Yes, that would make a lot of sense, and be (probably) simpler than trying to deal with two sets of control-plane features.

ipspace:
> Note that all the VLAN tests pass - perhaps MTU on VLANs could use a more specific test, so it gets flagged?

OSPF is the only protocol that reliably complains about the MTU mismatch. You could try to use dataplane tricks, but you know how "reliable" those are.

ipspace:
> Thinking more about this, it may be better to model this as 2 devices e.g. `vpp-bird` and `vpp-frr` as opposed to a 'vpp' device with a control_plane parameter.

Wait before moving forward with this. I plan to implement "device plugins" for cases where we need more functionality (beyond post-transform quirks). I'll try it out on the "kind" device, and if it works, it might be easier to implement feature merge etc.

sdktr:
> You also forgot to add VPP to a lot of "supported platforms" tables, but if neither your AI friend nor you care about that, neither do I

Are all the docs tables not generatable by the feature yamls? Seems odd that those are handled separately?

ipspace:
> Are all the docs tables not generatable by the feature yamls?

You're absolutely right, but not yet. So many ideas, so little time...

> Seems odd that those are handled separately?

Ah. the beauties of organic growth ;)) 🤷🏻‍♂️

PR #3493, opened 2026-06-17
Doc: Improve command description in README.md

### LINKED ISSUE #3489: [DOC] Make netlab create first class citizen in README.md
### Document we need to fix

https://github.com/ipspace/netlab/blob/dev/README.md

### What's wrong

I feel like we should mention netlab create on its own, as even if you do not want to install all the other packages netlab requires to make running virtual instances it provides a great learning tool for understanding config differences between devices.

**netlab create** : Creates configuration files, this generates config for any described device in `topology.yml` under the `node_files` directory, will work without any virtualization providers allowing you to review configuration for each device.

ipspace:
Closes #3489


ipspace:
What do you think about this change @snuffy22

DanPartelly (approved):
I think the original PR text reads better than Copilot's suggestion.

ipspace:
> I think the original PR text reads better than Copilot's suggestion.

Thank you ;)

PR #3494, opened 2026-06-18
Virtualbox removal from codebase

snuffy22:
This might be a bit presumptuous.

Thoughts? 

Nothing appeared to break in my initial testing


ipspace:
Maybe it's time to clean up that bit of history. The only worry I have is that someone is still using VirtualBox with netlab (I could write a blog post asking about that), but maybe we can pull the plug in the next release, similar to what we did with Ubuntu 20.04.

Any thoughts @ssasso @DanPartelly @jbemmel?

One more thing: VirtualBox was particularly useful in pre-WSL days (today, I'd tell Windows users to use WSL, which is effectively a Ubuntu VM) and on x86 Macs (my imaginary friend claims the last Apple model with the x86 CPU was launched in 2019 and discontinued in 2023). Even there, it's not hard to run Ubuntu VM on x86 Macs.

jbemmel:
> Maybe it's time to clean up that bit of history. The only worry I have is that someone is still using VirtualBox with netlab (I could write a blog post asking about that), but maybe we can pull the plug in the next release, similar to what we did with Ubuntu 20.04.
> 
> Any thoughts @ssasso @DanPartelly @jbemmel?

I used to work in VirtualBox, but I've stopped doing that about 2 years ago. There was a noticeably increasing level of friction even at the time.

I say remove it yesterday

DanPartelly:
I've never used VirtualBox in conjunction with Netlab. I seriously doubt any Netlab users rely on VirtualBox support in the wild.  Just remove it.

ipspace (approved):
Nice job. Had to fix a few documentation issues (like broken links); you might want to install sphinx (use requirements-dev.txt to install it) and do "make clean html" to check doc changes in the future.

Also, I have no idea how you managed to delete three random files. I restored them.

PR #3495, opened 2026-06-18
Add BFD support for RouterOS7

snuffy22:
I saw that BFD work had mostly been done for RouterOS7 

This adds the OSPF v2/v3 support.

Will not be much extra to add for BGP if requested.

To do multihop BFD will need a little bit more effort though.

Was tested with the `bfd.yml` in the `ospfv2` and `ospfv3`

This work would follow the 7.18 to 7.21.4

### DIFF
--- a/docs/module/bfd.md
+++ b/docs/module/bfd.md
@@ -18,7 +18,7 @@ BFD is supported on these platforms:
 | Dell OS10             | ✅  | ✅  |  ❌  |
 | Junos[^Junos]         | ✅[❗](caveats-junos) | ✅  | ✅  |
 | Mikrotik RouterOS 6   |  ❌  | ✅  | ✅  |
-| Mikrotik RouterOS 7   |  ❌  |  ❌  |  ❌  |
+| Mikrotik RouterOS 7   | ✅  | ✅  |  ❌  |
 | Nokia SR Linux        | ✅  | ✅  | ✅  |
 | Nokia SR OS[^SROS]    | ✅  | ✅  | ✅  |
 | VyOS                  | ✅[❗](caveats-vyos) | ✅  | ✅  |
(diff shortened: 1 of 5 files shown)


snuffy22:
I am now partially confused, are there two ways of adding support for things?

Should any nerd knobs etc be implemented in the `netsim/extras` plugins like `bgp.session` instead?

Cheers

ipspace:
> I am now partially confused

We can probably fix that ;) but I need to know what's confusing ;))

> are there two ways of adding support for things?

It depends ;))

> Should any nerd knobs etc be implemented in the `netsim/extras` plugins like `bgp.session` instead?

Some of that is modularity, some of it is history. We try not to implement too many things in the core of a technology module because then the configuration templates (as well as supporting code) become hard to understand. That's why a bunch of BGP session options ended in `bgp.session` plugin, and OSPF area parameters ended in `ospf.areas` plugin. Also, sometimes we expect that things won't be implemented for many platforms, and so they end in a plugin. As always, all constants eventually become variables, and plugins support more platforms than we thought 🤷🏻‍♂️ Oh, the beauties of organic growth 😜

In your particular case, you're probably wondering why BFD-with-BGP is configured in `bgp.session` while its OSPF counterpart is configured in the main OSPF module. IIRC (but hey, Git history goes all the way back ;)), we already had BGP session parameters configured in `bgp.session`, and putting BFD support there seemed like the best fit at that moment.

Hope this helps (and apologies for the confusion !), Ivan

snuffy22:
Hi @ipspace , its understandable, I've been guilty myself.

Would be best idea be to then implement when possible into the module if it exists?

In terms of using the 'plugin', do the devices automatically inherit the plugins when they run?
Ah, i missed it the top of the topology file lists plugins to run..

Cheers

ipspace:
> Would be best idea be to then implement when possible into the module if it exists?

As always, the correct answer is "it depends" 🤷🏻‍♂️

> In terms of using the 'plugin', do the devices automatically inherit the plugins when they run? 

Sort of, but you don't have to worry about that. If a device needs plugin-related configuration, the plugin code adds the plugin name to the node **config** list, so the per-device template (in the **extra/_plugin_** directory) is used sometime late in the configuration process.

ipspace (approved):
Thank you! Fixed a few nits, added documentation. Merging ;)

PR #3497, opened 2026-06-19
Fix older cEOS version empty ping failure validation

### LINKED ISSUE #3496: cEOS failed ping validation can be misclassified when stdout is empty
## Summary

When validating expected ping failures on Arista cEOS, `netlab validate` can classify the test as failed with `Ping ... did not fail` even though the data plane is isolated and a manual ping fails.

## Observed behavior

In `vrf/12-multi-vrf-bgp.yml`, the inter-VRF isolation validation using clab.image: harsha258/ceos:4.32.0F reported:

```text
[inter_vrf] Ping-based isolation test between blue and red VRF [ node(s): r1 ]
[FAIL]      Node r1: Ping to r4 did not fail
Input data: {"stdout": ""}
Plugin expression: ping('r4',expect='fail')
```

Manual checks from the same cEOS node showed the ping is unreachable, but the SSH validation path received empty stdout.

## Expected behavior

For `expect='fail'`, an empty stdout result from cEOS ping should be treated as a failed ping, not as a successful ping.

## Notes

Successful cEOS pings return normal output containing `80 bytes from ...`, while this failed-ping case can produce empty stdout through the validation path.

jbemmel:
Fixes #3496

### DIFF
--- a/netsim/validate/eos.py
+++ b/netsim/validate/eos.py
@@ -46,6 +46,8 @@ def valid_ping(
     msg += f' size {pkt_len}'
 
   if expect == 'fail':
+    if not _result.stdout.strip():
+      return msg+' failed as expected'
     for kw in ("0 packets received","0 received","unreachable"):
       if kw in _result.stdout:
         return msg+' failed as expected'


ipspace:
In principle, I hate slapping random fixes into the code base to address niche issues, and this seems like one of those.

First, what "older version" are we talking about? I had no problems for ages, and I'm not particularly fast at upgrading cEOS. Also, what's wrong with downloading not-so-ancient version from Arista's website?

Also, the **netlab validate** tests execute Docker or SSH commands, and examine the outputs only when those commands are completed. So I have to wonder under what scenario **ping** executed as an SSH command (that's what we're using on cEOS) returns no output.

jbemmel:
> First, what "older version" are we talking about? I had no problems for ages, and I'm not particularly fast at upgrading cEOS. Also, what's wrong with downloading not-so-ancient version from Arista's website?

I used harsha258/ceos:4.32.0F

I cannot download from the Arista website because I do not have a "corporate address" and ```@gmail.com``` isn't accepted

As documented in #3496: In vrf/12-multi-vrf-bgp.yml, the inter-VRF isolation validation using clab.image: harsha258/ceos:4.32.0F reported:

[inter_vrf] Ping-based isolation test between blue and red VRF [ node(s): r1 ]
[FAIL]      Node r1: Ping to r4 did not fail
Input data: {"stdout": ""}
Plugin expression: ping('r4',expect='fail')

ipspace:
OK, didn't take me long to figure out what the problem was: that particular version of cEOS prints the "ping: connect: Network is unreachable" to stderr, not stdout. Maybe we should be fixing the root cause instead?

ipspace:
Replaced by #3502 (this one is dancing around the symptom, not fixing the root cause)

PR #3498, opened 2026-06-19
Feature: bird vrf support including bgp, ospf and route leaking

jbemmel:
* Create Linux VRF tables (including any module specific script in initial/bird-clab.j2)
* Make BGP and OSPF VRF aware
* Add VRF route leaking (including static routes)
* Update documentation

Test results:
```
./device-module-test -d bird -p clab vrf
Pre-test cleanup: 
23:10:18 vrf/01-multi-vrf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:10:28 vrf/02-multi-vrf-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:10:39 vrf/03-multi-vrf-loopback.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:10:43 vrf/04-multi-vrf-loopback-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:10:50 vrf/11-multi-vrf-ospf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:11:12 vrf/12-multi-vrf-bgp.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:12:32 vrf/13-multi-vrf-bgp-loopback.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:12:48 vrf/14-multi-vrf-unnumbered.yml: create(FAIL) 
23:12:49 vrf/15-multi-vrf-mixed.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:13:10 vrf/16-vrf-bgp-community.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:13:20 vrf/21-multi-vrf-ospfv3.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:13:42 vrf/22-multi-vrf-bgp-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:13:58 vrf/23-multi-vrf-mixed-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:14:21 vrf/31-vrf-common-hosts.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:14:30 vrf/32-vrf-common-hosts-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:14:40 vrf/33-vrf-common-ospf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
23:15:01 vrf/34-vrf-common-mixed.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
```


jbemmel:
Test vrf/14-multi-vrf-unnumbered.yml can be fixed once #3475 is merged

ipspace (changes requested):
This will need a significant rework. The hacks like "include everything else in initial.j2" suck.

Also, for all other devices, we configure routing protocol instances within the VRF configuration; here, you do it in the main OSPF/BGP template. While I understand why you went down that route, it will make for an "interesting" experience when someone has to fix this code, and the two similar experiences I already had to go through persuaded me to avoid the third one at all costs 🤷🏻‍♂️

I know you're facing the challenge of doing data-plane configuration and daemon configuration in the same template, and I don't have a good answer for it yet... but it has to be better than this.

jbemmel:
> I know you're facing the challenge of doing data-plane configuration and daemon configuration in the same template, and I don't have a good answer for it yet... but it has to be better than this.

Currently, the issue is that a ```vrf``` key under ```daemon_config``` overrides the ```ansible/templates/vrf/bird-clab.j2``` script - hence the initial hack.

Somehow we need to make the ```daemon_config``` namespace distinct from regular module scripts, allowing both

jbemmel:
* Rebased from dev to get datapath wait functionality
* Refactor FRR VRF datapath script such that BIRD can import as-is
* Remove default router role (VRF test cases need fixing: #3506)
* Allow daemon config files to include "@conf" such that they can co-exist with a datapath script for the same module
* Refactor BGP, OSPF and VRF config templates separating out the Jinja2 macro's
* Align VRF logic with other devices (import BGP/OSPF macros)

jbemmel:
* Rebased from dev once more to get per-module CP config support for daemons
* replaced @conf with ```-daemon``` in ```extra_daemon_config```

```./device-module-test -d bird -p clab vrf``` still passes (except vrf-unnumbered)

File: netsim/ansible/templates/vrf/frr.data-plane.j2 [hunk 1]
```diff
@@ -9,6 +9,6 @@ ip link set {{vname}} up
 
 # Move interfaces and loopbacks to vrfs
 {% for i in interfaces if i.vrf is defined %}
-sysctl -qw net.ipv6.conf.{{ i.ifname }}.keep_addr_on_down=1
+sysctl -qw net/ipv6/conf/{{ i.ifname }}/keep_addr_on_down=1
```
jbemmel, line 12: Similar issue in https://github.com/ipspace/netlab/blob/cfd0faea9c4030125cb124372ec4f2a55f72e8ac/netsim/ansible/templates/vlan/frr.j2#L27?

ipspace:
Looks like we're almost there. I ran the whole set of integration tests, and (no surprise there) a few unexpected ones failed. When doing a large change, it's worth testing beyond the immediate feature you think you implemented.

* `bgp.session/04-default-originate` is failing the VRF part because the VRF default route is missing.
* VRF Lite with VLANs test (vlan/52) had a problem with **sysctl** on **ethX.Y** interfaces. Fixed.
* Inter-VRF static routes work (`routing/22-static-inter-vrf`), but the VRF-to-default ones do not.

I'm perfectly OK disabling BGP default route functionality for the moment (obviously it never worked as it should), documenting a caveat on VRF-to-global static routes and merge this.

Thoughts?

jbemmel:
> Thoughts?

I saw your sysctl patch and found other places in the codebase where this might be a problem. It tends to be in areas of combinations of features that we currently don't test ( e.g. VRFs with VLANs and IPv6 )

I'm ok with documenting the tests that currently fail, and then fixing those in follow-up PRs (instead of trying to fix them all in this one big PR)

netsim/ansible/templates/vrf/frr.data-plane.j2
ipspace: That's not a problem (not saying it should not be fixed) because SVI interface names don't have dots in them.

PR #3500, opened 2026-06-19
Fix: Unify bird Dockerfiles and logging

ipspace:
* Use the same command to start BIRD in all Dockerfiles
* Install the same tools into all BIRD containers
* Enable STDOUT/STDERR logging on BIRD daemon


jbemmel (approved):
Perhaps we should consider defining a single base image Dockerfile.base that contains the unified parts:
* Same base image (ubuntu:24.04 or debian:bookworm-slim)
* Same tools and CMD

Then derive child images which define only the different install scripts for the various flavors

We may want to move all docker related files into a separate daemons/bird/docker/ subdir

ipspace:
> Perhaps we should consider defining a single base image Dockerfile.base that contains the unified parts:

Good suggestions. They definitely make sense, but would probably require changes in clab/build.py, and I'm not sure it's worth doing as long as we're dealing with a single daemon.

jbemmel:
> Good suggestions. They definitely make sense, but would probably require changes in clab/build.py, and I'm not sure it's worth doing as long as we're dealing with a single daemon.

Alternatively, we could consolidate all 3 Dockerfiles into a single Dockerfile.j2. If sw_version is "bird2" (default) we install the Ubuntu package, etc.

PR #3501, opened 2026-06-19
BIRD: Start daemon after data-plane configuration

ipspace:
When the BIRD configuration file refers to an interface that still needs to be created with data-plane configuration scripts, we have to delay the start of the daemon until the data-plane scripts are done.

This commit contains the implementation framework for BIRD daemon (with VXLAN being the first use case; coming in next PR) that could be used (if needed) with other daemons:

* The netlab modules that do data-plane configuration are identified in the initial.dataplane_config device feature
* The daemon start command has to be in netlab_start_daemon group variable
* The device quirk calls a shared function that checks for dataplane configuration modules.

When the shared 'dataplane config module check' function finds a node that needs delayed start of a daemon, it:

* Modifies the container command to start a 'wait for dataplane config' script
* Modifies 'clab.config_templates' dictionary to generate the init script
* Stores the list of prerequisite dataplane scripts in netlab_dp_module group variable

The 'dataplane-wait' script (generated like any other container file) checks whether the dataplane config scripts created /var/run/_mod_.done files and executes the 'netlab_start_daemon' command only when all the required files are present.


ipspace:
@jbemmel -- this is the least-intrusive approach to the "delay the start of the daemon" challenge I could think of.

netsim/devices/_common.py
```diff
@@ -25,3 +26,27 @@ def requires_plugin(node: Box, plugin: str, topology: Box) -> None:
       more_hints=f'Add "{plugin}" to the topology "plugin" list',
       category=log.MissingDependency,
       module=node.device)
+
+def check_daemon_dataplane_config(node: Box, topology: Box) -> None:
```
jbemmel, line 30: As this is clab specific, I would put this code under providers/clab

netsim/devices/bird.py
```diff
@@ -4,11 +4,12 @@
...
-from ._common import check_indirect_static_routes
+from ._common import check_daemon_dataplane_config, check_indirect_static_routes
 
 
 class Bird(_Quirks):
 
   @classmethod
   def device_quirks(self, node: Box, topology: Box) -> None:
     check_indirect_static_routes(node)
+    check_daemon_dataplane_config(node,topology)
```
jbemmel, line 15: Instead of having each daemon include this line of code, consider checking for ```dataplane_config``` in clab provider code and then run the check

netsim/devices/_common.py
```diff
@@ -25,3 +26,27 @@ def requires_plugin(node: Box, plugin: str, topology: Box) -> None:
...
+  """
+  Checks whether a daemon uses a module that requires additional dataplane configuration
+
+  If such a module is found, the function puts plumbing in place to ensure the control-plane
+  daemon is started only when the dataplane configuration is finished. It is assumed that
```
jbemmel, line 35: Rather than assume this and rely on the script in daemons/bird, I would suggest putting the script under ```templates/provider/clab``` and then bind-mount it in as a shared read-only file.

jbemmel (changes requested):
Some refactoring suggestions, VPP will need the same logic so putting it under bird is suboptimal

netsim/devices/_common.py
ipspace: This is really device-specific (as in "used by a single device at the moment"), so I don't want to pollute clab code with yet another niche exception.

netsim/devices/bird.py
ipspace: Considered that, and decided it's better not to add single-device complexity to that code.

netsim/devices/_common.py
ipspace: Good idea. Done. However, the script cannot be a shared file as it uses node-specific variables

netsim/daemons/bird/dataplane-wait.j2
```diff
@@ -0,0 +1,22 @@
...
+      break
+    fi
+  done
+
+  if [ -z "$missing" ]; then
+    break
+  fi
+
+  sleep 1
+done
```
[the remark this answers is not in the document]
ipspace: Implemented

netsim/devices/_common.py
```diff
@@ -25,3 +26,27 @@ def requires_plugin(node: Box, plugin: str, topology: Box) -> None:
...
+  features = a_devices.get_device_features(node,topology.defaults)    # Get device features
+  dp_config = features.initial.get('dataplane_config',[])             # Do we have to take care of dataplane configs?
+  dp_module = [ m for m in node.get('module',[]) if m in dp_config ]  # Does this node use any of those modules?
+  if not dp_module:                                                   # Nope? Cool, we're done
+    return
+
+  node.netlab_dp_module = dp_module                                   # Remember the modules (we'll need them later)
+  n_clab = node.clab                                                  # We also need to change clab data
+  n_clab.cmd = '/etc/dataplane-wait.sh'                               # Entry point is now the "wait for dataplane config" script
+  n_clab.config_templates['dataplane-wait'] = '/etc/dataplane-wait.sh:sh'
```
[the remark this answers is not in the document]
ipspace: Nice thought, but we'll rely on the developer having some common sense (or figuring it out the hard way).

PR #3502, opened 2026-06-20
Fix: Check returned STDERR and STDOUT in validation tests

### LINKED ISSUE #3496: cEOS failed ping validation can be misclassified when stdout is empty

ipspace:
Some versions of Arista EOS return error messages on stderr. SSH client sends those messages to (local) stderr, and the validation code did not get them. This fix combines stderr and stdout returned by SSH (or docker exec) into a single string that can be checked by the validation code.

Fixes #3496

### DIFF
--- a/netsim/cli/external_commands.py
+++ b/netsim/cli/external_commands.py
@@ -77,6 +77,11 @@ def log_command(cmd: typing.Union[str,list], status: str) -> None:
 CAPTURED_STDOUT: str = ''
 CAPTURED_STDERR: str = ''
 
+def get_combined_output() -> str:
+  global CAPTURED_STDOUT, CAPTURED_STDERR
+  stderr = CAPTURED_STDERR.strip() + "\n" if CAPTURED_STDERR else ""
+  return stderr + CAPTURED_STDOUT
+
 def run_command(
     cmd : typing.Union[str,list],
     check_result : bool = False,
--- a/netsim/cli/connect.py
+++ b/netsim/cli/connect.py
@@ -8,7 +8,7 @@ import sys
 import typing
 from enum import IntEnum
 
-from .external_commands import run_command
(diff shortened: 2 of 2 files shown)


netsim/cli/external_commands.py
```diff
@@ -77,6 +77,11 @@ def log_command(cmd: typing.Union[str,list], status: str) -> None:
 CAPTURED_STDOUT: str = ''
 CAPTURED_STDERR: str = ''
 
+def get_combined_output() -> str:
+  global CAPTURED_STDOUT, CAPTURED_STDERR
```
jbemmel, line 81: global not needed here, only when the value is changed

jbemmel (approved):
nit

netsim/cli/external_commands.py
ipspace: Thank you, I know. I still prefer to make it explicit.

PR #3503, opened 2026-06-20
For consideration: Unify BIRD Dockerfile variants

jbemmel:
A single Docker template to build all flavors. Could host a single wait-for-config init script


ipspace:
I like the idea. The documentation needs polishing, and a few commands should be executed in all scenarios. Will fix.

ipspace:
Also, I would suggest we use `--sw-version v2` and `--sw-version v3` to get nicer container tags. Right now, the default v3 tag is `netlab/bird:bird3`, with that change it would become `netlab/bird:v3`

ipspace:
@jbemmel -- Would love to know how you managed to install BIRD v3 on top of debian:bookworm-slim 🤔

jbemmel:
> @jbemmel -- Would love to know how you managed to install BIRD v3 on top of debian:bookworm-slim 🤔

I thought I tested that, but apparently I didn't. It should be ```debian:trixie-slim```:

```
                          IMAGE                ID             DISK USAGE   CONTENT SIZE   EXTRA
Ubuntu:24.04        netlab/bird:latest   aded73134c26        236MB         73.9MB
debian:trixie-slim  netlab/bird:latest   f2c05ee96e3f        145MB         37.4MB
```

ipspace:
> I thought I tested that, but apparently I didn't. It should be `debian:trixie-slim`:

Thank you. Fixed, created a PR.

PR #3509, opened 2026-06-21
BIRD: Anycast gateway support

ipspace:
* Data-plane configuration is identical to FRR
* There is no control-plane configuration
* FRR template was split into DP/CP components to make it reusable for BIRD

Based on work by @jbemmel in #3475


netsim/ansible/templates/gateway/frr.data-plane.j2
```diff
@@ -0,0 +1,51 @@
+sysctl -w net.ipv6.conf.all.enhanced_dad=0
+sysctl -w net.ipv6.conf.default.enhanced_dad=0
+{% for intf in interfaces if intf.gateway.protocol|default('none') == 'vrrp' %}
```
jbemmel: If this thing generates content only when VRRP is enabled, don't you think it belongs in ```vrrp-config.j2```?

jbemmel (changes requested):
Would suggest to move vrrp bits

netsim/ansible/templates/gateway/frr.data-plane.j2
ipspace: Hey, I just copied your approach ;) Of course, we can further split the files.

PR #3511, opened 2026-06-21
Fix: Tested device in VRF integration tests must be a router

### LINKED ISSUE #3506: [BUG] VRF integration tests fail to set role 'router' for DUT
For example: ```netlab create vrf/01-multi-vrf.yml -d bird -p clab```

```host_vars/rtr/topology.json```:
```
...
"role": "host"
```

This causes IP forwarding to be disabled on devices that do not have the default 'router' role.

(hence my solution to set the default role to 'router', but if that is undesirable we'll have to fix the tests)

ipspace:
Also: rename 'rtr' to 'dut' in initial VRF tests for consistency

Closes #3506

### DIFF
--- a/tests/integration/vrf/01-multi-vrf.yml
+++ b/tests/integration/vrf/01-multi-vrf.yml
@@ -16,14 +16,14 @@ groups:
     device: linux
     provider: clab
   routers:
-    members: [ rtr ]
+    members: [ dut ]
     module: [ vrf ]
 
 vrfs:
   red:
-    links: [ rtr-h1, rtr-h3 ]
+    links: [ dut-h1, dut-h3 ]
   blue:
-    links: [ rtr-h2, rtr-h4 ]
+    links: [ dut-h2, dut-h4 ]
 
 validate:
   red:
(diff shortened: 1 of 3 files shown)


jbemmel (approved):
I'm not crazy about the fuzzy semantics of "role: router". It seems to be mostly "does this device need a loopback interface" but in this scenario it is about enabling IP forwarding

The use of the vRf module implies a need for IP forwarding, perhaps the check that looks at "role==router" should have "or 'vrf' in module". Alternatively, the 'vrf' module could imply 'role=router'. Or maybe we should model ```ip_forwarding``` explicitly, and enable it in devices and certain modules

ipspace:
> I'm not crazy about the fuzzy semantics of "role: router". It seems to be mostly "does this device need a loopback interface" but in this scenario it is about enabling IP forwarding

OK. This is the very last time I'm explaining this concept, and there's nothing fuzzy about it.

Routers do packet forwarding, hosts don't. Routers usually have multiple interfaces and a loopback interface (for CP stability reasons), hosts usually have a single interface. I know a couple of RFCs and blog posts that nicely explain the difference ;))

Control-plane-only nodes (route servers, route reflectors, OSPF database collectors...) do not need packet forwarding or loopback interfaces (unless they have multiple uplinks), thus `role: host`

> The use of the vRf module implies a need for IP forwarding

In your opinion. It's perfectly valid to have a route server running multiple VRFs, or a DHCP or SYSLOG server running in multiple VRF instances (_netlab_ cannot do that yet, but you get the idea).

PR #3512, opened 2026-06-21
Support per-module DP script + CP config for daemons

ipspace:
Some modules (VRF, EVPN) require data-plane scripts to configure new interfaces or routing tables and control-plane (daemon) configuration

This commit adds that functionality demonstrated with placeholder files for BIRD VRF support.

A module with DP/CP configuration should be listed in:

* features.initial.dataplane_config list (to ensure the data plane config is done before the daemon is started)
* features.extra_daemon_config dictionary to create an extra config entry/bind

The data-plane configuration template MUST be in the module directory of ansible/templates tree (named _device_.j2); the control-plane template MUST be in daemons/_device_ directory named _mod_-daemon.j2.

### DIFF
--- a/netsim/ansible/templates/vrf/bird.j2
+++ b/netsim/ansible/templates/vrf/bird.j2
@@ -1,0 +1,7 @@
+#!/bin/sh
+#
+set -e
+#
+# Placeholder: VRF data-plane configuration for BIRD
+#
+touch /var/run/vrf.done
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -31,7 +31,9 @@ clab:
     initial:
       roles: [ host, router ]
       config_mode: [ sh ]
-      dataplane_config: [ vxlan ]
+      dataplane_config: [ vxlan, vrf ]
+      extra_daemon_config:
+        vrf: /etc/bird/vrf.mod.conf
 libvirt:                        # Not yet available on libvirt or virtualbox
   image:
 virtualbox:
@@ -81,3 +83,4 @@ features:
     roles: [ host, router ]
   vxlan:
     vtep6: true
+  vrf: {}                           # Placeholder, has to be replaced with the real VRF code
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -83,4 +83,4 @@ features:
     roles: [ host, router ]
   vxlan:
     vtep6: true
-  vrf: {}                           # Placeholder, has to be replaced with the real VRF code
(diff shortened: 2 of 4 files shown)


ipspace:
@jbemmel -- I think this is the minimum fix to get the split DP/CP configuration functionality you need for VRFs. Let's hope we won't have to sort the _daemon_config entries.

jbemmel:
> @jbemmel -- I think this is the minimum fix to get the split DP/CP configuration functionality you need for VRFs. Let's hope we won't have to sort the _daemon_config entries.

The bird config depends on a particular order in ```daemon_config```, VRF must come before BGP IIRC. We can leave it implicit, or sort according to module dependency order (i.e. honor config_before)

ipspace:
> The bird config depends on a particular order in `daemon_config`

Not if we append the 'vrf-daemon' entry to the _daemon_config dict ;)

> VRF must come before BGP IIRC.

Nope.

https://github.com/ipspace/netlab/blob/a55258b0d70a5bd5a326c0bd5906c562d7788644/netsim/modules/vrf.yml#L4

We're good to go (for the moment).

jbemmel:
> Nope.
> 
> https://github.com/ipspace/netlab/blob/a55258b0d70a5bd5a326c0bd5906c562d7788644/netsim/modules/vrf.yml#L4
> 
> We're good to go (for the moment).

I was talking about:
```
daemon_config:
  bird: /etc/bird/bird.conf
  bfd: /etc/bird/bfd.conf
  vrf@conf: /etc/bird/vrf.mod.conf  <- this needs to come before
  bgp: /etc/bird/bgp.mod.conf
```

but it's easily fixed, no worries

ipspace:
> I was talking about:
> 
> ```
> daemon_config:
>   bird: /etc/bird/bird.conf
>   bfd: /etc/bird/bfd.conf
>   vrf@conf: /etc/bird/vrf.mod.conf  <- this needs to come before
>   bgp: /etc/bird/bgp.mod.conf
> ```

... and I told you it's wrong to configure VRF BGP instances in the BGP template 🤔 Now you know why 😎

jbemmel:
> ... and I told you it's wrong to configure VRF BGP instances in the BGP template 🤔 Now you know why 😎

The draft config may have been sub-optimal, but as a "device config provider" I do prefer to have visible control over the order in which the .conf files get included/rendered.

Maybe something like:
```
daemon_config:
  bird: /etc/bird/bird.conf
  bfd: /etc/bird/bfd.conf
  bgp: /etc/bird/bgp.mod.conf
  vrf: /etc/bird/vrf.mod.conf:conf
```

(instead of extra_daemon_config, similar to :sh and :ns qualifications)

PR #3514, opened 2026-06-22
Fix: Use Debian Trixie as the base image for BIRD containers

ipspace:


jbemmel (approved):

ipspace:
FWIW, I built all four container variants and tested then with a simple BGP test, so at least BIRD is correctly compiled/installed and does not crash (immediately).

PR #3515, opened 2026-06-22
Fix: Install procps in BIRD containers

jbemmel:
When using anycast gateway on bird, the scripts use ```sysctl```

This is one way to fix, another option might be to use ns provisioning


ipspace (approved):
Thank you!

PR #3517, opened 2026-06-23
Doc: Defining the meaning of True/False values in dictionary attributes

ipspace:
This commit adds an in-depth explanation of how 'False' and 'True' attribute values are transformed into dictionaries for attributes that are defined as dictionaries.

The included coverage test covers multiple scenarios, including:

* Creating a sub-dictionary from a scalar value
* Using 'True' and 'False' values at various levels in the attribute hierarchy
* Using node-level 'False' value to override a group setting

### DIFF
--- a/docs/dev/validation-alt.txt
+++ b/docs/dev/validation-alt.txt
@@ -98,13 +98,11 @@ attributes:
 
 ### Dictionaries Specified as True/False Values
 
-Sometimes we want an attribute that is defined as a dictionary specified as *True* (default value(s)) or *False* (do not use this attribute). You could achieve that with `_alt_types` (see above) and custom code transforming *True*/*False* values into desired attribute values, or you could get the job done with a slightly more convoluted data type definition and a bit more careful Jinja2 template.
+Sometimes we want an attribute that is expected to be a dictionary specified as *True* (enable the functionality with no extra parameters) or *False* (disable the functionality). You could achieve that with `_alt_types` (see above) and custom code transforming *True*/*False* values into desired attribute values, or you could get the job done with a slightly more convoluted data type definition.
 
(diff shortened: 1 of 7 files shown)


ipspace:
@jbemmel -- The coverage test closely mirrors what we need for **ospf.gr** attribute. If you're OK with this approach, I'll make the necessary changes to #3490; I've been sending you down misleading trails too many times 😢

docs/dev/validation-alt.txt
```diff
@@ -95,3 +95,68 @@ attributes:
         ibgp: '{value}'
         ebgp: '{value}'
 ```
+
+### Dictionaries Specified as True/False Values
+
+Sometimes we want an attribute that is defined as a dictionary specified as *True* (default value(s)) or *False* (do not use this attribute). You could achieve that with `_alt_types` (see above) and custom code transforming *True*/*False* values into desired attribute values, or you could get the job done with a slightly more convoluted data type definition and a bit more careful Jinja2 template.
+
+_netlab_ data validation code transforms a *False* value of an attribute that should be a dictionary into an empty dictionary. That behavior is hard-coded too deeply in the code to be changed; we have to live with it. The only way to disable it is to use `_alt_types: [ bool ]`, but then you'll have to deal with the *False* value in your code.
```
jbemmel: What if we were to add this:
```
defaults.ospf.attributes.node:
  gr:
    type: dict
    _keys:
      timer: int
    true_value: {}
    false_value: remove
```

This would allow us to keep the hard-coded reality, but still simplify the config templates for new features like ```gr```

ipspace: That would make even more sense ;), and it's probably a pretty simple change somewhere in the validation logic. Now, do I ask AI to do the change, or spend an hour trying to figure out where to add that one line? Decisions, decisions...

On a more serious note, I would prefer `false_value:` (= None) instead of string `remove`, or even better `_remove_when_false: True`. OK?

jbemmel: > `_remove_when_false: True`. 

OK

PR #3520, opened 2026-06-24
Make OSPFv3 validation helpers VRF aware

jbemmel:
Change the VLAN VRF lite test to use dual-stack ipv4/ipv6 so we test both combinations


ipspace:
Thanks for the fix, but you (yet again) mixed unrelated things into the same PR:

* The template bug is really bad. I cherry-picked those changes into 'dev' branch, so when you'll rebase this one, you'll have a single-topic PR.
* I agree that it would be nice to have VRF-aware OSPFv3 plugin tests (but that should be a standalone PR, which is now this PR ;)
* Finally, what's wrong with adding IPv6 AF to the existing VRF+VLAN test? I don't care much about the device(s) that will fail to pass that test because the vendor decided in their infinite wisdom not to support OSPFv3 or VRF-aware OSPFv3.

jbemmel:
Change from single topic "issues I hit when adding vlan vrf ipv6 test" to just the ospfv3 validation helpers and the modified dual-stack test that uses them

ipspace (approved):
Looks good to me ;) Will run the OSPFv2/OSPFv3 integration tests just to be on the safe side and merge it.

PR #3524, opened 2026-06-24
Fix: Reduce BIRD inter-VRF route leaking pipes

### LINKED ISSUE #3523: [BUG] BIRD inter-VRF route leaking is exceedingly verbose
The BIRD inter-VRF route leaking implementation creates a full-mesh of **pipe** protocols between VRFs even when there's no need for them. at the very minimum, the config template should check whether any of the import RT of the target VRF matches one of the export route targets of the source VRF.

jbemmel:
Create BIRD pipe protocols only for VRF pairs with matching import/export route targets to avoid unnecessary full-mesh route leaking configuration.

Fixes #3523 

```./device-module-test -d bird -p clab vrf``` all pass

### DIFF
--- a/netsim/daemons/bird/vrf-daemon.j2
+++ b/netsim/daemons/bird/vrf-daemon.j2
@@ -122,18 +122,21 @@ protocol static static_vrf_leak_{{ vname }}_{{ dst_name }}_{{ sr_af }} {
 {# Route leaking between VRF tables based on transformed import/export route targets #}
 {% for src_name,src_data in vrfs|default({})|dictsort %}
 {%   for dst_name,dst_data in vrfs|default({})|dictsort if dst_name != src_name %}
-{%     for rt in dst_data.import|default([]) if rt in src_data.export|default([]) %}
-{%       if loop.first %}
-{%         for _af in src_data.af if _af in dst_data.af %}
+{%     set import_rt = dst_data.import|default([])|select('in',src_data.export|default([]))|list %}
+{%     if import_rt %}
+{%       set bird_import_rt = import_rt|map('replace',':',', ')|join('), (rt, ') %}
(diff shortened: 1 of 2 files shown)


ipspace (changes requested):
* You cannot assume the route leaking is symmetrical (yeah, I know...). You need A=>B and B=>A checks
* Instead of convoluted expressions, just iterate over import RT of destination VRF and check whether any of them is in the export RT of the source VRF and generate the pipe protocol on the first iteration of the loop.
* Once you know you definitely need route leaking (because you checked RTs), you no longer need import/export filters in the pipe protocol.

Am I missing something?

jbemmel:
> * You cannot assume the route leaking is symmetrical (yeah, I know...). You need A=>B and B=>A checks

Would it be worth adding a test that checks asymmetrical leaking?

ipspace:
> Would it be worth adding a test that checks asymmetrical leaking?

No ;) In most cases, if you configure import/export RTs correctly, things just work. BIRD is a special case where you have to twiddle the low-level details.

ipspace:
@jbemmel -- Am I right in assuming that if I pipe A => B and B => C, then A happens to land in C?

jbemmel:
> @jbemmel -- Am I right in assuming that if I pipe A => B and B => C, then A happens to land in C?

Yes, we still need a filter to prevent transitive leaks. Split horizon

jbemmel:
> No ;) In most cases, if you configure import/export RTs correctly, things just work. BIRD is a special case where you have to twiddle the low-level details.

I added one and found an unrelated issue in SR Linux templates. Let me know if you want me to remove it from the PR

ipspace (commented):
Better than before, but the code is horrible. Perl is readable compared to some of this stuff :( I'm merging it (because it's an improvement), but then I'll have a serious word or two with it.

PR #3525, opened 2026-06-24
Fix bird vrf default static routes

### LINKED ISSUE #3522: [BUG] BIRD VRF-to-default static routes are not created
BIRD VRF daemon config does not create VRF-to-default static routes. The convoluted loop that creates inter-VRF static routes does not check routes with non-VRF destination table:

https://github.com/ipspace/netlab/blob/6f1399ae597e0f341e451ea83e5777f6c628c53e/netsim/daemons/bird/vrf-daemon.j2#L86

22-static-inter-vrf:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/routing/22-static-inter-vrf.yml)
* [Test results](https://tests.netlab.tools/bird/clab/routing/22-static-inter-vrf.yml-validate.log)

jbemmel:
Fixes #3522

I tried building a list of ['default'] plus VRF names, but the resulting logic becomes harder to read (IMHO)

The iteration over ```routing.static``` is a bit complicated, perhaps a helper macro with callback could simplify

### DIFF
--- a/netsim/daemons/bird/routing.j2
+++ b/netsim/daemons/bird/routing.j2
@@ -10,7 +10,7 @@ route {{ sr_data[af] }} {{ sr_nh }}{{ sr_intf }};
 # Global static routes
 #
 {% for sr_af in ['ipv4','ipv6'] %}
-{%   for sr_data in routing.static if sr_af in sr_data and 'vrf' not in sr_data %}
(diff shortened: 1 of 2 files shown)


ipspace:
Replaced by #3538

PR #3526, opened 2026-06-24
Fix bird vrf bgp default originate

### LINKED ISSUE #3521: [BUG] BIRD does not originate the default route over VRF BGP sessions
BIRD needs a local default route to advertise default route in BGP sessions. That default route is added into the global routing table but not in VRF routing tables.

https://github.com/ipspace/netlab/blob/6f1399ae597e0f341e451ea83e5777f6c628c53e/netsim/daemons/bird/bgp.j2#L6

04-default-originate:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/bgp#session/04-default-originate.yml)
* [Test results](https://tests.netlab.tools/bird/clab/bgp#session/04-default-originate.yml-validate.log)

jbemmel:
Fixes #3521

### DIFF
--- a/netsim/daemons/bird/bgp.j2
+++ b/netsim/daemons/bird/bgp.j2
@@ -3,7 +3,9 @@
 {% if 'router_id' in bgp %}
 router id {{ bgp.router_id }};
 {% endif %}
-{{ bgp_config.bgp_default_originate_routes(bgp) }}
+{% for _af in ['ipv4','ipv6'] %}
+{{ bgp_config.bgp_default_originate_routes(bgp,_af) }}
+{% endfor %}
 {{ bgp_config.bgp_advertise_list('bgp_advertise',bgp.advertise) }}
 {{ bgp_config.bgp_prefixes_function('bgp_prefixes',bgp,'bgp_advertise',bgp.import|default([]),True) }}
 {{ bgp_config.bgp_export_filters('bgp_export_','bgp_prefixes') }}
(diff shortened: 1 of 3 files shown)


ipspace:
Hey, what if I simplify the whole thing and create default static routes (when needed) in bgp.session plugin? Then we wouldn't have to have the extra stuff in the template.

The only drawback would be that those default routes would get redistributed into all routing protocols as soon as you do static-to-RP redistribution, but that probably happens in most cases today anyway.

ipspace:
Here's the proof-of-concept code for the idea: #3527

Have to add more external nodes to the coverage test to check all possible paths, but I think I'm almost there...

jbemmel:
> Hey, what if I simplify the whole thing and create default static routes (when needed) in bgp.session plugin? Then we wouldn't have to have the extra stuff in the template.

Some platforms still require things like ```_default.ios.j2: default-information originate``` so I guess we cannot fully implement default route origination as a plugin-only feature - but we can get close

ipspace:
> Some platforms still require things like `_default.ios.j2: default-information originate`

Most platforms do, and that's OK.

> so I guess we cannot fully implement default route origination as a plugin-only feature - but we can get close

The goal was to simplify the configuration templates and make them use standard features (static routes) instead of reinventing the special wheels.

ipspace:
Replaced by #3527

PR #3527, opened 2026-06-25
Add static default routes when originating BGP default routes

ipspace:
Some devices (Junos, BIRD) need a default route in the routing table to originate a BGP default route. To simplify the configuration templates, the bgp.session plugin adds the necessary static routes and activates the 'routing' module on the affected node(s).

Working:
* Global and VRF static routes are added
* Deduplication of added static routes

Still missing:
* Dual-stack/IPv6 checks
* Using existing defaults

### DIFF
--- a/netsim/extra/bgp.session/plugin.py
+++ b/netsim/extra/bgp.session/plugin.py
@@ -201,9 +201,9 @@ def process_default_requests(ndata: Box, topology: Box) -> None:
 
         # No per-AF default route in the target VRF yet, add it
         #
-        sr_final = sr_data                        # Assume we can use the static route data as-is
+        sr_final = get_box(sr_data)               # Assume we can use the static route data as-is
         if ngb_vrf:                               # But if we have a VRF neighbor, we have to add VRF info
-          sr_final = get_box({'vrf': ngb_vrf}) + sr_data
+          sr_final.vrf = ngb_vrf
         append_to_list(ndata,'routing.static',sr_final)
 
 '''
--- a/netsim/extra/bgp.session/plugin.py
+++ b/netsim/extra/bgp.session/plugin.py
@@ -3,7 +3,7 @@ import typing
 from box import Box, BoxList
 
 from netsim import api, modules
-from netsim.data import append_to_list
+from netsim.data import append_to_list, get_box
 from netsim.utils import log
 from netsim.utils import routing as _bgp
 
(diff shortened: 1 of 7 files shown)


netsim/extra/bgp.session/plugin.py
```diff
@@ -156,6 +156,52 @@ def process_bfd_requests(ndata: Box, topology: Box) -> None:
...
+      continue                                    # Nope? Cool
+
+    route_list = [                                # Find relevant static routes based on neighbor AFs
+      BGP_DEFAULT[af] for af in log.AF_LIST if af in ngb
+    ]
+    if not route_list:
+      continue
+
+    append_to_list(ndata,'module','routing')      # Activate the routing module in the node
+    append_to_list(topology,'module','routing')   # ... assuming the developer setting BGP features was sane ;)
```
jbemmel, line 181: Hope this doesn't activate the routing module on all nodes in the topology?

ipspace: Of course not. This is done late in the process (well after the topology.module is copied into node.module). The same two calls are used to add static routes to hosts.

jbemmel: In that case, what breaks if we don’t set this? Why is this necessary?

ipspace: > In that case, what breaks if we don’t set this? Why is this necessary?

The static routes are not configured 🤷🏻‍♂️

netsim/extra/bgp.session/plugin.py
```diff
@@ -156,6 +156,56 @@ def process_bfd_requests(ndata: Box, topology: Box) -> None:
...
+  if _bgp.get_device_bgp_feature('default_originate',ndata,topology) != 'static':
+    return                                        # The node does not need static routes to originate default
+
+  for ngb in _bgp.neighbors(ndata):
+    if not 'default_originate' in ngb:            # Does the neighbor need a BGP default route?
+      continue                                    # Nope? Cool
+
+    route_list = [                                # Find relevant static routes based on neighbor AFs
+      BGP_DEFAULT[af] for af in log.AF_LIST if af in ngb
+    ]
```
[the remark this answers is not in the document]
ipspace: Probably not applicable, but I nevertheless rewrote the SR creation code

PR #3531, opened 2026-06-25
Fix srlinux vrf leaking policies and decouple them from BGP EVPN

jbemmel:
Inspired by recent work on Bird route leaking, I revisited SR Linux route leaking policies and realized they had some issues.
Primarily, they assumed route leaking only works with BGP EVPN

Fixes:
* Move general policies from BGP module to bgp macro, such that a vrf-only scenario also instantiates them
* Replace BGP community based approach with native route leaking policies
* Enable dynamic host routes in VRF route leaking scenario's
* Tweak redistribution policies

I included the test that illustrates the use case which triggered the PR, feel free to remove it from the final merge

Tested:
```./device-module-test -d srlinux -p clab vrf```
```./device-module-test -d srlinux -p clab bgp```
```./device-module-test -d srlinux -p clab evpn```

### DIFF
--- a/netsim/ansible/templates/bgp/srlinux.j2
+++ b/netsim/ansible/templates/bgp/srlinux.j2
@@ -1,29 +1,4 @@
-updates:
-
-{#
-  Define default import/export policies for iBGP and eBGP: accept all
-#}
-- path: /routing-policy/policy[name=accept_all]
-  value:
-    default-action:
-      policy-result: accept
-
-- path: /routing-policy/community-set[name=ibgp-mark]
-  value:
-    member: [ "65536:0:65536" ]
-
-- path: /routing-policy/policy[name=ibgp-mark]
-  value:
-    default-action:
-      policy-result: reject
-    statement:
-    - name: mark-ibgp-routes
-      action:
-        bgp:
-          communities:
-            add: ibgp-mark
-        policy-result: accept
-
 {% from "srlinux.macro.j2" import bgp_config,bgp_shortcut with context %}
+updates:
 {{ bgp_config('default',bgp.as,bgp.router_id,bgp,{ 'af' : af }) }}
 {{ bgp_shortcut(bgp,'default') }}
--- a/netsim/ansible/templates/bgp/srlinux.macro.j2
+++ b/netsim/ansible/templates/bgp/srlinux.macro.j2
@@ -1,3 +1,29 @@
+{% macro bgp_common_policies() %}
+{#
+  Define default import/export policies for iBGP and eBGP: accept all
+#}
+- path: /routing-policy/policy[name=accept_all]
+  value:
+    default-action:
+      policy-result: accept
+
+- path: /routing-policy/community-set[name=ibgp-mark]
+  value:
+    member: [ "65536:0:65536" ]
+
+- path: /routing-policy/policy[name=ibgp-mark]
+  value:
+    default-action:
+      policy-result: reject
+    statement:
+    - name: mark-ibgp-routes
+      action:
+        bgp:
+          communities:
+            add: ibgp-mark
+        policy-result: accept
+{% endmacro %}
+
 {% macro bgp_export_prefix(vrf,prefix) %}
 - path: /routing-policy/prefix-set[name={{vrf}}_bgp_advertise]
   value:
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: netsim/ansible/templates/routing/srlinux/redistribute.j2 [hunk 1]
```diff
@@ -42,14 +42,20 @@
 {%   endif %}
 {%   for s_proto in p_data.import|default([]) %}
 {%     for srl_proto in netlab_match_protomap[s_proto] 
-         if (srl_proto != 'bgp-evpn' or evpn_active) and
-            (srl_proto != 'bgp-ipvpn' or (vrf != 'default' and mpls.vpn is defined)) %}
+         if (srl_proto != 'bgp-evpn' or evpn_active) and srl_proto != 'bgp-ipvpn' %}
```
ipspace: As I suspected -- route leaking works, but you broke MPLS/VPN stuff, and this is probably the root cause. Don't change things for no good reason.

Will change it back and retest.

netsim/ansible/templates/bgp/srlinux.j2
```diff
@@ -1,29 +1,4 @@
-updates:
```
ipspace, line 1: I have no problems with moving this to the macros file, however... (see the VRF comment)

netsim/ansible/templates/bgp/srlinux.macro.j2
```diff
@@ -180,6 +206,7 @@
     for loops are there to ensure the AF-IBGP export policy is defined
     only once)
 #}
+{{ bgp_common_policies() }}
```
ipspace, line 209: I can't see a way where the VRF BGP code would be invoked without the generic BGP code, so maybe the **bgp_common_policies** should be called just from the main BGP template?

Was there a specific reason why you (or your AI friend) decided this move was necessary?

ipspace (approved):
After reverting the BGP-IPVPN-related logic, the MPLS/VPN tests that worked before pass. Will open an issue for the MPLS/VPN route leaking failure.

It would also be nice to hear what made you recreate the common BGP policies in the VRF template.

netsim/ansible/templates/bgp/srlinux.macro.j2
jbemmel: The entire trigger for the PR started with the included VRF topology - for some reason it did not trigger the "SR Linux only supports route leaking with BGP EVPN" quirk, and instead errored out:
```
TASK [Update SRL vrf node configuration (template=/home/jbemmel/Projects/3rd-party/netlab/tests/integration/node_files/dut/vrf)] *************************************************************
fatal: [dut]: FAILED! => {"changed": false, "id": "2026-06-27 06:32:10:859827", "msg": "Error in /routing-policy/policy[name=blue_bgp_export]/statement[name=bgp]/action/bgp/communities/remove:\n    Leafref destination '.routing-policy.community-set{.name==\"ibgp-mark\"}' does not exist\nError in /routing-policy/policy[name=green_bgp_export]/statement[name=bgp]/action/bgp/communities/remove:\n    Leafref destination '.routing-policy.community-set{.name==\"ibgp-mark\"}' does not exist\nError in /routing-policy/policy[name=red_bgp_export]/statement[name=bgp]/action/bgp/communities/remove:\n    Leafref destination '.routing-policy.community-set{.name==\"ibgp-mark\"}' does not exist"}
```

Hence I started by moving the policy such that it would get included in a VRF-without-BGP scenario. Now that route leaking no longer depends on BGP policies for marking, it can be removed

netsim/ansible/templates/routing/srlinux/redistribute.j2
jbemmel: There was a "transient" reason as I was working through this, and I wanted to test for regressions but I couldn't because I no longer have access to the MPLS license. Sorry for the hickup

netsim/ansible/templates/bgp/srlinux.macro.j2
ipspace: > The entire trigger for the PR started with the included VRF topology - for some reason it did not trigger the "SR Linux only supports route leaking with BGP EVPN" quirk, and instead errored out:

Interesting. It would be nice to figure out the exact code path that caused that, but for the moment let's just leave those policies in place ;)

Thank you!

PR #3534, opened 2026-06-28
Add Validation functions BGP BFD for FRR/EOS

snuffy22:
Made the validation plugin be able to take BFD information similar to the OSPF one.

Works with both EOS and FRR.

Not entirely sure if is correct how I made change to the DUT with enabling BFD in the topology.

```
(py3-snuff) netlab@netlab:~/snuffy-netlab/tests/integration/bgp.session$ netlab validate
[WARNING] Topology source file(s) have changed since the lab has started
[INFO]    Reading validation tests from /home/netlab/snuffy-netlab/tests/integration/bgp.session/13-bfd.yml
[session_v4] Check IPv4 EBGP sessions with DUT [ node(s): x1 ]
[PASS]       x1: Neighbor 10.1.0.1 (dut) is in state Established
[PASS]       Test succeeded in 0.1 seconds

[bgp_bfd_v4] Check IPv4 EBGP sessions with DUT [ node(s): x1 ]
[PASS]       x1: All specified BGP neighbor parameters have the expected values
[PASS]       Test succeeded in 0.1 seconds

[bfd_v4]     Check BFD peer on X1 [ node(s): x1 ]
[PASS]       Validation succeeded on x1
[PASS]       Test succeeded in 0.1 seconds

[bfd_v4_dut] Check BFD peer on DUT [ node(s): dut ]
[PASS]       Validation succeeded on dut
[PASS]       Test succeeded in 0.2 seconds

[SUCCESS]    Tests passed: 4
```

### DIFF
--- a/netsim/validate/bgp/eos.py
+++ b/netsim/validate/bgp/eos.py
@@ -78,7 +78,7 @@ def valid_bgp_neighbor(
 
   return f'Neighbor {n_addr} ({n_id}) is in state {data[n_addr].peerState}'
 
-def show_bgp_neighbor_details(ngb: list, n_id: str, af: str='ipv4', *, activate: str = '', bfd: bool = False, **kwargs: typing.Any) -> str:
+def show_bgp_neighbor_details(ngb: list, n_id: str, af: str='ipv4', *, activate: str = '', **kwargs: typing.Any) -> str:
   n_addr = _common.get_bgp_neighbor_id(ngb,n_id,af)
   global af_lookup
   if not activate:
@@ -125,7 +125,7 @@ def valid_bgp_neighbor_details(
 
   result = f'The neighbor {n_addr} ({n_id}){act_err} is in BFD state {found.bfdState}'
   if not found.bfdState == 3:
-      raise Exception(f'{result} ( expected 3 - Up )')
+      raise Exception(f'{result} (expected 3 - Up)')
 
   return result
 
--- a/netsim/validate/bgp/eos.py
+++ b/netsim/validate/bgp/eos.py
@@ -78,6 +78,57 @@ def valid_bgp_neighbor(
 
   return f'Neighbor {n_addr} ({n_id}) is in state {data[n_addr].peerState}'
 
+def show_bgp_neighbor_details(ngb: list, n_id: str, af: str='ipv4', *, activate: str = '', bfd: bool = False, **kwargs: typing.Any) -> str:
+  n_addr = _common.get_bgp_neighbor_id(ngb,n_id,af)
+  global af_lookup
+  if not activate:
+    return f'bgp neighbors {n_addr} | json'
+
+  if activate not in af_lookup:
+    raise Exception(f'Unsupported address family {activate}')
+
+  return f'bgp {af_lookup[activate]} neighbors {n_addr} | json'
+
+def valid_bgp_neighbor_details(
+      ngb: list,
+      n_id: str,
+      af: str = 'ipv4',
+      state: str = 'Established',
+      vrf: str = 'default',
+      activate: str = '',
+      intf: str = '',
+      bfd: bool = False) -> str:
+  _result = global_vars.get_result_dict('_result')
+  n_addr = _common.get_bgp_neighbor_id(ngb,n_id,af)
+
+  data = check_vrf_data(_result,vrf,'peerList','BGP peers')
+
+  act_err = f' in address family {activate}' if activate else ''
+  found = next(item for item in data if item.peerAddress == n_addr)
+
+  if not found:
+    result = f'The router has no BGP neighbor with {af} address {n_addr} ({n_id}){act_err}'
+    if state == 'missing':
+      return result
+    else:
+      raise Exception(result)
+
+  if not state == found.state:
+    result = f'The neighbor {n_addr} ({n_id}){act_err} is in state {found.state}'
+    if state == 'missing' and data[0].state != 'Established':
+      return result
+    else:
+      raise Exception(f'{result} (expected {state})')
+
+  if not bfd:
+    return result
+
+  result = f'The neighbor {n_addr} ({n_id}){act_err} is in BFD state {found.bfdState}'
+  if not found.bfdState == 3:
+      raise Exception(f'{result} ( expected 3 - Up )')
+
+  return result
+
 """
 BGP prefix checks, starting with 'get a BGP prefix from JSON results'
 """
--- a/netsim/validate/bgp/eos.py
+++ b/netsim/validate/bgp/eos.py
@@ -95,7 +95,7 @@ def valid_bgp_neighbor_details(
 
   data = check_vrf_data(_result,vrf,'peerList','BGP peers')
 
-  found = next(item for item in data if item.peerAddress == n_addr)
+  found = next((item for item in data if item.peerAddress == n_addr),None)
   if not found:
     raise Exception(f'The router has no BGP neighbor with {af} address {n_addr} ({n_id})')
 
@@ -104,6 +104,8 @@ def valid_bgp_neighbor_details(
   if bfd is not None:
     bfd_state = 3 if bfd else 2
     bfd_s_txt = 'Up' if bfd else 'Down'
+    if 'bfdState' not in found:
+      raise Exception(f'We are not running BFD with neighbor {n_addr} ({n_id})')
     if found.bfdState != bfd_state:
       raise Exception(f'The neighbor {n_addr} ({n_id}) is in BFD state {found.bfdState}' +\
                       f' (expected {bfd_state} - {bfd_s_txt})')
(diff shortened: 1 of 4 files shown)


netsim/validate/bgp/eos.py
```diff
@@ -78,6 +78,57 @@ def valid_bgp_neighbor(
 
   return f'Neighbor {n_addr} ({n_id}) is in state {data[n_addr].peerState}'
 
+def show_bgp_neighbor_details(ngb: list, n_id: str, af: str='ipv4', *, activate: str = '', bfd: bool = False, **kwargs: typing.Any) -> str:
```
ipspace: No need to specify BFD parameter in the "show" call which only generates the command to execute. kwargs will take care of that.

netsim/validate/bgp/eos.py
```diff
@@ -78,6 +78,57 @@ def valid_bgp_neighbor(
...
+      return result
+    else:
+      raise Exception(f'{result} (expected {state})')
+
+  if not bfd:
+    return result
+
+  result = f'The neighbor {n_addr} ({n_id}){act_err} is in BFD state {found.bfdState}'
+  if not found.bfdState == 3:
+      raise Exception(f'{result} ( expected 3 - Up )')
```
ipspace: Nit: no need for spaces within brackets ;)

netsim/validate/bgp/frr.py
```diff
@@ -82,14 +82,16 @@ def valid_bgp_neighbor(
 
   return f'Neighbor {n_addr} ({n_id}) is in state {data[n_addr].state}'
 
-def show_bgp_neighbor_details(ngb: list, n_id: str, af: str = 'ipv4', **kwargs: typing.Any) -> str:
+def show_bgp_neighbor_details(ngb: list, n_id: str, af: str = 'ipv4', bfd: bool = False, **kwargs: typing.Any) -> str:
```
ipspace: As before, no need for bfd parameter; kwargs will take it.

netsim/validate/bgp/frr.py
```diff
@@ -101,6 +103,13 @@ def valid_bgp_neighbor_details(
     if data[k] != v:
       raise Exception(f'{k} expected value {v} actual {data[k]}')
 
+  if bfd:
+    if data.peerBfdInfo:
+      if data.peerBfdInfo.status != "Up":
+        raise Exception(f'{k} expected value UP actual {data.peerBfdInfo.status}')
+    else:
+      raise Exception(f'Neighbor data structure does not contain attribute peerBfdInfo')
```
ipspace: I would reword this along the lines of "No BFD information for BGP peer {n_id}"

tests/integration/bgp.session/13-bfd.yml
```diff
@@ -9,20 +9,20 @@ defaults.sources.extra: [ ../wait_times.yml, ../warnings.yml ]
...
   probes:
-    device: eos
+    device: frr
     provider: clab
     members: [ x1 ]
 
 nodes:
   dut:
     bgp.as: 65000
+    bgp.bfd: True
```
ipspace: Wrong. You see, the point is that DUT runs BFD with one peer but not the other.

tests/integration/bgp.session/13-bfd.yml
```diff
@@ -9,20 +9,20 @@ defaults.sources.extra: [ ../wait_times.yml, ../warnings.yml ]
 
 groups:
   probes:
-    device: eos
+    device: frr
```
ipspace: We probably need EOS here. See below...

tests/integration/bgp.session/13-bfd.yml
```diff
@@ -9,20 +9,20 @@ defaults.sources.extra: [ ../wait_times.yml, ../warnings.yml ]
...
     members: [ x1 ]
 
 nodes:
   dut:
     bgp.as: 65000
+    bgp.bfd: True
   x1:
     bgp.as: 65100
-    config: [ static_bfd ]
+    bgp.bfd: True
```
ipspace: IIRC, the reason for this tweak (we could easily configure BFD on X1) is to ensure that X1 is willing to run BFD with DUT (it has a BFD neighbor configured), but will not try to initiate it (so we know DUT has to actively run BFD for BGP neighbors).

tests/integration/bgp.session/13-bfd.yml
```diff
@@ -34,13 +34,37 @@ validate:
...
     show:
       eos: "bfd peers | json"
+      frr: "bfd peers json"
     valid:
       eos: >-
         vrfs.default.ipv4Neighbors["10.1.0.1"].peers.Ethernet1.types.normal.peerStats["10.1.0.2"].status == "up"
+      frr: >-
+        result[0].status == "up"
+
+  bfd_v4_dut:
```
ipspace: You cannot check anything on the tested device. It can be any one of the 20 or so platforms we support.

ipspace (changes requested):
While I'm perfectly fine with checking BFD status with a plugin, I'm wondering whether we need this as the integration test checks whether the tested device starts BFD, so we cannot configure BFD-for-BGP, so we won't get BFD info in the BGP neighbor anyway.

Having said that, it would be nice to have BFD-with-BGP checking functionality, but I think the parameter has to be be three-state: None (not interested in BFD), False (BFD must NOT be running) and True (BFD MUST be running).

tests/integration/bgp.session/13-bfd.yml
snuffy22: No problems,  the test from the DUT was left over, as i wanted to test both at once rather than having re up/down the environment to swap.

tests/integration/bgp.session/13-bfd.yml
snuffy22: There is only 1 peer in the `bgp.session` version of this BFD test topology currently...

netsim/validate/bgp/frr.py
snuffy22: noted

jbemmel:
Took the liberty to add a test for https://github.com/FRRouting/frr/issues/22097

ipspace:
> Took the liberty to add a test for [FRRouting/frr#22097](https://github.com/FRRouting/frr/issues/22097)

Why on earth would we test every device against an FRR bug that has already been fixed? Nope.

jbemmel:
> > Took the liberty to add a test for [FRRouting/frr#22097](https://github.com/FRRouting/frr/issues/22097)
> 
> Why on earth would we test every device against an FRR bug that has already been fixed? Nope.

Because it’s a quick minimal test that checks whether BFD actually works as intended. 

If there's a way to only run it if dut is FRR then we could add that if you prefer. I think it might be useful

ipspace:
> Because it’s a quick minimal test that checks whether BFD actually works as intended.

It's not our job to check whether implementations "work as intended". We have to do our best to configure them correctly.

> If there's a way to only run it if dut is FRR then we could add that if you prefer. I think it might be useful

If you feel like creating a test suite for FRR, please go ahead. However, although I report everything we find to my FRR contacts, that's not the job of netlab integration tests.

tests/integration/bgp.session/13-bfd.yml
snuffy22: fixed

tests/integration/bgp.session/13-bfd.yml
snuffy22: reverted.. i totally misread the 'static_bfd' part of the name..

tests/integration/bgp.session/13-bfd.yml
snuffy22: reverted

snuffy22:
I have updated my part, anything else required?

ipspace:
> I have updated my part, anything else required?

My mental energy to review it 🤔 The code looks OK, but...

As I already explained, in the test we want to check whether the tested device (DUT) starts BFD over BGP session. If DUT has BFD configured, but not enabled on the neighbor, and the probe starts a BFD session for its BGP neighbor, the test might succeed (the probe device is happy that it has a BFD session). There's a reason I made the test as convoluted as it looks to you.

The changes to the integration test are thus moot. Tangentially, you can **never make any custom changes to the tested device** or **execute any testing logic on it**, or we'd have to implement them for every device we want to test; clearly something that we will NOT do.

Regardless, let's assume we want to add BFD checking functionality to "bgp_neighbor_details" even though it will not be used in the integration test. In that case, it would be nice to have the three-state behavior I mentioned in the initial review:

> I think the parameter has to be three-state: None (not interested in BFD), False (BFD must NOT be running) and True (BFD MUST be running).

I can easily add the three-state logic and merge the code, but it will remain unused functionality from the integration test perspective. On a tangential note, we never documented the plugin functions, so even if someone wants to use this, they'd have to dig into the source code to discover it exists.

ipspace (approved):
Fixed a few things, added a platform integration test that checks plugin functionality.

Nice job, thank you!

PR #3538, opened 2026-06-29
Rewrite the BIRD VRF static routing configuration

### LINKED ISSUE #3522: [BUG] BIRD VRF-to-default static routes are not created

ipspace:
Complete rewrite of the BIRD VRF static routing config with modularization of VRF config template:

* Move data-plane, route leaking and static routing VRF configurations into separate templates to increase readability
* Add 'vrfs' and 'routing.static' guards around those templates, so we no longer need '|default(...)' every time we use a data structure
* Intra-VRF static routes are generated in a separate loop (to have them in the VRF static route template)
* Inter-VRF static routes are generated with a double loop that includes global tables as source or destination.
* The global-to-VRF static routes are generated in the VRF static route template as we need the VRF tables defined before we can use 'igp table' parameter. The 'routing' configuration contains only the global static routes.


ipspace:
@jbemmel -- Had to rewrite the whole inter-VRF static routing section, but now everything works ;)

jbemmel (approved):
Couldn't have done it better myself ;)

Good improvement, indeed much more readable. BIRD configuration has never been this easy to follow I'm sure

ipspace:
> Couldn't have done it better myself ;)

Thank you ;)

PR #3539, opened 2026-06-29
Implement indirect static routes on BIRD

ipspace:


netsim/daemons/bird/routing.j2
```diff
@@ -2,7 +2,9 @@
 {% macro config_sr(sr_data,af) %}
 {%   set sr_intf = ' dev "'+sr_data.nexthop.intf+'"' if 'intf' in sr_data.nexthop else '' %}
 {#   set sr_intf = '' #}
-{%   set sr_nh = 'unreachable' if 'discard' in sr_data.nexthop else 'via ' + sr_data.nexthop[af] %}
+{%   set sr_nh = 'unreachable' if 'discard' in sr_data.nexthop else 
+                    'recursive ' + sr_data.nexthop[af] if 'intf' not in sr_data.nexthop else
```
jbemmel, line 6: Should this be a feature flag, e.g. ```routing.static.indirect_next_hop: True```? Or do we assume that all platforms support this?

jbemmel (approved):
Minor question if there are other platforms that have similar quirks

netsim/daemons/bird/routing.j2
ipspace: Most platforms support this; for those that don't, we have a quirks function (see, for example, linux.py)

PR #3541, opened 2026-06-29
Log FRR container output to docker logs

jbemmel:
Relay FRR file logging to the container stdout stream for clab Docker nodes, enabling debugging when ```frr.debug``` is set

Sample use case:
```
netlab up bgp.session/13-bfd.yml -d frr -p clab -s nodes.dut.clab.image=quay.io/frrouting/frr:10.5.1 -s frr.debug="bfd peer"
```

With this PR, ```docker logs clab-bgpsession-dut``` shows the BFD events working for 10.6.1 but failing for 10.5.1

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -232,7 +232,6 @@ vtysh -f /tmp/config
 # Redirect FRR's log to the container's stdout so 'docker logs' works
 #
 DOCKER_LOGS_STDOUT="/proc/1/fd/1"
-touch {{ frr_log_path }} && chown frr:frr {{ frr_log_path }}
 tail -n 0 -F {{ frr_log_path }} >${DOCKER_LOGS_STDOUT} 2>&1 &
 {% endif %}
 !
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -131,6 +131,14 @@ echo "service integrated-vtysh-config" >/etc/frr/vtysh.conf
 # Ensure FRR's log directory exists before enabling file logging
 #
 mkdir -p /var/log/frr && chown frr:frr /var/log/frr
+{% if clab is defined %}
+#
+# Redirect FRR's log to the container's stdout so 'docker logs' works
+#
+FRR_LOG_OUTPUT="/proc/$(pgrep -f '^/bin/bash /usr/lib/frr/docker-start$')/fd/1"
+touch /var/log/frr/frr.log && chown frr:frr /var/log/frr/frr.log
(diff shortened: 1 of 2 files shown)


netsim/ansible/templates/initial/frr.j2
```diff
@@ -131,6 +131,14 @@ echo "service integrated-vtysh-config" >/etc/frr/vtysh.conf
 # Ensure FRR's log directory exists before enabling file logging
 #
 mkdir -p /var/log/frr && chown frr:frr /var/log/frr
+{% if clab is defined %}
+#
+# Redirect FRR's log to the container's stdout so 'docker logs' works
+#
+FRR_LOG_OUTPUT="/proc/$(pgrep -f '^/bin/bash /usr/lib/frr/docker-start$')/fd/1"
```
ipspace: ~~You might want to do some error checking here, for example, did we get a usable PID.~~

As the top process in a container is always the "init" process with PID 1, the FD you're looking for is `/proc/1/fd/1`;)

netsim/ansible/templates/initial/frr.j2
```diff
@@ -131,6 +131,14 @@ echo "service integrated-vtysh-config" >/etc/frr/vtysh.conf
...
+touch /var/log/frr/frr.log && chown frr:frr /var/log/frr/frr.log
```
ipspace: Move the whole block after FRR configuration is already done, and you won't need this

netsim/ansible/templates/initial/frr.j2
```diff
@@ -131,6 +131,14 @@ echo "service integrated-vtysh-config" >/etc/frr/vtysh.conf
...
+tail -n 0 -F /var/log/frr/frr.log >${FRR_LOG_OUTPUT} 2>&1 &
```
ipspace: You sure you don't need 'nohup' here?

Also, execute this at the very end of the initial configuration, preceded by `set +e`, so we won't get configuration errors if this concoction crashes.

ipspace (changes requested):
I guess there's not much harm in doing this, so let's do it (with a few safeguards ;)

netsim/ansible/templates/initial/frr.j2
jbemmel: I made an effort to lookup PID 7, but both work so PID 1 is simpler

netsim/ansible/templates/initial/frr.j2
jbemmel: I left out the set +e, after removing the touch it's only a ```tail```. I checked that ```nohup``` is not needed (bash has huponexit disabled by default)

netsim/ansible/templates/initial/frr.j2
ipspace: PID 1 is always there. If it dies, container crashes.

PR #3542, opened 2026-06-29
Set missing BIRD VRF feature flags

jbemmel:
Now that BIRD supports VRFs, we should be explicit about its capabilities.

TBD which integration test would catch these, if any

### DIFF
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -46,6 +46,7 @@ features:
     ipv6_lla: false             # Bird supports dynamic neighbors using 'range', but not active discovery based on RAs
     rfc8950: true
     local_as: true
+    vrf_local_as: true
     local_as_ibgp: true
     community:
       standard: [ standard, large ]
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -65,7 +65,7 @@ features:
     confederation: true
     timers: true
   ospf:
-    import: [ bgp, connected, static ]
+    import: [ bgp, connected, static, vrf ]
     password: true
     priority: true
     timers: true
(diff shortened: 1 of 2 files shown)


netsim/daemons/bird.yml
```diff
@@ -65,7 +65,7 @@ features:
     confederation: true
     timers: true
   ospf:
-    import: [ bgp, connected, static ]
+    import: [ bgp, connected, static, vrf ]
```
ipspace, line 69: Valid point, but it needs to be done for both protocols.

Also, you might want to go through other VRF-related options like bgp.vrf_local_as.

tests/integration/vrf/15-multi-vrf-mixed.yml
```diff
@@ -41,8 +41,10 @@ nodes:
     vrfs:
       red:
         ospf.router_id: 10.100.0.1
+        ospf.import: [ bgp, connected ]
```
ipspace: Absolutely not. This would crash the test for the devices that do not support configurable VRF route redistribution.

ipspace (changes requested):
VRF-related options must be enabled for more than just OSPF redistribution, and the integration test must stay as-is.

tests/integration/vrf/15-multi-vrf-mixed.yml
jbemmel: ok, so do you prefer a new separate test, or leave it untested?

ipspace: Would you check what **ospf.import** is set to in the transformed data for that integration test?

ipspace (approved):
Thank you. Running integration tests just to be on the safe side.

tests/integration/vrf/15-multi-vrf-mixed.yml
jbemmel:
```
import:
  bgp: 
    auto: true
  connected:
    auto: true
```

jbemmel: So I think I just made explicit what was already implicit anyway

PR #3543, opened 2026-06-29
Filter static node next hops by link when present

jbemmel:
Honor ```nexthop.link``` when resolving static routes via a node so topologies with parallel links can select the intended adjacency.

The current data schema already allows ```nexthop.link``` but doesn't seem to use it?

I need this for a topology involving VPN tunnels across uplinks, to filter out the duplicate routes and only select one.

### DIFF
--- a/netsim/modules/routing/static.py
+++ b/netsim/modules/routing/static.py
@@ -209,6 +209,7 @@ of the 'nexthop' data structure.
 def resolve_node_nexthop(sr_data: Box, node: Box, topology: Box) -> Box:
   nh = data.get_empty_box()
   nh_vrf = sr_data.nexthop.vrf if 'vrf' in sr_data.nexthop else sr_data.get('vrf',None)
+  nh_link = sr_data.nexthop.get('link',None)
 
   node_found = False
   for intf in node.interfaces:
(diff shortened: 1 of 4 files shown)


netsim/modules/routing/static.py
```diff
@@ -219,6 +220,10 @@ def resolve_node_nexthop(sr_data: Box, node: Box, topology: Box) -> Box:
         continue
 
       node_found = True
+      if nh_link:
```
ipspace: The link name is copied into the interface data. Why do you fetch it from the links?

tests/topology/input/rp-static.yml
```diff
@@ -55,5 +58,6 @@ nodes:
 
 links:
 - c-p
-- c-p
+- name: c-p-backup
```
ipspace, line 62: If you want to test this feature, add another c-p link, so we'll be sure only one of them is picked.

ipspace (changes requested):
Thank you, nice touch. Almost there ;)

tests/integration/evpn/topology-defaults.yml
```diff
@@ -3,4 +3,3 @@ evpn.start_transit_vni: 10000
 paths.plugin: [ "topology:.", "topology:../plugin", "package:extra" ]
 
 groups.switches.role: router            # All EVPN switches have to be routers
```
ipspace, line 5: I guess this is the result of me doing a force-push into dev. Please rebase on dev.

netsim/modules/routing/static.py
jbemmel: Changed, but noting that ```intf.name``` is confusing - ```intf.link``` or ```intf.linkname``` would be better IMHO

ipspace (approved):
Looks good to me ;)

PR #3548, opened 2026-06-30
BIRD EVPN MAC VRF support

ipspace:
Based on #3475 by @jbemmel

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -417,7 +417,7 @@ These devices support additional control-plane protocols or BGP address families
 | --------------------- | :-: | :-: | :-: | :-: |
 | Arista EOS            | ✅  | ✅  | ✅  | ✅  |
 | Aruba AOS-CX          | ✅  | ✅  | ✅  | ✅  |
-| BIRD                  |  ❌  |  ❌  |  ❌  | ✅  |
+| BIRD                  |  ❌  | ✅  |  ❌  | ✅  |
 | Cisco IOS XE[^XE]     | ✅  | ✅  | ✅  | ✅  |
 | Cisco IOSv/IOSvL2     | ✅  |  ❌  | ✅  | ✅  |
 | Cisco IOS XR[^XR]     |  ❌  | ✅  | ✅  | ✅  |
--- a/netsim/daemons/bird/bgp.j2
+++ b/netsim/daemons/bird/bgp.j2
@@ -3,6 +3,12 @@
 {% if 'router_id' in bgp %}
 router id {{ bgp.router_id }};
 {% endif %}
+{% for n in bgp.neighbors|default([]) if n.evpn|default(false) %}
+{%   if loop.first %}
+evpn table evpntab;
+{%   endif %}
+{% endfor %}
+
 {{ bgp_config.bgp_advertise_list('bgp_advertise',bgp.advertise) }}
 {{ bgp_config.bgp_prefixes_function('bgp_prefixes',bgp,'bgp_advertise',bgp.import|default([]),True) }}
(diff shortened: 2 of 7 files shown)


docs/platforms.md
```diff
@@ -417,7 +417,7 @@ These devices support additional control-plane protocols or BGP address families
 | --------------------- | :-: | :-: | :-: | :-: |
 | Arista EOS            | ✅  | ✅  | ✅  | ✅  |
 | Aruba AOS-CX          | ✅  | ✅  | ✅  | ✅  |
-| BIRD                  |  ❌  |  ❌  |  ❌  | ✅  |
+| BIRD                  |  ❌  | ✅  |  ❌  | ✅  |
```
jbemmel, line 420: BIRD does support BFD for BGP, unrelated to this PR

netsim/daemons/bird/bgp.macros.j2
```diff
@@ -161,6 +161,17 @@ protocol bgp bgp_{{ ('vrf_' + vrf_name + '_') if vrf_name else '' }}{{ n.name }}
   };
 {%       endif %}
 {%     endfor %}
+{%     if n.evpn|default('') == af %}
+  evpn {
+    table evpntab;
+    import all;
+    export all;
+{%     if n.type == 'ebgp' %}
```
jbemmel: ```if 'ebgp' in n.type``` to include ```confed_ebgp```

netsim/daemons/bird/bgp.j2
```diff
@@ -3,6 +3,12 @@
 {% if 'router_id' in bgp %}
 router id {{ bgp.router_id }};
 {% endif %}
+{% for n in bgp.neighbors|default([]) if n.evpn|default(false) %}
```
jbemmel, line 6: ```if bgp.neighbors|default([])|selectattr('evpn','defined')```

jbemmel (approved):
How could ~~you~~I forget about confederated EVPN EBGP scenarios ;)

_Confederated EBGP EVPN designs are not an absolute inevitability for most networks, but they become highly compelling at massive scale because they bridge the gap between the limitations of pure IBGP and pure EBGP:

The Problem: Pure IBGP scales poorly due to the massive CPU/memory burden placed on Route Reflectors by EVPN's chatty control plane. Pure EBGP scales well but strips or alters critical BGP attributes (like Next-Hop) at AS boundaries, which breaks advanced EVPN features like ESI multi-homing and seamless data-plane tunneling.

The Solution: BGP Confederations break a massive network into small, manageable internal sub-ASns. This provides the scalability and blast-radius isolation of EBGP while natively preserving the internal routing attributes of IBGP.

Ultimately, while unnecessary for standard enterprises, the design is functionally driven by the need to scale EVPN's massive control plane across hyperscale or multi-site data centers without breaking its underlying architecture._

netsim/daemons/bird/bgp.j2
ipspace: You love making Jinja2 resemble Perl, right? 😜

netsim/daemons/bird/bgp.macros.j2
ipspace: Of course. Fixed.

That's what happens when you blindly copy other people's code 🤷🏻‍♂️

ipspace:
> How could ~you~I forget about confederated EVPN EBGP scenarios ;)

You're absolutely right. What a mistake 😂

> _Ultimately, while unnecessary for standard enterprises, the design is functionally driven by the need to scale EVPN's massive control plane across hyperscale or multi-site data centers without breaking its underlying architecture._

Where did you get this wonderful pile of BS?

PR #3549, opened 2026-06-30
Add frr.logfile attribute change the logging file path

ipspace:


ipspace:
@jbemmel -- I got inspired by #3541 and decided to make the FRR log file location configurable ;) If you like the idea, we'd have to rebase #3541 after this is merged and use `frr.logfile` for your `tail -f` command.

jbemmel:
To be honest, I am not sure what other path value a user might want besides the default, but it doesn't hurt to make it configurable. I agree it should be in the data model, not in Python

jbemmel:
I would approve this if I could, but for some reason that's blocked

PR #3554, opened 2026-06-30
WIP: GRE tunnel plugin

ipspace:
Already done:

* Attributes, basic logic
* Most of the utility functions
* Transformation coverage test

Still missing:

* Tunnel source/destination calculations
* Device templates
* Documentation

### DIFF
--- a/netsim/devices/ios.yml
+++ b/netsim/devices/ios.yml
@@ -137,6 +137,8 @@ features:
       vrf: True
       inter_vrf: True
       discard: True
+  tunnel:
+    gre: [ vrf ]
   vlan:
     model: router
     svi_interface_name: BVI{bvi}
--- a/netsim/extra/tunnel.gre/defaults.yml
+++ b/netsim/extra/tunnel.gre/defaults.yml
@@ -1,0 +1,23 @@
+# GRE tunnels attributes and defaults
+#
+---
+attributes:
+  link:
+    tunnel:
+      af:                     # Transport address family
+        { type: str, valid_values: [ ipv4, ipv6] }
+      mode:                   # Every tunnel plugin must define tunnel.mode link attribute
+        _required: True
+        type: str
+        valid_values:
+          gre:                # ... and add its tunnel type as a valid value
+      source:                 # Define the source interface of the tunnel
+        type: dict
+        _keys:
+          ifindex: int        # You could specify ifindex if you know it
+          type:               # Source interface type (used to specify loopbacks)
+            { type: str, valid_values: [ loopback ] }
+          link:
+            name: str         # You could also specify the link name
+            role: id          # ... or link role
+      vrf: str                # Specify the transport VRF
--- a/netsim/extra/tunnel.gre/plugin.py
+++ b/netsim/extra/tunnel.gre/plugin.py
@@ -17,7 +17,7 @@ def pre_link_transform(topology: Box) -> None:
   for link in _tunnel.links(topology,'gre'):
     if len(link.interfaces) != 2:
       log.error(
-        'A GRE tunnel must have exactly two nodes attached to it (link {link._linkname})',
+        f'A GRE tunnel must have exactly two nodes attached to it (link {link._linkname})',
         category=log.IncorrectAttr,
         module='tunnel.gre')
 
--- a/netsim/extra/tunnel.gre/plugin.py
+++ b/netsim/extra/tunnel.gre/plugin.py
@@ -1,0 +1,112 @@
+
+from box import Box
+
+from netsim import api
+from netsim.augment import links as _links
+from netsim.augment import nodes as _nodes
+from netsim.utils import log
+from netsim.utils import tunnel as _tunnel
+
+_config_name = 'tunnel.gre'
+
+def pre_link_transform(topology: Box) -> None:
+  '''
+  pre_link_transform hook: set tunnel link type, check whether GRE tunnels are P2P
+  '''
+  _tunnel.set_tunnel_type(topology)
+  for link in _tunnel.links(topology,'gre'):
+    if len(link.interfaces) != 2:
+      log.error(
+        'A GRE tunnel must have exactly two nodes attached to it (link {link._linkname})',
+        category=log.IncorrectAttr,
+        module='tunnel.gre')
+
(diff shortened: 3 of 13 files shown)


netsim/extra/tunnel.gre/defaults.yml
```diff
@@ -0,0 +1,15 @@
...
+attributes:
+  link:
+    tunnel:
+      mode:                   # Every tunnel plugin must define tunnel.mode link attribute
+        type: str
+        valid_values:
+          gre:                # ... and add its tunnel type as a valid value
+      source:                 # Define the source interface of the tunnel
+        ifindex: int          # You could specify ifindex if you know it
+        linkname: str         # You could also specify the link name
```
jbemmel: ```link.name```?

jbemmel:
Could we use ```extra/tunnel/gre``` as path, so all tunnel plugins are together?

netsim/extra/tunnel.gre/defaults.yml
ipspace: Sure, I could go for that ;)

ipspace:
> Could we use `extra/tunnel/gre` as path, so all tunnel plugins are together?

No idea, never tested it. Will try it out

netsim/extra/tunnel.gre/defaults.yml
ipspace: Implemented. Thank you!

ipspace:
@jbemmel, @DanPartelly -- The tunnel.gre plugin is ready. Based on @jbemmel suggestion, I changed the source attribute names to **tunnel.source.link.something**. Will figure out how to put 'tunnel.gre' plugin in 'tunnel/gre' directory as a separate PR.

netsim/extra/tunnel.gre/defaults.yml
```diff
@@ -0,0 +1,23 @@
+# GRE tunnels attributes and defaults
+#
+---
+attributes:
+  link:
+    tunnel:
+      af:                     # Transport address family
+        { type: str, valid_values: [ ipv4, ipv6] }
```
jbemmel, line 8: spacing

netsim/extra/tunnel.gre/plugin.py
```diff
@@ -0,0 +1,112 @@
...
+      if 'name' in intf and '->' in intf.name and 'GRE' not in intf.name:
+        intf.name += ' [GRE tunnel]'
+
+  if log.get_error_count():                                 # Has someone reported an error?
+    return                                                  # Might have been us, no reason to continue
+
+  # Second pass -- after computing tunnel source for all GRE interfaces, set tunnel destinations
+  #
+  for node in node_iflist:                                  # Process only nodes with tunnel interfaces
+    ndata = topology.nodes[node]                            # Get node data (we might need it)
```
jbemmel, line 87: unused?

netsim/utils/tunnel.py
```diff
@@ -0,0 +1,130 @@
...
+      continue
+
+    if t_vrf is None:                             # No VRF means 'global routing table'
+      if 'vrf' in intf:                           # ... so skip non-global interfaces
+        continue
+    elif t_vrf != intf.get('vrf',None):           # Otherwise, check the VRF name
+      continue
+
+    if t_name and t_name != intf.get('name',None):
+      continue                                    # Mismatch in interface name
```
jbemmel, line 84: link name - that's the confusing part I noted earlier

netsim/utils/tunnel.py
```diff
@@ -0,0 +1,130 @@
...
+          continue                                # Mismatch in interface role
+      else:                                       # No interface role, let's check link role
+        link = _links.get_link_by_index(topology,intf.linkindex)
+        if link is None or t_role != link.get('role',None):
+          continue                                  # Mismatch in link role
+
+    if t_af not in intf:                          # Is the interface enabled for the desired tunnel transport AF
+      continue
+
+    if not isinstance(intf[t_af],str):            # Cannot run tunnels from unnumbered interfaces
```
jbemmel, line 98: Should generate a (suppressible) warning?

jbemmel (approved):
Some minor tweaks, I think we will find that much of the code is applicable to other tunnel types as well but we can refactor as we go

DanPartelly (approved):
Thank you Ivan.

netsim/devices/ios.yml
```diff
@@ -137,6 +137,8 @@ features:
       vrf: True
       inter_vrf: True
       discard: True
+  tunnel:
```
jbemmel, line 140: Given that this is a plugin, shouldn't this be in ```extra/tunnel.gre/defaults.yml``` similar to e.g. EVPN multihoming feature flags?

I do realize we haven't always been consistent, e.g. ```bgp.session``` features are declared in devices as well.

ipspace: A while ago, I moved all plugin device features into the main device features file. There's definitely a PR somewhere, and most probably related discussion etc.

netsim/extra/tunnel.gre/defaults.yml
ipspace: Dang, missed it before merging. Maybe we have to update yamllint defaults ;)

netsim/extra/tunnel.gre/plugin.py
ipspace: As I said, we _might_ need it ;) Interestingly, none of the code checking tools complain about setting a variable value and not using it afterwards; they only complain about unused variable.

netsim/utils/tunnel.py
ipspace: Yeah, of course I agree, but that ship has sailed years ago...

netsim/utils/tunnel.py
ipspace: Let's see how this evolves. We could either generate errors in here or in the caller function. We could be more precise here, but would have more context in the caller.

netsim/utils/tunnel.py
jbemmel: The comment should say "Mismatch in link name"

ipspace: IIRC I tested and you can specify 'name' on an individual interface, which would make this exactly what it says

PR #3560, opened 2026-07-01
Draft: Add source-build Dockerfile for FRR container images

jbemmel:
Enable ```netlab clab build frr``` to compile FRR from GitHub sources on debian:trixie-slim, matching the layout and startup of the official image.

This allows us to test the latest features and bug fixes, like that EVPN IPv6 issue

I tried enabling the wait-for-dataplane logic such that the MPLS label setting gets applied before Zebra starts, but the workflow for FRR is different; initial uses ```vtysh``` which requires FRR to be running.

"Solved" in the Dockerfile for now, pending possible refactoring.

```devices/frr/Dockerfile.j2``` might be a better location for this?

### DIFF
--- a/netsim/daemons/frr/Dockerfile.j2
+++ b/netsim/daemons/frr/Dockerfile.j2
@@ -1,0 +1,119 @@
+{%- set _sw_version = sw_version|default("master") -%}
+{%- set _base_image = "debian:trixie-slim" -%}
+{%- set _runtime_tools = "iputils-ping iproute2 procps" -%}
+{%- set _clab = defaults.devices.frr.clab -%}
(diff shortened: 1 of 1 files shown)


ipspace:
What's wrong with https://blog.ipspace.net/2026/04/frr-pr-testing/

jbemmel:
> What's wrong with https://blog.ipspace.net/2026/04/frr-pr-testing/

Well, it takes 5 minutes versus 2 for my build script, and I was anticipating putting in other things like Wireguard. But you're right that there are other ways to achieve the same - hence 'Draft'

ipspace:
> Well, it takes 5 minutes versus 2 for my build scriptđ

I would still prefer the "official" build process, plus you can start the build from any commit in their repo.

> and I was anticipating putting in other things like Wireguard.

For that, create your own Dockerfile, start with the official FRR container and add stuff. I don't think that needs to be integrated into netlab.

> But you're right that there are other ways to achieve the same - hence 'Draft'

👍

PR #3561, opened 2026-07-01
Feature/frr GRE tunnel support

jbemmel:
Building out the GRE tunnel support #3554, couple of highlights:

* First plugin that requires a kernel module for Linux devices (with .-key issues); code adds a check for config keys such that kernel modules are only loaded if the plugin is actually used by at least 1 node
* ipv6 support anticipated, not tested yet

### DIFF
--- a/netsim/providers/clab/labops.py
+++ b/netsim/providers/clab/labops.py
@@ -109,11 +109,16 @@ def load_kmods(topology: Box) -> None:
 
     # At this point, we have device-specific dictionary mapping netlab modules into kernel modules
     #
-    for m in (['initial']+ndata.get('module',[])):          # Now iterate over all the netlab modules the node uses
-      if m not in kdata:                                    # ... and if the netlab modules does not need kernel modules
-        continue                                            # ... move on
-      for kmod in kdata[m]:                                 # Next, add individual kernel modules in the kdata entry
-        append_to_list(kmod_list,m,kmod)                    # ... to the module-specific list of kernel mdules
+    for m in (['initial']+ndata.get('module',[])+ndata.get('config',[])):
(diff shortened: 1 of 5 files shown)


ipspace:
> Building out the GRE tunnel support #3554, couple of highlights:

Please rebase, the PR is showing tons of modified files from #3554

> * First plugin that requires a kernel module for Linux devices (with .-key issues); code adds a check for config keys such that kernel modules are only loaded if the plugin is actually used by at least 1 node

Interesting ;) Will review after the rebase.

netsim/providers/clab/labops.py
```diff
@@ -109,10 +109,15 @@ def load_kmods(topology: Box) -> None:
...
     #
-    for m in (['initial']+ndata.get('module',[])):          # Now iterate over all the netlab modules the node uses
-      if m not in kdata:                                    # ... and if the netlab modules does not need kernel modules
-        continue                                            # ... move on
-      for kmod in kdata[m]:                                 # Next, add individual kernel modules in the kdata entry
+    for m in (['initial']+ndata.get('module',[])+ndata.get('config',[])):  # Now iterate over all the netlab modules the node uses
+      if m in kdata:
+        kmods = kdata[m]
+      else:
+        kmod_key = m.replace('.','@')                       # For plugins like tunnel.gre, use the @-as-. hack
```
ipspace: I think you could just replace '.' with '@' and check that (no module has a dot in its name for good reasons). No need to have convoluted logic like what you did here.

ipspace (changes requested):
Nice job. There's just a minor detail that (I think) makes code unnecessarily verbose.

PR #3566, opened 2026-07-02
[VyOS] GRE tunnel implementation

ssasso:
Main configuration performed in 'initial' module, due to #3564 

Requires fixing #3563 for test '01-gre.yaml' to work.

With a local fix:

Test 01:
```
[SUCCESS] Lab devices configured


Test the GRE tunnels. The two tested devices have a GRE tunnel and run
OSPF over it. The probe should receive an OSPF route advertised from the
remote device (DUT_R1) and be able to ping its loopback.

[adj]     Check OSPF adjacencies [ node(s): r3 ]
[WAITING] Waiting for OSPF adjacency process to complete (retrying for 30 seconds)
[PASS]    r3: OSPFv2 neighbor 10.0.0.2 is in state Full/-
[PASS]    Test succeeded in 5.0 seconds

[pfx]     Check for DUT_R1 loopback prefix being propagated to R3 [ node(s): r3 ]
[WAITING] Waiting for SPF to do its magic (retrying for 30 seconds)
[PASS]    r3: The prefix 10.0.0.1/32 is in the OSPF topology
[PASS]    Test succeeded in 9.8 seconds

[ping]    Check end-to-end connectivity [ node(s): r3 ]
[PASS]    r3: Ping to 10.0.0.1 succeeded
[PASS]    Test succeeded in 0.2 seconds

[SUCCESS] Tests passed: 3
```

Test 02:
```
[SUCCESS] Lab devices configured


Test the GRE tunnels. The two tested devices have a GRE tunnel in a
transport VRF and run (global) OSPF over it. The probe should receive
an OSPF route advertised from the remote device (DUT_R1) and be able
to ping its loopback.

[adj]     Check OSPF adjacencies [ node(s): r3 ]
[PASS]    r3: OSPFv2 neighbor 10.0.0.2 is in state Full/-
[PASS]    Test succeeded in 0.2 seconds

[pfx]     Check for DUT_R1 loopback prefix being propagated to R3 [ node(s): r3 ]
[WAITING] Waiting for SPF to do its magic (retrying for 30 seconds)
[PASS]    r3: The prefix 10.0.0.1/32 is in the OSPF topology
[PASS]    Test succeeded in 14.6 seconds

[ping]    Check end-to-end connectivity [ node(s): r3 ]
[PASS]    r3: Ping to 10.0.0.1 succeeded
[PASS]    Test succeeded in 0.2 seconds

[SUCCESS] Tests passed: 3
```

PR #3569, opened 2026-07-02
Feature/tunnel wireguard for FRR

jbemmel:
* IPv4 or IPv6
* default routing table or in VRF
* WireGuard package dynamically installed
* keys auto-generated if WireGuard tools are installed
* MTU auto-derived and validated

Tested:
```
./device-module-test -d frr -p clab tunnel *wire*
./device-module-test -d frr -p libvirt tunnel *wire* (needed the default route fixes)
```

Notes:
* PR generalizes Linux package installation to support both apk and apt-get; needed for FRR containers
* The ipv6 link local address is needed for OSPFv3, made this a util method in routing
* ```add_linux_packages``` is a candidate for general util method


netsim/extra/tunnel/wireguard/__init__.py
```diff
@@ -0,0 +1,380 @@
...
+      input=f'{private_key}\n',
+      text=True,
+      stderr=subprocess.DEVNULL).strip()
+  except (FileNotFoundError, subprocess.CalledProcessError, OSError):
+    log.fatal(
+      'Cannot derive WireGuard public key (install wireguard-tools, '
+      'or configure tunnel.private_key and tunnel.public_key)',
+      module='tunnel.wireguard')
+
+def generate_keypair() -> tuple[str, str]:
```
snuffy22, line 31: Do we need to use the wg-tools callout's in the generation of keys if they dont exist?

As its rather simple to just let the python handle this with, which i think you partially had in one of your earlier commits?
```
import base64
from cryptography.hazmat.primitives.asymmetric.x25519 import X25519PrivateKey
from cryptography.hazmat.primitives import serialization
```

jbemmel: We could, my thinking was to let WireGuard tools generate the keys for WireGuard tunnels so it's likely to work, but I'm not married to them

snuffy22: Benefit of it not using those tools is the plugin can be more easily used for other OSs that also might not have the `wg-tools` or similar package. (thinking of OpenBSD/RouterOS or similar that do have wg implementations)

Plus this appears to be the only use of calling external program from within this python code.

ipspace: > Benefit of it not using those tools is the plugin can be more easily used for other OSs that also might not have the `wg-tools` or similar package.

As the wg-tools are executed on the netlab hosts, the availability of the package on lab devices is irrelevant.

@jbemmel -- I don't care how you generate the keys, apart from being nervous seeing "cryptography.**hazmat**' in source code 🤔

netsim/extra/tunnel.wireguard/plugin.py
```diff
@@ -0,0 +1,380 @@
...
+  Utility function: get the name of a link referenced in an interface.
+  Return 'unknown' when the link cannot be found
+  '''
+  link = _links.get_link_by_index(topology,linkindex)
+  return link._linkname if link is not None else 'unknown'
+
+def _interface_has_af(intf: Box, af: str) -> bool:
+  return af in intf and isinstance(intf[af],str)
+
+def _interface_matches_constraints(
```
ipspace: And why exactly did you reinvent the utility function? There's a reason I put it into `utils`, not into `tunnel.gre`

netsim/extra/tunnel.wireguard/plugin.py
```diff
@@ -0,0 +1,380 @@
...
+  return 'ipv6' if has_ipv6 else 'ipv4'
+
+def get_wireguard_source(
+      ndata: Box,
+      intf: Box,
+      topology: Box,
+      peer_node: str) -> typing.Optional[Box]:
+  '''
+  Find the tunnel underlay interface. When the source is not pinned to a
+  specific interface, prefer the underlay link connected to the tunnel peer.
```
ipspace: Usually we build VPN tunnels _over_ something, but OK, let's assume you have to provider P2P encryption. We could add another parameter to the link selection instead of reinventing the wheel.

netsim/extra/tunnel.wireguard/plugin.py
```diff
@@ -0,0 +1,380 @@
...
+  '''
+  Set an IPv6 link-local address on a WireGuard tunnel interface.
+
+  WireGuard devices are ARPHRD_NONE (link/none), so the kernel forces
+  addr_gen_mode to 'none' and never assigns a link-local address (random and
+  stable-privacy modes are ignored). OSPFv3 needs one to form adjacencies, so
+  we derive a deterministic address that the initial config script assigns when
+  the interface is created.
+  '''
+  if intf.get('tunnel.af') != 'ipv6' or 'ipv6' not in intf:
```
ipspace: If the wireguard tunnels are not dual-stacked (I know I could check), then you have to throw an error, not just check the AF, shrug, and return. If they are dual-stacked, then I don't understand why you're checking tunnel.af (which impacts the underlay interface selection)

netsim/extra/tunnel.wireguard/plugin.py
```diff
@@ -0,0 +1,380 @@
...
+    return
+
+  if not isinstance(intf.ipv6,str):
```
ipspace: How will this work with LLA-only wireguard interfaces?

ipspace (changes requested):
It looks like you (or your AI friend) copied the `utils/tunnel.py` code, resulting in duplicate functionality. Please fix that.

And yes, I know you'll tell me something about AFs, but we can deal with that.

netsim/extra/tunnel.wireguard/plugin.py
jbemmel: I instructed AI to implement a Wireguard tunnel plugin similar to the GRE tunnel plugin - it took this a bit too literally

netsim/extra/tunnel.wireguard/plugin.py
jbemmel: Now using ```get_tunnel_source```

netsim/extra/tunnel.wireguard/plugin.py
jbemmel: ```get_tunnel_source``` excludes non-str IP addresses in the same way

jbemmel:
PR now uses the common ```get_tunnel_source``` proposing a minor change in the AF filtering (i.e. if ```tunnel.af``` is set, use it to filter valid interfaces) and adding the MTU of the source interface in the response

ipspace:
I would suggest you use the new utility functions from #3576, which might make your code cleaner (and with less duplicate code). Also, the new source selection function should give you all the information you need to figure out AF and MTU.

tests/integration/tunnel/03-wireguard.yml
```diff
@@ -0,0 +1,75 @@
...
+- group: core
+  ospf: False
+  role: core
+  pool: core
+  mtu: 1560
+  members: [ dut_r1-core, dut_r2-core ]
+- group: site
+  members: [ dut_r2-r3 ]
+- dut_r1:
+    tunnel.private_key: dwdtCnMYpX08FsFyUbJmRd9ML4frwJkqsXf7pR25LCo=
```
snuffy22, line 53: Maybe as part of this test, one side should have  only a private key to cover your code that generates the keys?

jbemmel: I considered that (even leaving all keys out) but the tests intentionally specify all the keys such that we do not need to install the wireguard tools on the (CI/CD) host. For the integration tests, we mainly need to test the tunnel functionality itself

jbemmel:
> I would suggest you use the new utility functions from #3576, which might make your code cleaner (and with less duplicate code). Also, the new source selection function should give you all the information you need to figure out AF and MTU.

ok, updated:
* Moved IPv6 link-local address generation to a FRR device quirk (rather than the Wireguard tunnel plugin)
* Moved tunnel device creation to FRR initial script (also for GRE -> requires initializing the full tunnel config)
* Suggested change for ```tunnel_source``` to yield each source interface such that specific plugins can configure each interface if needed

```add_linux_packages``` doesn't belong in the Wireguard plugin, suggested alternative:
```
# tunnel/wireguard/defaults.yml (plugin declares dependency)
netlab_linux_packages:
  tunnel@wireguard:
    wireguard-tools: wg
```
with logic in ```augment/devices.py``` to merge it per device that runs the included modules or plugins

File: netsim/ansible/templates/initial/frr.j2 [hunk 7]
```diff
@@ -49,7 +49,11 @@ else
   echo deb https://deb.frrouting.org/frr $(lsb_release -s -c) $FRRVER > /etc/apt/sources.list.d/frr.list
   apt-get update -qq && apt-get install -qq frr frr-pythontools bridge-utils ethtool
 fi
-{% elif netlab_mgmt_vrf|default(True) %}
+{% else %}
+{%   if netlab_linux_packages|default({}) %}
```
ipspace: Why would you install packages only in FRR container but not in FRR VM?

netsim/devices/frr.py
```diff
@@ -24,10 +25,25 @@ def check_stp_on_trunks(node: Box, topology: Box) -> None:
       more_data=err_data,
       node=node)
 
+def wireguard_ipv6_link_local(node: Box) -> None:
```
ipspace: What happens when someone decides to run Wireguard on a Linux or BIRD node? This really belongs to the plugin.

netsim/ansible/templates/initial/frr.j2
```diff
@@ -90,24 +95,32 @@ fi
...
-# Create loopbacks, stub and lag/bond devices
+# Create loopbacks, stub, lag/bond and tunnel devices
 #
-{% for i in netlab_interfaces if i.type in ['loopback','stub','lag','bond'] %}
+{% for i in netlab_interfaces if i.type in ['loopback','stub','lag','bond','tunnel'] %}
 if [ ! -e /sys/class/net/{{ i.ifname }} ]; then
   if [ ! -e /sys/devices/virtual/net/{{ i.ifname }} ]; then
 {%   if i.type in ['lag','bond'] %}
 {{     create_bond_dev(i,node_provider) }}
+{%   elif i.tunnel.mode|default('') == 'gre' %}
```
ipspace: No. Why are you "fixing" a problem that does not exist, and why are you doing that in an unrelated PR?

netsim/ansible/templates/initial/frr.j2
```diff
@@ -90,24 +95,32 @@ fi
...
+{%     set af = i.tunnel.af|default('ipv4') %}
+{%     set gre_mode = 'ip6gre' if af == 'ipv6' else 'gre' %}
+    ip tunnel add {{ i.ifname }} mode {{ gre_mode }} \
+      local {{ i.tunnel._source[af] }} remote {{ i.tunnel._destination[af] }} \
+      dev {{ i.tunnel._source.ifname }} ttl 255
 {%   else %}
-    ip link add {{ i.ifname }} type dummy
+    ip link add {{ i.ifname }} type {{ 'wireguard' if i.tunnel.mode|default('') == 'wireguard' else 'dummy' }}
```
ipspace: No. We have a nice solution in #3566. Why do you think it makes sense to hard-code "wireguard" in main FRR initial template just to fix your immediate challenge?

netsim/extra/tunnel/gre/frr.j2
```diff
@@ -1,13 +1 @@
...
-{% for intf in netlab_interfaces if intf.tunnel.mode|default('') in ['gre'] %}
-{%   set af = intf.tunnel.af|default('ipv4') %}
-{%   set gre_mode = 'ip6gre' if af == 'ipv6' else 'gre' %}
-ip tunnel add {{ intf.ifname }} mode {{ gre_mode }} \
-  local {{ intf.tunnel._source[af] }} remote {{ intf.tunnel._destination[af] }} \
-  dev {{ intf.tunnel._source.ifname }} ttl 255
-ip link set dev {{ intf.ifname }} up
-{% endfor %}
-exit 0
+# Empty file, GRE tunnels are created during initial FRR device configuration
```
ipspace: Please stop fixing imaginary problems

netsim/extra/tunnel/wireguard/frr.j2
```diff
@@ -0,0 +1,103 @@
...
+    listen-port "${listen_port}" \
+    ${wg_fwmark} \
+    peer "${peer_key}" \
+    allowed-ips "${peer_allowed}" \
+    endpoint "${peer_endpoint}" \
+    persistent-keepalive "${keepalive}"
+
+  ip link set dev "${ifname}" mtu "${mtu}" up
+
+  wait_wg_peer "${ifname}" "${peer_key}"
```
ipspace: How do you expect this to work between a FRR VM and a FRR container? Also, what happens if the two containers are configured in different batches?

netsim/extra/tunnel/_p2p.py
```diff
@@ -75,6 +74,24 @@ def tunnel_source(
       if t_name:
         _tunnel.set_tunnel_name(intf,t_name)
 
+      yield ndata,intf
+
+def tunnel_source(
```
ipspace: Why was this change necessary?

ipspace (changes requested):
Please stop "fixing" imaginary problems, hard-coding stuff where include files work, and changing things for no documented reason.

There are probably other things that have to be fixed; I had enough.

ipspace:
> `add_linux_packages` doesn't belong in the Wireguard plugin, suggested alternative:

Why not? It's either there or in the device quirks.

> with logic in `augment/devices.py` to merge it per device that runs the included modules or plugins

So we'll modify core logic to address the needs of one device for one plugin? Well, no.

Even if it were more common, it's still applicable only to Linux and friends (FRR, BIRD).

netsim/ansible/templates/initial/frr.j2
jbemmel: As requested, I moved device creation to the initial template (including 'tunnel' similar to 'bond'). This in turn broke the GRE plugin, because now tunnel devices got created twice

I can refactor things and use the per-plugin include pattern instead

netsim/extra/tunnel/_p2p.py
jbemmel: For the Wireguard plugin I needed to customize each tunnel beyond what the generic code does.

There are many ways to do that, I could simply iterate again too. I thought this might be a nice way to do it - but if you don't like it, we can change it

netsim/ansible/templates/initial/frr.j2
ipspace: How about creating tunnels in the custom config templates? Why do you think you have to create it during the initial device configuration? FRR seems patient...

netsim/extra/tunnel/_p2p.py
ipspace: Premature optimization at its best... Let's add a generator, and a stub function, with all the accompanying docstrings, just so we don't need an extra for loop 🤦‍♂️ in code that will be executed once every other blue moon.

netsim/ansible/templates/initial/frr.j2
jbemmel: Both clab and libvirt providers pass the integration tests, what led you to believe this would only apply to FRR container?

netsim/devices/frr.py
jbemmel: ok, moved back

netsim/ansible/templates/initial/frr.j2
ipspace: Stupid me. We install Linux packages unconditionally in libvirt path and check whether netlab_linux_packages is defined in clab path. Maybe we should unify them once this gets merged?

jbemmel: It's the asymmetric redundant ```if``` condition that may have confused you, I'll make it the same

jbemmel:
* Moved ipv6 link local back to the plugin
* Reverted changes to tunnel/__init__.py
* Applied VyOS initial plugin for loop include pattern (separate PR #3585)

jbemmel:
_Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away..._

netsim/extra/tunnel/wireguard/frr.j2
```diff
@@ -0,0 +1,65 @@
...
+  fi
+}
+#
+{% for intf in netlab_interfaces if intf.tunnel.mode|default('') in ['wireguard'] %}
+if [ ! -e /sys/class/net/{{ intf.ifname }} ]; then
+  ip link add {{ intf.ifname }} type wireguard
+fi
+{{ linux_intf_sysctl(intf,role,activate=False) }}
+{%   if intf._ipv6_link_local is defined %}
+ip -6 addr add {{ intf._ipv6_link_local }} dev {{ intf.ifname }} scope link
```
ipspace: I apologize for turning this into a never-ending saga, but now that we consistently apply `sysctl -qw net/ipv6/conf/tun0/addr_gen_mode=3` to interfaces, the tunnel interface gets a LLA address and OSPFv3 adjacency is established without the "manual LLA" trick.

ipspace (changes requested):
Please try taking out the LLA hack. The current setup works for me without adding manual LLA to the tunnel interface.

netsim/extra/tunnel/wireguard/frr.j2
jbemmel: I could swear I tested that before resorting to the manual trick, but you're right it works beautifully now. Thanks for double checking

netsim/extra/tunnel/wireguard/__init__.py
```diff
@@ -0,0 +1,163 @@
...
+  '''
+
+  # Use shared P2P tunnel function to check feature support
+  #
+  node_iflist = _p2p.feature_check(topology,t_mode='wireguard',t_desc='WireGuard tunnels')
+  _p2p.tunnel_source(topology,node_iflist,t_name='WireGuard')
+
+  for node in node_iflist:
+    ndata = topology.nodes[node]
+    for intf in _tunnel.interfaces(ndata,'wireguard'):
```
ipspace, line 156: You could have used node_iflist[node] here, but it doesn't matter.

ipspace (approved):
Works like a charm ;) Thank you!

PR #3570, opened 2026-07-03
Add package:extra to templates lookup dirs

ssasso:
Required for #3566


ipspace (approved):
Thank you! (and thanks a million for making this an extra PR).

PR #3573, opened 2026-07-03
JunOS: GRE tunnel on vSRX and vJunos-router

ssasso:
Tested and working wih both.

vSRX ofc requires to be in packet-based mode (or have firewall zones/rules definition).


ipspace (approved):
As expected, it all works. Thank you, merging...

PR #3575, opened 2026-07-03
Convert built-in plugins to module names

ipspace:
The built-in plugins are now structured as Python modules (for example, bgp.session plugin is in netsim/extra/bgp/session directory). That change required:

* Additional plugin loading code (if the "traditional" plugin cannot be found, we try to load a module). As the 'package:extra' element has been removed from the search path, this should not impact the correct operation.
* The device configuration templates are no longer in 'a.b' directory under 'extra' directory but in 'a/b' one. Extra elements with the 'custom_slash_config' variable were added to custom template file names to cope with that
* Removal of additional mypy checking of 'plugin.py' files (the were all renamed to '__init__.py')

Breaking change:

* The 'proxy-arp' plugin has been renamed to 'proxy_arp'.


jbemmel:
> * The 'proxy-arp' plugin has been renamed to 'proxy_arp'.

Perhaps add a ```_redirect``` for this?

ipspace:
> > * The 'proxy-arp' plugin has been renamed to 'proxy_arp'.
> 
> Perhaps add a `_redirect` for this?

If we no longer load built-in plugins as individual source files, then it's impossible to have a 'proxy-arp' plugin that would redirect to 'proxy_arp'. One can, however, add that to your lab topology that uses proxy-arp plugin, or even create a user- or system plugin doing the redirect.

ipspace:
YOLO, merging...

PR #3576, opened 2026-07-04
Refactor initial tunnel implementation

ipspace:
* Move utils/tunnel.py to extra/tunnel/__init__.py
* Return interface list from 'select source interface' function, making if useful in environments with additional requirements
* Make tunnel.af check optional to support dual-stack-capable tunnels
* Move most of GRE processing to _p2p.py utility function, assuming it can be used (or adapted) to other tunnel implementations.

Also:

* Move all link-related utility functions to augment/links and move them to the top of that module

### DIFF
--- a/netsim/augment/links.py
+++ b/netsim/augment/links.py
@@ -21,6 +21,61 @@ from . import addressing, devices
 VIRTUAL_INTERFACE_TYPES: typing.Final[typing.List[str]] = [
   'loopback', 'tunnel', 'lag', 'svi' ]
 
+# Utility functions
+#
+# - get_link_by_index: returns a link given its linkindex
+# - get_linkname: returns a link name given its linkindex
+# - get_next_linkindex: returns a unique linkindex for a new link
+# - set_linkindex: Assign unique linkindex to every link
+# - set_linkname: Assign names to all links without them
+
+def get_link_by_index(topology: Box, idx: int) -> typing.Optional[Box]:
+  '''
+  Find a link from its linkindex
+  '''
+  for link in topology.links:
+    if link.linkindex == idx:
+      return link
+  return None
+
+def get_linkname(topology: Box, linkindex: int) -> str:
+  '''
+  Get the name of a link referenced in an interface.
+
+  Always returns string ('unknown' when the link cannot be found)
+  '''
+  link = get_link_by_index(topology,linkindex)
+  return link._linkname if link is not None else 'unknown'
+
+def get_next_linkindex(topology: Box) -> int:
+  '''
+  Get a new linkindex -- either the last linkindex + 1 or default value
+  '''
+  if not topology.links:
+    topology.links = []
+    return topology.defaults.get('link_index',1)
+
+  return topology.links[-1].linkindex + 1
+
+def set_linkindex(topology: Box) -> None:
+  '''
+  set_linkindex -- set link index for each link
+  '''
+  linkindex = topology.defaults.get('link_index',1)
+  for link in topology.links:
+    link.linkindex = linkindex
+    linkindex = linkindex + 1
+
+def set_linknames(topology: Box) -> None:
+  '''
+  set_linknames -- set link name if not defined
+  '''
+  for cnt,link in enumerate(topology.links):
+    if link.get('group'):
+      link._linkname = f'links[{link.group}]'
+    elif not '_linkname' in link:
+      link._linkname = f'links[{cnt+1}]'
+
 '''
 We cannot definitely say whether a node name is valid before the component
 expansion is complete. We can, however, do basic sanity checks.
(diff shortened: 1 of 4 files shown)


netsim/augment/links.py
```diff
@@ -21,6 +21,61 @@
...
+  return None
+
+def get_linkname(topology: Box, linkindex: int) -> str:
+  '''
+  Get the name of a link referenced in an interface.
+
+  Always returns string ('unknown' when the link cannot be found)
+  '''
+  link = get_link_by_index(topology,linkindex)
+  return link._linkname if link is not None else 'unknown'
```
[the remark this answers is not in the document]
ipspace: We're assuming that each link has a _linkname

netsim/extra/tunnel/gre/__init__.py
```diff
@@ -5,7 +5,9 @@
 from netsim.augment import links as _links
 from netsim.augment import nodes as _nodes
 from netsim.utils import log
-from netsim.utils import tunnel as _tunnel
+
```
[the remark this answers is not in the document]
ipspace: Fixed (even though ruff did NOT fail 🤷‍♂️)

ipspace:
@jbemmel -- Looking at your wireguard code, there was a lot of duplication with the GRE code, so I refactored the whole thing and created a bunch of (hopefully useful) _p2p functions.

Also, the "select source interface" now returns a list of viable interfaces (and ignores AF if it's not set) which you can then use to do smarter interface selection.

netsim/extra/tunnel/__init__.py
```diff
@@ -7,12 +7,19 @@
...
+from ...data import get_empty_box, global_vars
+from ...utils import log
 
 
+def init(topology: Box) -> typing.NoReturn:
+  '''
+  Just in case someone has a great idea to use the 'tunnel' plugin, we
+  have to set them straight
+  '''
+  log.fatal("You cannot use the 'tunnel' plugin. Use a plugin for the tunnel mode you're interested in")
```
jbemmel, line 21: Could list the available tunnel types explicitly, or provide an example like "Use 'tunnel.gre' or similar"

jbemmel (approved):
no comment

netsim/extra/tunnel/__init__.py
ipspace: That wouldn't be exactly trivial, and this is just a safeguard against someone making a typo. I would say RTFM applies in that case ;)

netsim/extra/tunnel/gre/__init__.py
```diff
@@ -21,92 +20,17 @@ def pre_link_transform(topology: Box) -> None:
...
-          f'Cannot get {intf.tunnel.af} tunnel source for link {get_linkname(topology,intf.linkindex)} on node {node}',
-          more_data=f'Tunnel source data: {intf.tunnel.source if "tunnel.source" in intf else "none"}',
-          category=log.MissingDependency,
-          module='tunnel')
-        continue
-
-      if 'name' in intf and '->' in intf.name and 'GRE' not in intf.name:
-        intf.name += ' [GRE tunnel]'
+  node_iflist = _p2p.feature_check(topology,t_mode='gre',t_desc='GRE tunnels')
+  _p2p.tunnel_source(topology,node_iflist,default_af='ipv4',t_name='GRE')
```
jbemmel, line 31: The current code requires the user to explicitly specify ```tunnel.af: ipv6``` in an IPv6-only setup. I would suggest we infer ```tunnel.af``` from the available address families on the selected interface, preferring ```ipv4``` if both are available. This behavior should be consistent across tunnel plugins

ipspace: You know that uneasy feeling I had when looking at #3541?

I get a similar feeling considering auto-AF for GRE tunnels. I can't tell you what could break, but I'll stay with the current "ipv4 is the default" behavior.

jbemmel: > I get a similar feeling considering auto-AF for GRE tunnels. I can't tell you what could break, but I'll stay with the current "ipv4 is the default" behavior.

I have no problem with that, just thinking that we might want to do that like this?
```
attributes:
  link:
    tunnel:
      af:                     # Transport address family
        { type: str, valid_values: [ ipv4, ipv6 ], _default: ipv4 }
```
rather than bury it inside Python logic?

ipspace: And what would happen when someone uses tunnel.gre and tunnel.wireguard in the same topology?

jbemmel: Nothing, this section would be under GRE not affecting Wireguard

ipspace: Are you absolutely sure about that? Want to try it out first?

PR #3577, opened 2026-07-05
[FRR] Bugfix log redirect only when debug options are set

jbemmel:
To avoid issues with SELinux deny policies, only redirect when debugging options are set (i.e. off by default)

Addresses #3574


ipspace (approved):
Looks good to me. Thank you!

PR #3583, opened 2026-07-06
Use node/group variables for OpenBSD username/password

snuffy22:
Allow for users to have different  `ansible_ssh_user`  and `ansible_ssh_pass`  for OpenBSD Vagrant image.

### DIFF
--- a/netsim/templates/provider/libvirt/openbsd-domain.j2
+++ b/netsim/templates/provider/libvirt/openbsd-domain.j2
@@ -1,8 +1,8 @@
     {{ name }}.vm.synced_folder ".", "/vagrant", id: "vagrant-root", disabled: true
 
     {{ name }}.ssh.insert_key = false
-    {{ name }}.ssh.username = "openbsd"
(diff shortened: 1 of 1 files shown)


ipspace:
Good catch. This thing needs to be fixed consistently, so I changed the `define-domain.j2` configuration template to set SSH username/password when defined.

Looking at the code, it should work for all devices; running initial configuration integration test across all devices I'm testing (which will take a while), then I'll create a separate PR and ask everyone involved with libvirt device definitions for their opinions.

ipspace:
> Good catch. This thing needs to be fixed consistently, so I changed the `define-domain.j2` configuration template to set SSH username/password when defined.

... and failed miserably on Arista EOS :(( More details coming in a blog post. I hate the world that can't get something as trivial as SSH authentication to work.

ipspace (approved):
I would prefer having "if ansible_user/ansible_ssh_pass is defined" guards around the code, but it works with the default settings, so we're good to go.

PR #3586, opened 2026-07-06
VyOS: reflect new directory structure for GRE

ssasso:
@ipspace related to what @jbemmel reported in #3566

### DIFF
--- a/netsim/ansible/templates/initial/vyos.j2
+++ b/netsim/ansible/templates/initial/vyos.j2
@@ -14,7 +14,7 @@ set system host-name 'vyos-{{ inventory_hostname | replace("_","-") }}'
 {% include 'vyos.vlan.j2' %}
 {% endif %}
 {% for x_dp in config|default([]) %}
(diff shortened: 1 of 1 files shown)


ipspace:
Hey, thanks a million, but I have a more generic fix ready for a PR (will send it over) which should work also with user-modified templates.

Anyway, do GRE tunnels work for you without this fix, or has my server gone to the la-la land?

ssasso:
> Anyway, do GRE tunnels work for you without this fix, or has my server gone to the la-la land?

no, without this fix the proper template was not included (but I had to clean my "old" directory structure because I do a rsync from dev to host where I run the stuff). maybe it is the same case as yours :)

ipspace:
> no, without this fix the proper template was not included (but I had to clean my "old" directory structure because I do a rsync from dev to host where I run the stuff). maybe it is the same case as yours :)

That was the first thing I did -- the old directories are no longer present. I also ran the tests with a different netlab copy from a "clean" dev branch. Anyway, good to know the fix is mandatory, PR coming...

PR #3587, opened 2026-07-06
Fix: Recognize a.b or a/b in included VyOS dataplane templates

ipspace:
VyOS initial device configuration includes interface configuration code from plugin configuration templates. The plugin name in the node data is sometimes in 'a.b' format (for example, 'tunnel.gre'), whereas the new directory structure uses slashes (i.e. 'tunnel/gre').

This fix tries both names to allow users to override plugin dataplane templates in their own 'tunnel.gre' directory.

### DIFF
--- a/netsim/ansible/templates/initial/vyos.j2
+++ b/netsim/ansible/templates/initial/vyos.j2
@@ -13,8 +13,18 @@ set system host-name 'vyos-{{ inventory_hostname | replace("_","-") }}'
 {% if vlans is defined %}
 {% include 'vyos.vlan.j2' %}
 {% endif %}
+{#
+    Include interface-related parts of plugin configuration
+#}
 {% for x_dp in config|default([]) %}
-{%  include x_dp + '/vyos.initial.j2' ignore missing %}
+{#
+    With the restructuring of plugin directories, we have to try
+    plugin name with '.' and '/'. Fortunately, Jinja2 provides a
+    nifty mechanism to do that.
+#}
+{%  set dp_initial = x_dp + '/vyos.initial.j2' %}
+{%  set dp_alt = x_dp.replace('.','/') + '/vyos.initial.j2' %}
+{%  include [ dp_initial, dp_alt ] ignore missing %}
 {% endfor %}
 
 {% if loopback.ipv4 is defined %}


ipspace:
@ssasso -- Are you OK if I extend the same approach to VLANs and VRFs? That would clean up the "initial" directory.

ssasso:
> @ssasso -- Are you OK if I extend the same approach to VLANs and VRFs? That would clean up the "initial" directory.

Yes please

netsim/ansible/templates/initial/vyos.j2
```diff
@@ -13,8 +13,18 @@ set system host-name 'vyos-{{ inventory_hostname | replace("_","-") }}'
...
-{%  include x_dp + '/vyos.initial.j2' ignore missing %}
+{#
+    With the restructuring of plugin directories, we have to try
+    plugin name with '.' and '/'. Fortunately, Jinja2 provides a
+    nifty mechanism to do that.
+#}
+{%  set dp_initial = x_dp + '/vyos.initial.j2' %}
+{%  set dp_alt = x_dp.replace('.','/') + '/vyos.initial.j2' %}
+{%  include [ dp_initial, dp_alt ] ignore missing %}
 {% endfor %}
```
jbemmel, line 28:
```
{% for x_dp in config|default([]) %}
{%   for dir in [ x_dp, x_dp.replace('.', '/') ] %}
{%     include dir + '/vyos.initial.j2' ignore missing %}
{%   endfor %}
{% endfor %}```

Comment (ipspace) [reply netsim/ansible/templates/initial/vyos.j2:28]: ... and if the server behaves as weirdly as mine does, you'll get two copies of the same config commands. That's why I went for a list in the "include" directive.

**Comment by ipspace** [comment]
> Yes please

Will do after merging this one. Does it work for you?

File: netsim/ansible/templates/initial/vyos.j2 (hunk 1 above)
Comment (jbemmel) [reply netsim/ansible/templates/initial/vyos.j2:28]: Understood, in that case:
```
{% macro include_plugin_initial_configs(device_name) -%}
  {% set template = '/' + device_name + '.initial.j2' %}
  {% for x_dp in config|default([]) %}
    {% include [ x_dp + template, x_dp.replace('.', '/') + template ] ignore missing %}
  {% endfor %}
{%- endmacro %}
```

so we can use it for `frr` too?

ipspace: You're reading my mind. The next step will be a generic (macro) solution that will work with all devices, and I'll use it to clean up a few VLAN/VRF templats.

However, I still think we don't need it for FRR ;)) but we already have that discussion in #3585

ssasso:
> > Yes please
> 
> Will do after merging this one. Does it work for you?

Yes absolutely, no problem

netsim/ansible/templates/initial/vyos.j2
jbemmel: > However, I still think we don't need it for FRR ;)) but we already have that discussion in #3585

I guess I'll have to disagree on multiple fronts then, perhaps it's more effective that way ;)

PR #3589, opened 2026-07-07
FRR: Unify Linux interface sysctl settings

ipspace:
FRR configuration templates applied inconsistent interface sysctl settings to physical and VLAN interfaces (and none to gre tunnels).

This commit creates a shared macro that sets Linux interface sysctl parameters (primarily IPv6), MTU, and interface state. That macro can be used by all Linux-based devices (BIRD, FRR, Linux) in the future (its current use is limited to FRR).

Also:
* Had to change the search paths for the custom config templates to include the main config templates, or the custom configs (plugins) wouldn't be able to use macros from the main templates.

### DIFF
--- a/netsim/ansible/templates/initial/frr.j2
+++ b/netsim/ansible/templates/initial/frr.j2
@@ -1,4 +1,5 @@
 {% from "initial/linux/create-bond.j2" import create_bond_dev %}
+{% from "initial/linux/intf-sysctl.j2" import linux_intf_sysctl with context %}
 #!/bin/bash
 #
 set -e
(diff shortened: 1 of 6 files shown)


netsim/defaults/paths.yml
```diff
@@ -20,6 +20,7 @@ custom:                         # Custom configuration templates
   - "~/.netlab"
   - "/etc/netlab"
   - "package:extra"
+  - "package:ansible/templates" # Give custom configs access to main configuration macros
```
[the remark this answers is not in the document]
ipspace: Good catch ;) Fixed.

netsim/ansible/templates/initial/frr.j2
```diff
@@ -1,4 +1,5 @@
 {% from "initial/linux/create-bond.j2" import create_bond_dev %}
+{% from "initial/linux/intf-sysctl.j2" import linux_intf_sysctl with context %}
```
jbemmel: Looks like it should work also without context? All parameters are explicit

netsim/ansible/templates/initial/frr.j2
```diff
@@ -105,21 +106,12 @@ if [ ! -e /sys/class/net/{{ i.ifname }} ]; then
...
-{%   else %}
-{%     set ra_flag = '0' if role == 'router' else '1' %}
-sysctl -qw net.ipv6.conf.{{ i.ifname }}.autoconf={{ ra_flag }}
-sysctl -qw net.ipv6.conf.{{ i.ifname }}.accept_ra={{ ra_flag }}
-{%   endif %}
-{%   if i.mtu is defined %}
-ip link set dev {{ i.ifname }} mtu {{ i.mtu }}
-{%   endif %}
-ip link set {{ i.ifname }} {{ "down" if i.shutdown|default(False) else "up" }}
+{{ linux_intf_sysctl(i,role|default('router')) }}
```
jbemmel: ```role``` should always be set due to ```role: router``` in ```frr.yml```, no need to re-define that default

netsim/defaults/paths.yml
```diff
@@ -20,6 +20,8 @@ custom:                         # Custom configuration templates
   - "~/.netlab"
   - "/etc/netlab"
   - "package:extra"
+  extra_dirs:
```
jbemmel, line 23: Call this ```extra_j2_dirs``` to clarify intent?

jbemmel (approved):
Should work, but has some superfluous elements

netsim/defaults/paths.yml
ipspace: Meh 🤷‍♂️

PR #3592, opened 2026-07-07
Draft feature: Add multi-topology file merge support

jbemmel:
Allow netlab create and related commands to accept multiple topology files that are merged into a single lab definition. 

Includes an example GRE integration test split into underlay and overlay topologies.

Context: #3588


jbemmel:
Closing for now

PR #3594, opened 2026-07-07
Add BIRD OSPF default route origination support.

jbemmel:
Export default routes via OSPF filters with E1/E2 metrics, use static routes for always-originate, and restrict conditional origination to BGP and static sources to avoid re-advertising OSPF-learned defaults.

### DIFF
--- a/docs/module/ospf.md
+++ b/docs/module/ospf.md
@@ -55,7 +55,7 @@ The following table describes the per-platform support of individual router-leve
 | ------------------------ |:-:|:-:|:-:|:-:|:-:|
 | Arista EOS               | ✅| ✅| ✅| ✅| ✅|
 | Aruba AOS-CX             | ✅| ✅| ✅| ✅| ✅|
-| BIRD                     | ✅| ✅| ✅| ✅| ❌ | 
+| BIRD                     | ✅| ✅| ✅| ✅| ✅|
 | Cisco ASAv               | ✅| ✅| ❌ | ❌ | ✅|
 | Cisco IOSv/IOSvL2        | ✅| ✅| ✅| ✅| ✅|
 | Cisco IOS XE[^18v]       | ✅| ✅| ✅| ✅| ✅|
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -68,6 +68,7 @@ features:
     timers: true
   ospf:
     import: [ bgp, connected, static, vrf ]
+    default: true
     password: true
     priority: true
     timers: true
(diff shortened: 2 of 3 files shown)


ipspace (changes requested):
Have to go through the OSPF template again, but it looks OK and if there are minor details, I can fix them before merging.

However, if you have to add a static route to generate the default, it would be really nice to make it a floating static route (low preference/high admin distance/whatever it's called), so that if we get a different default route from somewhere else, we prefer that one.

Also, you need the (floating) static default route only when **ospf.default.always** is True (https://netlab.tools/module/ospf/#specifying-external-default-route)

jbemmel:
> However, if you have to add a static route to generate the default, it would be really nice to make it a floating static route (low preference/high admin distance/whatever it's called), so that if we get a different default route from somewhere else, we prefer that one.

OSPF routes from lab ospfv2 (tests/integration):
```
probe (FRR) — OSPF route table
N    10.0.0.3/32           [0] area: 0.0.0.0  → lo
N    172.16.0.0/24         [10] area: 0.0.0.0 → eth1
N    172.16.1.0/24         [10] area: 0.0.0.0 → eth2
R    10.0.0.1              [10] ASBR via 172.16.0.1 (dut_a)
R    10.0.0.2              [10] ASBR via 172.16.1.2 (dut_c)
N E2 0.0.0.0/0             [10/40] via 172.16.0.1, eth1   ← default from dut_a
dut_a (Bird) — OSPF routes (+ floating static default)
0.0.0.0/0     unreachable [ospf_default_global_v2] (1)   ← floating static, pref 1
              * selected: kernel via 192.168.121.1 (10)
10.0.0.3/32   * I [ospf_global_v2] (150/10) via 172.16.0.3
172.16.0.0/24   I [ospf_global_v2] (150/10) on eth1
172.16.1.0/24 * I [ospf_global_v2] (150/20) via 172.16.0.3
Neighbor: 10.0.0.3 Full/PtP on eth1.

dut_c (Bird) — OSPF routes
0.0.0.0/0     * E2 [ospf_global_v2] (150/20/40) [10.0.0.1] via 172.16.1.3
10.0.0.3/32   * I  [ospf_global_v2] (150/10) via 172.16.1.3
172.16.0.0/24 * I  [ospf_global_v2] (150/20) via 172.16.1.3
172.16.1.0/24   I  [ospf_global_v2] (150/10) on eth1
xf has no OSPF (BGP only). The floating static on dut_a is present at preference 1 and is losing to the kernel default, as intended.
```

Your wish is my PR ;)

ipspace (approved):
The "OSPFv3 VRF default route" test is failing for me; the "::/0" route (previously advertised by BGP neighbor) is stuck in the kernel and gets readvertised via OSPFv3.

I simplified your "default route origination" test; if the device has any default route, it should advertise it in OSPF, and as OSPF routes do not get readvertised in OSPF, that's safe (and it doesn't affect the OSPFv3 VRF failure anyway).

I think the problem is the "import" statement in the VRF kernel table. I will merge this PR, change that parameter, and then rerun the VRF tests.

ipspace:
> Your wish is my PR ;)

😂😂 -- thank you, made my day ;)

PR #3595, opened 2026-07-07
Add BIRD OSPF graceful restart support

jbemmel:
## Summary
- Render BIRD OSPF `graceful restart` settings from `ospf.gr.restart` and `ospf.gr.helper`
- Declare IPv4/IPv6 OSPF GR capability for the `bird` daemon
- Document BIRD in the OSPF graceful restart platform table

## Test plan
- [x] `netlab create` on a topology with `device: bird` and `ospf.gr.restart` / `ospf.gr.helper`
- [x] Verify generated `ospf.mod.conf` contains `graceful restart on` or `graceful restart aware`
- [x] Confirm `ospf.gr` validation passes for bird nodes with IPv4 and IPv6

### DIFF
--- a/docs/module/ospf.md
+++ b/docs/module/ospf.md
@@ -114,6 +114,7 @@ The following devices support OSPF graceful restart:
 
 | Operating system | OSPFv2 | OSPFv3 |
 | ---------------- | :--: | :--: |
+| BIRD             |  ✅  |  ✅  |
 | Fortinet FortiOS |  ✅  |  ✅  |
 | FRR              |  ✅  |  ✅  |
 
--- a/netsim/daemons/bird.yml
+++ b/netsim/daemons/bird.yml
@@ -69,6 +69,7 @@ features:
   ospf:
     import: [ bgp, connected, static, vrf ]
     default: true
+    gr: [ ipv4, ipv6 ]
     password: true
     priority: true
     timers: true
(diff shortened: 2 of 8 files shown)


jbemmel:
BIRD establishes the adjacency quickly, but the GR helper seems to need a few moments more:
```
netlab connect dut birdcl

show protocols
debug ospf_global_v3 { states,routes,events }
echo all
```

Added a 5-second extra wait for BIRD only, which stabilizes the test. Tried to find a more explicit state to watch for, but no luck

Also improved the GR reset logic at the end, to enable repeated GR tests

ipspace (changes requested):
* Please rebase, we have merge conflicts.
* If you decided to change the GR logic, wouldn't it be better to do `pkill -9 ospf6d` in `gr_stop` step?

jbemmel:
* Rebased
* Use pkill -9 and make both cases consistent; document one-shot nature

ipspace (approved):
OSPFv2 consistently fails for me, OSPFv3 works, so I'll merge it and play with the timing.

jbemmel:
> OSPFv2 consistently fails for me, OSPFv3 works, so I'll merge it and play with the timing.

For me it was the other way around, I did test it again and on my setup it works. We may need to add the delay on both cases

ipspace:
> For me it was the other way around

You added delay for OSPFv3, but not for OSPFv2

> I did test it again and on my setup it works.

The devil is in the (timing) details. Looks like my server is a bit faster than yours, so the test gets to that point before BIRD is ready.

> We may need to add the delay on both cases

Done.

PR #3598, opened 2026-07-09
Fix missing 'f' for Python f-strings

snuffy22:
Went hunting for these missing 'f' f-strings after seeing a few in augment.

Was able to still run netlab and do the general tests, 
Though have not tested trying to reach all these now fixed f-string locations.


ipspace (approved):
Thank you, really appreciated ;)

They all look good to my untrained eye ;)) Let me run Copilot over them and then merge this PR.

PR #3599, opened 2026-07-09
OSPFv3 default route integration tests must be run with routers

ipspace:
Some devices require a default route in their routing table to originate an OSPF default route. If we decide to add a static default route to support that functionality, then it's better to make it a floating static route (so a real default route can take over).

However, in the IPv6 world, a "better" default route could be the RA default route, so we have to set 'role: router' on tested devices to ensure they do NOT listen to RA updates.


ipspace:
@jbemmel -- this was the root cause of the problems with #3594

jbemmel:
> @jbemmel -- this was the root cause of the problems with #3594

I keep forgetting that the default BIRD role isn't ```router``` - to me that is the root cause of the root cause...

PR #3624, opened 2026-07-11
Fix: IOL L2 cannot configure BFD for OSPFv3

ipspace:


jbemmel:
Could be modeled in device features 

‘’’
bfd: [ ipv4 ]
’’’

ipspace:
> Could be modeled in device features

That's true. However, it seems that BFD works for BGP over IPv6, so it's really just an OSPFv3 glitch.

Have to add IPv6 checks to `13-bfd.yml`

jbemmel:
> That's true. However, it seems that BFD works for BGP over IPv6, so it's really just an OSPFv3 glitch.

Yes, I meant:
```
ospf:
  bfd: [ ipv4 ]
```

We currently have ```bfd``` top level and under ```bgp```, but not under ```ospf```

ipspace:
> We currently have `bfd` top level and under `bgp`, but not under `ospf`

Got it. However, as all other devices for which we implemented BFD work with OSPF (now that I fixed "a few" things) and everything but IOL L2 works with OSPFv3 (when OSPFv3 works), I might just leave it as a quirk.

PR #3642, opened 2026-07-16
Update Tunnel tests

snuffy22:
First go at writing some more complex tests.

Added the following tests:
* IPv6 GRE 
* IPv6 over IPv4 GRE (using link locals)
* IPv4 over IPv6 GRE

Renumbered tests to allow for more GRE ones.

As for my topology files, I'm sure there is a better way that what I have done but hopefully I got the idea across..
The X over Y I wanted to ensure only the tunnel + R3 had X protocol on them.

Test results for FRR
```
(py3-snuff) netlab@netlab:~/snuffy-netlab/tests/integration$ ./device-module-test -p clab -d frr tunnel/
Pre-test cleanup:
16:32:02 tunnel/01-gre.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:32:26 tunnel/02-gre-vrf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:32:49 tunnel/03-gre-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:33:13 tunnel/04-gre-ipv4-over-ipv6.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:33:37 tunnel/05-gre-ipv6-over-ipv4.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:34:03 tunnel/10-wireguard.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
16:34:36 tunnel/11-wireguard-vrf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
```


ipspace:
Hey, I told you (in an email) that I'll make the existing GRE test dual-stacked and that you should submit just the OpenBSD part. Let me reiterate (again) that I don't want to run too many integration tests, particularly when devices like vjunos-router take minutes to boot (for each test).

The dual-stack GRE test is in #3648. Please use it to test your OpenBSD implementation; I can't merge it yet because I want to have a check for IPv6 transport AF in there as well.

snuffy22:
No problems, I have made a EOS/RouterOS7 one as well..
Will test all 3 and put up a PR after testing.

jbemmel:
> Let me reiterate (again) that I don't want to run too many integration tests, particularly when devices like vjunos-router take minutes to boot (for each test).

I’m thinking it may be time for a hierarchy of test sets. I think it would be valuable to have some additional detailed tests for combinations of certain features, but you would not want to run all of them every integration test cycle.

Another aspect is that we may be able to run multiple test scenarios in sequence and/or in parallel. If we orchestrate carefully and standardize our test topology a bit more, we could either reconfigure sequences of tests one after another, or create combined configs that can do all - say using different vlans, vrfs, etc.

Figuring out elegant models and mechanisms for this would be an interesting challenge, and very valuable imho

ipspace:
> I’m thinking it may be time for a hierarchy of test sets. I think it would be valuable to have some additional detailed tests for combinations of certain features, but you would not want to run all of them every integration test cycle.

We already have that. Anything that does not start with `[0-9]` can only be triggered manually.

> Another aspect is that we may be able to run multiple test scenarios in sequence and/or in parallel.

Already doing that. Still doesn't help that certain devices take forever to boot.

> If we orchestrate carefully and standardize our test topology a bit more, we could either reconfigure sequences of tests one after another, or create combined configs that can do all - say using different vlans, vrfs, etc.

Of course we could write a giant test that would test "everything" in one go, but then (A) it would be fun trying to figure out what went wrong and (B) the test would be a nightmare to maintain. I've been there before, and have no intention of revisiting that particular chamber of hell.

> Figuring out elegant models and mechanisms for this would be an interesting challenge, and very valuable imho

When you find someone with the willpower to reimplement the existing tests, and reliably maintain the resulting Rube Goldberg machine forever, do let me know.

PR #3643, opened 2026-07-16
Acls

DanPartelly:
initial ACL implementation.

- both ipv4 and ipv6 ACLs
- support for matching on protocol,  source address, destination address
- support for any and host keywords as booleans
- support for matching on ports and port ranges with eq, lt, gt, neq operators for ports, and in, not_in for ranges
- not_in operator is unconditionally expanded into two consecutive lt/gt entries. 
- support for established and log keywords
- addresses can be specified manually, or using any of the following:  a named prefix , an addressing pool or the  tuples  node_id/interface name and node_id / link role 
- support for descriptions. entries containing descriptions are automatically expanded in two entries. One for the description, and one for the acl entry proper. The description entry retains the sequence number of the original entry. For the original entry seq is incremented by one . One NOSs which support sequenced comments, they will be properly displayed. 
- initial support for Cisco iosv, iosvl2, iol, ioll2 and Arista ceos
- ipv4 / ipv6 tests using netcat


netsim/modules/routing.yml
```diff
@@ -243,6 +270,91 @@ _top:                               # Modification of global defaults
...
+          type: int
+          min_value: 0
+          max_value: 65535
+          _required: True
+        max:
+          type: int
+          min_value: 0
+          max_value: 65535
+          _required: True
+      _invalid_with: { op: [eq, gt, lt, neq] }
```
jbemmel: indent

DanPartelly:
This is largely complete now. Needs tests and doc.

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,46 @@
...
+{%- else -%}any{%- endif -%}
+{%- endmacro %}
+
+{% macro r_acl_port(p) -%}
+{%- if p.port_range is defined %} range {{ p.port_range.min }} {{ p.port_range.max }}
+{%- elif p.port is defined %} {{ p.port_op | default('eq') }} {{ p.port }}
+{%- endif -%}
+{%- endmacro %}
+
+{%if routing.acl is defined %}
```
jbemmel: indent (nit)

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,46 @@
...
+{% for name, acl in routing.acl.items() %}
+!
+{% set cmd = "ipv6" if acl.af == "ipv6" else "ip" %}
+{{cmd}} access-list extended {{ name }}_{{ acl.af }}
+{% for e in acl.value %}
+{% set est = "established" if e.established is defined else "" %}
+{% set log = "log" if e.log is defined else "" %}
+{% if e.description is defined %}
```
jbemmel: nested if should be indented by 2 for readability

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,46 @@
+{% macro r_acl_addr(a, af) -%}
+{%- if a.any is defined and a.any -%}any
+{%- elif a.host is defined and a.host -%}host {{ (a.ipv4 if af == 'ipv4' else a.ipv6) | ipaddr('address') }}
+{%- elif af == 'ipv4' and a.ipv4 is defined -%}{{ a.ipv4 | ipaddr('network') }} {{ a.ipv4 | ipaddr('hostmask') }}
+{%- elif af == 'ipv6' and a.ipv6 is defined -%}{{ a.ipv6 }}
+{%- else -%}any{%- endif -%}
```
jbemmel: This is a bit odd, seems like an inconsistency - fail?

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,46 @@
...
+!
+{% set cmd = "ipv6" if acl.af == "ipv6" else "ip" %}
+{{cmd}} access-list extended {{ name }}_{{ acl.af }}
+{% for e in acl.value %}
+{% set est = "established" if e.established is defined else "" %}
+{% set log = "log" if e.log is defined else "" %}
+{% if e.description is defined %}
+  {{e.sequence}} remark {{e.description}}
+{% else %}
+  {{ e.sequence}} {{ e.action}} {{e.protocol}} {{r_acl_addr(e.src, acl.af)}} {{ r_acl_port(e.src)}} {{ r_acl_addr(e.dst, acl.af) }} {{ r_acl_port(e.dst)}} {{est}} {{log}}
```
jbemmel: inconsistent whitespace...

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,46 @@
...
+{% endif %}
+{% endfor %}
+{% endfor %}
+{% endif %}
+!
+{% for intf in interfaces %}
+interface {{ intf.ifname }}
+{%if intf.routing.acl.in.ipv4 is defined %}
```
jbemmel: This could be a double for loop over [ip,ipv6] and [in,out]

netsim/devices/ios.yml
```diff
@@ -137,8 +137,10 @@ features:
       vrf: True
       inter_vrf: True
       discard: True
+    acl:
+      not_in: False
```
jbemmel: I realize I was the one suggesting this, but if none of the current platforms support it we could also simply leave it out for now, and then add it whenever we find a platform that does

jbemmel (changes requested):
Indentation and whitespace should be made at least internally consistent and ideally in line with the rest of .j2 scripts

You may want to omit ```not_in``` support for now, and then add it whenever there is a need later. There is little point in calling out things that are not supported by any device.

netsim/devices/ios.yml
DanPartelly: I am not aware of any device supporting not_in directly. When you mentioned the operator, I believed one of your platforms supports it directly. Ill remove the flag.  And all in all, it's a nice feature, so yeah, thanks for suggesting it.

DanPartelly: resolved

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,44 @@
+{% macro r_acl_addr(a, af) -%}
+{%- if a.any is defined and a.any %} any
+{%- elif a.host is defined and a.host %} host {{ (a.ipv4 if af == 'ipv4' else a.ipv6) | ipaddr('address') }}
+{%- elif af == 'ipv4' and a.ipv4 is defined %} {{ a.ipv4 | ipaddr('network') }} {{ a.ipv4 | ipaddr('hostmask') }}
+{%- elif af == 'ipv6' and a.ipv6 is defined %} {{ a.ipv6 }}
+{%- endif -%}
+{%- endmacro %}
```
jbemmel: I wonder if this '{%-' is needed since you have '-%}' on the endif above

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,44 @@
...
+
+{% if routing.acl is defined %}
+{%  for name, acl in routing.acl.items() %}
+{%    set cmd = "ipv6" if acl.af == "ipv6" else "ip" %}
+{%    set seq = "sequence " if acl.af == "ipv6" else "" %}
+{%    set ext = " extended" if acl.af == "ipv4" else "" %}
+{{cmd}} access-list{{ext}} {{name}}_{{ acl.af }}
+{%    for e in acl.value %}
+{%      set est = " established" if e.established is defined else "" %}
```
jbemmel, line 22: These 2 can be moved down into the else branch, closer to where they are used

netsim/ansible/templates/routing/ios.j2
```diff
@@ -40,15 +41,15 @@ ip community-list {{ c_value.cl_type }} {{ c_name }} {{ c_line.action }} {{ c_li
...
 {%   set leak_global = ' global' if sr_data.nexthop.vrf is defined and sr_data.nexthop.vrf is none else '' %}
 {%   if 'ipv4' in sr_data %}
-ip route {{ cmd_vrf }}{{ 
+ip route {{ cmd_vrf }}{{
     sr_data.ipv4|ansible.utils.ipaddr('network') }} {{
-    sr_data.ipv4|ansible.utils.ipaddr('netmask') }} {{ 
-    cmd_intf }}{{ 
+    sr_data.ipv4|ansible.utils.ipaddr('netmask') }} {{
+    cmd_intf }}{{
     'Null0' if 'discard' in sr_data.nexthop else sr_data.nexthop.ipv4 }}{{ leak_global }}
```
jbemmel, line 48: is ```sr_data.nexthop.ipv4``` always defined?

DanPartelly: Not my code, I have no idea.

netsim/modules/routing/acl.py
```diff
@@ -0,0 +1,263 @@
...
+
+def expand_acl(p_name: str, o_name: str, node: Box, topology: Box) -> typing.Optional[list]:
+  ctx = validation_context(p_name=p_name, idx=0, af_ipv4=True, protocol=0, established=False)
+  acl_def = node.routing[o_name][p_name]
+  acl_list = acl_def.value
+  ctx.af_ipv4 = acl_def.af == "ipv4"
+  expansion: list = []
+
+  for idx, entry in enumerate(list(acl_list)):
+    ctx.protocol = resolve_protocol(entry.protocol)
```
jbemmel, line 207: Could skip if 0 (unmapped protocol)

netsim/modules/routing/acl.py
```diff
@@ -0,0 +1,263 @@
...
+
+def resolve_interface_acl_references(node: Box, topology: Box) -> None:
+  routing = topology.get("routing")
+  if routing is None:
+    return
+
+  global_acls = routing.get("acl", {})
+  node_acls = node.get("routing", {}).get("acl", {})
+
+  for intf in node.get("interfaces", []):
```
jbemmel, line 240: Platform allows one to define ACLs on loopback interfaces too, although they may not affect traffic.
Should document that ACLs on loopbacks are not supported

tests/integration/routing/26-acl-ipv4.yml
```diff
@@ -0,0 +1,134 @@
...
+    nohup nc  -l -k -p 100 >/dev/null 2>&1 &
+  h2: |
+    #!/bin/sh
+    for p in 30 100 150 151 ; do
+        nohup nc -l -k -p $p >/dev/null 2>&1 &
+    done
+
+validate:
+  ping_h2:
+    description: "h1->h2 ICMP echo must succed."
```
jbemmel: succeed

tests/integration/routing/27-acl-ipv6.yml
```diff
@@ -0,0 +1,147 @@
...
+      dst.port_range.max: 150
+      dst.port_op: in
+    - action: permit
+      protocol: tcp
+      src.node: h1
+      src.interface: eth1
+      dst.prefix: s_pfx
+      dst.port_range.min: 50
+      dst.port_range.max: 150
+      dst.port_op: not_in
```
jbemmel, line 59: I thought we removed this? Should trigger an error?

jbemmel (changes requested):
Few minor nits and optional alternatives

netsim/modules/routing/acl.py
DanPartelly: I consider routed  SVI acls, loopbacks, and their ilk out of scope for the first PR. And whatever we support local traffic filtering or not, we shall discuss in the next PR.  Hopefully the second PR will come before 27.08 release.

tests/integration/routing/27-acl-ipv6.yml
DanPartelly: I removed the flag only , and I choose to expand the operator ina gt/lt pair for all supported devices.  I like  it, so Ill keep the operator itself.

netsim/modules/routing/acl.py
DanPartelly: Protocol can be specified both as string or as a int  by the user.  The purpose of the function is to map the string to an int, or return the integer itself. The string is checked by the data model validation step, so this lookup table should be in sync with that.  The return value is kinda wrong , since 0 is a valid IANA protocol, but also harmless, since it's never checked for 0.

So far, the result of this function is only used to validate if the protocol is TCP for established or TCP/UDP for port operation. Validation is done later in 

`validate_acl_address_entry(p_entry: Box, ctx: validation_context) -> None:`

before this function is called, we also resolve the IP addreses by various means in expand_acl_address_entry(). then validation is called which does more than checking protocol, it ensures you have a valid ipv4 / ipv6 at this step as well

so skipping the loop would be the wrong thing to do 

Or maybe I misunderstand what you wanna say ?

jbemmel: Just pointing out that the function returns 0 on error, that should lead to skipping the entry or an error 

You may want to validate the input, instead of producing an acl and then validating that (unless I misunderstood)

DanPartelly: the protocol entry is validated in the data model with valid_values. and the lookup table is supposed to be kept in sync ,so that can never fail. 

But now I get your point, Ill fail loudly for the case someone changes the data model in the future.

DanPartelly: resolved in [6c9b0f4](https://github.com/ipspace/netlab/pull/3643/commits/6c9b0f4ed9a6a8f12f44bff23c63cb19df8aeda2)

netsim/ansible/templates/routing/acl.ios.j2
```diff
@@ -0,0 +1,48 @@
...
+{%    set cmd = "ipv6" if acl.af == "ipv6" else "ip" %}
+{%    set seqcmd = "sequence " if acl.af == "ipv6" else "" %}
+{%    set ext = " extended" if acl.af == "ipv4" else "" %}
+{{cmd}} access-list{{ext}} {{name}}_{{ acl.af }}
+{%    for e in acl.value %}
+{%      set est = " established" if e.established is defined else "" %}
+{%      set log = " log" if e.log is defined else "" %}
+{%      set seq = "" if netlab_device_type in ['ios','iosvl2'] else e.sequence %}
+{%      if e.description is defined %}
+  {{seqcmd}}{{seq}} remark {{e.description}}
```
DanPartelly, line 26: This will fail on iosv, iosvl2 for af ipv6.

DanPartelly:
@jbemmel If you have time, please check if all issues have been resolved in a way that is acceptable for you. And thank you for the time invested in reading all this.

netsim/ansible/templates/routing/acl.ios.j2
DanPartelly: The else branch won't ever be reached due to checks in Python code.  It is removed now

netsim/ansible/templates/routing/acl.ios.j2
DanPartelly: removed.

netsim/ansible/templates/routing/acl.ios.j2
DanPartelly: hopefully all indents are OK by now.

netsim/ansible/templates/routing/acl.ios.j2
DanPartelly: hopefully all indents are OK by now.

netsim/modules/routing/acl.py
DanPartelly: Lack of support for loopbacks was documented.

ipspace (changes requested):
I know I'm late to the party, and if you want to merge this as-is, we can do that as well, but I'd fundamentally change the data model:

* Keep ACLs multi-protocol
* Infer the "host" attribute from ipv4/ipv6 entries (or 32/128 prefixes)
* Split ACLs into IPv4 and IPv6 ACLs internally (similarly to what we do with route maps)
* Apply multi-AF ACLs to interfaces and deal with IPv4/IPv6 details internally.

@DanPartelly -- if you're OK with these ideas, I would do the necessary rewrites. In any case, the device templates wouldn't change; the device data model would stay as it is now (because we need all those details when configuring device ACLs)

DanPartelly:
@ipspace I am OK with those changes. They improve end user experience and make more please to express ACLs.

DanPartelly:
@ipspace I seen that you merged this. We manage the changes you want into another PR ? As I said above they improve end user experience and we should have them

ipspace:
> We manage the changes you want into another PR?

Yes. This saga was long enough ;)

> As I said above they improve end user experience and we should have them

Working on them, starting with #3718

PR #3652, opened 2026-07-17
fix(sr): render SRGB (segment-routing global-block) under router isis on FRR

roc-ops:
## Problem

`netsim/ansible/templates/sr/frr.j2` renders the SRGB (`segment-routing global-block`, derived from `sr.srgb.start`) **only** inside the `{% if 'ospfv2' in sr.protocol %}` block. Under IS-IS SR the SRGB is silently dropped — `sr.srgb.start` is a no-op for `isisd`, so a lab that sets a custom SRGB gets FRR default global block instead of the requested one.

## Fix

Mirror the existing guarded block into the `router isis` section. FRR `isisd` accepts `segment-routing global-block <lower> <upper>` with the same syntax as `ospfd`.

```diff
 {% if 'isis' in sr.protocol %}
 router isis {{ isis.instance }}
   segment-routing on
+{%   if sr.srgb.start is defined %}
+  segment-routing global-block {{ sr.srgb.start }} {{ sr.srgb.start + sr.srgb.size|default(8000) - 1 }}
+{%   endif %}
 {%   if isis.router_id is defined %}
```

## Test

Added `sr.srgb.start: 16000` to `tests/integration/sr/01-isis-ipv4-p.yml` so the IS-IS SRGB rendering path is exercised.

## Verification

Offline Jinja render of the template:
- IS-IS case now emits `segment-routing global-block 16000 23999` under `router isis`.
- OSPF case unchanged (no regression).

### DIFF
--- a/netsim/ansible/templates/sr/frr.j2
+++ b/netsim/ansible/templates/sr/frr.j2
@@ -15,6 +15,9 @@ cat >/tmp/config <<'CONFIG'
 {% if 'isis' in sr.protocol %}
 router isis {{ isis.instance }}
   segment-routing on
+{%   if sr.srgb.start is defined %}
+  segment-routing global-block {{ sr.srgb.start }} {{ sr.srgb.start + sr.srgb.size|default(8000) - 1 }}
+{%   endif %}
 {%   if isis.router_id is defined %}
   mpls-te router-address {{ isis.router_id }}
 {%   endif %}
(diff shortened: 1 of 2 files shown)


netsim/ansible/templates/sr/frr.j2
```diff
@@ -15,6 +15,9 @@ cat >/tmp/config <<'CONFIG'
 {% if 'isis' in sr.protocol %}
 router isis {{ isis.instance }}
   segment-routing on
+{%   if sr.srgb.start is defined %}
+  segment-routing global-block {{ sr.srgb.start }} {{ sr.srgb.start + sr.srgb.size|default(8000) - 1 }}
```
jbemmel, line 19: Not directly related to this PR, but I think it would be nice to have the sr module calculate the end label, instead of each template

ipspace:
Thanks a million. I will try to add an SRGB check to the integration test (to check that we're really configuring it) and then merge your PR.

ipspace:
@roc-ops -- Thanks again, and welcome to the netlab contributors club ;))

PR #3655, opened 2026-07-17
GRE: Transport address family check

ipspace:
Cisco IOL L2 image does not support GRE tunnels over IPv6, so the GRE tunnel plugin checks the transport AF support.

The "supported transport AF" check was added to the _p2p.feature_check shared function and is thus available to any tunnel plugin.


snuffy22:
This also applies for EOS ;)

PR #3665, opened 2026-07-19
Add Static routing to RouterOS7 

snuffy22:
Create the needed templates for RouterOS 7 to do static routing.

Does all types
* Static Global
* Static VRF
* Static Inter-VRF 
* Discard Routes

Passes all the static tests except the OSPFv3 because we do not yet support it.
```
(py-net) netlab@LAPTOP-8380MASB:~/snuffy-netlab/tests/integration$ ./device-module-test -p clab -d routeros7 -l '2[0-5]-static' routing
Pre-test cleanup:
12:59:03 routing/20-static.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
13:00:29 routing/21-static-vrf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
13:01:55 routing/22-static-inter-vrf.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
13:03:18 routing/23-static-indirect.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
13:04:43 routing/24-static-indirect-inter-vrf.yml: create(FAIL)
13:04:43 routing/25-static-discard.yml: create(ok) up(ok) config(ok) validate(ok) cleanup(ok) OK
```

### DIFF
--- a/docs/module/routing-static.txt
+++ b/docs/module/routing-static.txt
@@ -21,6 +21,7 @@ _netlab_ supports static routes on these platforms:
 | FRR                 | ✅ | ✅ | ✅ | ✅ |
 | Junos               | ✅ | ✅ | ✅ | ❌ |
 | Linux               | ✅ |  ❌ |  ❌ |  ❌ |
+| Mikrotik RouterOS 7 | ✅ | ✅ | ✅ | ✅ |
 | Nokia SR Linux      | ✅ | ✅ | ✅ |  ❌ |
 | Nokia SR OS[^SROS]  | ✅ | ✅ | ✅ |  ❌ |
 | OpenBSD             | ✅ | ✅ |  ❌ |  ❌ |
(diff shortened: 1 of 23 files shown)


snuffy22:
Reviewed why OSPFv3 was not working for VRF, just a small copy/paste n value tweak to get working.

Advise if we want to break that out into another change.

Also migrated the target-scope to 41 to ensure any static indirect happens after at all other protocols.

ipspace:
> Reviewed why OSPFv3 was not working for VRF, just a small copy/paste n value tweak to get working.

Nice job. However, now you configure OSPFv3 for IPv4-only networks (like vrf/11-multi-vrf-ospf). Please look at the bigger picture, not just whether the tests pass. We cannot possibly check all ways in which things can be misconfigured. For example, we do not configure OSPFv3 on probes in OSPFv2-only scenarios just to detect whether someone configured OSPFv3 when it's not needed. OK, maybe we should...

> Advise if we want to break that out into another change.

Please do.

> Also migrated the target-scope to 41 to ensure any static indirect happens after at all other protocols.

That's an interesting concept. I've never seen anything like that before.

snuffy22:
Will create the new change with just vrf ospfv3 (with af check) in new PR  as requested

ipspace (approved):
Everything works ;) Thank you!

PR #3673, opened 2026-07-20
Fix: Ensure Vagrant uses libvirt as provider for destroy command

snuffy22:
Fixes Vagrant preferring `virtualbox` for WSL Linux instances

We already tell Vagrant to use `libvirt` for bringing up virtual machines via the `-p libvirt` flag.

Unfortunately that flag does not exist for `vagrant destroy`.

Let us set environment variable `VAGRANT_DEFAULT_PROVIDER=libvirt` before running `vagrant destroy` so it uses correct provider.


ipspace (approved):
Thank you!

PR #3677, opened 2026-07-21
Fix: Move FRR 'netlab_linux_packages' to libvirt-only setup

ipspace:
The wireguard support (#3569) added Linux package installation to the clab path in the initial configuration script. Unfortunately, nobody (not the AI, not the principal PR author, not the reviewer) thought about the impact of that change -- packages needed to install FRR on Debian VM were now installed every time the FRR container was started.

This fix moves the 'packages we need to make FRR VM work' group var to 'libvirt' part of FRR device definition, as they're not needed on FRR containers.

### DIFF
--- a/netsim/devices/frr.yml
+++ b/netsim/devices/frr.yml
@@ -16,10 +16,6 @@ group_vars:
   ansible_network_os: frr
   ansible_python_interpreter: auto_silent
   netlab_initial: always
-  netlab_linux_packages:
-    gnupg: gpg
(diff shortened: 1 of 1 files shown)


jbemmel:
Could be an idea to track a reference deployment workflow for a topology with an FRR container in CI/CD, and measure the time it takes to deploy. Or maybe track this as part of the existing integration tests, and include it in the reporting - then flag if there is a regression

PR #3679, opened 2026-07-21
Add IP Infusion OcNOS device (best-effort)

### LINKED ISSUE #3676: Native post-deployment validation for a device whose CLI has no one-shot command mode over SSH (e.g. OcNOS cmlsh) — proposal + design question
## The gap

The validation framework reaches a device through `netsim/cli/connect.py::connect_to_node()`, which dispatches to `docker_connect` (when `ansible_connection == 'docker'`) or otherwise `ssh_connect`. That covers container devices and any NOS whose SSH accepts a one-shot command and returns parseable output.

It doesn't cover a NOS that is **not** a container **and** whose SSH lands in a restricted CLI with **no one-shot command mode**. The concrete case is **IP Infusion OcNOS** (vrnetlab VM): an SSH session drops into `cmlsh`, and `ssh <node> "show …"` doesn't run the command and return output — it just opens the interactive shell. So `valid_<test>()` never receives a usable `_result`.

Importantly, this is **not** "OcNOS has no non-interactive command execution." It does — but only via paths the current validation transports can't drive:
- its **Ansible collection** (`ipinfusion.ocnos`), and
- **NETCONF**, via OcNOS's `custom-cmlsh-show-rpc` (executes a `show` command and returns the output).

netlab maps `netconf` to an interactive SSH session rather than a real NETCONF client, and there's no Ansible-based execution path in the validation transports, so neither of OcNOS's non-interactive mechanisms is reachable. Config **generation** for such a device is fine (network_cli); it's specifically the **validation read path** that has nowhere to go — today it can only be checked indirectly from a neighbor, which can't assert on the device under test itself.

## What I'd like your input on (before a PR)

I have a working prototype, but I don't want to guess at the shape you'd accept — this is the kind of core decision you asked to raise first.

**The general problem:** how should post-deployment validation retrieve `show` output from a non-container device whose only non-interactive channels are Ansible or NETCONF-RPC?

**Design questions:**

1. **Do you want this covered in core at all**, or should such a device rely on the SuzieQ source / a probe-node pattern / something else I've overlooked?
2. **Placement.** Which boundary fits netlab best?
   - (a) A new **non-interactive execution transport** as a peer of `docker_connect`/`ssh_connect` in `connect_to_node()` — I prototyped an Ansible-based one (`ansible <node> -m <module>` against the generated inventory); it's generic to any Ansible-drivable NOS, leaves interactive `netlab connect` on SSH unchanged, and as a bonus makes non-interactive `netlab connect --show` work for these devices.
   - (b) A new **validation source** (peer of SuzieQ) in the validation framework.
   - (c) A **NETCONF execution path** using OcNOS's `custom-cmlsh-show-rpc` — standard protocol, but device-specific rather than general.
3. **Opt-in / naming.** However this lands, a device would need to declare it uses the non-interactive path and which module/RPC to invoke — I'll follow whatever naming convention you prefer.
4. **Dependency.** This makes the device's Ansible collection (or a NETCONF client) a validation-time dependency for that device. Acceptable, or a concern?

Context: this surfaced while building OcNOS as a proper netlab device — the device config templates already pass the generic module integration suites; the only open piece is this validation read path. Happy to share the prototype in whatever shape (a/b/c) you'd accept.

roc-ops:
# Add IP Infusion OcNOS (best-effort) and a general `ansible` validation action

Closes #3676.

This PR contributes two things that stand on their own but ship together because the first needs the second:

1. **IP Infusion OcNOS** as a new containerlab device, with configuration templates across the standard module set and state validators for the routing control plane.
2. A new, device-agnostic **`ansible` validation action** for `netlab validate`, so a device whose CLI cannot be driven non-interactively can still have its state read and asserted.

---

### Why

`netlab validate` reads device state either by running a `show`/`exec` command over the node connection, or from an external source (`suzieq`). Both assume the platform can return command output to a non-interactive caller. Some network OSes only expose their operational data through a vendor Ansible module — their interactive CLI rejects non-interactive command execution — so neither existing path can reach them, and today such a device cannot be validated at all.

The `ansible` action closes that gap. It is a validation **data source**, a peer of `netsim/cli/validate/suzieq.py` — **not** a connection method. `netsim/cli/connect.py` is unchanged (byte-identical to `dev`); interactive `netlab connect` is untouched. As suggested on #3676, it is implemented as a first-class, reusable action rather than a one-off device plugin, so any platform in the same situation can adopt it.

### How it works

- A device declares the module it validates through:
  ```yaml
  netlab_validate:
    ansible_module: ipinfusion.ocnos.ocnos_command
  ```
- A test can invoke it directly in a `validate` entry:
  ```yaml
  validate:
    adj:
      device: ocnos
      ansible: show ip ospf neighbor
      valid: "'Full' in stdout"
  ```
  or a device validation plugin can supply the command from an `ansible_<test>()` function and assert on the result in `valid_<test>()`.
- `netsim/cli/validate/ansible.py` runs the module ad-hoc against the node using the netlab-generated inventory, and returns parsed **JSON** when the command emits it (`... | json`), otherwise the raw CLI text in `stdout` for a text assertion to match.

The action is registered everywhere a validation action is expected — the topology schema (`attributes.yml`), the augment/validation checks (`augment/validate.py`), the test and plugin dispatchers — so `ansible:` is accepted and executed like any built-in action, and documented in `docs/topology/validate.md`.

---

## The IP Infusion OcNOS device

A native containerlab device driven by the `ipinfusion.ocnos` Ansible collection. Because OcNOS's `cmlsh` has no non-interactive exec mode, it uses the new `ansible` action for state validation (this is the concrete first consumer described above).

### Module coverage

Configuration templates are provided across the standard module set:

- **Addressing / base:** IPv4, IPv6 (incl. link-local), device MTU, BFD
- **IGP:** OSPFv2 and OSPFv3 (priority, timers, cleartext/MD5 auth, default origination), IS-IS (circuit type, redistribution), RIPv2
- **BGP:** IPv4/IPv6 address families, local-AS, multihop, timers, authentication, standard communities, route policy
- **Routing policy:** static routes (incl. per-VRF), prefix lists, AS-path lists, route maps
- **L2 / fabric:** VLANs (L3-switch model, SVIs, native-routed trunks), LAG (LACP + static), STP (STP/RSTP/MSTP, per-port), VXLAN, EVPN (VXLAN/MPLS transport, IRB, LAG multihoming)
- **Services / other:** VRF (OSPFv2/BGP/IS-IS), first-hop redundancy (VRRP v2/v3), DHCP relay and client, MPLS LDP + L3VPN, Segment Routing (IS-IS, IPv4), SRv6, GRE

Module-level exceptions and platform quirks (e.g. OSPF NSSA type-7 ranges, GRE tunnel line-protocol on the 7.0 image, and DHCP relay not yet covered end-to-end on the test host) are recorded in `docs/caveats.md` and the per-module support tables rather than silently claimed.

### State validation

Validators are provided for the modules whose assertions read state directly off the OcNOS device — **OSPFv2, OSPFv3, BGP and IS-IS** — following the same `ansible_<test>()` / `valid_<test>()` signatures as the FRR/EOS validators. Other modules are exercised by netlab's standard cross-device checks (reachability and neighbor state verified from the peer nodes in each integration topology), which need no device-specific validator.

---

## Verification

Verified on real OcNOS containerlab topologies (`vrnetlab/ipinfusion_ocnos:7.0.0-262`), with an FRR node on each side of the device under test. The device-side checks below are the ones that exercise the OcNOS validators through the `ansible` action:

| Module | Device-side checks (`ansible` action)                                  | Topology total |
|--------|------------------------------------------------------------------------|----------------|
| OSPFv2 | 3/3 — two neighbors Full + OSPF route to a peer loopback               | 5/5 pass |
| OSPFv3 | 3/3 — two neighbors Full + OSPFv3 route to a peer loopback             | 5/5 pass |
| BGP    | 3/3 — two EBGP sessions Established + received advertised prefix       | 5/5 pass |
| IS-IS  | 5/5 — two adjacencies Up + L1 level checks + IS-IS route to a loopback | 7/7 pass |

The direct `ansible:` action (no plugin) was additionally confirmed end-to-end against the device.

Both the configuration the templates render and the show commands the validators parse were checked against the vendor's OcNOS-SP 7.0 CLI documentation — command forms, and the output columns the validators key on (`show ip`/`ipv6 ospf neighbor`, `show ip bgp summary`, `show clns neighbors`, `show ip`/`ipv6 route`).

---

## Files

**`ansible` validation action (core, device-agnostic)**

- `netsim/cli/validate/ansible.py` — new validation source (peer of `suzieq.py`)
- `netsim/cli/validate/tests.py`, `utils.py`, `plugin.py` — dispatch + action registration
- `netsim/augment/validate.py`, `netsim/defaults/attributes.yml` — accept `ansible:` as a first-class validation action
- `docs/topology/validate.md`, `docs/release/26.07.md` — documentation

**OcNOS device**

- `netsim/devices/ocnos.yml` — device settings, feature matrix, `netlab_validate.ansible_module` opt-in
- `netsim/ansible/templates/**/ocnos.j2`, `netsim/extra/**/ocnos.j2` — configuration templates
- `netsim/validate/{ospf,bgp,isis}/ocnos.py` + `netsim/validate/ocnos.py` — state validators
- `netsim/ansible/tasks/{deploy-config,fetch-config,readiness-check}/ocnos.yml` — device tasks
- `docs/platforms.md`, `docs/caveats.md`, `docs/install/ocnos.md` — device documentation
- `tests/integration/platform/ocnos/` — integration topology and notes

---

## Support level

OcNOS is contributed at **best-effort** support. I don't expect it to be part of the regular upstream integration runs; the supported-module matrix and its documented exceptions live in `docs/platforms.md` and `docs/caveats.md`, and I'm happy to maintain the device and address issues against it.

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -436,6 +436,21 @@ ansible_httpapi_port: 80
 * An OSPFv3 ABR running FRR release 10.3 does not originate summary external routes from NSSA areas
 * As of FRRouting release 10.6.0, the FRR implementation of EVPN for VXLAN-over-IPv6 is incompatible with the Arista EOS implementation due to different encodings of the PMSI_TUNNEL_ATTRIBUTE.
 
+(caveats-ocnos)=
+## IP Infusion OcNOS
+
+* OcNOS ships a commercial NOS image; there is no public Vagrant/Containerlab box. Obtain the `vrnetlab/ipinfusion_ocnos` container image (or an equivalent OcNOS VM) from IP Infusion and build/import it yourself.
+* OcNOS's CLI (`cmlsh`) is a restricted shell with **no non-interactive exec mode** -- a plain `ssh device 'show ...'` fails with `Try 'cmlsh --help'`. Configuration deploy and collection use the **ipinfusion.ocnos** Ansible collection (`ocnos_config` / `ocnos_facts`) over `ansible_connection: network_cli`, which drives `cmlsh` interactively through dedicated cliconf/terminal plugins. Install the collection with `ansible-galaxy collection install ipinfusion.ocnos`.
+* Because of the `cmlsh` restriction above, *netlab*'s SSH and `docker exec` validation transports cannot issue `show` commands directly against an OcNOS node. OcNOS instead uses the **`ansible` validation action**, which fetches show output through the `ipinfusion.ocnos.ocnos_command` module (declared in `netsim/devices/ocnos.yml` as `netlab_validate.ansible_module`); `netlab validate` works against OcNOS on both the device under test and attached FRR/Linux probes.
+* An OSPF (v2 or v3) network-membership change on an already-running process (a new area or network added to an existing `router ospf`/`router ipv6 ospf`) makes OcNOS's `commit` return a non-empty informational notice (`Use "clear ip[v6] ospf process" command to take effect`). The `ocnos_config` Ansible module treats any non-empty commit response as a failure, which aborts the rest of that node's play -- any modules queued after `ospf` in the same **netlab initial** run (for example `isis`, `bgp`, `vrf`) are silently skipped for that node. Run **netlab initial** a second time after the first OSPF network/area change on a node; the second pass re-sends the now-unchanged OSPF config (no new membership, no notice) and lets the remaining modules deploy normally. This is a one-time cost per fresh OSPF network change, not a per-boot cost.
+* The main loopback interface name is `lo`; the kernel `lo` owns `127.0.0.1`, so *netlab*'s loopback address is configured as a `secondary` address. Additional *netlab* loopbacks become `loopback<N>` (OcNOS does not support `loopback0`).
+* OcNOS declares DHCP client and relay support and renders the corresponding configuration (cross-checked against the OKF CLI reference), but the DHCP **relay datapath** is not exercised by the integration suite on a containerlab-only host -- the `dhcp/11-ipv4-relay` test needs a libvirt-based `dnsmasq` server probe. OcNOS is therefore intentionally omitted from the [DHCP support table](platform-services-support) until that end-to-end path is covered.
+* GRE tunnel configuration renders for OcNOS, but the tunnel line protocol does not come up on the tested OcNOS 7.0 image, so OcNOS is not listed among the [tunnel.gre](plugin-tunnel-gre) supported platforms.
+* VLANs use the Customer Bridge (VLAN-aware bridge) model: `bridge 1 protocol ieee vlan-bridge` + `vlan database`, with SVIs named `vlan1.<vlan-id>`.
+* Per-VRF OSPF uses a positional VRF name, not a `vrf` keyword: `router ospf <process-id> <vrf-name>`.
+* LACP aggregates are named `po<N>`; static (non-LACP) port channels are named `sa<N>`.
+* OSPFv3 is not supported inside a VRF: the `router ipv6 vrf ospf <name>` command commits and the interface area binding is accepted, but the process never actually attaches to the VRF (`show ipv6 ospf` reports zero areas indefinitely). This is a proven NSM-level gap, not a template limitation.
+
 (caveats-junos)=
 ## Common Junos caveats
 
--- a/docs/install/ocnos.md
+++ b/docs/install/ocnos.md
@@ -1,0 +1,116 @@
+# Installing IP Infusion OcNOS
+
+netlab runs **IP Infusion OcNOS** as a [containerlab](clab.md)-provisioned device: a vrnetlab-packaged
+OcNOS VM. Only the **clab** provider is supported (no Vagrant box).
+
+## Container image
+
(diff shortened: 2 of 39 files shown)


ipspace (changes requested):
Thanks a million for the PR!

Not sure what timezone you're in, so just a short comment (more later today or tomorrow): we're pretty strict about having one functionality/change per PR, as it makes things easier to track down when you're trying to figure out why something was changed.

Considering that, please submit the validation-using-ansible changes as a separate PR. Doesn't have to be anything more complex than creating a new branch, picking the relevant files from this branch into that one, and submitting that one as a PR.

Thank you, Ivan

roc-ops:
Thanks Ivan — done. The `ansible` validation action is now its own PR: #3684 (closes #3676). I've retitled this one to just the IP Infusion OcNOS device. Since the OcNOS validators use that action, this branch is stacked on the action branch for now, and I'll rebase it down to `dev` — dropping the action files from this diff — as soon as #3684 merges. Appreciate the review, and looking forward to the rest of your comments.

ipspace:
@roc-ops -- I will look at this one when we get #3681 and #3680 merged.

PR #3680, opened 2026-07-21
Add SONiC containerlab (docker-sonic-vs) support

roc-ops:
# Add SONiC containerlab (docker-sonic-vs) support

## Summary

netlab already ships a `sonic` device, but only as a **libvirt** Vagrant box (the Azure/VM
SONiC image) with `minimal` support declared — initial configuration and BGP. This PR adds
SONiC support under **containerlab** using the community
[`docker-sonic-vs`](https://github.com/sonic-net/sonic-buildimage) image, and expands the
declared/validated SONiC module coverage well beyond *initial + BGP*.

The new module coverage is attached to a child device **`sonic_clab` (parent: `sonic`)** rather
than to the existing libvirt device, because that coverage was developed and validated against
the `docker-sonic-vs` containerlab image and its access model specifically — a single monolithic
container with FRR `vtysh` running directly in it, configured over Ansible's `docker` connection
plugin. The libvirt SONiC VM has a different access model (FRR runs in a nested `bgp`
sub-container reached via `docker exec bgp vtysh`), so the templates and config-push path here
are not the ones exercised on the VM. Scoping the new support to the child device keeps each
device advertising only what has actually been validated on its image; it is not a claim that
the VM is incapable of these features.

`sonic_clab` inherits interface/loopback naming and the base BGP feature set from `sonic`,
adds a `clab:` block, and provides its own `initial`/`bgp`/`vlan`/`lag`/`vxlan`/`vrrp.version`
templates where the monolithic image requires it. Every other module is delegated to the in-tree
FRR templates via the inherited `ansible_network_os: frr`.

**No netlab-core changes.** `netsim/cli/`, `netsim/augment/`, `netsim/attributes.yml`,
`netsim/providers/`, and `netsim/modules/` are byte-identical to `dev`. `mypy -p netsim` reports
*Success: no issues found in 339 source files* and the transformation test suite passes
*284 passed, 1 skipped, 0 failed* (run in a clean venv from `requirements.txt` +
`requirements-dev.txt`).

## Design note

Added as a child device `sonic_clab` (parent: `sonic`), following the existing
`iol`/`ioll2`, `csr`/`cat8000v`, `vjunos-router`/`vjunos-switch` precedent — same NOS, a
different deployment image/shape and validated feature set, expressed as a separate device so the
new coverage attaches to the containerlab image without changing the existing libvirt `sonic`
device. Happy to rename or restructure if you'd prefer a different approach (e.g. a different
child-device name, or modelling it as a provider/image variant of `sonic`) — just let me know.

## Connection / config-deployment model

Configuration is deployed over Ansible's built-in **`docker` connection plugin** (`docker exec`),
not `network_cli` — the same pattern netlab already uses for the in-tree `frr` containerlab
device. `docker-sonic-vs` ships no running `sshd` and no cliconf-compatible CLI (the only SONiC
cliconf Ansible ships, `dellemc.enterprise_sonic`, targets Dell's licensed Management-Framework
CLI, which the community image does not contain). Routing config is rendered as FRR `vtysh`
scripts; `initial`/`vlan`/`lag` use SONiC's native `config` CLI plus a deterministic
CONFIG_DB→kernel sync where the VS orchestration agents don't program the datapath.

FRR daemons that ship disabled in `/etc/frr/daemons` are enabled per configured module during
initial configuration.

## Supported modules

`ospf` (incl. OSPFv3, `ospf.areas`), `isis` (v4/v6), `bgp` (+ `bgp.session`, `bgp.policy`,
`bgp.originate`, `bgp.domain`, `ebgp.multihop`), `ripv2`, `bfd`, `vrf` (incl. route-leaking,
per-VRF OSPFv2/OSPFv3/IS-IS/BGP), `vlan` (L3 switch / SVIs), `lag`, `vxlan`, `evpn` (VXLAN and
MPLS transport, symmetric IRB / L3VNI, `evpn.multihoming`), `mpls` (LDP + L3VPN), `sr` (SR-MPLS
via IS-IS), `srv6`, `gateway`/`vrrp.version`, `routing` (static/prefix/aspath/community/policy),
and `tunnel.gre`.

Support level: **best-effort** — broad module coverage, each module carrying a device
integration topology under `tests/integration/platform/sonic_clab/`; the image is user-supplied
and the device is not (yet) wired into the automated per-release integration matrix.

## Verification and provenance

**Live-verified for this submission** against a real `docker-sonic-vs:latest` container on
containerlab, config + validation over `docker exec`:

| Module | How verified | Result |
|--------|--------------|--------|
| OSPF   | `netlab validate` (native docker-exec) | Tests passed: 2/2 (both neighbors Full) |
| BGP    | `netlab validate` (native docker-exec) | Tests passed: 2/2 (eBGP Established both ways) |
| IS-IS  | `netlab validate` (native docker-exec) | Tests passed: 2/2 (adjacency Up both ways) |
| VRF    | per-VRF OSPF neighbor state | Full in both directions in VRF `red` |
| VLAN   | SVI-to-SVI datapath ping | 0% loss; access port `master Bridge state forwarding` (CONFIG_DB→kernel bridge sync confirmed) |

**Supported, with integration topologies included but not re-run for this submission:** the
remaining modules (EVPN / symmetric-IRB / `evpn.multihoming`, MPLS L3VPN, SR-MPLS, SRv6, RIPv2,
BFD, `tunnel.gre`, the BGP plugins, and the IPv4/IPv6 dual-stack variants) each ship a topology
under `tests/integration/platform/sonic_clab/` (several with native `validate:` sections) that was
exercised during the device's development. They are presented as supported, not as re-verified in
this PR.

## Caveats / limits

Documented in `docs/caveats.md` (`caveats-sonic-clab`) and flagged in the platform support
tables:

* You supply your own `docker-sonic-vs:latest` image; netlab does not ship it.
* **MPLS / SR-MPLS** labs need a one-time host prerequisite: `sudo modprobe mpls_router
  mpls_iptunnel` (loads `net.mpls` into every container netns → real kernel label FIB). Flagged
  with a caveat marker on the MPLS/SR-MPLS cells in the support tables.
* **SRv6** is control-plane + kernel-plane only on this image: locator/SIDs are advertised in the
  IS-IS LSDB and installed as real kernel `seg6local` routes, but end-to-end SRv6-OAM datapath
  (ping to a `uN` End SID) is not resolved — the same open item seen with FRR/IS-IS SRv6
  elsewhere. Marked with a caveat in the support table.
* Genuinely unsupported on this image (probed, not merely untested): DHCP relay/client (no
  `dhcrelay`/`dhcp6relay` binary), real STP port-blocking (no `stpd` daemon), and the
  `mlag.vtep` active-active EVPN datapath (no `mclagd`). These are shown as ✗/absent in the
  tables, not claimed.

## Files

* `netsim/devices/sonic_clab.yml` — device definition (parent `sonic`, `clab:` block, features).
* `netsim/ansible/templates/{initial,bgp,vlan,lag,vxlan}/sonic_clab.j2`,
  `netsim/extra/vrrp/version/sonic_clab.j2` — device-specific templates.
* `netsim/ansible/tasks/{deploy-config,readiness-check}/sonic_clab.yml` — docker-exec deploy.
* `netsim/validate/{,ospf,bgp,isis,route}/sonic_clab.py` — FRR validation re-exports.
* `tests/integration/platform/sonic_clab/` — integration topologies + README.
* `docs/platforms.md`, `docs/caveats.md`, `docs/labs/sonic.md`, `docs/release/26.07.md` — docs.
* `netsim/extra/vrrp/version/defaults.yml` — registers the `none` meta-device for the
  `vrrp.version` plugin so `tests/check-integration-tests.sh` can transform the `10-vrrp-version`
  topology under `-d none`, mirroring the existing `evpn.multihoming` plugin. This is the only
  change outside the `sonic_clab` device + docs.


docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+  This matches how *netlab* already drives the in-tree `frr` *containerlab* device and the
+  libvirt `sonic` device above (which has a full sshd) -- neither uses `network_cli` for this
+  vtysh-delegated render-then-push style of configuration, sshd or not.
+* `docker-sonic-vs` ships `sshd` and host keys but starts neither `sshd` nor a login user; the
+  initial configuration bootstraps both (`admin`/`YourPaSsWoRd`, matching the `sonic` device's
+  own credentials) purely for interactive access (`netlab connect`, ad-hoc troubleshooting) --
+  SSH plays no part in configuration deployment.
+* Most FRR daemons ship disabled in `/etc/frr/daemons` by default (to save resources); the
+  initial configuration enables the ones the configured modules need and restarts FRR once.
+* MPLS/SR-MPLS labs need a one-time host prerequisite before `netlab up`: `sudo modprobe
```
jbemmel: We have a mechanism for this, see ```kmods```: https://github.com/ipspace/netlab/blob/9acc322669a504fb466f59f287ec7a72f7f1eb09/netsim/devices/frr.yml#L42

Since Sonic runs FRR inside, it may be better to inherit the FRR settings where possible, rather than redefine them

netsim/ansible/templates/initial/sonic_clab.j2
```diff
@@ -0,0 +1,206 @@
...
+       VLAN/PortChannel member the command errors out ("Cannot configure the IPv6 link
+       local mode!"), breaking re-deploys #}
+config interface ipv6 disable use-link-local-only {{ l.ifname }}
+{%   endif %}
+{% endfor %}
+#
+# Configure interfaces
+#
+config hostname {{ inventory_hostname.replace("_","-") }}
+{% for l in netlab_interfaces|default([]) if l.type == 'lag' %}
```
jbemmel: There is a recently added mechanism for running module specific initialization during init, see VyOS templates for an example

netsim/ansible/templates/initial/sonic_clab.j2
```diff
@@ -0,0 +1,206 @@
...
+{# PortChannels must exist before `config interface ip add PortChannel<N>` runs --
+   same create-before-address ordering as VLANs below #}
+if ip link show {{ l.ifname }} >/dev/null 2>&1; then
+  echo {{ l.ifname }} already exists
+else
+  config portchannel add {{ l.ifname }}
+fi
+{% endfor %}
+{% if vlans is defined %}
```
jbemmel: This belongs in the vlan module, you could use ```vlan/sonic.initial.j2``` similar to vyos

netsim/ansible/templates/initial/sonic-clab.j2
```diff
@@ -0,0 +1,206 @@
...
+if vtysh -c 'show running' >/dev/null; then
+  echo vtysh is ready
+else
+  echo 'giving vtysh some more time :('
+  sleep 5
+fi
+#
+# And now let's configure the interfaces
+#
+cat >/tmp/netlab-initial.frr <<CONFIG
```
jbemmel, line 168: This should probably reuse FRR scripting

netsim/ansible/templates/lag/sonic-clab.j2
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
jbemmel, line 1: should be named ```sonic-clab.j2``` I believe (i.e. dash)

netsim/ansible/templates/vlan/sonic-clab.j2
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
jbemmel, line 1: ```sonic-clab.j2```

netsim/ansible/templates/vxlan/sonic-clab.j2
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
jbemmel, line 1: If possible reuse FRR scripts

netsim/devices/sonic_clab.yml
```diff
@@ -0,0 +1,166 @@
...
+  netlab_frr_daemons:            # module -> FRR daemons to enable in /etc/frr/daemons (default: no)
+    bgp: [ bgpd ]
+    ospf: [ ospfd, ospf6d ]
+    isis: [ isisd ]
+    bfd: [ bfdd ]
+    ripv2: [ ripd ]
+    vrf: [ bgpd ]
+    gateway: [ vrrpd ]
+    mpls: [ ldpd ]
+clab:
```
jbemmel: add empty ```kmods:``` key to enable kernel module loading support

netsim/devices/sonic_clab.yml
```diff
@@ -0,0 +1,166 @@
...
+    areas: true
+    timers: true
+    priority: true
+    password: true
+    bfd: true
+  isis:
+    circuit_type: true
+    unnumbered: { ipv4: true, ipv6: true, network: true }
+    import: [ static ]
+  mpls:
```
jbemmel: It's up to Ivan, but I imagine it might be better to break this up into smaller PRs: First add initial support, then add OSPF, BGP, etc.

netsim/extra/vrrp/version/defaults.yml
```diff
@@ -6,6 +6,9 @@ devices:
     features.gateway.vrrp.version: True
   dellos10:
     features.gateway.vrrp.version: True
+  # the meta device used by tests/check-integration-tests.sh has to accept the plugin
+  none:
```
jbemmel: This belongs in a separate PR

netsim/extra/vrrp/version/sonic_clab.j2
```diff
@@ -0,0 +1,16 @@
...
+# vrrp.version plugin — FRR vrrpd supports both VRRP v2 and v3. netlab
+# renders this as a separate per-node config step; FRR requires the vrrp instance
+# to be shut down before its version can change, so: shutdown -> version -> no shutdown.
+set -e
+cat >/tmp/netlab-vrrp-version <<CONFIG
+{% for intf in interfaces if intf.gateway.protocol|default('') == 'vrrp' %}
+interface {{ intf.ifname }}
+  vrrp {{ intf.gateway.vrrp.group }} shutdown
+  vrrp {{ intf.gateway.vrrp.group }} version {{ gateway.vrrp.version|default(3) }}
+  no vrrp {{ intf.gateway.vrrp.group }} shutdown
```
jbemmel: The user could have requested that the interface be shutdown, the logic should check that

netsim/extra/vrrp/version/sonic_clab.j2
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+# vrrp.version plugin — FRR vrrpd supports both VRRP v2 and v3. netlab
+# renders this as a separate per-node config step; FRR requires the vrrp instance
```
jbemmel: afaik none of the existing devices supports changing VRRP versions mid-flight, it's a nice detail but also a bit overkill?

jbemmel (changes requested):
Thanks a lot for your sizable contribution, it's appreciated!

Overall, I think our preference would be to reuse FRR logic and templates as much as possible, rather than reimplement the same thing in parallel. You already do that in some places like validation scripts, since Sonic is running standard FRR (I presume) the configuration should be fairly similar (if not identical)

I also think a 50 file PR is perhaps a bit too big of a chunk to accept in one go, but I'll let Ivan comment

roc-ops:
Thanks for the thorough review — the "reuse FRR logic and templates rather than reimplement in parallel" principle is exactly right, and I've taken it as the primary lens for this revision, not just the individually-flagged lines. Summary of what collapsed into FRR reuse vs what stays genuinely SONiC-specific, then the per-comment detail.

**Reused from FRR (no parallel implementation):**
- **Routing config** — every routing module (bgp, ospf, isis, vrf, mpls, sr, srv6, evpn, ripv2, gateway, routing, tunnel.gre) renders through the in-tree `<module>/frr.j2` templates via the inherited `ansible_network_os: frr`. In particular the `bgp/sonic_clab.j2` template is gone — SONiC's BGP is standard FRR, so it now uses `bgp/frr.j2` (docker-sonic-vs, unlike the Azure/libvirt image, does not pre-seed a BGP AS, so the `no router bgp` reset it carried wasn't needed).
- **Validation** — collapsed the four per-module validators into a single `validate/sonic_clab.py` that re-exports the FRR checks (`from netsim.validate.frr import *`).
- **Kernel modules** — `clab.kmods` instead of a manual `modprobe`.

**Genuinely SONiC-specific (kept, because it isn't FRR):**
- Interface / VLAN / PortChannel objects created through SONiC's `config` CLI (config_db).
- FRR daemons enabled per configured module (docker-sonic-vs ships them `=no` in `/etc/frr/daemons`).
- Config deployed over the Ansible `docker` connection (docker exec) — the image has no sshd/cliconf CLI.
- A deterministic CONFIG_DB→kernel sync for the objects the VS orchestration agents don't program (routed-port IPs, VLAN members, VXLAN).

Per comment:

**kmods (initial.j2 modprobe / `sonic_clab.yml` kmods key):** Adopted. The device now declares `clab.kmods` mirroring `frr.yml` (`initial: [ 'vrf?' ]`), which opts it into netlab's clab kernel-module loader; the system-wide `providers.clab.kmods` then auto-loads `mpls_router`/`mpls_iptunnel` (and vrf/vxlan) per module. Verified by unloading those modules from the host and booting the MPLS/L3VPN test with no manual setup — netlab loads them and the L3VPN datapath (LDP transport + BGP VPN label, real kernel label FIB) comes up 0% loss. The manual-modprobe caveat is removed.

**Module init hooks (LAG PortChannel loop, VLAN creation):** Moved out of the monolithic initial template into the per-module init hooks `lag/sonic_clab.initial.j2` and `vlan/sonic_clab.initial.j2`, pulled into the initial config by `extra_module_initial(['lag','vlan'])` at the correct ordering point (create-before-address) — the same mechanism VyOS uses. Verified live: PortChannel + LACP up with OSPF over the aggregate, and the VLAN SVI-to-SVI datapath at 0% loss.

**Reuse FRR scripting (initial.j2 interface block) and vxlan template:** Assessed against the FRR templates. The routing and EVPN *control* planes already reuse FRR entirely (`bgp/frr.j2` + zebra `advertise-all-vni`). What remains in these two templates is the part that genuinely isn't FRR: the FRR base initial template brings interfaces up with `ip link`, whereas SONiC brings them up through config_db (the `config` CLI), and there's no shared include to reuse for that. The vxlan template is only the kernel L2VNI/L3VNI data-plane setup that docker-sonic-vs's orchagent doesn't program from config_db; it parallels `vxlan/frr.j2`'s kernel approach but keeps the config_db-aligned `Vlan<id>` bridge naming (so `vrf/frr.data-plane.j2`'s enslave-by-ifname finds the SVI) plus idempotency guards. I've kept those as SONiC deltas and documented why; happy to converge further if you'd prefer.

**vrrp.version (separate PR / buggy shutdown logic / overkill):** Agreed on all three — dropped entirely from this PR. Removed the template, the test, and the `gateway.vrrp.version` feature flag, and reverted the `extra/vrrp/version/defaults.yml` registration, so this PR now touches zero files outside the `sonic_clab` device + docs. Base gateway/VRRP support is retained.

**Rename `sonic_clab.j2` → `sonic-clab.j2` (dash):** The dash is trivial for the templates, but it surfaces a latent core issue that made me hold off: `netsim/cli/validate/plugin.py` evaluates `validate_{device}.{action}_{plugin}`, and for a dash device that parses as subtraction — `eval("validate_sonic-clab.show_ospf_neighbor")` becomes `validate_sonic - clab.show_ospf_neighbor`, where `clab...` is a topology Box → "'Box' object is not callable", which breaks native `netlab validate`. No existing dash-named device (vjunos-router/-switch, vptx) ships validators, so this path has never been exercised. The device deploys and forwards fine under a dash name — only the validate eval breaks. To keep this PR free of core changes I've left it as `sonic_clab` for now, but I'm glad to do the rename together with a ~2-line fix that sanitizes the device name at the two build sites in that eval (`str.replace('-','_')`) — in this PR or a separate core PR, whichever you and Ivan prefer.

**Break into smaller PRs:** Agreed it's a sizable change. I'll leave the split call to you/Ivan — happy to restructure into staged PRs (e.g. device+initial first, then per-module) if that's easier to review.

docs/labs/sonic.md
```diff
@@ -19,3 +19,22 @@ During the box-building process, you might have to disable ZTP or clean up the i
 .. include:: sonic.txt
    :literal:
 ```
+
+(labs-sonic-clab)=
+## Using the containerlab Provider (docker-sonic-vs)
```
ipspace: Change to "Using the SoNIC containers"

docs/labs/sonic.md
```diff
@@ -19,3 +19,22 @@ During the box-building process, you might have to disable ZTP or clean up the i
...
+
+Apart from the libvirt Vagrant box above, SONiC can run under **containerlab** using the community
+`docker-sonic-vs` image via the `sonic_clab` device (parent: `sonic`). No box build is needed --
+supply your own `docker-sonic-vs:latest` image (build it from the
+[sonic-buildimage](https://github.com/sonic-net/sonic-buildimage) `docker-sonic-vs` target, or pull a
```
ipspace: It might be a good idea to provide the download URL

docs/release/26.07.md
```diff
@@ -14,6 +14,7 @@
 * The [**WireGuard tunnel** plugin](plugin-tunnel-wireguard) supports WireGuard tunnels on FRR.
 * The [**bgp.session** plugin](plugin-bgp-session) and the [OSPF module](module-ospf) support graceful restart on Arista EOS, BIRD, FortiOS, and FRR
 * The [**bgp.policy** plugin](plugin-bgp-policy) supports the **bgp.role** attribute on FRR and BIRD.
+* SONiC can run under *containerlab* with the community `docker-sonic-vs` image via the new [`sonic_clab` device](caveats-sonic-clab), expanding SONiC coverage well beyond the libvirt device (OSPF/OSPFv3, IS-IS, BGP with the bgp.session/policy/originate/domain and ebgp.multihop plugins, RIPv2, BFD, VRF, VLAN, LAG, VXLAN/EVPN including symmetric IRB and multihoming, MPLS L3VPN, SR-MPLS, SRv6, VRRP, and GRE tunnels).
```
ipspace: This goes into 26.08

docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+(caveats-sonic-clab)=
+## Sonic (containerlab)
+
+A separate device (`sonic_clab`, parent `sonic`) for the community `docker-sonic-vs` image
+running under *containerlab*, distinct from the `sonic` device above (which targets the
+Azure/libvirt SONiC VM). The two images have different internal architectures and are not
+interchangeable: `docker-sonic-vs` is a single monolithic container (FRR's `vtysh` runs
+directly in it), while the VM runs FRR inside a nested `bgp` sub-container.
+
+* You supply your own `docker-sonic-vs:latest` image (build it from the
```
ipspace: This is in the "labs/sonic" document. Just add a link to that document

docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+  [sonic-buildimage](https://github.com/sonic-net/sonic-buildimage) `docker-sonic-vs` target, or
+  pull a prebuilt one); *netlab* does not ship or distribute it.
+* Configuration is deployed over Ansible's built-in `docker` connection plugin (`docker exec`),
```
ipspace: Too many details (but as before, you can move them to labs/sonic.md document. "it uses docker exec not network_cli" is good enough.

docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+  This matches how *netlab* already drives the in-tree `frr` *containerlab* device and the
+  libvirt `sonic` device above (which has a full sshd) -- neither uses `network_cli` for this
+  vtysh-delegated render-then-push style of configuration, sshd or not.
+* `docker-sonic-vs` ships `sshd` and host keys but starts neither `sshd` nor a login user; the
+  initial configuration bootstraps both (`admin`/`YourPaSsWoRd`, matching the `sonic` device's
+  own credentials) purely for interactive access (`netlab connect`, ad-hoc troubleshooting) --
+  SSH plays no part in configuration deployment.
+* Most FRR daemons ship disabled in `/etc/frr/daemons` by default (to save resources); the
+  initial configuration enables the ones the configured modules need and restarts FRR once.
+* MPLS/SR-MPLS get a real kernel MPLS label FIB (`ip -M route`) with no manual setup: because
```
ipspace: If you think it's needed, move it to labs/sonic.md. No need to explain what works as expected in caveats ;)

docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+  SONiC runs FRR, the device reuses netlab's FRR kernel-module handling (the `clab.kmods`
+  declaration), so `netlab up` loads `mpls_router`/`mpls_iptunnel` on the host automatically when
+  an MPLS or SR lab is deployed, the same way the in-tree `frr` containerlab device does.
+* Module coverage is broad and FRR-delegated (`ospf`, `bgp`, `isis`, `vrf`, `bfd`, `mpls`, `sr`,
```
ipspace: We have the "platforms.md" document for this.

docs/caveats.md
```diff
@@ -652,6 +652,55 @@ See also [](caveats-sros) caveats for further details.
...
+  VRF, and VLAN (including the SVI-to-SVI kernel-bridge datapath) were live-verified on
+  `docker-sonic-vs:latest` for this submission; the remaining topologies -- including the MPLS
+  L3VPN datapath (LDP transport + BGP VPN label, real kernel label FIB) and the EVPN-VXLAN
+  symmetric-IRB (L3VNI) datapath -- were validated during the device's development, not re-run for
+  this submission.
+* `srv6` is control-plane + kernel-plane only on this image: the SRv6 locator and End/End.X SIDs
+  are advertised in the IS-IS LSDB and installed as real kernel `seg6local` routes (`ip -6 route`),
+  but end-to-end SRv6-OAM datapath (ping to a `uN` End SID) is not resolved on `docker-sonic-vs`
+  -- the same open item as the FRR/IS-IS SRv6 result on other platforms.
+* Genuinely unsupported on this image (not just untested): DHCP relay/client (no
```
ipspace: If things don't work, turn them off in features.

ipspace (changes requested):
I didn't go into the details (yet ;). However, I think that:

1. There is no need for a new device. You can define extra (or disabled) features in clab.features and extra parameters in clab.group_vars.
2. It would be nice to have a way to get a recent working SoNIC VM, maybe declare `frr` as SoNIC parent device (so we get all FRR control-plane features), and then turn off the features that are not supported in SoNIC VM or container.
3. You have to use "sonic_clab" templates because that's how you defined your device. Using "sonic" device and have different clab templates ("sonic-clab") would be a better alternative (see (1)).
4. The platforms are tested with the existing tests/integration. There is no need to write new platform integration tests.

roc-ops:
Thanks — I've restructured it as you suggested, and **your architecture measurably beats what I had**.

`sonic_clab` is gone. `sonic` now declares `parent: frr`, inherits the FRR control-plane features, and turns off what SONiC doesn't support; the container/VM differences live in `clab.features`, `clab.group_vars` and `sonic-clab.j2` templates. The per-platform integration tests are removed.

**Results on the full integration suite, same harness throughout:**

| | baseline | my separate-device version | **inheriting `frr`** |
|---|---:|---:|---:|
| PASS | 39 | 93 | **114** |
| FAIL | 81 | 34 | 26 |

**+21 over the structure I was defending.** Inheriting FRR's feature surface reached tests my hand-maintained declarations never did — OSPF is now rendered entirely by the inherited `ospf/frr.j2`, with no SONiC OSPF template at all.

**Three things worth reporting back, since they weren't obvious:**

**1. Inheriting a parent brings its deployment mechanism, not just its features.** Three settings had to be overridden or the device wouldn't configure at all:

* `clab.node.config_templates` — FRR bind-mounts `/etc/frr/daemons`, which undoes the in-place edit `initial/sonic-clab.j2` makes to the daemons file that `docker-sonic-vs` ships.
* `netlab_config_mode` / `netlab_default_shebang` — FRR selects the native `sh` config mode with a `#!/usr/bin/vtysh -f` shebang. That bypasses the SONiC deploy task entirely and feeds SONiC's bash `config` script to vtysh; the symptom is `RuntimeError: Unable to connect to redis - Connection refused`, and the rendered file silently changes extension from `.cfg` to `.sh`.
* `netlab_mgmt_vrf` — FRR sets it, SONiC does not.

None of these is a feature flag, so a features-only review would not catch them. Also worth knowing: `{}` does **not** clear an inherited key — a bare `config_templates:` does.

**2. Inheritance converts under-declaration into over-declaration.** 19 tests began attempting things they previously refused; 14 passed, 6 did not. I have explicitly disabled four capabilities FRR declares that SONiC does not have — anycast gateway (`gateway.protocol: [ vrrp ]`), the `host` and `bridge` roles, and RFC 9234 `bgp.role` — each in its own commit with the reason. Those tests are now correctly refused rather than failing.

**3. The device now has both `libvirt:` and `clab:` blocks** where `sonic_clab` had only clab, so a topology that does not pin a provider can select libvirt and fail on `KVM is not installed`. I have documented it rather than working around it, but you may have a view on whether that wants handling in core.

ipspace:
> Thanks — I've restructured it as you suggested, and **your architecture measurably beats what I had**.

Thank you!

> `sonic_clab` is gone. `sonic` now declares `parent: frr`, inherits the FRR control-plane features, and turns off what SONiC doesn't support; the container/VM differences live in `clab.features`, `clab.group_vars` and `sonic-clab.j2` templates.

Perfect.

> **Results on the full integration suite, same harness throughout:**
> 
> baseline	my separate-device version	**inheriting `frr`**
> PASS	39	93	**114**
> FAIL	81	34	26

I'm usually not merging things that don't pass the integration tests. I can look at what's going on and maybe tweak a few things, but to do that, I'd appreciate having the "download from this URL" mention in the documentation.

Speaking about that, there are still a few unresolved documentation comments.

> **Three things worth reporting back, since they weren't obvious:**

Thanks a million. If I ever decide to add an "inheriting from another device" section to developer documentation, these will definitely go in there.

> **3. The device now has both `libvirt:` and `clab:` blocks** where `sonic_clab` had only clab, so a topology that does not pin a provider can select libvirt and fail on `KVM is not installed`.

That's a common behavior you can experience with many devices 🤷🏻‍♂️

ipspace:
Had to make a few changes to make the code work with SONiC build 2605 (I'd expect people to pull down recent containers):

* sudo no longer exists; check it before modifying /etc/sudoers.d
* I was fighting with FRR way too long until I decided it's better to just restart it (it's not like that would take a long time) and then wait a few seconds. Miraculously, all worked
* As I already had the "wait for something to work" code, I applied it to "config hostname" command (which accidentally checks whether redis works), and all of a sudden, we can use script-based deployments 🎉

I did not change the deployment settings yet (your call), but I'd strongly recommend we stay as far away from Ansible as feasible (plus, the error messages are way easier to decode).

Running full set of integration tests, let's see how that works out ;)

ipspace (approved):
OK, let's move this forward. I'm merging it, but there are tons of things that still need to be fixed:

* Using Linux interfaces (and **ip** commands) to configure dataplane features makes no sense. The only reason to have a SONiC device is to try out features that SONiC supports, and that can be used on real hardware. So, whatever is configurable with **config** commands is fair game, but nothing more.
* Ideally, we'd use **config** commands to configure FRR as well, and only enable the features that can be configured that way. Looking at the "config" command help, it looks like SONiC only supports BGP, so maybe we should limit ourselves to that (please note that I reused FRR scripts because I wanted to invest minimal time in having a working SONiC device).
* The VLAN and LAG modules failed so many tests that I disabled them. Please figure out what works; I'll help you figure out whether that can be covered with device features or quirks, and then we can turn them back on. However, using a mix of **config** and **ip** commands that the current VLAN configuration template uses is NOT ok.

Minor details:

* With the minor changes I made, we can go to the "direct script execution" deployment model and bypass Ansible.
* It looks like the interface MTU is not set correctly
* The failed OSPF test is caused by an old FRR version; I'll add a caveat.
* Once we agree on what makes sense to support, we have to update the "platforms.md" support tables, as well as the support tables in individual modules.

With all that in mind, I'm merging this PR to move things forward. Thanks again for all the work you invested into this feature!

ipspace:
One more thing: test results are @ https://tests.netlab.tools/_html/sonic

roc-ops:
Thanks for merging it, and for going hands-on with the container — the readiness gate on `config hostname` was the piece I was missing, and script-based deployment is in.

Your direction is clear and I agree with it: if it can't be done with **config** commands, it isn't really SONiC support. Taking that as the rule going forward.

Two consequences I'll act on:

**VLAN 1 is a good test of the rule.** Chasing the `Vlan1 is default VLAN` error, I found *every* config_db path for it is closed — `config vlan add 1`, `config vlan member add -u 1`, `config interface ip add Vlan1`, and `config interface ipv6 enable use-link-local-only Vlan1` are all rejected, while `config vlan add 2` works. I had a kernel-bridge workaround for it; under your rule that's the wrong answer, so I've dropped it. It looks like a case for a device feature or quirk that refuses VLAN 1 outright rather than working around it.

**I'll take you up on the VLAN/LAG offer.** Rather than argue for turning them back on, I'll work out what's actually achievable with **config** commands only — per operation (access, trunk, native, SVI/IRB) and per LAG operation — and come back with a list of what can pass under that restriction and what can't be expressed at all. I expect that means rewriting the VLAN and LAG templates rather than patching them, since the current ones mix `config` and `ip`, which is the thing you're objecting to.

**On the MTU** — that one surprises me, because interface MTU was the single biggest fix in this work (it was why every OSPF adjacency sat in `ExStart`). I'm checking what's actually programmed versus intended on a current build; if `ip link set ... mtu` is itself outside the rule, it needs re-expressing through `config interface mtu` with a readiness gate like the one you used.

**On scope** — you raised that `config` may only really cover BGP. If that's right it shrinks this device a lot, and I'd rather know your view before investing in the VLAN/LAG rewrite. Worth deciding that first, since it determines whether the rest is worth doing.

I'll leave `platforms.md` and the per-module tables until the scope question is settled, so they only get written once.

ipspace:
> Your direction is clear and I agree with it: if it can't be done with **config** commands, it isn't really SONiC support. Taking that as the rule going forward.

👍

> **VLAN 1 is a good test of the rule.** Chasing the `Vlan1 is default VLAN` error, I found _every_ config_db path for it is closed

We don't deal with "VLAN 1 is broken" scenario in the core, but it's so common that we have a "check_tagged_vlan_1" common quirk for switches. Some devices with router configuration mindset (= create subinterfaces, connect them to a bridge) also have problems; you might want to check IOSv and RouterOS7 quirks.

Also: check the very last column in https://tests.netlab.tools/_html/coverage.vlan ("VLAN 1 in trunks"). Devices that fail it correctly (with "not supported") usually do it via device quirk.

> **I'll take you up on the VLAN/LAG offer.** Rather than argue for turning them back on, I'll work out what's actually achievable with **config** commands only — per operation (access, trunk, native, SVI/IRB) and per LAG operation — and come back with a list of what can pass under that restriction and what can't be expressed at all.

Perfect.

> I expect that means rewriting the VLAN and LAG templates rather than patching them, since the current ones mix `config` and `ip`, which is the thing you're objecting to.

Exactly.

> **On the MTU** — that one surprises me, because interface MTU was the single biggest fix in this work (it was why every OSPF adjacency sat in `ExStart`). I'm checking what's actually programmed versus intended on a current build

That approach could result in a never-ending whack-a-mole exercise. Do whatever you can with the **config** commands, and if it doesn't work, we'll add a testing/documentation caveat. If they manage to fix it (like FRR team fixed the OSPFv3 prefixes on loopbacks), the testing caveat will eventually "miraculously" disappear.

> if `ip link set ... mtu` is itself outside the rule, it needs re-expressing through `config interface mtu` with a readiness gate like the one you used.

Correct. Please do not use **ip** commands. Also, who knows how the SONiC **config** commands stomp on interface settings and how that might change.

As for readiness gate -- once the **config** commands work, we should be good to use them, right?

> **On scope** — you raised that `config` may only really cover BGP. If that's right it shrinks this device a lot, and I'd rather know your view before investing in the VLAN/LAG rewrite.

Well, you should do a VLAN/LAG rewrite anyway ;) Using dataplane **ip** commands makes no sense as they probably won't do much good on hardware platforms.

From the control-plane perspective, I think we should give users a fair assessment of what SONiC can do out of the box (= BGP). As it uses FRR in the background, and we can configure that, we could document how to enable additional control-plane features if the user wants them, with an added warning that those features configure FRR directly, which might defy the whole purpose of having SONiC.

Alternatively, we could eventually go for a configurable control plane, like what @jbemmel did with VPP.
 
> I'll leave `platforms.md` and the per-module tables until the scope question is settled, so they only get written once.

👍

PR #3681, opened 2026-07-21
Add Arrcus ArcOS as a containerlab device (best-effort)

roc-ops:
# Add Arrcus ArcOS as a containerlab device (best-effort)

This PR adds **Arrcus ArcOS** as a *netlab* device. ArcOS runs as a **native**
containerlab node (`kind: arrcus_arcos`) — there is no vrnetlab packaging and no
Vagrant box. Support level is **best-effort**.

## Connection and validation model — no core change

ArcOS ships the official `arrcus.arcos` Ansible collection (the project-recommended
`network_cli` model), but every published version (2.0.13–2.0.18) hangs against the
tested container image: its cliconf/terminal plugins issue a `config terminal` /
`arcos_cli` mode-switch that this build's `confd_cli` rejects. Configuration is
therefore deployed over Ansible's built-in **`docker` connection plugin** (`docker
exec` running `confd_cli` `load merge` / `commit`) — the same pattern
`sonic_clab` already uses. The device keeps a commented-out `network_cli` block to
switch to once a working collection version ships.

`netlab validate` reads device state over the **same** docker-exec path
(`show <path> | display json | confd_cli`, parsed as OpenConfig JSON by
`netsim/validate/**/arcos.py`). This is netlab's standard device-side show-command
validation — **no** `ansible` validation action and **no** change to any core file.
`netsim/cli/connect.py`, `netsim/cli/validate.py`, `netsim/augment/validate.py`,
`netsim/attributes.yml`, `netsim/templates/provider/clab/clab.j2`, and `tests.py` are
untouched (byte-identical to `dev`). This is a pure device addition.

The tested image boots with SSH/NETCONF/gNMI disabled and refuses to enable
interfaces until the factory admin-user password is changed; both are handled with
the native `netlab_start_exec` group_var (containerlab's post-start `exec:`) — no
custom deploy code.

## Module coverage

`initial`, `ospf` (v2/v3), `bgp`, `isis`, `vrf` (+ospf/isis/bgp), `vlan`, `lag`,
`gateway` (VRRP), `dhcp`/relay, `bfd`, `routing` (static/prefix-set/policy),
`mpls` (LDP), `sr` (SR-MPLS via IS-IS), `srv6`, `vxlan`, and `evpn` (L2VNI).

Honest limitations are documented in `docs/caveats.md` (e.g. EVPN symmetric IRB/L3VNI,
VRF route-target leaking, and OSPFv2 area authentication are genuinely unsupported and
not declared; MPLS/SR/SRv6 bring up the control plane only on the native container;
DHCP relay renders but its datapath is not exercised on a clab-only host, so ArcOS is
omitted from the DHCP support table).

## Verification

Verified on real ArcOS containers (image `arcos:8.2.1A.P2`), containerlab provider,
against this branch:

| Test | Result |
| ---- | ------ |
| `02-ospf` — OSPFv2 adjacency | 2/2 PASS (FULL both directions) |
| `03-bgp` — eBGP session + prefixes | 4/4 PASS (ESTABLISHED + peer loopbacks in RIB) |
| `04-isis` — IS-IS adjacency + reachability | 4/4 PASS |
| `05-routing` — static route | 2/2 PASS; kernel FIB (`proto static`) installed and loopback-to-loopback ping 0% loss |
| combined `ospf`+`bfd`+`vrf` | deploys clean; per-VRF OSPF reaches FULL; BFD session UP both directions |

`netlab initial` commits every module cleanly (idempotent on re-run) and `netlab
collect` retrieves the running configuration. The rendered CLI and the show/get paths
used by the validators were cross-checked against the vendor ArcOS S8.5.1A CLI
reference.

## Files

* Device: `netsim/devices/arcos.yml`
* Deploy/collect tasks: `netsim/ansible/tasks/{deploy-config,fetch-config}/arcos.yml`
* Templates: `netsim/ansible/templates/*/arcos.j2` (16 modules)
* Validators: `netsim/validate/arcos.py` + `netsim/validate/{ospf,bgp,isis,routing}/arcos.py`
  (adds the `netsim/validate/routing/` package)
* Integration tests: `tests/integration/platform/arcos/`
* Docs: `docs/platforms.md`, `docs/caveats.md`, `docs/install/arcos.md`,
  `docs/release/26.07.md`

This PR also removes `netsim/templates/provider/libvirt/arcos-domain.j2`, an orphaned
libvirt domain template from a much older, never-completed ArcOS device (no
`netsim/devices/arcos.yml` ever existed for it); the new device is containerlab-only.

ArcOS is a commercial NOS, so users supply their own container image (as with
`sros`/`dellos10`/`arubacx`).


ipspace (changes requested):
Before going into the details: I think you should change the configuration deployment method.

It looks like you're generating a bash script that takes a config and runs `confd_cli` to deploy it. If that's the case, you don't need Ansible -- netlab can generate bash scripts,  map them to the container, and then execute them inside the container. You can also map templates into container files to create scripts (like what you have in the **shell** command) that are then executed.

FRR uses the "run a bash script" approach together with "let's add a shebang when needed" trick. cRPD maps a template to a container script to generate the deployment script (`/config/netlab/netlab-config.sh`).

More @ https://netlab.tools/dev/config/deploy/#linux-configuration-scripts

ipspace:
Next major item: we don't have per-platform integration tests (what you added in `tests/integration/platform`). Please remove those.

The device implementations are tested with regular integration tests (see https://netlab.tools/dev/integration-tests/). If you want to test the validation plugins, you'll find some tests in `tests/platform-integration/validate`.

netsim/ansible/templates/bgp/arcos.j2
```diff
@@ -0,0 +1,61 @@
...
+#}
+network-instance default protocol BGP b1
+ global as {{ bgp.as }}
+ global router-id {{ bgp.router_id }}
+ global afi-safi IPV4_UNICAST
+{% if bgp.ipv6|default(false) %}
+ global afi-safi IPV6_UNICAST
+{% endif %}
+{% for a in bgp.advertise|default([]) %}
+{% if a.ipv4 is defined %}
```
ipspace: One more generic comment -- I strongly recommend you use the same indentation trick we use in other Jinja2 templates to make the templates more readable. Unfortunately, that's the best I could come up with, and it looks like some vendors picked it up as well.

For example:

```
{% for x %}
{%   if y is defined %}
{%   endif %}
{% endfor %}

**Comment by roc-ops** [comment]
Thanks for the review, Ivan — all three addressed.

**1. Deploy method.** Switched the ArcOS device from the Ansible deploy task to netlab's native container-script (`sh`) config mode, following cRPD: a wrapper `netsim/templates/provider/clab/arcos/netlab-config.j2` is mapped to `/config/netlab/netlab-config.sh`, and each module is rendered with a `#!/config/netlab/netlab-config.sh` shebang and executed via `docker exec`, which hands the rendered config to the wrapper to load through `confd_cli` (`load merge`/`commit`). `netsim/ansible/tasks/deploy-config/arcos.yml` is removed; no Ansible is involved in deployment. I kept the docker connection only for the validation/collect path (the tested build has SSH/NETCONF/gNMI disabled). Re-verified on a real `arcos:8.2.1A.P2` container: `netlab up` deploys via the mapped script and OSPF reaches FULL / iBGP ESTABLISHED.

**2. Platform tests.** Removed `tests/integration/platform/arcos/` entirely — the device is exercised by the regular integration tests.

**3. Jinja2 indentation.** Applied the readable nested-indentation convention (tag keyword indented with the block) across the ArcOS templates.

Rebased onto current `dev`.

File: docs/labs/arcos.md [hunk 2]
```diff
@@ -0,0 +1,54 @@
+# Installing Arrcus ArcOS
```
ipspace, line 2: This file should be in docs/labs. The "docs/install" directory is for netlab installation, the "docs/labs" directory is for providers and devices.

docs/release/26.07.md
```diff
@@ -14,6 +14,7 @@
 * The [**WireGuard tunnel** plugin](plugin-tunnel-wireguard) supports WireGuard tunnels on FRR.
 * The [**bgp.session** plugin](plugin-bgp-session) and the [OSPF module](module-ospf) support graceful restart on Arista EOS, BIRD, FortiOS, and FRR
 * The [**bgp.policy** plugin](plugin-bgp-policy) supports the **bgp.role** attribute on FRR and BIRD.
+* [Arrcus ArcOS](https://arrcus.com/) is now supported (best-effort) as a native containerlab node, with configuration deployment and validation over a `docker exec` + `confd_cli` path.
```
ipspace: This should be in 26.08. 26.07 is already out.

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
 * Device configurations that contain `no lldp transmit` or `no lldp receive` configuration command trigger configuration reload failures due to an Arista EOS bug ([more details](https://github.com/ipspace/netlab/issues/2577)). These commands are thus automatically removed from collected device configurations.
 
+(caveats-arcos)=
+## Arrcus ArcOS
+
+* ArcOS ships a commercial NOS image; there is no public Vagrant/Containerlab box. ArcOS runs as a **native** containerlab node (`kind: arrcus_arcos`) -- no vrnetlab packaging. Provide your own container image and tag it as in `netsim/devices/arcos.yml` (`clab.image`), or override `clab.image` in your topology.
+* The tested container image boots with **SSH, NETCONF, and gNMI all disabled** (its baked-in startup config fails to load a built-in prefix-set). *netlab* bootstraps the node with the native `netlab_start_exec` group_var (containerlab's post-start `exec:`), which enables `ssh-server`, changes the factory-default admin-user password (ArcOS refuses to enable interfaces until this is done), and creates an AAA user. No custom deploy code and no SSH is used to run it.
```
ipspace: I would just point to the labs/arcos.md document instead of repeating the points already made there.

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
+* `netlab validate` reads ArcOS state over the **same** docker-exec path, running `show <path> | display json | confd_cli` and parsing the OpenConfig JSON (`netsim/validate/**/arcos.py`). This uses netlab's standard device-side show-command validation -- **no** `ansible` validation action, and no SSH/NETCONF/gNMI.
+* **MTU:** the `interface <if> mtu <n>` leaf is not a pass-through -- ArcOS programs the real kernel netdev MTU (and the value it advertises in OSPF DBD packets) to `<n> - 26` on every commit. `netsim/ansible/templates/initial/arcos.j2` therefore sets the leaf to `l.mtu + 26` so ArcOS lands on netlab's intended MTU; this alone makes an ArcOS<->FRR OSPF adjacency converge, with no containerlab/core change.
```
ipspace: I would just say that initial device configuration sets ArcOS to lab topology layer-3 MTU (interface.mtu) + 26.

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
+* **Static routing validation** reads the committed route from the running-config JSON rather than an operational RIB: this virtual image does not populate an OpenConfig STATIC operational-state/RIB tree. A present, complete static-route entry proves netlab rendered it and the device accepted and committed it; kernel FIB install and forwarding were verified separately.
```
ipspace: I wouldn't go into so many details. I would just say something along the lines of "netlab configures static routes and they are used for traffic forwarding, but the tested ArcOS image does not report them in OpenConfig RIB tree"

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
+* Several modules **render and commit** but their datapath is **not exercised** by the container image: SR-MPLS, MPLS/LDP, and SRv6 bring up the control plane only (the native container has no kernel MPLS platform-labels); DHCP relay renders but its relay datapath needs a real DHCP-server probe, so ArcOS is intentionally omitted from the [DHCP support table](platform-services-support).
```
ipspace: Just say "while netlab configures SR-MPLS, MPLS/LDP, and SRv6, the dataplane does not work in the container image" or some such. You could also disable these features and describe how to enable them in `labs/arcos.md`(let's keep caveats short, they're too long anyway) in case someone wants to use netlab with a physical device).

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
+* **Genuinely unsupported**, not merely untested (and therefore not declared in `features`): EVPN symmetric IRB / L3VNI (a route-type-5 IP-prefix route is never originated -- only the L2VNI / distributed-anycast-gateway form is declared); VRF route-target import/export leaking (no rd/RT/L3VPN model on this image); OSPFv2 area-interface authentication; routing-policy large/extended-community set-actions (ArcOS has only an inline standard-community set action); redistribution sources other than static; and the `bgp.session` and other dotted plugins.
```
ipspace: I would skip this. Things don't work (for whatever reason), they are not implemented, they are not enabled in features, so you can't use them. Let's move on.

This is nothing special; many devices are like this.

docs/labs/arcos.md
ipspace, line 2: Also, add "(build-arcos)=" or "(labs-arcos)=" anchor so you don't have to hyperlink to a file name.

docs/platforms.md
```diff
@@ -112,6 +113,7 @@ You cannot use all supported network devices with all virtualization providers.
 | Virtual network device | Vagrant<br />[Libvirt](labs/libvirt.md) | [Containerlab](labs/clab.md) |
 | ------------------ | :-: | :-: |
 | Arista vEOS        |  [✅](build-eos)  |  [✅](build-ceos)  |
+| Arrcus ArcOS       |  ❌  |  ✅  |
```
ipspace: Add a link to the anchor at the top of labs/arcos.md

docs/caveats.md
```diff
@@ -56,6 +56,19 @@ nodes:
...
+* Configuration is deployed with netlab's native containerlab **"sh" config mode**: each module is rendered into `/config/netlab/NN-<module>.sh` and executed via `docker exec`, loading config through `confd_cli` (`load merge` / `commit`) via the mapped wrapper `netsim/templates/provider/clab/arcos/netlab-config.j2` -- the same mapped-script pattern Juniper cRPD uses, with no Ansible in the deploy path. Validation/collect use `ansible_connection: docker` (`docker exec` running `confd_cli`). ArcOS ships the official `arrcus.arcos` `network_cli` collection (the project-recommended model), but every published version (2.0.13-2.0.18) hangs against this image: its cliconf/terminal plugins issue a `config terminal` / `arcos_cli` mode-switch that this build's `confd_cli` rejects, so it is not used.
```
ipspace: I try to keep caveats short. They should explain how things work or what doesn't, but don't have to go into all the underlying details.

For example, would keep this one short saying "configuration is deployed with bash script executed within the ArcOS container" or some such, maybe adding "there is no dependency on Ansible, as the arrcus.arcos network_cli plugin does not work with the ArcOS release xxx" as a separate bullet.

However, hese detailed implementation/testing notes are really valuable, so it would be a shame to just shorten/remove them. Move them to a new section at the end of "labs/arcos.md", maybe "Implementation and Testing Notes".

netsim/ansible/templates/bgp/arcos.j2
```diff
@@ -0,0 +1,61 @@
...
+  iBGP session addressed at the peer's loopback stays IDLE forever -- ArcOS does not auto-source
+  from the interface with the best route to the peer.
+
+  The bgp.session plugin (per-neighbor password/gtsm/timers/passive/bfd) is a separate template
+  and is not yet supported.
+#}
+network-instance default protocol BGP b1
+ global as {{ bgp.as }}
+ global router-id {{ bgp.router_id }}
+ global afi-safi IPV4_UNICAST
```
ipspace: You might want to use "if bgp.ipv4 is defined" as well.

netsim/ansible/templates/bgp/arcos.j2
```diff
@@ -0,0 +1,61 @@
...
+{% if bgp.ipv6|default(false) %}
```
ipspace: Nit: Nothing wrong with this, but it would be easier to use "if bgp.ipv6 is defined". It is never set to False.

netsim/ansible/templates/bgp/arcos.j2
```diff
@@ -0,0 +1,61 @@
...
+{% for a in bgp.advertise|default([]) %}
+{%   if a.ipv4 is defined %}
+ global afi-safi IPV4_UNICAST network {{ a.ipv4 }}
+{%   endif %}
+{%   if a.ipv6 is defined %}
+ global afi-safi IPV6_UNICAST network {{ a.ipv6 }}
+{%   endif %}
+{% endfor %}
+{% for n in bgp.neighbors|default([]) %}
+{%   if n.ipv4 is defined %}
```
ipspace: Hint: Instead of repeating the same code snippet (which can lead to nasty problems once you start adding functionality), you could just add an inner loop iterating of "ipv4","ipv6" and have a local dictionary (or even device group variable) to map AF into afi-safi instance name (or use something like `n_af|upper+"_UNICAST"`)

netsim/ansible/templates/bgp/arcos.j2
```diff
@@ -0,0 +1,61 @@
...
+ neighbor {{ n.ipv4 }} peer-as {{ n.as }}
+{%     if n.local_as is defined %}
+ neighbor {{ n.ipv4 }} ebgp-local-as local-as {{ n.local_as }}
+{%     endif %}
+{%     if n.type|default("") == "ibgp" and n._source_intf is defined and n._source_intf.ipv4 is defined %}
```
ipspace: Nits:

A) _source_intf is defined only when needed (IBGP or multihop EBGP), so there's no need to check BGP neighbor type.
B) The way we set up Jinja2 environment, you don't have to check every component in the object tree. `n._source_intf.ipv4 is defined` or `n._source_intf[n_af] is defined` (if you iterate over AFs) works.

netsim/ansible/templates/dhcp/arcos.j2
```diff
@@ -0,0 +1,23 @@
...
+#}
+{% for l in interfaces|default([]) if l.dhcp is defined and l.dhcp.relay is defined %}
+{%   if l.dhcp.relay.ipv4|default([]) %}
+relay-agent dhcp interface {{ l.ifname }} enable true
+relay-agent dhcp interface {{ l.ifname }} helper-address [ {% for s in l.dhcp.relay.ipv4 %}{{ s }} {% endfor %}]
+{%     if l.dhcp.vrf is defined and l.dhcp.vrf %}
+relay-agent dhcp interface {{ l.ifname }} server-vrf {{ 'default' if l.dhcp.vrf == 'global' else l.dhcp.vrf }}
+{%     endif %}
+{%   endif %}
+{#- DHCPv6 relay: SAME per-interface shape, "relay-agent dhcpv6" instead of "relay-agent dhcp"
```
ipspace, line 14: Nit: Yet again, you could iterate over AFs and then use `{% set dhcp_kw = 'dhcp' if d_af == 'ipv4' else 'dhcpv6' %}` and use `{{ dhcp_kw }}` in the commands to ensure the IPv4 and IPv6 configurations stay in sync.

netsim/ansible/templates/evpn/arcos.j2
```diff
@@ -0,0 +1,60 @@
...
+   * default-NI BGP gets L2VPN_EVPN globally + per EVPN neighbor (one-liners OK).
+   * iBGP-over-loopback EVPN sessions REQUIRE `transport local-address <loopback>` or they stay
+     IDLE (same landmine documented in templates/bgp/arcos.j2).
+   * each EVPN VLAN's auto-created L2VLAN NI gets advertise-mac-routes + a MAC-VRF BGP instance
+     carrying RD/RT -- this block MUST be mode-descent (the flat one-liner form aborts:
+     "network-instance vlanN protocol BGP vlanN: element does not exist"). BGP instance name
+     matches the main bgp template's ("b1").
+#}
+network-instance default protocol BGP b1 global afi-safi L2VPN_EVPN
+{% for n in bgp.neighbors|default([]) if n.evpn|default(false) and n.ipv4 is defined %}
```
ipspace, line 22: Nit: n.evpn is set only when the neighbor is usable (has the desired AF), and it's set to the transport AF, so you can directly use `n[n.evpn]`to get the neighbor IP.

netsim/ansible/templates/evpn/arcos.j2
```diff
@@ -0,0 +1,60 @@
...
+network-instance default protocol BGP b1 neighbor {{ n.ipv4 }} afi-safi L2VPN_EVPN
+{%   if n._source_intf is defined and n._source_intf.ipv4 is defined %}
```
ipspace, line 25: This seems to duplicate the BGP configuration.

netsim/ansible/templates/gateway/arcos.j2
```diff
@@ -0,0 +1,12 @@
+{#
+  ArcOS first-hop redundancy -- VRRP. Config path hangs off the interface PRIMARY
+  address: interface <if> subinterface 0 ipv4 address <primary> vrrp vrrp-group <vrid>
+  virtual-address <vip> | priority <p> | preempt ... | advertisement-interval ...
+  Anycast gateway is not declared -- untested on this image.
+#}
+{% for l in interfaces|default([]) if l.gateway is defined and l.gateway.protocol|default('') == 'vrrp' and l.ipv4 is defined and l.ipv4 is string %}
```
ipspace: Nit: `l.gateway.protocol|default('')` works even when `l.gateway` is not defined.

netsim/ansible/templates/initial/arcos.j2
```diff
@@ -1,29 +1,74 @@
...
  enable true
  subinterface 0
-   {% if 'ipv4' in loopback %}
-   ipv4 address {{ loopback.ipv4 | ansible.netcommon.ipv4('address') }} prefix-length 32
-   {% endif %}
-   {% if 'ipv6' in loopback %}
-   ipv6 address {{ loopback.ipv6 | ansible.netcommon.ipv6('address') }} prefix-length 64
-   {% endif %}
+{% if loopback.ipv4 is defined %}
+   ipv4 address {{ loopback.ipv4 | ansible.netcommon.ipv4("address") }} prefix-length 32
```
ipspace: Unless ArcOS requires the system loopback to have /32 prefix, please use the actual prefix length.

netsim/ansible/templates/initial/arcos.j2
```diff
@@ -1,29 +1,74 @@
...
+{% endif %}
+{% if loopback.ipv6 is defined %}
+   ipv6 address {{ loopback.ipv6 | ansible.netcommon.ipv6("address") }} prefix-length 128
```
ipspace: Same as IPv4 -- unless /128 prefix is mandatory, please use the actual prefix length (or document this as a caveat).

netsim/ansible/templates/initial/arcos.j2
```diff
@@ -1,29 +1,74 @@
...
+  netsim/augment/links.py for every device), the leaf must be set to l.mtu + 26, not l.mtu --
+  setting the leaf to the bare intended value visibly "commits clean" but silently under-shoots
+  the real kernel/OSPF-advertised MTU by 26. This device-side compensation alone makes an
+  ArcOS<->FRR OSPF adjacency converge (both sides agree on 1500); no containerlab/core change is
+  required.
+#}
 system hostname {{ inventory_hostname }}
 !
-!
 interface loopback0
```
ipspace: This can probably be just {{ loopback.ifname }}

netsim/ansible/templates/initial/arcos.j2
```diff
@@ -1,29 +1,74 @@
...
-   ipv6 address {{ loopback.ipv6 | ansible.netcommon.ipv6('address') }} prefix-length 64
-   {% endif %}
+{% if loopback.ipv4 is defined %}
+   ipv4 address {{ loopback.ipv4 | ansible.netcommon.ipv4("address") }} prefix-length 32
+{% endif %}
+{% if loopback.ipv6 is defined %}
+   ipv6 address {{ loopback.ipv6 | ansible.netcommon.ipv6("address") }} prefix-length 128
+{% endif %}
 !
+{% for l in interfaces|default([]) if l.type|default('') == 'loopback' and l.ifname != 'loopback0' %}
```
ipspace: System loopback is never in `interfaces`. However, if you need **all** interfaces (including system loopback), then you should use `netlab_interfaces` (don't ask, long story ;)

netsim/ansible/templates/initial/arcos.j2
```diff
@@ -1,29 +1,74 @@
...
+interface {{ l.ifname }}
+ enable true
+ subinterface 0
+{%   if l.ipv4 is defined and l.ipv4 is string %}
```
ipspace: No IPv6 on loopbacks? If that's a limitation, it should definitely go under "caveats".

OTOH, if that's not the case, I would configure all interfaces in a single loop, using `for ... if` condition to select L3 interfaces, and maybe with tests on interface type inside the loop to configure interface-type-specific tweaks, to ensure all interface configs are in sync.

netsim/ansible/templates/isis/arcos.j2
```diff
@@ -0,0 +1,80 @@
...
+#}
+{% set typ = isis.type|default("level-2") %}
+{% set caps = {"level-1":"LEVEL_1","level-2":"LEVEL_2","level-1-2":"LEVEL_1_2"} %}
+{% set do_l1 = typ in ["level-1","level-1-2"] %}
+{% set do_l2 = typ in ["level-2","level-1-2"] %}
+{% if isis.net is defined %}
+network-instance default protocol ISIS i1
+ global net {{ isis.net }}
+ global level-capability {{ caps[typ]|default("LEVEL_2") }}
+ global af IPV4 UNICAST enabled true
```
ipspace, line 31: Maybe use `if isis.af.ipv4 is defined` here, otherwise you'd run IPv4 topology on an IPv6-only box.

netsim/ansible/templates/isis/arcos.j2
```diff
@@ -0,0 +1,80 @@
...
+{%   if isis.af.ipv6|default(false) %}
+ global af IPV6 UNICAST enabled true
```
ipspace, line 34: BTW; is this multi-topology IPv6? We expect IS-IS devices to use multi-topology for IPv6.

netsim/ansible/templates/isis/arcos.j2
```diff
@@ -0,0 +1,80 @@
...
+{%   endif %}
+{%   if do_l1 %}
+ level 1 enabled true
+{%   endif %}
+{%   if do_l2 %}
+ level 2 enabled true
+{%   endif %}
+ interface loopback0 enabled true
+ interface loopback0 passive true
+ interface loopback0 af IPV4 UNICAST enabled true
```
ipspace, line 45: As before, consider IPv6-only deployments (and I will stop repeating this comment, it probably applies to the rest of the PR as well).

netsim/ansible/templates/isis/arcos.j2
```diff
@@ -0,0 +1,80 @@
...
+    (control plane; FIB install is container-limited, same as SR-MPLS). Two ArcOS
+    rules: (1) the ISIS instance must be COMMITTED before the table-connection references it ->
+    #COMMIT split; (2) ISIS dst rejects `default-import-policy` and the BARE table-connection
+    redistributes nothing (implicit reject) -> an explicit accept import-policy is required.
+    connected(DIRECTLY_CONNECTED) has no addressable src-instance (same gap as BGP import) ->
+    feature declares import: [ static ]. -#}
+{%   if isis.import.static is defined %}
+#COMMIT
+{%     set _pol = (isis.import.static.policy ~ "-ipv4") if (isis.import.static is mapping and isis.import.static.policy is defined) else "isis-redist-accept-ipv4" %}
+{%     if _pol == "isis-redist-accept-ipv4" %}
```
ipspace: We have IPv4 and IPv6 routes. You should either redistribute both, or write some more caveats.

netsim/ansible/templates/routing/arcos.j2
```diff
@@ -0,0 +1,77 @@
...
+{%         endif %}
+{%       endfor %}
+ exit
+!
+{%     endfor %}
+{%   endfor %}
+{% endif %}
+{% if routing.policy|default({}) %}
+{%   for p_name, p_entries in routing.policy.items() %}
+routing-policy policy-definition {{ p_name }}-ipv4
```
ipspace: This seems to be IPv4-only. We can do both protocols.

netsim/ansible/templates/routing/arcos.j2
```diff
@@ -0,0 +1,77 @@
...
+{%       endif %}
+  actions {{ 'accept-route' if p_entry.action|default('permit') == 'permit' else 'reject-route' }}
+  exit
+{%     endfor %}
+ exit
+!
+{%   endfor %}
+{% endif %}
+{% if routing.static|default([]) %}
+network-instance default protocol STATIC default
```
ipspace: Yet again, seems to be the exact same configs for IPv4 and IPv6, so maybe use an inner loop.

netsim/ansible/templates/vlan/arcos.j2
```diff
@@ -0,0 +1,41 @@
...
+interface {{ l.ifname }} ethernet switched-vlan native-vlan {{ vlans[l.vlan.native].id }}
+{%     endif %}
+interface {{ l.ifname }} ethernet switched-vlan trunk-vlans [ {% for vname in l.vlan.trunk if vname != l.vlan.native|default("") %}{{ vlans[vname].id }} {% endfor %}]
+{%   endif %}
+{% endfor %}
+{% for l in interfaces|default([]) %}
+{%   if l.type|default("") == "svi" %}
+interface {{ l.ifname }} enabled true
+interface {{ l.ifname }} autostate false
+{%     if l.ipv4 is defined and l.ipv4 %}
```
ipspace: What about IPv6 over VLANs? Also: MTU?

netsim/ansible/templates/vrf/arcos.j2
```diff
@@ -0,0 +1,88 @@
...
+ global router-id {{ vdata.ospf.router_id }}
+ area {{ varea3 }}
+{%       for l in vdata.ospf.interfaces|default([]) if l.ipv6 is defined %}
+  interface {{ l.ifname }}
+{# same fix as templates/ospf/arcos.j2: only fall back to l.type == 'p2p' when
+   ospf.network_type wasn't explicitly set, so an explicit "broadcast" isn't overridden #}
+{%         set nt = l.ospf.network_type|default('point-to-point' if l.type|default('') == 'p2p' else '') %}
+{%         if nt == 'point-to-point' %}
+   network-type POINT_TO_POINT_NETWORK
+{%         endif %}
```
ipspace, line 107: What about cost, passive interfaces, and the like? We usually define macros in OSPF/BGP/IS-IS templates and then reuse them in VRFs

netsim/ansible/templates/ospf/arcos.j2
```diff
@@ -1,18 +1,65 @@
...
+  interface loopback0
+{%   endif %}
+{{ area_interfaces('ipv4') }}
+{% endif %}
+{% if ospf.router_id is defined and ospf.af.ipv6|default(false) %}
+{%   set area3 = '0' if (ospf.area|default('0.0.0.0')) == '0.0.0.0' else ospf.area %}
+network-instance default protocol OSPF3 p{{ pid }}
+ global router-id {{ ospf.router_id }}
+ global log-adjacency-changes LOG_ADJ_ENABLE_DETAILED
+ area {{ area3 }}
```
ipspace: I don't see where you define other areas

ipspace (changes requested):
Lots of nits and a few major showstoppers:

A) Changes to documentation (file placement, caveats)
B) IPv4/IPv6 feature parity
C) Correct configuration of IPv4-only, dual-stack and IPv6-only environments. I'm pretty sure we're not testing every potential error, so incorrect configurations could pass the integration tests just because we never thought that someone could do X.
D) Global/VRF feature parity. We expect people to use same/similar configuration for global and VRF routing protocols, and thus do not check every detail like support for areas or OSPF cost in VRF OSPF instances.
E) OSPF areas

Also, while I cannot test stuff (and wouldn't even bother with a vendor that does not have a way to download images for non-customers), I would strongly recommend you run the full set of integration tests. I'm sure some of them (like OSPF multi-area test) would fail.

roc-ops:
Thanks Ivan — I took the integration-test recommendation seriously and ran the full set. **You were right about multi-area OSPF.**

**Results: 25 → 77 of 181 tests passing**, measured before and after with the same harness and stock addressing, no regressions. Every deploy failure eliminated (24 → 0).

**On (E), OSPF areas — your prediction was correct, and here is the mechanism.** `ospf/ospfv2/02-areas` fails. The ABR forms adjacencies in every area and floods correct inter-area Type-3 summaries for transit (e.g. /30) networks, but it summarises a **/32 loopback host route with a `/0` network mask** — read from a neighbour's database, the ABR's own summary LSA for its `10.0.0.1/32` loopback arrives as `Link State ID: 10.0.0.1 ... Network Mask: /0`, and the neighbour mis-installs it as a `0.0.0.0/0` default. The same LSA set carries the /30 transit prefixes with the correct mask, so it is specific to host routes. OSPFv3 gets it wrong differently — it advertises the loopback with the **interface prefix length instead of /128**, so the prefix is present but not as a host route. The same thing happens **inside a VRF**. Net effect: a loopback placed in a non-backbone area is not reachable from other areas; loopbacks in area 0 and all inter-area transit prefixes are unaffected. This is documented in `docs/caveats.md` rather than silently dropped, and `ospf.areas` parameters are refused outright rather than half-rendered.

**On (B) and (D), IPv4/IPv6 and global/VRF parity** — both axes were swept, and gaps were either implemented or declared honestly. Several capabilities that were previously declared unsupported turned out to work, and are now implemented and verified on the device rather than from `show running-config`:

* **STP** — Rapid PVST only (the `stp enabled-protocol` leaf enumerates `NONE | RAPID_PVST`, so no 802.1D/RSTP/MSTP). Cross-vendor root election verified against an EOS peer. `features.stp.port_type` is deliberately **not** declared, because the flag is a bare boolean and ArcOS has no per-port equivalent of `network` (bridge assurance is global).
* **BGP large communities** — match, set and delete. These were previously rendered as a comment and silently dropped.
* **VRF static routes** and **discard/blackhole static routes**.
* **OSPF conditional default origination** — the platform can do it via a table-connection; what it cannot do is `always`, and `cost`/`type` are accepted and then break redistribution, so `features.ospf.default` stays undeclared and the reason is documented.
* **Routing-policy prefix lists** — per-entry actions were being flattened so that a list with a `deny` entry filtered everything.

Others were removed from the feature list because the platform genuinely cannot do them.

**On (A)** — the per-platform integration tests are gone, as you asked. The documentation placement I have not fixed yet; it will be in the next push.

**On (C)** — dual-stack is covered by the suite and passes. IPv6-only and IPv4-only topologies I have **not** systematically verified, so I would rather say that than claim it.

**One process question.** Addressing all of this has grown the branch considerably — it now touches 17 template modules. Would you prefer it staged as a series of smaller dependent PRs (device skeleton → IGP → BGP/routing-policy → L2 → overlay → remaining modules), or kept as one? Happy either way; I would rather ask than force-push a restructure onto a review you are part-way through.

ipspace:
> **Results: 25 → 77 of 181 tests passing**, measured before and after with the same harness and stock addressing, no regressions. Every deploy failure eliminated (24 → 0).

Does that mean there are zero tests that fail in **config** or **validate** phase, or just that the configurations generated through the templates are not invalid?

> **On (E), OSPF areas — your prediction was correct, and here is the mechanism.** `ospf/ospfv2/02-areas` fails.

I'm guessing this hints at what the above answer might be ;) Anyway, would love to know the exact details.

> Both axes were swept, and gaps were either implemented or declared honestly.

There are still several unresolved documentation comments. Even if you feel like the whole thing is ready to be deployed, please address them.

> **On (C)** — dual-stack is covered by the suite and passes. IPv6-only and IPv4-only topologies I have **not** systematically verified, so I would rather say that than claim it.

There are IPv4-only and IPv6-only BGP/OSPF tests; it's just that in many cases, we don't check whether the tested device has also enabled the other address family.

> **One process question.** Addressing all of this has grown the branch considerably — it now touches 17 template modules. Would you prefer it staged as a series of smaller dependent PRs (device skeleton → IGP → BGP/routing-policy → L2 → overlay → remaining modules), or kept as one?

Keep it as one PR. Once the documentation is polished, I'll merge it (I can't test it anyway 🤷🏻‍♂️)

roc-ops:
> Does that mean there are zero tests that fail in **config** or **validate** phase, or just that the configurations generated through the templates are not invalid?

Good question, and the honest answer is "the first for **config**, definitely not for **validate**". The full breakdown of the 181:

| | |
|---|---:|
| PASS (deployed and all checks green) | **77** |
| FAIL (deployed, one or more validation checks failed) | **27** |
| WARN (deployed, checks green, validation emitted a warning) | 8 |
| refused at transformation as unsupported | 69 |
| **failed the config phase** | **0** |

So: zero deploy failures — every topology that transforms gets a configuration ArcOS accepts and commits. But **27 tests still fail validation**, which is the number I should have led with rather than the 77. It went from 24 config-phase failures to none, and from 45 validation failures to 27.

> I'm guessing this hints at what the above answer might be ;) Anyway, would love to know the exact details.

`ospf/ospfv2/02-areas` is one of the 27, and the mechanism is a genuine ArcOS defect rather than a template problem.

The ABR accepts interfaces in several areas, forms adjacencies in each, and floods correct inter-area Type-3 summaries for **transit** prefixes. What it gets wrong is a **/32 loopback host route**: read from a neighbour's database, the ABR's own summary LSA for its `10.0.0.1/32` loopback arrives as

```
Link State ID: 10.0.0.1 ... Network Mask: /0
```

and the neighbour installs it as a `0.0.0.0/0` default. The same LSA set carries the /30 transit prefixes with the correct mask, so it is specific to host routes.

OSPFv3 gets it wrong differently — it advertises the loopback with the **interface prefix length instead of /128**, so the prefix is present but not as a host route. That is the quieter of the two failures. The same thing happens **inside a VRF**: a per-VRF instance acting as ABR shows the identical `Network Mask: /0` for the VRF loopback while summarising that VRF's transit prefixes correctly.

Net effect: a loopback placed in a **non-backbone** area is unreachable from other areas; loopbacks in area 0 and all inter-area transit prefixes are fine. It is written up in `docs/caveats.md`, and `ospf.areas` parameters are refused outright rather than half-rendered.

> There are IPv4-only and IPv6-only BGP/OSPF tests; it's just that in many cases, we don't check whether the tested device has also enabled the other address family.

That is exactly the hole, and there was one on our side. Rendering ten topologies across the modules in IPv4-only / dual-stack / IPv6-only turned up an **IPv6-only node emitting an empty OSPFv2 instance** — router-id and log-adjacency-changes, no area, no interfaces, on a node whose addresses are all IPv6. ArcOS accepts and keeps it without complaint, and no test catches it: the render succeeds, the commit succeeds, and the IPv6-only topologies that do fail, fail on their own IPv4 assumptions and would fail identically with the block absent.

Cause was `ospf/arcos.j2` using `ospf.af.ipv4|default(true)` — the only `|default(true)` address-family gate in the ArcOS templates, against `|default(false)` in the OSPFv3 guard, all six IS-IS guards and the VRF guards. Fixed, with the mirror cases verified: IPv4-only still emits its OSPFv2 block, dual-stack still emits both, and a render diff over 47 topologies × 3 modes shows the only change anywhere is the removal of those empty stanzas.

I am now sweeping the rest of the PR for the same class, per your note that it probably applies more widely, and working through the outstanding documentation comments.

roc-ops:
Pushed. This is the first time the branch has been updated since your review — you had only ever seen the initial two commits, so there are 90 on top of that, plus a merge of current `dev`. Summary of what changed and what I verified.

## Documentation (the merge condition)

* `docs/caveats.md` ArcOS section cut from **3973 words to 343**, 57 lines to 33. Nine caveats kept — the ones a user actually hits — and the rest either dropped or moved into `docs/labs/arcos.md` where the installation and bootstrap detail belongs.
* Added a `(build-arcos)` anchor and linked it from the platform table, so the "best-effort / no public image" story is reachable from where a user starts.
* Release note moved to `26.08`.

## The address-family sweep

You said it probably applied to the rest of the PR, and you were right — there was a second instance. I rendered **every module the device declares in IPv4-only, dual-stack and IPv6-only: 211 topologies, 633 combinations**, and read the rendered configuration rather than only scoring tests.

Two shapes were hunted:

**"A block emitted for a family the node does not have" — one more defect.** `vrf/arcos.j2` guarded the per-VRF OSPFv2 block only on `vdata.ospf is defined`, while the per-VRF OSPFv3 block sixteen lines below already checked `vdata.ospf.af.ipv6`. An IPv6-only VRF got a stranded `network-instance <vrf> protocol OSPF p-<vrf>` carrying a router ID and nothing else. Same class as the `ospf/arcos.j2` one, missed the first time because no topology in that first pass had an IPv6-only VRF.

Verified the same way as before: 88 topologies × 3 modes, base vs branch, gives 16 differing combinations and **all 16 are the identical two-line removal with zero lines added anywhere** — 28 stanza headers and their 28 router-id lines, across five VRF names in nine topologies. All 54 OSPF/VRF/VLAN integration tests render identically before and after.

**"A block emitted for only one family when both are present" — zero hits.** Nowhere in the 633 combinations does a dual-stack render carry fewer constructs of a family than the corresponding single-family render. That is the shape your `routing/arcos.j2:28` and `isis/arcos.j2:70` comments were pointing at, and it does not occur on the branch.

On your five specific pointers: four of them were already addressed by `e58e18decb` (IPv4/IPv6 parity) and `010d468f8e` (IPv6 addresses + MTU on SVIs), which landed after your review — so those line numbers no longer land on the code they describe. Each is now confirmed by a render: an IPv6-only IS-IS node emits zero IPv4 constructs; dual-stack IS-IS redistribution emits both `STATIC ISIS IPV4` and `STATIC ISIS IPV6`; static routes render both families from one loop; a dual-stack SVI renders the MTU leaf and both addresses. IS-IS is still single-topology, and both the caveat and `platforms.md` still say so.

## One thing I did not change, because I do not think it is mine to decide

`routing/arcos.j2` renders every routing policy twice, `-ipv4` and `-ipv6`, unconditionally, so a single-family node carries one dead policy definition. Gating that looks like a two-line fix, and I believe it is unsafe: `bgp/07-ebgp-localas` sets `bgp.originate: 172.42.42.0/24` on a node with no IPv4 interface, so its BGP instance legitimately carries `afi-safi IPV4_UNICAST` even in the IPv6-only render. Gating the policy on interface address families would drop the `-ipv4` policy and leave that reference dangling. The current behaviour is deliberate, documented in the template header, and reference-safe — but it is a real wart, and if you would rather gate it on something else I will take the direction.

## Merge of current `dev`

Two files conflicted, both documentation. Your new GRE bullets went under *Other Arista EOS Caveats* where they belong, with the ArcOS section after them so the file stays alphabetical; the `26.08` release note took our entry into your skeleton.

Re-verified after the merge: **140 topologies, 93 transform, 47 refused as unsupported, zero hard failures.** That includes the new ACL module — ArcOS refuses `routing/30-acl-ipv4` and `31-acl-ipv6` cleanly with *does not support 'acl' objects*, and likewise the new `routing/06-weight` and `11-match-aspath`, rather than rendering something wrong.

The 27 validation failures from the earlier breakdown are unchanged and still documented as caveats; none of them are config-phase failures.

docs/caveats.md
```diff
@@ -62,6 +62,39 @@ GRE tunnel caveats:
...
+
+ArcOS is a commercial NOS; there is no public container image. See the [ArcOS installation
+notes](build-arcos) for the image, the first-boot bootstrap, the supported configuration modules,
+and validation. Configuration is deployed with a bash script executed within the ArcOS container.
+
+* Initial device configuration sets the ArcOS interface MTU to the lab topology layer-3 MTU
+  (**interface.mtu**) plus 26.
+* _netlab_ configures static routes and they are used for traffic forwarding, but the tested ArcOS
+  image does not report them in the OpenConfig RIB tree.
+* While _netlab_ configures SR-MPLS, MPLS/LDP, and SRv6, the dataplane does not work in the
```
ipspace, line 76: In this case, I would disable MPLS, SR-MPLS, and SRv6 support, unless you want to use ArcOS as a route reflector. Sadly, we don't have a device feature covering that yet (like we have in EVPN).

Not a showstopper, you can do it after the marge.

docs/caveats.md
```diff
@@ -62,6 +62,39 @@ GRE tunnel caveats:
...
+  a non-backbone area is not reachable from other areas. Loopbacks in area 0 and all inter-area
+  transit prefixes are unaffected.
+* IS-IS IPv6 is single-topology; the tested image has no multi-topology (RFC 5120) knob. Devices
+  running multi-topology IS-IS -- which is how _netlab_ configures every other device -- will not
+  install ArcOS IPv6 routes, so IPv6 IS-IS works only in an all-ArcOS IS-IS domain. IPv4 IS-IS
+  interoperates normally.
+* ArcOS computes the IPv4 VRRPv3 checksum over the IPv4 pseudo-header and has no knob to disable
+  it, while _netlab_ configures every other device the other way, so IPv4 VRRP does not interoperate
+  with other _netlab_ devices. A single ArcOS first-hop router works. IPv6 VRRP is unaffected.
+* With static VXLAN flooding, the container image withdraws the head-end-replication entry shortly
```
ipspace, line 91: In this case, you can declare that VXLAN needs EVPN (see SR Linux), ensuring the VXLAN-only topologies generate an error. Another "not a showstopper, fix later" remark.

docs/caveats.md
```diff
@@ -62,6 +62,39 @@ GRE tunnel caveats:
...
+  after programming it, so an ArcOS VTEP cannot originate BUM traffic (receiving works). Use
+  `vxlan.flooding: evpn`.
+* A discard (blackhole) static route pointing to a directly-connected host is displaced by the
```
ipspace, line 94: Interesting. The first time I've encountered something like this. Any chance you could play with admin distance (or whatever it is called) of static routes? As before, not a showstopper ;)

ipspace (approved):
The documentation looks good, and I trust you that the config templates are in the best shape they can be. Let's move this forward; we can always fix the details later.

PR #3684, opened 2026-07-22
Add a general `ansible` validation action

### LINKED ISSUE #3676: Native post-deployment validation for a device whose CLI has no one-shot command mode over SSH (e.g. OcNOS cmlsh) — proposal + design question

roc-ops:
# Add a general `ansible` validation action

Closes #3676.

`netlab validate` reads device state either by running a `show`/`exec` command over the node connection, or from an external source (`suzieq`). Both assume the platform can return command output to a non-interactive caller. Some network OSes only expose their operational data through a vendor Ansible module — their interactive CLI rejects non-interactive command execution — so neither existing path can reach them, and today such a device cannot be validated at all.

This PR adds an **`ansible` validation action** to close that gap. It is a validation **data source**, a peer of `netsim/cli/validate/suzieq.py` — **not** a connection method. `netsim/cli/connect.py` is unchanged; interactive `netlab connect` is untouched. As discussed on #3676, it is implemented as a first-class, reusable action rather than a device-specific plugin, so any platform in the same situation can adopt it.

## How it works

- A device declares the module it validates through, e.g.:
  ```yaml
  netlab_validate:
    ansible_module: ipinfusion.ocnos.ocnos_command
  ```
- A test can invoke it directly in a `validate` entry:
  ```yaml
  validate:
    adj:
      device: <device>
      ansible: show ip ospf neighbor
      valid: "'Full' in stdout"
  ```
  or a device validation plugin can supply the command from an `ansible_<test>()` function and assert on the result in `valid_<test>()`.
- `netsim/cli/validate/ansible.py` runs the module ad-hoc against the node using the netlab-generated inventory, and returns parsed **JSON** when the command emits it (`... | json`), otherwise the raw CLI text in `stdout` for a text assertion to match.

The action is registered everywhere a validation action is expected — the topology schema (`attributes.yml`), the augment/validation checks and their hint strings (`augment/validate.py`, `hints.yml`), and the test/plugin dispatchers — so `ansible:` is accepted and executed like any built-in action, and it's documented in `docs/topology/validate.md`.

## Files

- **new** `netsim/cli/validate/ansible.py` — the validation source (peer of `suzieq.py`)
- `netsim/cli/validate/tests.py`, `utils.py`, `plugin.py` — dispatch + action registration
- `netsim/augment/validate.py`, `netsim/defaults/attributes.yml`, `netsim/defaults/hints.yml` — accept `ansible:` as a first-class validation action (schema, checks, and user-facing hints)
- `docs/topology/validate.md` — documentation
- `tests/coverage/errors/validation-checks.log` — updated golden for the new action's error text

## Notes

The first consumer is the IP Infusion OcNOS device (submitted separately), whose restricted CLI (`cmlsh`) returns a non-zero exit on every non-interactive SSH command and so can't be reached by the standard `show`/`exec` transports. The action itself is device-agnostic. Verified live against a real OcNOS lab (OSPFv2/OSPFv3, BGP and IS-IS validators passing through the action).

### DIFF
--- a/netsim/cli/validate/ansible.py
+++ b/netsim/cli/validate/ansible.py
@@ -1,0 +1,106 @@
+#
+# Ansible functions for the netlab validate command
+#
+# The 'ansible' validation action fetches device state by running a show command
+# through an Ansible module (network_cli), for devices whose CLI cannot be driven
+# by netlab's SSH/docker validation transports -- e.g. IP Infusion OcNOS, whose
+# restricted `cmlsh` rejects every non-interactive SSH command. It is a validation
+# data *source*, a peer of suzieq.py (netsim/cli/validate/suzieq.py), not a device
+# connection method.
+#
+# A test selects it with an `ansible` action (or a plugin `ansible_<test>` function)
+# that supplies the show command; the device supplies the module in
+#   defaults.devices.<device>.netlab_validate.ansible_module
+# The result is parsed as JSON when the command emits JSON (e.g. `... | json`), so a
+# `valid:` expression can use structured data; otherwise the raw CLI text is returned
+# as `_result.stdout` for a plugin valid_<test>() to screen-scrape.
+#
+import json
+import shlex
+import subprocess
+import typing
+
+from box import Box
+
+from ... import data
+from ...utils import log
+from . import report, utils
+
+
+def get_result(v_entry: Box, n_name: typing.Optional[str], topology: Box, verbosity: int) -> Box:
+  err_value = data.get_box({'_error': True})
+  if not n_name:
+    return err_value
+  node = topology.nodes[n_name]
+
+  v_cmd = utils.get_exec_list(v_entry, 'ansible', node, topology)   # command tokens
+  if not v_cmd:
+    log.error(
+      f'Test {v_entry.name}: no ansible show command for {n_name}/{node.device}',
+      category=log.MissingValue, module='validation')
+    return err_value
+  command = ' '.join(v_cmd)
+
+  a_module = topology.defaults.devices[node.device].netlab_validate.ansible_module
+  if not a_module:
+    log.error(
+      f'Device {node.device} has no netlab_validate.ansible_module; cannot use the ansible validation action',
+      category=log.MissingValue, module='validation')
+    return err_value
+
+  # Build the module arguments. Networking "<os>_command" modules take the show
+  # command in a `commands` argument, passed as a single quoted string (ansible
+  # ad-hoc rejects a bare list literal: "does not support raw params"). The plain
+  # command/shell modules take the bare command line instead; other argument shapes
+  # can be added here later if a device needs them.
+  if str(a_module).split('.')[-1] in ('command', 'shell'):
+    module_args = command
+  else:
+    module_args = 'commands=%r' % command
+
+  # Reuse the netlab-generated inventory (hosts.yml + ansible.cfg in the lab dir).
+  cmd = ['ansible', n_name, '-m', str(a_module), '-a', module_args, '--one-line']
+  cmd_str = shlex.join(cmd)
+
+  # On any failure, surface the exact command so the user can run it by hand to see
+  # what went wrong (mirrors execute_netlab_config in netsim/cli/validate/devices.py).
+  def _fail(msg: str, detail: typing.Optional[str] = None) -> Box:
+    hint = f'Run the command manually to inspect the device output:\n  {cmd_str}'
+    if detail:
+      hint = f'{str(detail).rstrip()}\n\n{hint}'
+    report.log_failure(msg, topology, more_data=hint)
+    return err_value
+
+  if verbosity >= 3:
+    print(f'Preparing to execute {command!r} on {n_name} via Ansible ({a_module})')
+
+  try:
+    out = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
+  except Exception as ex:                                   # noqa: BLE001
+    return _fail(f'Ansible action failed for "{command}" on {n_name}', str(ex))
+
+  raw = out.stdout or ''
+  # ansible --one-line: "<host> | SUCCESS => { ...json... }" (or FAILED!/UNREACHABLE!)
+  if ' => ' not in raw:
+    return _fail(f'Ansible action returned no parseable result for "{command}" on {n_name}',
+                 raw or out.stderr)
+  status, _, payload = raw.partition(' => ')
+  try:
+    a_result = json.loads(payload)
+  except Exception as ex:                                   # noqa: BLE001
+    return _fail(f'Cannot parse the Ansible result for "{command}" on {n_name}', str(ex))
+  if 'SUCCESS' not in status:
+    return _fail(f'Ansible command "{command}" failed on {n_name}', a_result.get('msg') or raw)
+
+  if verbosity >= 3:
+    print(f'Executed {command!r}, got {a_result}')
+
+  stdout = a_result.get('stdout', '')
+  text = '\n'.join(stdout) if isinstance(stdout, list) else str(stdout)
+
+  # If the device emitted JSON (e.g. `show ... | json`), hand structured data to the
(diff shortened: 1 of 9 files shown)


netsim/cli/validate/ansible.py
```diff
@@ -0,0 +1,94 @@
...
+
+  v_cmd = utils.get_exec_list(v_entry, 'ansible', node, topology)   # command tokens
+  if not v_cmd:
+    log.error(
+      f'Test {v_entry.name}: no ansible show command for {n_name}/{node.device}',
+      category=log.MissingValue, module='validation')
+    return err_value
+  command = ' '.join(v_cmd)
+
+  a_module = topology.defaults.devices[node.device].netlab_validate.ansible_module
```
ipspace, line 44: This was my only real questionable point in this PR. On one hand, it would be nice to place this into an existing dictionary (for example, into`features`, i.e. `features.validate.ansible_module` or some such). OTOH, we already use the `device.netlab_validate` dictionary to extend wait times (although `netlab_validate` is so far never used in device definition).

I don't have a strong opinion either way, so whatever works for you is fine ;)

netsim/cli/validate/ansible.py
```diff
@@ -0,0 +1,94 @@
...
+    log.error(
+      f'Device {node.device} has no netlab_validate.ansible_module; cannot use the ansible validation action',
+      category=log.MissingValue, module='validation')
+    return err_value
+
+  # Run an ad-hoc module against the node, reusing the netlab-generated inventory
+  # (hosts.yml + ansible.cfg in the lab directory). Pass the command as a single
+  # quoted string -- ansible ad-hoc rejects a bare list literal ("does not support
+  # raw params").
+  cmd = ['ansible', n_name, '-m', str(a_module), '-a', 'commands=%r' % command, '--one-line']
```
ipspace: We might have to support a more generic command format. While most networking "something_command" modules use "commands" argument, the "command" module uses none. However, we can change that later if needed.

netsim/cli/validate/ansible.py
```diff
@@ -0,0 +1,94 @@
...
+  if verbosity >= 3:
+    print(f'Preparing to execute {command!r} on {n_name} via Ansible ({a_module})')
+
+  try:
+    out = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
```
ipspace, line 78: We have "cli/external_commands/run_command" wrapper around subprocess.run in case you want to use that one. Maybe it's simpler to just go with "subprocess.run" here.

netsim/cli/validate/ansible.py
```diff
@@ -0,0 +1,94 @@
...
+  except Exception as ex:                                   # noqa: BLE001
+    report.log_failure(f'Ansible action failed for "{command}" on {n_name}',
+                       more_data=str(ex), topology=topology)
```
ipspace: It might make sense to include the actual command executed in "more_data", so the user can run the command manually to see what's wrong -- similar to how we do it in "execute_netlab_config" in devices.py (same directory)

netsim/cli/validate/ansible.py
```diff
@@ -0,0 +1,94 @@
...
+    return err_value
+
+  raw = out.stdout or ''
+  # ansible --one-line: "<host> | SUCCESS => { ...json... }" (or FAILED!/UNREACHABLE!)
+  if ' => ' not in raw:
```
ipspace, line 84: I knew the output would be messy ;)

ipspace (changes requested):
Mostly nits, the only thing that would be nice to have is a more informative printout on failure.

roc-ops:
Thanks for the quick review — glad it's mostly nits! All four addressed:

**1. More informative failure output.** Good call. The failure path now surfaces the exact ad-hoc `ansible …` command in `more_data` (mirroring `execute_netlab_config` in `netsim/cli/validate/devices.py`), so you can copy-paste and re-run it by hand. Each failure branch — subprocess error, no parseable output, JSON parse error, and a module `FAILED!`/`UNREACHABLE!` — now prints a clean one-line reason plus the runnable command instead of dumping the raw `--one-line` blob. (And yes, the old output was messy — thanks for the nudge 🙂.)

**2. subprocess wrapper.** I looked at `cli/external_commands.run_command` but kept the plain `subprocess.run` here: the action needs the module's stdout *even when ansible exits non-zero* (a `FAILED!`/`UNREACHABLE!` still returns a parseable `… => {json}` payload we read the error `msg` from), then inspects that payload rather than a bool/exit code. `run_command`'s default `check=True` raises on non-zero (losing stdout), and the `return_exitcode` + `CAPTURED_STDOUT`-global route is more convoluted than a direct `capture_output=True` — so I went with `subprocess.run`, as you said was fine.

**3. Generic command format.** Handled the common split now (low-effort): `<os>_command` modules get `commands='<cmd>'`, while the plain `command`/`shell` modules (no `commands=` arg) get the bare command line, selected by the module's last name component, with a comment that other arg shapes can be added later.

**4. `ansible_module` placement.** Kept it in `device.netlab_validate.ansible_module` — it sits alongside the existing `netlab_validate.<test>.wait` device settings used across the integration tests, so grouping it there felt consistent. Happy to move it to `features.validate` if you'd prefer; just say the word.

ipspace (approved):
Thank you!

Issue #3657, opened 2026-07-17
[BUG] IPv6 GRE test is failing on vjunos-router


ipspace:
01-gre:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/tunnel/01-gre.yml)
* [Test results](https://tests.netlab.tools/vjunos-router/clab/tunnel/01-gre.yml-validate.log)

Worst case, we could configure vJunos-router to do GRE over IPv4 only

ssasso:
I identified at least a couple of issues.

First one is MTU mismatch. Well, in OSPFv3 DBD packet junos is reporting interface MTU as 0, while FRR reports 1400.

<img width="567" height="212" alt="Image" src="https://github.com/user-attachments/assets/2098093c-2d14-486e-b96a-c8f87cf659b0" />

<img width="570" height="292" alt="Image" src="https://github.com/user-attachments/assets/b95501d0-73fa-4c52-a48e-0e002a0d6fb1" />

```
r2# show interface tun0
Interface tun0 is up, line protocol is up
  Link ups:       2    last: 2026/07/23 13:48:01.69
  Link downs:     1    last: 2026/07/23 13:48:01.69
  vrf: default
  Description: r2 -> dut [GRE tunnel]
  index 9 metric 0 mtu 1400 speed 0 txqlen 1000
  flags: <UP,LOWER_UP,POINTOPOINT,RUNNING,NOARP>
  MPLS  Not specified by CLI
  Multicast config is Not specified by CLI
  Shutdown config is Disabled by CLI
  Type: GRE over IP
  HWaddr: ac:16:01:02
  inet 172.23.0.2/24
  inet6 fe80::5efe:ac16:102/64
  inet6 2001:db8:1::2/64
  inet6 fe80::b59a:1040:bb7a:4b42/64
  Interface Type GRE
  Interface Slave Type None
  VTEP IP: 172.22.1.2 , remote 172.22.0.2
  Link Interface eth1
  protodown: off
  Parent interface: eth1
  ND advertised reachable time is 0 milliseconds
  ND advertised retransmit interval is 0 milliseconds
  ND advertised hop-count limit is 64 hops
  ND router solicit rcvd: 1
  ND router advertisements sent: 523 rcvd: 10
  ND router advertisements are sent every 5 seconds
  ND router advertisements lifetime tracks ra-interval
  ND router advertisement default router preference is medium
  Hosts use stateless autoconfig for addresses.
  Neighbor address(s):
[...]

PR #3691, opened 2026-07-24
JunOS: removing gre over ipv6, add caveats for ospfv3

ssasso:
Related to #3657 .

@ipspace if you already have an idea on how to implement the FRR's config tweak during the test I'll let you implement this


ipspace (approved):
Thank you. Adding modified test, then I'll merge it.

PR #3692, opened 2026-07-24
Stabilize MTU validation after PMTU cache updates

jbemmel:
## Summary
- Flushes the cEOS probe's IPv4/IPv6 route (PMTU) cache between the under-MTU success checks and the Frag Needed / Packet Too Big expectations in `tests/integration/initial/04-mtu.yml`.
- Adds `wait: ping_long` on the IPv6 under-MTU check for the same path.

## Why
After an oversized DF ping, Linux/cEOS learns and caches the path MTU. The next oversized ping then fails locally with "message too long" instead of eliciting a fresh ICMP unreachable from the DUT, so the PTB validation becomes flaky or fails even when the DUT is correct. Flushing the cache forces a new discovery attempt and keeps the test asserting DUT behavior.

## Test plan
- [ ] Run `tests/integration/initial/04-mtu.yml` against a DUT that previously showed intermittent PTB failures
- [ ] Confirm `v4_1400_ptb` / `v6_1400_ptb` still match Frag Needed / Packet Too Big (not local message-too-long)

### DIFF
--- a/tests/integration/initial/04-mtu.yml
+++ b/tests/integration/initial/04-mtu.yml
@@ -89,6 +89,13 @@ validate:
       eos: ping t1400 df-bit size 1400
     valid: |
       'bytes from' in stdout
+  flush_v4_1400:
+    description: Flush IPv4 PMTU cache on p1
+    nodes: [ p1 ]
+    exec:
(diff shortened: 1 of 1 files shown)


ipspace (approved):

Issue #3698, opened 2026-07-26
Replace sudo -E with explicit variable passing


jbemmel:
As mentioned here, even though it may not be the root cause of that particular issue, Netlab currently relies on ```sudo -E``` to start containerlab, which does not work as expected in some environments ( e.g. recent Ubuntu )

```
netsim/providers/clab.yml:start: sudo -E containerlab deploy --reconfigure -t clab.yml
netsim/providers/clab.yml:stop: sudo -E containerlab destroy --cleanup -t clab.yml
netsim/providers/libvirt.yml:  start: sudo -E containerlab deploy --reconfigure -t clab-augment.yml
netsim/providers/libvirt.yml:  stop: sudo -E containerlab destroy --cleanup -t clab-augment.yml
```
A better design would be to explicitly list the environment variables Netlab will pass on to subprocesses like containerlab

> I think I may have seen warnings about ‘’’sudo -E’’’ not being able to pass on the environment - will keep an eye out
> 
> ```bash
> jbemmel@jvb-laptop:~$ export MYVAR=test
> env | grep MYVAR
> sudo -E env | grep MYVAR
> MYVAR=test
> sudo: preserving the entire environment is not supported, '-E' is ignored
> ```
> 
> OS: 
> ```
> jbemmel@jvb-laptop:~$ uname -a
> Linux jvb-laptop 7.0.0-28-generic #28-Ubuntu SMP PREEMPT_DYNAMIC Sun Jun 21 01:01:36 UTC 2026 x86_64 GNU/Linux
> ```
> 
> Root cause in my case:
> ```
> sudo --version
> sudo-rs 0.2.13-0ubuntu1
> ```
> [sudo-rs](https://github.com/trifectatechfoundation/sudo-rs) is a Rust incarnation of ```sudo``` which does not support -E:
> 
> ```markdown
> Some other notable restrictions to be aware of:
> 
> Some functionality is not supported, such as preventing shell escapes using INTERCEPT and storing config in LDAP using sudoers.ldap, and cvtsudoers. This includes sudo -E, which is [discouraged](https://www.sudo.ws/docs/troubleshooting/#why-does-sudo-modify-the-commands-environment) in original sudo.
> ``` 

 _Originally posted by @jbemmel in [#3690](https://github.com/ipspace/netlab/issues/3690#issuecomment-5085174443)_

joebb97:
For containerlab commands this list of environment variables should be sufficient I would think

[...]

PR #3704, opened 2026-07-27
Fix: Use sudo-less containerlab operation

ipspace:
Avoids the need for #3698

### DIFF
--- a/docs/release/26.08.md
+++ b/docs/release/26.08.md
@@ -33,6 +33,12 @@ Arista EOS:
 ## Breaking changes
 
 * We enabled the [Transparent Management _containerlab_ feature](https://containerlab.dev/manual/vrnetlab/#management-interface) on vrnetlab-based devices supporting it in July 2026. This change should not impact older containers, but will change the management IPv4 address on virtual machines running within some newly-built containers from 10.0.0.15 (default management IPv4 address) to the outside (container) management IPv4 address.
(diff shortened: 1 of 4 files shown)


ipspace:
This should work seamlessly for most users, as containerlab adds the current user to `clab_admins` group on first installation.

It could be a problem for people running netlab under multiple usernames, thus the "breaking change" release note.

/cc @sdargoeuves @steffann

docs/release/26.08.md
```diff
@@ -33,6 +33,12 @@ Arista EOS:
 ## Breaking changes
 
 * We enabled the [Transparent Management _containerlab_ feature](https://containerlab.dev/manual/vrnetlab/#management-interface) on vrnetlab-based devices supporting it in July 2026. This change should not impact older containers, but will change the management IPv4 address on virtual machines running within some newly-built containers from 10.0.0.15 (default management IPv4 address) to the outside (container) management IPv4 address.
+* We switched to [sudo-less *containerlab* operation](https://containerlab.dev/install/#sudo-less-operation). If your containerlab-based labs fail to start, check whether your username is in the `clab_admins` group with the `groups` command. To add your username to that group, use:
```
jbemmel, line 36: As a courtesy, Netlab could run this check itself when lab deployment fails. It’s a simple 1 line check

jbemmel (approved):
It’s optional, but I think it would be nice if Netlab would simply run the group membership check and explain how to fix it only if that seems to be the problem 

Other than that, I agree this is a better solution than passing a list of env parameters

Issue #3702, opened 2026-07-27
[BUG] Incorrect MTU calculation on Wireguard tunnel


ipspace:
When setting up a dual-stack Wireguard tunnel between RouterOS7 and FRR, the tunnel MTU on the RouterOS7 side is 20 bytes larger than the FRR tunnel MTU.

### Lab topology

```
---
defaults.device: frr

addressing:
  loopback:
    ipv6: 2001:db8:1::/48
  lan:
    ipv6: 2001:db8:2::/48
  p2p:
    ipv6: 2001:db8:3::/48

plugin: [ tunnel.wireguard ]

vrfs:
  transport:

nodes:
  dut:
    module: [ vrf ]
    device: routeros7
  r2:
    module: []
    device: frr

links:
- mtu: 1580
  dut:
    vrf: transport
  r2:
- dut:
    tunnel.private_key: dwdtCnMYpX08FsFyUbJmRd9ML4frwJkqsXf7pR25LCo=
    tunnel.public_key: hSDwCYkwp1R0i33ctD73Wg2/Og0mOBr066SpjqqbTmo=
    tunnel.vrf: transport
  r2:
    tunnel.private_key: XasIfmJKikt54X+Lg4AO5m87sSkmGLb9HC+LJ/+I4Os=
    tunnel.public_key: 3p7bfXt9wbTTW2HC7OQ1Nz+DQ8hbeGdNrfx+FG+IK08=
  tunnel.mode: wireguard
```

### Transformation Results

````
% ../netlab inspect --node dut 'interfaces[1].mtu'
[INFO]    Using lab topology file topology.yml
1520

% ../netlab inspect --node r2 'interfaces[1].mtu'
[INFO]    Using lab topology file topology.yml
1500
````

### Expected behavior

MTUs should match

PR #3707, opened 2026-07-27
Bugfix: Wireguard tunnel mtu calculation for devices with a default node MTU

jbemmel:
WireGuard derives tunnel MTU from the underlay (source MTU minus encapsulation overhead) only when the tunnel interface has no mtu yet. Core augmentation was copying node/device defaults (e.g. FRR’s mtu: 1500) onto every interface without an explicit MTU, including tunnels. That made an inherited default look like a user-set value, so derivation was skipped on FRR while RouterOS7 (no device MTU default) correctly computed underlay − overhead — peers disagreed by 20–60 bytes (#3702).

Skipping that copy for type: tunnel keeps tunnel MTU ownership with the tunnel plugins, matches how loopbacks already avoid inherited MTU, and still honors an explicit link/interface mtu on the tunnel provided by the user.

### DIFF
--- a/docs/links.md
+++ b/docs/links.md
@@ -412,7 +412,7 @@ links:
   mtu: 1500
 ```
 
-[^TMTU]: Tunnel interfaces are excluded. Tunnel interface MTU is the responsibility of tunnel plugins (for example, WireGuard derives it from the underlay MTU minus encapsulation overhead).
+[^TMTU]: Excluding tunnel interfaces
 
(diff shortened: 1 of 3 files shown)


ipspace (approved):
LGTM. Thank you!

ipspace:
@jbemmel -- not exactly unrelated: I know your AI friend is chatty, but it REALLY SHOULD be aware of the 72-character commit title rule (for example, https://www.gitkraken.com/learn/git/best-practices/git-commit-message#git-commit-message-structure). As the PR# is added to the commit message, the actual PR title should be a bit shorter than that.

PR #3708, opened 2026-07-27
Containerlab podman support

ipspace:
Core functionality:
* Remove 'runtime' from the list of special clab attributes
* Do not generate node 'runtime' attribute in clab.yml from the provider defaults
* Check whether the 'docker' command is available and whether the user runs clab as root (mandatory for podman)
* Set CLAB_RUNTIME environment variable to tell containerlab to use podman (otherwise it crashes even if all nodes use podman)

Support functionality:
* Add podman installation script and post-install test
* Documentation (defaults, installation, caveats)
* Updated 'netlab test' and 'netlab install' docs

Also:
* Sort tests displayed in 'netlab test' to put 'podman' at the very end (sometimes we get lucky ;)
* Removed references to Cumulus VX test environment

### DIFF
--- a/netsim/install/podman.sh
+++ b/netsim/install/podman.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+#
+# Install a specific version of Containerlab -- it's now set in 'netlab install' before calling this script
+#
+set -e
+REPLACE="--upgrade"
+IGNORE="--ignore-installed"
+#
+echo "Update the package list"
+. apt-get-update.sh
+echo "Install podman"
+$SUDO apt-get install -y podman podman-docker
+$SUDO systemctl enable --now podman.socket
+$SUDO systemctl start podman.socket
+#
+echo "Install containerlab version $CONTAINERLAB_VERSION"
+$SUDO bash "-c" "$(curl -sL https://get.containerlab.dev)" -- -v $CONTAINERLAB_VERSION
--- a/netsim/providers/clab/labops.py
+++ b/netsim/providers/clab/labops.py
@@ -154,6 +154,6 @@ def set_clab_runtime(topology: Box) -> None:
     return
 
   log.info(f'Setting containerlab runtime to {runtime}')
-  os.environ['CLAB_RUNTIME'] = runtime
-  if runtime == 'podman' and os.environ['USER'] != 'root':
+  os.environ['CLAB_RUNTIME'] = str(runtime)
+  if runtime == 'podman' and os.geteuid() != 0:
     log.fatal('You must run netlab as the root user when using podman')
--- a/netsim/providers/clab/labops.py
+++ b/netsim/providers/clab/labops.py
@@ -1,6 +1,7 @@
 #
 # Containerlab provider module
 #
+import os
 import pathlib
 
 from box import Box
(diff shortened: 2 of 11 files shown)


ipspace:
@joebb97 -- Could you test whether this works for you? It does for me on Ubuntu 26.04, but having a RHEL/Fedora data point would be nice.

@jbemmel -- note to self: DO NOT merge random untested off-the-cuff documentation remarks (https://github.com/ipspace/netlab/blame/41cd2a36aef9f719c1626d8027c222fd06500520/docs/labs/clab.md#L286)

jbemmel:
> note to self: DO NOT merge random untested off-the-cuff documentation remarks

Is it possible that it was working for me at the time, but the ecosystem has evolved since? I don't remember exactly, but I may have been running it as root.

netsim/providers/clab/labops.py
```diff
@@ -146,3 +147,13 @@ def load_kmods(topology: Box) -> None:
...
   log.exit_on_error()
+
+def set_clab_runtime(topology: Box) -> None:
+  runtime = topology.get('defaults.providers.clab.runtime',None)
+  if not runtime:
+    return
+
+  log.info(f'Setting containerlab runtime to {runtime}')
+  os.environ['CLAB_RUNTIME'] = str(runtime)
+  if runtime == 'podman' and os.geteuid() != 0:
```
jbemmel, line 158: Could be moved up 2 lines

netsim/install/podman.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+echo "Update the package list"
+. apt-get-update.sh
+echo "Install podman"
+$SUDO apt-get install -y podman podman-docker
+$SUDO systemctl enable --now podman.socket
+$SUDO systemctl start podman.socket
+#
+echo "Install containerlab version $CONTAINERLAB_VERSION"
+$SUDO bash "-c" "$(curl -sL https://get.containerlab.dev)" -- -v $CONTAINERLAB_VERSION
```
jbemmel, line 17: This should really be calling ```netsim/install/containerlab.sh``` to make sure all its dependencies are installed etc.

jbemmel (changes requested):
Avoid duplicating Containerlab install logic

joebb97:
I'll hold off on testing until the changes requested go in

netsim/install/podman.sh
ipspace: Nice thought, but that script installs Docker (because that's how we decided to use containerlab years ago), and the only line shared between the two scripts is the "get containerlab" line. I'm not going to create a new file to share a single line (and I know it's eventually going to bite me ;)

ipspace:
> I'll hold off on testing until the changes requested go in

I rechecked: the installation and the test procedure work on Ubuntu 26.04. If you experience problems on RHEL/Fedora, please let me know.

Also: thanks a million for all the research. Made my job way easier ;)

netsim/install/podman.sh
jbemmel: So you're basically saying there are more potential issues here, including an unfortunately named key install script that does more than one would expect based on its name...

It's all good - as long as we're aware ;)

PR #3711, opened 2026-07-27
DNS Client for Arista

sdargoeuves:
After following and checking what was done on https://github.com/ipspace/netlab/pull/3678 I wanted to add the config for Arista EOS.
I'd be happy to investigate other platforms based on the images I've got.


ipspace (approved):
Works ;) Thank you!

Had to tweak the integration test to add the EOS-specific **ping** command. Please add equivalent commands for other platforms if you implement further DNS clients.

PR #3716, opened 2026-07-29
Technology-specific tunnel interface names

ipspace:
Some devices (Junos, RouterOS7, OpenBSD) use different interface names for different tunnel types. This commit adds a generic "select interface name based on interface technology" function, and uses FRR as an implementation example.


ipspace:
@snuffy22 -- This is what you asked for to implement WG on OpenBSD. I invited you as a repo collaborator so I could add you as a reviewer.

snuffy22:
Yes please, that is what I was looking for.

Thanks alot

PR #3720, opened 2026-07-30
Catch KeyboardInterrupt when aborting netlab test cleanup

### LINKED ISSUE #3696: [BUG] netlab test command crashes on abort attempt
If the "netlab test" command fails, it might be interesting to inspect the lab topology, but trying to abort
the cleanup process with CTRL/C crashes the script (with KeyboardInterrupt exception). Catch that exception and provide a decent error messsage.

Sanjays2402:
Closes #3696

Pressing CTRL/C at the `Press RETURN to continue ->` prompt in `netlab test`'s cleanup step raised an uncaught `KeyboardInterrupt` and printed a Python traceback. The prompt now catches it and exits through `log.fatal` with a message noting the test directory was not cleaned up, matching how `netlab install` and `strings.confirm` already handle CTRL/C.

Regression test added in a new `tests/test_cli.py` (mocks `input()` to raise `KeyboardInterrupt`); it fails on the current code and passes with the fix. `pytest -k 'error_cases or cli'` is green and `mypy -p netsim` is unchanged.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.


ipspace (approved):
Thanks a million. One less thing to do ;)

I removed the regression test -- the only tests we run on the CLI commands are the platform integration tests.

PR #3721, opened 2026-07-30
Rewrite routing.acl functionality (part 1)

ipspace:
* Make ACLs AF-agnostic and generate per-AF entries in the _acl.af dictionaries
* Change most src/dst attributes to lists to allow the user to specify multiple values (addresses, pools...). Also, the attributes are no longer exclusive as we collect their values into ipv4/ipv6 lists
* Simplify interface ACL settings -- AF dict is optional and created based on interface AFs
* Replace 'any' attribute with 'prefix: any'
* Accept addresses or prefixes in ipv4/ipv6 attribute
* Remove 'host' attribute and create it based on whether an IP address has a slash (not that we would need it anyway)
* Set default 'prefix: [ any ]' value for src/dst

While generating the per-AF ACL entries for the target devices, the src/dst IP lists are expanded into a cartesian product, generating as many ACL entries as needed.

Also:
* Renumber integration tests to have a distinct first number (3)

Still missing:
* node/interface/role logic is broken
* Documentation

Out-of-scope: port matching logic


ipspace:
@DanPartelly -- this is as far as I got today. Have to fix the node/intf/role logic, and update documentation.

After we agree on this part, I'll work on the port-matching logic (next PR)

DanPartelly (approved):
All OK with me. Thank you for all this work.

PR #3723, opened 2026-07-31
Rewrite ACL port matching

ipspace:
Use <src|dst>.port.<op> parameters to specify port matching parameters These parameters are rewritten into port/port_op/port_range values expected by the configuration templates, and expanded into multiple ACL entries if needed (resulting in six levels of Cartesian product)


DanPartelly (approved):
Thank you. This ended up generating great usability improvements.

PR #3729, opened 2026-08-01
Bugfix: VPP/BIRD OSPFv3 loopback prefix advertisement (#3725)

### LINKED ISSUE #3725: [BUG] VPP/BIRD does not advertise loopback IPv6 prefix in OSPFv3
06-lb-prefix:
* [Test topology](https://github.com/ipspace/netlab/blob/dev/tests/integration/ospf/ospfv3/06-lb-prefix.yml)
* [Test results](https://tests.netlab.tools/vpp/clab/ospf/ospfv3/06-lb-prefix.yml-validate.log)

It's not that BIRD would advertise /64 instead of /128; it looks like it does not advertise the loopback prefix at all. IIRC, we had a problem with a lack of LLA on BIRD loopbacks.

jbemmel:
## Summary
- Run bird control-plane initial hooks (IPv6 LLA on the loopback) from VPP `netlab-start` in the dataplane netns, via shared `initial/bird.cp.j2`
- Advertise OSPFv3 `stubnet …/128` for all loopback-typed interfaces (not only `lo`), using each interface’s own address so multi-loopback nodes get correct prefixes

## Test plan
- [x] `netlab up ospf/ospfv3/01-network.yml -d vpp -p clab` passes
- [x] Bird lab with two loopbacks per node: both `/128` stubnets present and learned by the peer
- [x] `ospf/ospfv3/06-lb-prefix.yml` on VPP/clab (original failing case)

Fixes #3725

Note: We may want to extend ```06-lb-prefix.yml``` to test the case of multiple loopbacks

### DIFF
--- a/netsim/ansible/templates/initial/bird-clab.j2
+++ b/netsim/ansible/templates/initial/bird-clab.j2
@@ -2,8 +2,5 @@
 set -x
 {% include 'linux-clab.j2' +%}
 #
-{% if loopback.ipv6 is defined %}
-set +e
-ip addr add fe80::1/64 dev lo scope link
-{% endif %}
+{% include 'bird.cp.j2' +%}
 exit 0
--- a/netsim/ansible/templates/initial/bird.cp.j2
+++ b/netsim/ansible/templates/initial/bird.cp.j2
@@ -1,0 +1,10 @@
+{#
+  Bird control-plane initial hooks that must run in the routing netns.
+
+  Used by bird-clab.j2 and by VPP (dataplane netns) so OSPFv3 can advertise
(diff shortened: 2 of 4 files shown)


ipspace (approved):
Works great. Thank you!

ipspace:
> Note: We may want to extend `06-lb-prefix.yml` to test the case of multiple loopbacks

I think BIRD is the only case where we configure the extra prefix "manually". OTOH, if we use multiple loopback, we'll have problems with Junos, which hates that concept, so we'd have to adjust the tests for it. You can't win when dealing with networking vendors :(

PR #3731, opened 2026-08-02
docs(sonic): correct the docker-sonic-vs download instructions

roc-ops:
Follow-up to #3680: the `docker-sonic-vs` download instructions I wrote in that PR were wrong in several places. I had never actually followed them — the image was already on my test box — so this corrects them after walking the path end to end and pulling SONiC `202605` build 1166687.

**The route in the original text was right; five details were not.**

* **The branch example was two years stale.** `202405` was given as the example; the current list is `master`, `202605`, `202511`, `202505`, `202411`, `202405`, `202311`, `202305`.
* **Cancelled builds sit beside successful ones** and have Artifacts links too, so "pick the latest build" needs to say *successful* explicitly.
* **`target/docker-sonic-vs-asan.gz` sits directly next to `target/docker-sonic-vs.gz`** in the artifact listing — an AddressSanitizer build, easy to take by mistake.
* **The download URL is constructible** from branch and build ID, which beats scrolling a long listing; added with a working example.
* **The `docker load` guidance had both halves backwards.** The archive already carries `docker-sonic-vs:latest`, so there is nothing to retag — and more importantly, **`docker load` silently replaces an existing image with that tag**, printing `renaming the old one with ID ... to empty string` and leaving the previous image untagged. Anyone following the old text to compare two builds would have lost the image they had. That is now a warning with a tag-first workaround.

Also adds a one-liner for identifying which build you ended up with (`cat /etc/sonic/sonic_version.yml`), which turned out to matter when comparing behaviour between builds.

Documentation only — no code, no template, no device changes.

### DIFF
--- a/docs/labs/sonic.md
+++ b/docs/labs/sonic.md
@@ -51,10 +51,6 @@ once you know the branch and the build ID from the *Build History* page:
 https://sonic-build.azurewebsites.net/api/sonic/artifacts?branchName=202605&platform=vs&buildId=1166687&target=target%2Fdocker-sonic-vs.gz
 ```
 
-*containerlab* documents the same path for its
-[`sonic-vs` kind](https://containerlab.dev/manual/kinds/sonic-vs/), which uses this image. [sonic.software](https://SONiC.software/) is an unofficial index that is sometimes offered as an alternative, but it carries SONiC *installation* images (`sonic-vs.img`, used for the Vagrant box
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Thanks a million! Love the attention to details and the striving to make things perfect ;))

I removed the reference to the alternate image source (used for Vagrant boxes) as it's not relevant for the container instructions.

PR #3732, opened 2026-08-03
Check VLAN structures exist before use in RouterOS7

snuffy22:
Add guard to check `vlans` exist, found when attempting some EVPN tests.

### DIFF
--- a/netsim/ansible/templates/vlan/routeros7.initial.j2
+++ b/netsim/ansible/templates/vlan/routeros7.initial.j2
@@ -4,15 +4,18 @@
 {#
 # Created Switched VLANs interfaces
 #}
-{% for vname,vdata in vlans.items() if vdata.mode is defined and vdata.mode != 'route' %}
+{% if vlans is defined %}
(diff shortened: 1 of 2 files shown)


netsim/ansible/templates/vlan/routeros7.initial.j2
```diff
@@ -4,15 +4,18 @@
...
 
 {#
 # Created Routed VLANs interfaces
 #}
-{% for interface in interfaces 
-     if interface.vlan.access_id is defined
-        and interface.vlan.mode|default('') == 'route' %}
-/interface/vlan add name={{ interface.ifname }} vlan-id={{ interface.vlan.access_id }} interface={{ interface.parent_ifname }}
+{% for intf in interfaces
+       if intf.vlan is defined
```
ipspace, line 17: This is not needed (but doesn't hurt either) due to the way we handle "is defined".

ipspace (approved):
Thank you!

PR #3733, opened 2026-08-03
Integration test documentation updates

snuffy22:
Revisit the DUT renaming, fix minor inconsistencies.

One non description/message change:
Remove `x2` from node list for `bgp.policy` tests 10,11,12, as we want to check for `x2` loopback on `x1`.


ipspace (approved):
Thank you, much appreciated!

PR #3737, opened 2026-08-05
EXOS: add list of reserved name as Vlan quirk

sdargoeuves:
I had an issue where I accidentally used a VLAN name reserved on an EXOS device on a topology: `fabric`
Who would want to call a Vlan `fabric` or `account` or whatever else is on that long list? Apparently me!

Not sure the correct formatting for such list, I've left it as a long line, do we want to [ruff](https://github.com/ipspace/netlab/discussions/3653) this? On this one, it then looks like a awfully long list with 200+ lines, with one entry per line, pretty ugly from my point of view.

### DIFF
--- a/netsim/devices/exos.py
+++ b/netsim/devices/exos.py
@@ -4,6 +4,18 @@ from ..utils import log
 from . import _common, _Quirks, need_ansible_collection, report_quirk
 
 EXOS_VLAN_1_NAME = 'Default'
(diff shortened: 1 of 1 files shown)


ipspace (approved):
Do I want to know how you got the list?

Thanks a million, Ivan

netsim/devices/exos.py
```diff
@@ -4,6 +4,18 @@
...
+
+
+def check_reserved_vlans(node: Box, topology: Box) -> None:
+  for vname,_ in node.get('vlans',{}).items():
```
ipspace, line 11: A nit: as you don't need VLAN data, you could use `node.get('vlans',{}).keys()` or even just `node.get('vlans',{})`

sdargoeuves:
> Do I want to know how you got the list?

It wasn't brute force!
I should have added this link as a comment: https://documentation.extremenetworks.com/exos_32.7.1/GUID-F9E4E9B8-4D90-4DCB-B325-87347F797C7B.shtml

ipspace:
> I should have added this link as a comment: https://documentation.extremenetworks.com/exos_32.7.1/GUID-F9E4E9B8-4D90-4DCB-B325-87347F797C7B.shtml

ROFL. Looks like someone couldn't write a decent parser ;))

PR #3764, opened 2026-08-11
Add 'doc_url' parameter to log.error/warning

### LINKED ISSUE #3667: Enhance  log.warn.&  log.error
Make log.error and log.warn (but no log.fatal) take an optional keyword argument accepting a URL pointing to a valid entry in the Netlab documentation. 

(New user facility. We want as many of those as possible)

ipspace:
The 'doc_url' parameter can be used to point the user to a particular documentation section while displaying an error message. The hint is usually formatted as "see also: {url}" unless the caller also specifies the format string (doc_url_text).

Proof-of-concept implementation in some error messages in STP module.


DanPartelly (approved):
That's nice. Thanks.

PR #3765, opened 2026-08-12
Consider files created by configlets in check.config

### LINKED ISSUE #3650: check.config false-positives and drops files-plugin inline configlets
When the `files` and `check.config` plugins are enabled together, a valid inline `configlets:` entry referenced by a node `config:` list is reported as missing by `check.config` and never lands on the node.

### Repro (netlab 26.07)

```yaml
provider: clab
plugin: [ files, check.config ]
defaults.device: frr
configlets:
  hello:
    frr: "! hello world configlet"
nodes:
  r1: { config: [ hello ] }
  r2: {}
links: [ r1-r2 ]
```

`netlab create` output:

```
Warning in check.config: Custom configuration hello is missing for device frr (nodes r1)
UserWarning in configs: Cannot find hello configuration template for r1/device frr
```

-> `node_files/r1/` does **not** contain the configlet.

### Control

With `plugin: [ files ]` only (no check.config), `node_files/r1/` **does** contain the `hello` configlet. So `check.config` is the cause. Order-independent — `[files, check.config]` and `[check.config, files]` both reproduce.

### Root cause

`extra/check/config/__init__.py::cleanup()` checks config-reference existence at a point where the `files` plugin (which restructures configlets in `extra/files/__init__.py::init()` and writes them via its output hook) has not landed the configlet for the node.

### Impact

The `files` plugin is the documented escape hatch for vendor-proprietary config; enabling `check.config` silently breaks it.

### Suggested fix

Make `check.config` aware of files-plugin-generated configlets (check against the transformed files / custom-config list rather than on-disk state), or reconcile the hook phase ordering; at minimum document the incompatibility.

ipspace:
* The 'check.config' main loop checks files that will be created by the 'files' plugin (the 'files' list) when trying to figure out if it's safe to use a custom configuration template.
* The 'check.config' logic was moved to the 'post_quirks' plugin hook to be executed before provider post_transform hook which creates all sorts of hard-to-deal-with data (in clab case). This should be late enough in the process to have the final 'config' list, and it's also the very last plugin hook before clab provider does its stuff.

### DIFF
--- a/netsim/extra/check/config/__init__.py
+++ b/netsim/extra/check/config/__init__.py
@@ -1,13 +1,21 @@
+import typing
 
 from box import Box
 
 from netsim import data
+from netsim.augment import devices as a_devices
 from netsim.utils import files as _files
 from netsim.utils import log, strings
 
+_execute_after    = [ 'files' ]
 
-def cleanup(topology: Box) -> None:
+def post_quirks(topology: Box) -> None:
   missing = data.get_empty_box()
+  if 'files' in topology:
+    extra_files = [ x.path for x in topology.files if isinstance(x,Box) and 'path' in x ]
+  else:
+    extra_files = []
+
   for ndata in topology.nodes.values():
     if not 'config' in ndata:
       continue
@@ -18,9 +26,14 @@ def cleanup(topology: Box) -> None:
       f'{ndata.device}.j2' ]
 
     for cfg in list(ndata.config):
+      cfg_file: typing.Optional[str] = None
       for cfg_candidate in candidate_files:
-        cfg_file = _files.find_file(f'{cfg}/{cfg_candidate}',topology.defaults.paths.custom.dirs)
+        cfg_path = f'{cfg}/{cfg_candidate}'
+        if cfg_path in extra_files:
+          cfg_file = cfg_path
+          break
 
+        cfg_file = _files.find_file(cfg_path,topology.defaults.paths.custom.dirs)
         if cfg_file:
           break
 
--- a/netsim/extra/check/config/__init__.py
+++ b/netsim/extra/check/config/__init__.py
@@ -3,7 +3,6 @@ import typing
 from box import Box
 
 from netsim import data
-from netsim.augment import devices as a_devices
 from netsim.utils import files as _files
 from netsim.utils import log, strings
 
@@ -11,10 +10,9 @@ _execute_after    = [ 'files' ]
 
 def post_quirks(topology: Box) -> None:
   missing = data.get_empty_box()
+  extra_files: set = set()
   if 'files' in topology:
-    extra_files = [ x.path for x in topology.files if isinstance(x,Box) and 'path' in x ]
-  else:
-    extra_files = []
+    extra_files = { x.path for x in topology.files if isinstance(x,Box) and 'path' in x }
 
   for ndata in topology.nodes.values():
     if not 'config' in ndata:
--- a/tests/coverage/expected/check-config-files.yml
+++ b/tests/coverage/expected/check-config-files.yml
@@ -1,0 +1,161 @@
+files:
+- content: '! hello world configlet'
+  path: hello/frr.j2
+input:
+- coverage/input/check-config-files.yml
+- package:topology-defaults.yml
+links:
+- _linkname: links[1]
+  interfaces:
+  - ifindex: 1
+    ifname: eth1
+    ipv4: 10.1.0.1/30
+    node: r1
+  - ifindex: 1
+    ifname: eth1
+    ipv4: 10.1.0.2/30
+    node: r2
+  linkindex: 1
+  node_count: 2
+  prefix:
+    ipv4: 10.1.0.0/30
+  type: p2p
+name: input
+nodes:
+  r1:
+    _node_config:
+      hello: /etc/config/02-hello.sh:sh
+      initial: /etc/config/01-initial.sh:sh
+    af:
+      ipv4: true
+    box: quay.io/frrouting/frr:10.6.1
+    clab:
+      binds:
+      - source: node_files/r1/initial
+        target: /etc/config/01-initial.sh
+      - source: node_files/r1/hello
+        target: /etc/config/02-hello.sh
+      - source: node_files/r1/daemons
+        target: /etc/frr/daemons
+      - mode: ro
+        source: node_files/-shared-hosts
+        target: /etc/hosts
+      - source: node_files/r1/resolv
+        target: /etc/resolv.conf
+      config_templates:
+      - mode: sh
+        source: initial
+        target: /etc/config/01-initial.sh
+      - mode: sh
+        source: hello
+        target: /etc/config/02-hello.sh
+      - source: daemons
+        target: /etc/frr/daemons
+      - mode: shared
+        source: hosts
+        target: /etc/hosts
+      - source: resolv
+        target: /etc/resolv.conf
+      exec:
+      - sleep 1
+      kind: linux
+    config:
+    - hello
+    device: frr
+    hostname: clab-input-r1
+    id: 1
+    interfaces:
+    - ifindex: 1
+      ifname: eth1
+      ipv4: 10.1.0.1/30
+      linkindex: 1
+      mtu: 1500
+      name: r1 -> r2
+      neighbors:
+      - ifname: eth1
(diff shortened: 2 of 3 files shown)


ipspace:
@roc-ops -- could you please check whether this resolves #3650 for you?

roc-ops:
Tested, and yes — but with a wrinkle worth knowing: **#3650 no longer reproduces on `dev` without this PR either**, so what this PR is doing for us is preventing a regression rather than fixing a live bug.

The original repro from the issue, run on three trees with `__pycache__` cleared and identical invocations:

| tree | `check.config` false positive | configlet lands in `node_files/r1/` |
|---|---|---|
| `dev` (`ab85760ada`) | none | **yes** |
| `chk-config-files` (`fbf02df62d`) | none | **yes** |

So from our side this resolves #3650 and it can be closed.

**Why I think the two halves of the PR are load-bearing together.** On `dev` the check runs in `cleanup`, which is late enough that `hello/frr.j2` is already written, so `find_file` finds it on disk and there is nothing to false-positive on. Moving the check to `post_quirks` puts it *before* the file exists — at which point `find_file` fails and the old symptom would come back. The `topology.files` lookup is what stops that. That is my reading of the diff rather than something I instrumented, so treat it as a hypothesis about *why* it works, not a measurement.

**One extra case, since the hook moved.** Copilot flagged that `node.config` may still carry the group/template removal markers at the new hook point. I tried the obvious shape:

```yaml
plugin: [ files, check.config ]
configlets:
  hello:
    frr: "! hello world configlet"
groups:
  gx:
    members: [ r1, r2 ]
    config: [ hello ]
nodes:
  r1: {}
  r2: { config: [ "-hello" ] }
```

Both `dev` and this branch behave correctly and identically: the configlet lands on `r1`, not on `r2`, and neither emits a spurious warning. So that concern does not reproduce for the `-<name>` form. I did not try every marker shape.

**Unrelated observation, take it or leave it.** With the `files` plugin enabled, a genuinely missing config name (`config: [ definitely_not_there ]`) produces no `check.config` warning on *either* branch — the entry is gone from `node.config` before the check runs. The user still gets `UserWarning in configs: Cannot find definitely_not_there configuration template`, so nothing fails silently, and this is not a regression from your PR. Mentioning it only because it means `check.config` is effectively a no-op for missing names once `files` is in play, which may not be what you intend from a plugin whose whole job is that warning.

Thanks for picking this up.

ipspace:
> Tested, and yes — but with a wrinkle worth knowing: **#3650 no longer reproduces on `dev` without this PR either**, so what this PR is doing for us is preventing a regression rather than fixing a live bug.

This is what I get on the dev branch against the `tests/coverage/input/check-config-files.yml` topology:

```
[WARNING] check.config: Custom configuration hello is missing for device frr (nodes r1)
[HINT]    You will have to configure the missing functionality yourself
[INFO]    files: Created hello/frr.j2
[WRONG]   configs: Cannot find hello configuration template for r1/device frr
[HINT]    Use the '--debug template' option if you're troubleshooting custom configuration templates
[CONFIG]  r1: initial,daemons,hosts,resolv
[CONFIG]  r2: initial,daemons,hosts (shared),resolv
````

It works (obviously ;) with this PR.

> **Why I think the two halves of the PR are load-bearing together.** On `dev` the check runs in `cleanup`, which is late enough that `hello/frr.j2` is already written, so `find_file` finds it on disk and there is nothing to false-positive on.

Interestingly, I got similar error with your sample topology. The file is created in "output" hook, which is later.

> Moving the check to `post_quirks` puts it _before_ the file exists — at which point `find_file` fails and the old symptom would come back.

I had to move the check to "post_transform" otherwise the clab provider creates all sorts of data structures assuming the config will be there, and just removing it from "config" list doesn't do much. However, that's for FRR; for containers that use(d) SSH or Ansible for configuration you wouldn't have the same issues.

> **One extra case, since the hook moved.** Copilot flagged that `node.config` may still carry the group/template removal markers at the new hook point.

Yeah, that was the "oh shit" moment that sent me into yak-shaving. Three commits later, I had the "post_quirks" hook that is the right place to do this check.

> I tried the obvious shape:

Thank you!

> **Unrelated observation, take it or leave it.** With the `files` plugin enabled, a genuinely missing config name (`config: [ definitely_not_there ]`) produces no `check.config` warning on _either_ branch — the entry is gone from `node.config` before the check runs.

This is what I get in the **dev** branch:

```
[WARNING] check.config: Custom configuration xhello is missing for device frr (nodes r1)
[HINT]    You will have to configure the missing functionality yourself
[INFO]    files: Created hello/frr.j2
[WRONG]   configs: Cannot find xhello configuration template for r1/device frr
[HINT]    Use the '--debug template' option if you're troubleshooting custom configuration templates
[CONFIG]  r1: initial,daemons,hosts,resolv
[CONFIG]  r2: initial,daemons,hosts (shared),resolv
```

... and this is what I get with this PR:

```
[WARNING] check.config: Custom configuration xhello is missing for device frr (nodes r1)
[HINT]    You will have to configure the missing functionality yourself
[INFO]    files: Created hello/frr.j2
[CONFIG]  r1: initial,daemons,hosts,resolv
[CONFIG]  r2: initial,daemons,hosts (shared),resolv
[CREATED] provider configuration file: clab.yml
[CREATED] pickled transformed topology data into netlab.snapshot.pickle
[GROUPS]  group_vars for all
[GROUPS]  group_vars for frr
[HOSTS]   host_vars for r1, r2
[CREATED] minimized Ansible inventory hosts.yml
[CREATED] Ansible configuration file: ansible.cfg
```

... which is the expected behavior: warning, followed by completed transformation process.

Thanks again for double-checking this.

roc-ops:
You're right and I was wrong — my previous comment was testing against a contaminated directory. Correcting it for the record, since it's misleading as written.

**What happened.** My first run created `hello/frr.j2` in the lab directory. Between runs I was clearing `node_files/` but not `hello/`, so every subsequent run found the configlet already on disk and `find_file` succeeded. That is why I saw no warning. Worse, my very first run *did* show the warning — the true result — and I talked myself out of it by blaming a stale `__pycache__`. I corrected myself in the wrong direction and then built a mechanism story on top of it.

Re-run in a pristine directory each time, your results reproduce exactly:

| topology | `dev` (`ab85760ada`) | this PR |
|---|---|---|
| `config: [ hello ]` (configlet exists) | `check.config` warning → `Cannot find hello …` → **FATAL**, no `clab.yml` | clean, transformation completes |
| `config: [ xhello ]` (genuinely missing) | warning → **FATAL**, no `clab.yml` | warning, transformation **completes** |

So all three of my claims were artifacts of the same leftover directory:

* **"#3650 no longer reproduces on `dev`"** — wrong. It reproduces, and fatally.
* **"This PR prevents a regression rather than fixing a live bug"** — wrong. It fixes a live bug.
* **"`check.config` is a no-op for missing names once `files` is in play"** — wrong. It warns on both branches; the difference your PR makes is warn-and-continue instead of warn-and-abort, which is the correct behaviour.

**And my mechanism explanation was backwards.** I claimed `cleanup` runs late enough that the file is already written. You pointed out the file is created in the `output` hook, which is *later* than `cleanup` — so `find_file` legitimately fails on `dev`, which is exactly why the check needs `topology.files`. Thanks for the correction, and for the note about having to land on `post_transform`/`post_quirks` because the clab provider builds data assuming the config will be there — that ordering constraint wasn't visible to me from the diff.

**The removal-marker result needs correcting too** — I said it behaved "correctly and identically on both branches", and that was the polluted directory again. Re-run clean, group-inherited `config: [ hello ]` with `-hello` on one node:

* `dev` — aborts with the same fatal error; nothing lands on either node.
* this PR — completes, configlet on `r1`, not on `r2`, no spurious warning.

So the useful half of that result survives: on your branch the `-<name>` marker is handled correctly at the new hook point, which is what Copilot was worried about. It just isn't a both-branches comparison, because `dev` never gets far enough to have an opinion.

Sorry for the noise on a PR you'd already got right.

ipspace:
> You're right and I was wrong — my previous comment was testing against a contaminated directory.

No worries, happens to all of us all the time ;)

> Correcting it for the record, since it's misleading as written.

Thanks a million, Much appreciated.

> Sorry for the noise on a PR you'd already got right.

I much prefer double-checking to being confidently wrong, so thanks again for the checks.

PR #3771, opened 2026-08-14
Deprecate IOSv, describe its ancient SSH woes

ipspace:
Based on #3714:

* Change IOSv support level to 'minimal'
* Add a device caveat explaining the obsolesence of its SSH algorithms
* Expand the 'Cisco IOS SSH access' caveats to document the paramiko 5.0 changes and the potential downgrade path

### DIFF
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -205,7 +205,7 @@ These caveats are common to all Cisco IOS XE platforms:
 
 The Cisco IOSv SSH implementation uses RSA keys and older encryption algorithms that may not be enabled by default on newer Linux distributions.
 
-That wasn't a problem for Ansible users until October 2025, when the new version of the `ansible-pylibssh` package (installed with Ansible) was released. `ansible-pylibssh` release 1.3.0 uses `libssh` release 0.11.0, which [no longer supports legacy SSH algorithms](https://github.com/ipspace/netlab/discussions/2759).
+That wasn't a problem for Ansible users until October 2025, when the new version of the `ansible-pylibssh` package (installed with Ansible) was released. `ansible-pylibssh` release 1.3.0 bundles `libssh` release 0.11.0 with [disabled legacy SSH algorithms](https://github.com/ipspace/netlab/discussions/2759).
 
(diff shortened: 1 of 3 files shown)


docs/caveats.md
```diff
@@ -203,7 +203,7 @@ These caveats are common to all Cisco IOS XE platforms:
 (cisco-ios-ssh)=
 ### SSH Access to Cisco IOS/IOS-XE
 
-The Cisco IOS/IOS-XE SSH implementation uses RSA keys and older encryption algorithms that may not be enabled by default on newer Linux distributions.
+The Cisco IOSv SSH implementation uses RSA keys and older encryption algorithms that may not be enabled by default on newer Linux distributions.
 
 That wasn't a problem for Ansible users until October 2025, when the new version of the `ansible-pylibssh` package (installed with Ansible) was released. `ansible-pylibssh` release 1.3.0 uses `libssh` release 0.11.0, which [no longer supports legacy SSH algorithms](https://github.com/ipspace/netlab/discussions/2759).
```
DanPartelly, line 222: This is actually not true. libssh series 0.11 supports legacy algos, but they are disabled by default and requires configuration, The problem here is that if you install ansible-pylibssh with pip or whatever, you'll probably get a bundled libssh which may or may not not conform with host distribution management of crypto settings.

ipspace: OK, how would you want to rephrase it? How about "ansible-pylibssh includes bundled libssh release 0.11.0 with settings that disable older algorithms"?

DanPartelly: Sure that's OK.

DanPartelly (approved):
Good riddance.

netsim/devices/iosv.py
```diff
@@ -64,17 +64,32 @@ def use_paramiko(node: Box, topology: Box) -> None:
...
       report_quirk(
-        f"Changing Ansible network_cli connection SSH type to 'paramiko'",
+        f"Paramiko version {paramiko.__version__} does not support ancient SSH algorithms used in Cisco IOSv",
         node,
-        quirk='paramiko',
-        category=Warning,
-        more_hints=[ 'The installed version of ansible-pylibssh might not work with Cisco IOS devices' ])
+        quirk='paramiko_50',
+        category=log.MissingDependency,
+        more_hints=['Use "pip install --upgrade \'paramiko<5.0\'" to downgrade paramiko to version 4.0.0'],
```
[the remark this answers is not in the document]
ipspace: Intentional. Either we crash here, or we expose user to an Ansible troubleshooting sessions.

Issue #3760, opened 2026-08-11
DNS clients on Dell OS10


ipspace:
DNS server can run on dnsmasq, and DNS clients on IOS, EOS, and Linux-based platforms. It should be easy to add a few DNS client implementations, and the DNS server can run on (at least) Cisco IOS.

snuffy22:
I can make routeros7 DNS server/client templates if wanted, OpenBSD would probably make do with a `unbound` config which allows for creating domains while allowing recursion vs `nsd` which is authoritative only.

ipspace:
> I can make routeros7 DNS server/client templates if wanted

Perfect. Thank you!

ipspace:
I implemented the DNS clients on Junos, IOS XR, Nexus OS, and SR Linux. Doing it on SR OS (SR SIM) is too cumbersome (you have to configure the BOF data store). The remaining (more supported) platforms are now listed in the issue title. As we implement DNS clients for them, we'll remove them, and eventually close the issue.

As for DNS servers, it's too much hassle/noise (at least on Cisco IOS).

snuffy22:
@ipspace  we can now remove from subject Mikrotik (#3838), OpenBSD (#3837) and VyOS (#3839)

ipspace:
@snuffy22 -- That's good enough for me. Thanks for all the great work, closing the issue.

PR #3796, opened 2026-08-18
Aruba DNS & syslog

sdargoeuves:
ArubaCX: https://github.com/ipspace/netlab/issues/3760

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -292,7 +292,7 @@ The following system-wide features are configured on supported network operating
 | --------------------- | :------: | :--------: | :-----------------------: | :------------------------: | :------------------------: |
 | Arista EOS               | ✅  | ✅  | ✅  | ✅  | ✅  |
 | Arrcus ArcOS             | ✅  |  ❌  |  ❌  | ✅  | ✅  |
-| Aruba AOS-CX             | ✅  |  ❌  | ✅  | ✅  | ✅  |
+| Aruba AOS-CX             | ✅  |  ✅  | ✅  | ✅  | ✅  |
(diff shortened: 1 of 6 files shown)


ipspace (changes requested):
The DNS resolution through a VRF does not work for me. What version of ArubaCX are you using?

Also, considering so many other platforms cannot do DNS resolution through a VRF, maybe we disable it for ArubaCX as well?

ipspace:
OK, I figured out the details

```
ip dns domain-list cicd.local vrf oob_nms
ip dns server-address 172.16.0.2 vrf oob_nms
```

... configures DNS resolution _within the VRF_ -- DNS client runs in the VRF and is used for commands like **ping X vrf Y**. ArubaCX does not seem to have the "global DNS client using VRF for the transport" functionality. I'll change the feature flags, rerun the DNS-in-VRF test, and merge the PR.

ipspace (approved):
The modified feature flags work. Merging.

Thank you for the PR!

sdargoeuves:
Thanks for the correction. I have to admit I was relying on their documentation... next on my todo list:
- start using the integration tests
- **stop assuming** something will work just because you think it could, or you've read it somewhere in the vendor docs!

ipspace:
> Thanks for the correction. I have to admit I was relying on their documentation

I looked at that documentation and said "That's fishy. There's no way it does what I want it to do" ;)) Must have seen too many vendor docs for one lifetime ;)

> start using the integration tests

👍

> **stop assuming** something will work just because you think it could, or you've read it somewhere in the vendor docs!

That's always a dangerous assumption.

PR #3802, opened 2026-08-20
fix: memory & status info for Podman

sdargoeuves:
## Fix `netlab status --memory` for Podman runtime

When using `defaults.providers.clab.runtime: podman`, the `netlab status` and
`netlab status --memory` commands showed no container status or memory usage.

### Root cause

Podman's Docker-compatible CLI returns JSON differently from Docker:

| | Docker | Podman |
|---|---|---|
| `docker ps --format json` | one JSON object per line | JSON array |
| `Names` field | string | list of strings |
| `docker stats --format json` | one JSON object per line | JSON array |
| field names | `Name`, `MemUsage` | `name`, `mem_usage` |
| memory units | binary (`KiB`, `MiB`, `GiB`) | decimal (`kB`, `MB`, `GB`) |

### Changes

**`providers/clab/__init__.py`**

- `get_lab_status`: if `docker ps` output is a JSON array (Podman), normalize it
  to Docker's per-line format and unwrap `Names` from list to string before
  running the existing parsing logic unchanged.
- `add_memory_usage`: same normalization for `docker stats` — map Podman's
  `name`/`mem_usage` fields to Docker's `Name`/`MemUsage`, then parse and
  re-format the value in binary units (MiB/GiB) so the memory column is
  consistent across Docker, Podman, and libvirt VMs.

**`utils/strings.py`**

- `MEMORY_UNITS`: added decimal SI units (`KB`, `MB`, `GB`, `TB`) alongside
  the existing binary ones, with correct byte-to-KiB conversion factors.
- `parse_memory_size`: extended the regex to also match decimal unit suffixes
  (`[KMGT]B`).

### DIFF
--- a/netsim/providers/clab/__init__.py
+++ b/netsim/providers/clab/__init__.py
@@ -86,12 +86,18 @@ class Containerlab(_Provider):
       if not isinstance(status,str):
         return stat_box
       try:
-        for line in status.split('\n'):
+        status_str = status.strip()
+        if status_str.startswith('['):                        # Podman: normalize array to one-object-per-line
+          status_str = '\n'.join(
+            json.dumps({**item, 'Names': item['Names'][0] if isinstance(item.get('Names'), list) else item.get('Names','')})
+            for item in json.loads(status_str))
+
+        for line in status_str.split('\n'):
           if not line.startswith('{'):
             continue
           docker_stats = json.loads(line)
-          stat_box[docker_stats['Names']].status = docker_stats['Status']
-          stat_box[docker_stats['Names']].image = docker_stats['Image']
+          stat_box[docker_stats['Names']].status = docker_stats.get('Status','')
+          stat_box[docker_stats['Names']].image = docker_stats.get('Image','')
       except Exception as ex:
         log.error(f'Cannot get Docker status: {ex}',category=log.FatalError,module='clab')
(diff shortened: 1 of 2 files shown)


sdargoeuves:
I'm in unchartered territory regarding git/github. I've submitted the PR against your branch `nl-status-collecting` instead of `dev` as from my perspective, those 2 PR are linked.
Now github is proposing to stack those PR, not sure if I should or not.
Let me know if it's just easier to create the PR against `dev`!

ipspace:
> I'm in unchartered territory regarding git/github.

You're obviously not the only one lost :((

> Let me know if it's just easier to create the PR against `dev`!

I merged the other PR and deleted the branch (as always), and all of a sudden, this one is an orphan :(( I'm sorry for the mess; please resubmit it.

sdargoeuves:
> all of a sudden, this one is an orphan :(( I'm sorry for the mess; please resubmit it.

Don't worry it's all good, it was interesting to see what would happen! And it makes sense this PR became orphan as it was against a branch which no longer exists.
I'm not sure what is the _correct_ process to submit a PR against an un-merged branch, with changes on the same file. It doesn't happen enough to be something I want to investigate yet.

Long story short, I'll create the new PR :)

PR #3804, opened 2026-08-21
Fix: Podman cli status & memory

sdargoeuves:
## Fix `netlab status --memory` for Podman runtime

When using `defaults.providers.clab.runtime: podman`, the `netlab status` and
`netlab status --memory` commands showed no container status or memory usage.

### Root cause

Podman's Docker-compatible CLI returns JSON differently from Docker:

| | Docker | Podman |
|---|---|---|
| `docker ps --format json` | one JSON object per line | JSON array |
| `Names` field | string | list of strings |
| `docker stats --format json` | one JSON object per line | JSON array |
| field names | `Name`, `MemUsage` | `name`, `mem_usage` |
| memory units | binary (`KiB`, `MiB`, `GiB`) | decimal (`kB`, `MB`, `GB`) |

### Changes

**`providers/clab/__init__.py`**

- `get_lab_status`: if `docker ps` output is a JSON array (Podman), normalize it
  to Docker's per-line format and unwrap `Names` from list to string before
  running the existing parsing logic unchanged.
- `add_memory_usage`: same normalization for `docker stats` — map Podman's
  `name`/`mem_usage` fields to Docker's `Name`/`MemUsage`, then parse and
  re-format the value in binary units (MiB/GiB) so the memory column is
  consistent across Docker, Podman, and libvirt VMs.

**`utils/strings.py`**

- `MEMORY_UNITS`: added decimal SI units (`KB`, `MB`, `GB`, `TB`) alongside
  the existing binary ones, with correct byte-to-KiB conversion factors.
- `parse_memory_size`: extended the regex to also match decimal unit suffixes
  (`[KMGT]B`).

### DIFF
--- a/netsim/providers/clab/__init__.py
+++ b/netsim/providers/clab/__init__.py
@@ -86,12 +86,18 @@ class Containerlab(_Provider):
       if not isinstance(status,str):
         return stat_box
       try:
-        for line in status.split('\n'):
+        status_str = status.strip()
+        if status_str.startswith('['):                        # Podman: normalize array to one-object-per-line
+          status_str = '\n'.join(
+            json.dumps({**item, 'Names': item['Names'][0] if isinstance(item.get('Names'), list) else item.get('Names','')})
+            for item in json.loads(status_str))
+
+        for line in status_str.split('\n'):
           if not line.startswith('{'):
             continue
           docker_stats = json.loads(line)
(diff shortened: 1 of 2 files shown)


ipspace (approved):
Looks good to me, and still works with Docker ;) Thank you!

PR #3805, opened 2026-08-21
Add 'netmiko' device configuration method

ipspace:
This code adds support for 'netmiko' netlab_config_mode which invokes netmiko library to connect to the lab device(s) and send configuration commands from node_files.

Initial test case: Cisco IOS devices -- netmiko configuration method is allowed but not enabled. You have to enable it in user defaults

Changes made to 'cli/initial/deploy':
* Rename 'deploy_provider_config' to 'deploy_internal_config' (because it's no longer provider-specific)
* Add a lookup table of internal config mechanisms (currently: netmiko)
* Execute internal config mechanism and provider-specific config mechanism

Changes made to provider code:
* 'add_default_config_mode' is now a shared provider function
* clab code calls the shared function and adds 'netlab_config_exec' command only if the shared function returns True (the default config method is used)
* libvirt allows 'netlab_config_method' for 'none' device and for 'netmiko' value

Also included:
* Documentation update
* Platform integration test
* Initial configuration integration test

### DIFF
--- a/docs/platforms.md
+++ b/docs/platforms.md
@@ -262,13 +262,13 @@ _netlab_ uses Ansible playbooks and device-specific task lists to deploy device 
 
 [^cRBS]: The configuration deployment uses a custom **bash** script that calls **cli** command to execute **load merge** followed by **commit**. The custom script is used as the *shebang* interpreter for the configuration snippets.
 
-Several other devices can use alternate (faster) configuration methods that are not enabled by default; you have to set the **netlab_config_mode** group variable[^NCMGV] or node parameter to use them:
+Several other devices can use alternate (faster) configuration methods that are not enabled by default; you have to set the **netlab_config_mode** device group variable[^NCMGV] or node parameter to use them:
 
 | Device    | containerlab<br> deployment method |
 |-----------|------------------------------------|
 | Arista EOS | **sh**[^EOSSH] |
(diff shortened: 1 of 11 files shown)


ipspace:
@DanPartelly @sdargoeuves -- finally found the motivation to implement netmiko-based configuration deployment. You might be the most interested in this one (particularly as the first implementation is Cisco IOS)

Muddyblack:
This works with just 300 lines of code nice. Although I suppose this has many other PRs before to make this one here work.

This now works without ansible at all if netmiko is enabled?
I will be happy to test it on frr soon :P

ipspace:
> Although I suppose this has many other PRs before to make this one here work.

Exactly ;)

> This now works without ansible at all if netmiko is enabled?

For the supported devices (currently, Cisco IOS) and when enabled in user defaults.

> I will be happy to test it on frr soon :P

FRR has been "forever" configured with scripts that were triggered directly by **netlab initial** -- you didn't need Ansible to configure FRR, BIRD, and a few other things for quite a while.

Muddyblack:
I have to take a closer look on why I thought it still uses it 😅😅😅. But yes I remeber the part that frr uses scripts....

Muddyblack:
Ah yes because it still makes a groups vars folder etc. Ansible style a like made me think of that for a short time ups :D

PR #3809, opened 2026-08-23
Fix: Make routing.acl.in/out a link attribute

### LINKED ISSUE #3766: ACL on SVIs
We should be able to specify ACls directly on VLAN interface, and  would be also nice to specify prefixes as vlan-ids.

ipspace:
You could apply routing.acl to an interface, which prevented its use on VLANs (see #3766). The attribute was changed to a link attribute, so it can be used on a link (or an interface), a global VLAN definition or a node VLAN definition. The attached coverage test validates the global/node VLAN use case.

The documentation was updated to explain where the routing.acl attribute could be attached.

Closes #3766


DanPartelly (approved):
Thank you.

Issue #3829, opened 2026-08-26
[BUG] multiserver plugin: cross-server links generate invalid containerlab syntax


sdargoeuves:
# multiserver plugin: cross-server P2P links generate invalid containerlab syntax

**Versions:** netlab `dev`, containerlab 0.79.0

`netlab up` fails immediately when the `multiserver` plugin generates a topology
containing cross-server point-to-point links:

```
Failed to read topology file: yaml: unmarshal errors:
  line 67: field endpoints not found in type struct { Type string "yaml:\"type\"";
  links.LinkVxlanRaw "yaml:\",inline\"" }
  [... one error per cross-server link ...]
```

## Likely root cause (suggested by AI)

`netsim/templates/provider/clab/clab.j2:104-110` mixes two incompatible
containerlab link syntaxes in a single link definition:

```yaml
  - endpoints:                  # brief/veth syntax
    - "d1xfw01:eth1"
    - "host:vx10000"            # host device that is never created
    type: vxlan                 # typed-link syntax
    remote: "10.194.50.3"
    vni: 10000
    udp-port: 4789
```

A containerlab `type: vxlan` link takes a single `endpoint`, not an `endpoints`
list — the far end is described by `remote` + `vni`. See `LinkVxlanRaw` in
[links/link_vxlan.go](https://github.com/srl-labs/containerlab/blob/main/links/link_vxlan.go)
and `link-type-vxlan` in `schemas/clab.schema.json`.

## Suggested fix

Emit the `endpoint` mapping instead of the `endpoints` list:

```jinja
{%     for n in nodes.values() if not (n.unmanaged|default(False)) %}
{%       for nl in n.interfaces|default([]) if nl.linkindex|default(0) == l.linkindex %}
  - type: vxlan
    endpoint:
      node: "{{ n.name }}"
      interface: "{{ nl.clab.name|default(nl.ifname) }}"
    remote: "{{ l.clab.vxlan.remote }}"
    vni: {{ l.clab.vxlan.vni }}
    dst-port: {{ l.clab.vxlan.dstport }}
{%       endfor %}
{%     endfor %}
```

I also renamed `udp-port` to `dst-port`: containerlab marks
`udp-port` as deprecated in `LinkVxlanRaw` in favour of `dst-port`, which pairs
with the later `src-port` option. Both keys still work, `udp-port` is copied to
`dst-port`.

`dst-port` requires containerlab v0.70.0

## Testing

[...]

PR #3830, opened 2026-08-26
Fix/multiserver vxlan clab link syntax

sdargoeuves:
https://github.com/ipspace/netlab/issues/3829


ipspace (approved):
Works like a charm once I fixed the MTU issue on the worker server. Will add that to the documentation.

Thank you!

PR #3831, opened 2026-08-26
Generic prefix sets (used in ACL src/dst prefix)

### LINKED ISSUE #3810: Additional ACL prefix-generating attributes
> Resolving prefixes by the new linkid is something we should have.  (by @DanPartelly in [#3766](https://github.com/ipspace/netlab/issues/3766#issuecomment-5306178618))

@DanPartelly @jbemmel -- would it make more sense to have new attributes (**linkid**, **vlan**, **role**) or to reuse the **prefix** attribute and expand it to look up named prefixes, link IDs, link roles, and VLAN prefixes?

ipspace:
Based on the ideas discussed in #3810, this PR adds a new data type: generic prefix. It can be an IPv4/IPv6 address/prefix, or a named prefix, vlan prefix, address pool, link prefix, or a set of prefixes from links with specified role (I'm pretty sure we'll get further ideas ;).

It's currently used in ACLs, but could be integrated into prefix lists and static routes.

The PR includes:

* A new type validator and corresponding coverage test
* The 'eval_prefixset' code that transforms a named prefix into a list of ipv4/ipv6 addresses
* Updated documentation end-user and developer documentation

### DIFF
--- a/docs/dev/validation-types.txt
+++ b/docs/dev/validation-types.txt
@@ -15,6 +15,7 @@ Validator recognizes standard Python data types (**str**, **int**, **float**, **
 | **ipv6**       | An IPv6 address, prefix, integer (offset in a subnet), or bool (LLA only) |
 | **mac**        | MAC address in any format recognized by the `netaddr` library |
 | **named_pfx**  | [named prefix](named-prefixes) |
+| **prefixset**  | [a generic prefix](generic-routing-prefix-set) |
 | **net**        | IS-IS NET/NSAP |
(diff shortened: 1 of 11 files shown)


ipspace:
@DanPartelly -- this is the best I could do to implement #3810 without breaking existing functionality. In the end, I think it's not too shabby ;)

DanPartelly (approved):
If flows nice, and a pretty good improvement on UX.

ipspace:
> If flows nice, and a pretty good improvement on UX.

Thank you!

PR #3832, opened 2026-08-28
Ensure Vagrant provider is 'libvirt' when trying to package our own boxes

snuffy22:
We have earlier fixed issue with `netlab down` where you cannot specify the provider to use via CLI and must use Environment variable.

This fixes use when we are building our own vagrant boxes under WSL.


ipspace (approved):
Thank you!

PR #3842, opened 2026-09-01
Refuse MLAG peer-links on ESI-LAG-only devices (#3835)

### LINKED ISSUE #3835: lag: a device declaring ESI-LAG cannot decline MLAG (single OR gates both peer-link and M-side)
### Summary

`netsim/modules/lag.py` gates **both** the MLAG peer-link machinery and the M-side LAG check
through a single predicate:

```python
features.lag.mlag or features.evpn.multihoming.lag      # lag.py:77
```

consulted from `create_peer_links` (line 273) and the M-side check (line 226). So a device that
declares **ESI-LAG cannot decline MLAG** — the two capabilities are not separable.

### Concrete consequence

`docker-sonic-vs` ships no `mclagd`, and MLAG provably cannot work on it: the pair never forms,
the host aggregates with exactly one switch, and the second switch's port stays `NO-CARRIER`.
Yet `netlab create` accepts a full MLAG topology on it — purely because ESI-LAG is declared, which
**does** work there (live-verified: ESI advertised, DF elected, remote VTEP seen).

The result is a topology that builds, deploys, and silently does not do the thing it describes.

### Suggested fix, and what we measured

Gate `create_peer_links` on `features.lag.mlag` **alone**, leaving the M-side check on the OR. A
peer-link only ever exists for real MLAG — ESI-LAG coordinates over BGP EVPN Type-1/Type-4 and
builds none. `tests/integration/lag/21-evpn-multihoming.yml`'s own header says "No MLAG peer-link
needed".

Tested against both of netlab's own topologies on a device declaring both capabilities:

| change | `10-mlag.yml` | `21-evpn-multihoming.yml` |
|---|---|---|
| as-is | creates (wrongly) | creates |
| drop `lag.mlag.peer.ip` only | creates (wrongly) | creates |
| narrow the gate only | creates (wrongly) | creates |
| **narrow the gate AND `lag.mlag: False`** | **rejected, correctly** | **creates** |

The last row is the desired behaviour. Note it takes **both** halves: narrowing the gate alone
does not help a device that declares `lag.mlag.peer.ip`, because a `peer.ip` dict is truthy and
passes the narrowed check. So the fix is upstream's half, and the device author's half is to stop
declaring `lag.mlag`.

One further detail: on the peer-link path the error would still read *"does not support MLAG or
ESI-LAG"*, which is wrong in the ESI-LAG half once the check is narrowed.

Offered as a proposal to validate rather than a certainty — we read the call graph and ran the two
topologies, but did not run the variant against a device that only ever had MLAG.

jbemmel:
Peer-link creation used the same OR as the M-side check, so declaring ESI-LAG made netlab accept MLAG topologies the device cannot run.

Fixes #3835

Note: The check for ```'evpn.multihoming.lag'``` is a little ugly, but I don't see a better way to allow the plugin to take responsibility for the M-side; lag runs first. Adding another hook just for the plugin to set some other generic flag to bypass the M-side check seems overkill

### DIFF
--- a/netsim/modules/lag.py
+++ b/netsim/modules/lag.py
@@ -68,14 +68,36 @@ def check_lag_config(node: str, linkname: str, topology: Box) -> bool:
     return False
   return True
 
+def _supports_mlag(features: Box) -> bool:
+  return bool(features.lag.get('mlag',False))
+
+def _supports_esi_lag(features: Box) -> bool:
+  return bool(features.get('evpn.multihoming.lag',False))
+
 """
-check_mlag_support - check if the given node supports mlag
+check_mlag_support - check if the given node supports classic MLAG (peer-link)
 """
 def check_mlag_support(node: str, linkname: str, topology: Box) -> bool:
   _n = topology.nodes[node]
   features = devices.get_device_features(_n,topology.defaults)
-  if not (features.lag.get('mlag',False) or features.get('evpn.multihoming.lag',False)):
-    log.error(f'Node {_n.name} ({_n.device}) does not support MLAG or ESI-LAG, cannot be part of peerlink or M-side of LAG {linkname}',
+  if not _supports_mlag(features):
+    log.error(f'Node {_n.name} ({_n.device}) does not support MLAG, cannot be part of peerlink {linkname}',
+      category=log.IncorrectValue,
+      module='lag')
+    return False
+  return True
+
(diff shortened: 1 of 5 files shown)


ipspace:
@jbemmel -- Thanks a million!

@roc-ops -- Does this work for you?

PR #3850, opened 2026-09-02
Create hashicorp repository in a way that doesn't conflict

### LINKED ISSUE #3848: [BUG] Check for existing hashicorp in apt sources
If a user has already installed other Hashicorp products on Ubuntu, following the [official hashicorp installation guide](https://developer.hashicorp.com/terraform/tutorials/aws-get-started/install-cli), there is already a file `/etc/apt/sources.list.d/hashicorp.list`

`netlab install libvirt` creates a new file `vagrant.list` in the same directory with the same contents, which ends up creating a conflict in apt.

> E: Conflicting values set for option Signed-By regarding source https://apt.releases.hashicorp.com/ noble: /usr/share/keyrings/hashicorp-archive-keyring.gpg !

The only way to get past the installation is to delete the `vagrant.list` file.

`netlab install libvirt` should avoid creating a duplicate file that conflicts.

sc68cal:
Update the install script to stop conflicting with Hashicorp's official install guide, by replicating the same commands.

This fixes #3848

### DIFF
--- a/netsim/install/libvirt.sh
+++ b/netsim/install/libvirt.sh
@@ -30,6 +30,8 @@ echo
 echo "Install vagrant"
 echo ".. setting up Vagrant repository"
 set +e
+# Delete pre-3848 location for hashicorp keyring
+$SUDO rm /etc/apt/trusted.gpg.d/hashicorp-security.gpg 2>/dev/null
 $SUDO rm /usr/share/keyrings/hashicorp-archive-keyring.gpg
 # Delete old vagrant.list
 # See #3848
--- a/netsim/install/libvirt.sh
+++ b/netsim/install/libvirt.sh
@@ -30,6 +30,7 @@ echo
 echo "Install vagrant"
 echo ".. setting up Vagrant repository"
 set +e
+$SUDO rm /usr/share/keyrings/hashicorp-archive-keyring.gpg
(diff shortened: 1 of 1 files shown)


ipspace (changes requested):
Thanks a million for the fix. However, you are using a keyring (`/usr/share/keyrings/hashicorp-archive-keyring.gpg`) that is not installed by this script, so the subsequent `apt` operations fail. I could modify the `hashicorp.list` to not use that key (like we did before), or you could add the keyring installation step.

sc68cal:
Good catch. I will update that segment to install it the same way that it is documented by hashicorp

sc68cal:
To close the loop, this location for their GPG key is from their documentation, inside a tab for package based Linux installs.

image

sc68cal:
Feel free to squash these into a single commit upon merge

ipspace (approved):
After moving the cleanup commands (because the previous version of the script borked repository definition), it all works like a charm.

Thanks again! Ivan

ipspace:
Also, welcome to the contributors club ;))

sc68cal:
Thanks! I enjoy your blog, podcast, and other work so it's nice to be able to contribute a small bit back

PR #3861, opened 2026-09-04
Create dynamic document feature tables for bgp.session.md

snuffy22:
Add the dynamic table for `bgp.session.md`

Would like comments on how I handled the `remove_private_as`


ipspace (approved):
Looks pretty good, thank you. You removed the paragraphs above the tables, so I'll add them back before merging.

ipspace:
> Would like comments on how I handled the `remove_private_as`

Pretty good, apart from (A) "on" being translated into "True" (yeah, one has to love YAML) and crashing the extension, and not checking whether "remove_private_as" is True (which I fixed by cleaning up the device definitions).

Have to add "this is now you test changes to documentation" section

PR #3866, opened 2026-09-05
Create dynamic BGP feature tables for bgp.md

snuffy22:
Here is the transitioned bgp.md.

Had to add a caveats section to `srlinux` and `sros` device files to get the '!'

### DIFF
--- a/docs/module/bgp.md
+++ b/docs/module/bgp.md
@@ -53,27 +53,38 @@ Even more BGP features are implemented in the following plugins:
 (bgp-platform)=
 ## Platform Support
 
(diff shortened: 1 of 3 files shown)


ipspace (approved):
Thank you... and you REALLY SHOULD test stuff locally.

The import caveats failed because "bgp.import_caveats" resulted in ['vrf'], which is not a valid anchor. I had to restructure the caveats a bit.

Also, I removed the columns that were matched by all devices (because that's no fun) and added "this is supported by (almost) all platforms" list.

PR #3874, opened 2026-09-07
Use L4 hashing for Linux ECMP (#3770)

### LINKED ISSUE #3770: Feature: Support setting net.ipv4.fib_multipath_hash_policy for Linux / FRR ECMP hashing
When running ECMP topologies on Linux-based nodes or FRR routers in netlab, multi-flow traffic fails to load-balance across parallel paths as expected.

By default, the Linux kernel sets net.ipv4.fib_multipath_hash_policy = 0 (Layer 3 outer IP header hashing only). This creates major limitations during lab testing:

Unencapsulated IP Traffic: Standard TCP/UDP multi-stream tests (e.g., using iperf3 -P 4) between two hosts pin all streams to a single nexthop leg because outer IP addresses are identical. Enabling Layer 4 hashing (policy = 1) is required to split flows by L4 ports.

Encapsulated Traffic (VXLAN, GRE, IPIP): Tunnels sharing identical outer IP endpoints send all encapsulated traffic down a single path. Enabling inner Layer 3 hashing (policy = 2) is required to balance based on inner payload headers.

Currently, netlab does not provide an explicit mechanism or default provision step to adjust fib_multipath_hash_policy when deploying FRR or other Linux-based nodes.

Automated / Sensible Defaults:

Automatically set net.ipv4.fib_multipath_hash_policy = 1 (L4 hashing) on Linux router nodes when ECMP/multipathing is enabled.

Automatically promote to policy = 2 (or allow explicit toggle) when overlay modules like vxlan or gre tunnel are loaded.

jbemmel:
Linux hashes only the outer IP header by default, so multi-flow tests pin all streams to one nexthop. Set fib_multipath_hash_policy=1 on FRR, Linux, and Cumulus nodes.

Fixes #3770

### DIFF
--- a/netsim/ansible/templates/initial/cumulus.j2
+++ b/netsim/ansible/templates/initial/cumulus.j2
@@ -120,6 +120,7 @@ done
 {% for l in interfaces if l.type in ['lan','p2p','stub','lag','loopback'] and (l.ipv6 is defined or 'external' in l.role|default('')) %}
 sysctl -qw net.ipv6.conf.{{ l.ifname }}.disable_ipv6=0
 {% endfor %}
+sysctl -w net.ipv4.fib_multipath_hash_policy=1
 #
 # Enable FRR modules for {{ module|default('none') }}
 #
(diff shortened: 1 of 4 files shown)


ipspace (approved):
Thank you!

PR #3886, opened 2026-09-20
Cleanup netlab installation instructions

ipspace:
What started as a "add a few bits of WSL info" turned into a major cleanup of installation instructions:

* Prioritize Docker/containerlab over KVM/Vagrant, mention Podman
* Prioritize shell/netmiko over Ansible
* Remove duplicate information (for example, package installation instructions)
* Remove obsolete vagrant management network details
* Add WSL-specific info

Minor details:

* Grammarly-suggested fixes
* Use anchors instead of filenames
* Remove obsolete GCP instructions


ipspace:
What started as a simple "let's add some WSL info from @snuffy22" task turned into a major cleanup. I hope the instructions are more structured now than they were before.

Also, @snuffy22, I can't add you as a reviewer, but would love to hear your feedback.

Thank you!

docs/install/cloud.md
```diff
@@ -1,32 +1,16 @@
...
 
-If you want to run network devices as virtual machines, the public cloud has to support *nested virtualization*. Create a Ubuntu instance, verify that it has virtualization capabilities (`kvm-ok` is usually a good way to find that out), and follow [](ubuntu.md) instructions.
+If you want to run network devices as virtual machines, the public cloud has to support *nested virtualization*. Create an Ubuntu instance, verify it has virtualization capabilities (`kvm-ok` is usually a good way to check that), and follow the [](install-ubuntu) instructions.
 
-Nested virtualization seems to be supported on AWS bare-metal instances (they tend to be expensive), Oracle Cloud bare-metal instances, [some Azure instances](https://azure.microsoft.com/en-gb/blog/nested-virtualization-in-azure/), in Google Cloud, by Packet (bare-metal provider), and by Digital Ocean.
+Nested virtualization seems to be supported on AWS bare-metal instances (they tend to be expensive), Oracle Cloud bare-metal instances, [some Azure instances](https://azure.microsoft.com/en-gb/blog/nested-virtualization-in-azure/), in Google Cloud, by Packet (bare-metal provider), and by DigitalOcean.
```
ssasso: to be 100% accurate, AWS started supporting nested virtualization in some instances as well: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/amazon-ec2-nested-virtualization.html

ssasso (changes requested):
I would add the link for supported aws instances for nested virtualization (like you already did for azure), then all good :)

docs/install/cloud.md
ipspace: Thanks a million, will fix.

snuffy22:
All of it seems fine to me.

Not sure if it is worth mentioning that containers depend on host Linux kernel modules that might not be included (at least with WSL distros) such as MPLS.

As to why you might not have been able to add as reviewer.. I forgot accept the upgrade from contributor to collaborator and it expired.

ipspace:
> All of it seems fine to me.

Thank you!

> Not sure if it is worth mentioning that containers depend on host Linux kernel modules that might not be included (at least with WSL distros) such as MPLS.

Not all containers do. For example, Arista EOS and SR Linux do not rely on MPLS kernel module to do MPLS packet forwarding. Have to check whether it's documented somewhere.

> As to why you might not have been able to add as reviewer.. I forgot accept the upgrade from contributor to collaborator

I know ;))

> and it expired.

That was easy to fix ;)

PR #3888, opened 2026-09-21
EBGP propagation of SRv6 SIDs

ipspace:
* Supports global IPv4 BGP routes with SRv6 SID next hops where the SRv6 SID prefixes are advertised over EBGP
* Adds new allowed values
* Generates bgp.srv6 flag to streamline the config templates
* Implemented on FRR and IOS XR


ipspace:
@jbemmel -- Finally found time to implement the rest of the ideas from #3263 with a few extra quirks (that took hours to troubleshoot) for good measure

netsim/modules/srv6.py
```diff
@@ -56,8 +56,9 @@ def configure_bgp_for_srv6(node: Box, topology: Box) -> None:
     if not need_srv6:                                   # No SRv6 for this neighbor?
       continue                                          # ... cool, let's get out of here
 
-    if nb.type == 'ebgp':                               # Are we running SRv6 services with EBGP neighbor?
-      nb.srv6.next_hop_unchanged = True                 # Cool, but we have to take care of next hops
+    node.bgp.srv6 = True                                # BGP SRv6 is enabled on this node
+    if 'ebgp' in srv6.get('igp',[]):                    # Are we using EBGP with SRv6?
+      nb.srv6.next_hop_unchanged = True                 # We have to take care of unchanged next hops
```
jbemmel, line 61: ```nb.next_hop_unchanged``` ?

jbemmel (changes requested):
Lovely, check that next hop flag (was already in the source, may have been me) but otherwise looks good to me

ipspace:
> Lovely, check that next hop flag (was already in the source, may have been me) but otherwise looks good to me

It turns out that the next-hop thing is a red herring. Next hops change but the SID attached to the route does not, so the traffic always goes directly to the egress PE-router. I'll leave the "nb.next_hop_unchanged" logic there for the eventual Inter-AS option C MPLS/VPN use case.

jbemmel:
> I'll leave the "nb.next_hop_unchanged" logic there for the eventual Inter-AS option C MPLS/VPN use case.

My point was this:
```
jbemmel@jvb-laptop:~/Projects/3rd-party/netlab$ grep -r ".next_hop_unchanged" *
netsim/ansible/templates/srv6/frr.bgp.j2:{%   if n.next_hop_unchanged is defined %}
netsim/modules/srv6.py:      nb.srv6.next_hop_unchanged = True                 # We have to take care of unchanged next hops
```

We're setting 'nb.srv6.next_hop_unchanged' but reading 'nb.next_hop_unchanged', so it won't have any effect

ipspace:
> My point was this:

I know exactly what it was...

> We're setting 'nb.srv6.next_hop_unchanged' but reading 'nb.next_hop_unchanged', so it won't have any effect

... and I explained why it doesn't matter for the global IPv4 BGP case ;)

